PR #4, opened 2017-09-27
Implement a new Undo plugin

JiuqingSong:
Implement a new Undo plugin, this will fix:

1. Can undo the very first change
2. addUndoSnapshot() now can check duplicated content so that plugin developer can feel free to call this API and no worry about the duplication.
3. Support preserve snapshots so that when close the editor and reopen, undo will still work
4. The new Undo will be the default Undo and no need to specify it when creating editor now.

The original HtmlSnapshotUndo is deprecated, will be removed in next major change.


packages/roosterjs-editor-core/lib/undo/Undo.ts
```diff
@@ -0,0 +1,226 @@
...
+            this.addUndoSnapshot();
+            this.lastKeyPress = 0;
+        }
+    }
+
+    private onKeyPress(pluginEvent: PluginDomEvent): void {
+        let evt = pluginEvent.rawEvent as KeyboardEvent;
+        if (evt.metaKey) {
+            // if metaKey is pressed, simply return since no actual effect will be taken on the editor.
+            // this is to prevent changing hasNewContent to true when âŒ˜ + v to paste on Safari.
```
redmondbowang: The comment is messed up -> 'âŒ˜ + v'

packages/roosterjs-editor-core/lib/undo/Undo.ts
```diff
@@ -0,0 +1,226 @@
...
+                selectionRange &&
+                (!selectionRange.collapsed ||
+                    // If the selection contains image, we need to add undo snapshots
+                    containsImage(selectionRange.startContainer))
+            ) {
+                shouldTakeUndo = true;
+            }
+        } else if (this.hasNewContent && evt.which >= KEY_PAGEUP && evt.which <= KEY_DOWN) {
+            // PageUp, PageDown, Home, End, Left, Right, Up, Down
+            shouldTakeUndo = true;
```
MSJiaGuo: It may add a lot of snapshots if user keep press "Left"

packages/roosterjs-editor-core/lib/undo/Undo.ts
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        if (shouldTakeUndo) {
+            this.addUndoSnapshot();
+            this.lastKeyPress = 0;
+        }
+    }
+
+    private onKeyPress(pluginEvent: PluginDomEvent): void {
```
MSJiaGuo: What's the difference between keyDown and keyPress event? Why does some keys be handled in keyDown event, but some keys be handled in keyPress event?

packages/roosterjs-editor-core/lib/undo/UndoSnapshots.ts
```diff
@@ -0,0 +1,61 @@
...
+    private totalSize: number;
+    private currentIndex: number;
+
+    constructor(private maxSize: number = MAXSIZELIMIT) {
+        this.snapshots = [];
+        this.totalSize = 0;
+        this.currentIndex = -1;
+    }
+
+    public canMove(delta: number) {
```
redmondbowang: Do we need to add return types?

packages/roosterjs-editor-core/lib/undo/Undo.ts
```diff
@@ -0,0 +1,226 @@
...
+                this.onKeyDown(event as PluginDomEvent);
+                break;
+            case PluginEventType.KeyPress:
+                this.onKeyPress(event as PluginDomEvent);
+                break;
+            case PluginEventType.CompositionEnd:
+                this.clearRedoForInput();
+                break;
+            case PluginEventType.ContentChanged:
+                if ((<ContentChangedEvent>event).source != 'Undo' && !this.isRestoring) {
```
redmondbowang: What's the difference between "<ContentChangedEvent>event).source != 'Undo'" and "!this.isRestoring" ?  Doesn't event source is undo meaning it's restoring a snapshot?

packages/roosterjs-editor-core/lib/undo/snapshotUtils.ts
```diff
@@ -0,0 +1,139 @@
...
+    return htmlContent;
+}
+
+// Restore a snapshot
+export function restoreSnapshot(editor: Editor, snapshot: string): void {
+    editor.setContent(snapshot);
+
+    // Restore the selection and delete the cursor marker afterwards
+    updateSelectionToCursorMarkers(editor);
+    removeCursorMarkers(editor);
```
redmondbowang: I've seen cases where cursor marker still exists after some actions. Maybe errors are thrown in 'updateSelectionToCursorMarkers', should we make it like 'try{} finally{}'?

packages/roosterjs-editor-core/lib/undo/Undo.ts
```diff
@@ -0,0 +1,226 @@
...
+        }
+    }
+
+    /**
+     * Dispose this plugin
+     */
+    public dispose() {
+        this.editor = null;
+
+        if (!this.preserveSnapshots) {
```
redmondbowang: Don't really understand how this works.. The next time an editor instance is created, will it still use the same Undo instance? If user compose a new email, would he be seeing the old undo snapshots?

packages/roosterjs-editor-core/lib/undo/Undo.ts
Lego6245: Wouldn't the this.hasNewContent solve that? As long as that's getting set to false after a snapshot is created, then this doesn't matter.

It's also only checking for PageUp and PageDown right now. Is that intentional?

packages/roosterjs-editor-core/lib/undo/Undo.ts
Lego6245: keyDown -> Key "bottoms out". Will repeat infinitely if the key is held down, per OS behaviour

 keyPress -> a combination of the key going down and then coming up again. It's a single discrete action. There might be many keyDown events, but only one keyPress.

packages/roosterjs-editor-core/lib/undo/Undo.ts
redmondbowang: I guess the other keys are >= KEY_PageUp and <= KEY_DOWN, :D

packages/roosterjs-editor-core/lib/undo/Undo.ts
JiuqingSong: fixed

packages/roosterjs-editor-core/lib/undo/UndoSnapshots.ts
JiuqingSong: I think it is optional. Just added.

packages/roosterjs-editor-core/lib/undo/Undo.ts
JiuqingSong: Lego is right. In that case hasNewContent will be false.

And Bo is right too, these 8 keys are in the range of 33-40.

packages/roosterjs-editor-core/lib/undo/Undo.ts
JiuqingSong: There purpose here is to differentiate Input key and Function key. Such as Up/Down/Left/Right/..., they will not trigger keyPress but will trigger keyDown only. Here for KeyPress I will only handle a real input.

packages/roosterjs-editor-core/lib/undo/Undo.ts
JiuqingSong: Removed the check for 'Undo'. And remove the code to trigger ContentChanged event since setContent is already triggering one.

packages/roosterjs-editor-core/lib/undo/Undo.ts
JiuqingSong: This is on purpose. It is for reusing undo snapshots. For the same session with different editor instances, we share the undo snapshots. For different session, we should create new Undo object.

PR #5, opened 2017-09-29
support displayText in createLink

JiuqingSong:
Support display text in createLink


packages/roosterjs-editor-api/lib/format/createLink.ts
```diff
@@ -54,12 +66,9 @@ export default function createLink(editor: Editor, link: string, altText?: strin
         let originalUrl = linkData ? linkData.originalUrl : normalizedUrl;
         // On Edge/IE/Firefox (except Webkit based browsers like Chrome/Safari), when there is no selection,
```
MSJiaGuo: Please remove these unused comments

JiuqingSong: Fixed.

PR #7, opened 2017-10-06
New paste manager

JiuqingSong:
This change will simplify the paste manager code.

1. Get copied image and plain text from clipboard
2. If there is no plain text or empty, and there is image, paste the image (call onPasteComplete)
3. Otherwise, use hidden div to paste the html content
4. Process image and normalize content, then paste to body.

With this change, if pasting image, we don't need to use hidden div now.

Other changes:
1. replace `<BR>` with `<DIV></DIV> if <BR>` is the only html tag. This will fix the bug when paste text from notepad and toggle bullet in Edge.
2. make insertOptions optional since in most case we are using the same option there.
3. remove pasteHandler*.ts since we have a unified process now.


packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -234,7 +234,7 @@ export default class Editor {
     }
 
     // Insert content into editor
-    public insertContent(content: string, option: InsertOption): void {
+    public insertContent(content: string, option?: InsertOption): void {
```
redmondbowang: Why do you not set a default 'option' in this function? Could option be null?

JiuqingSong: I can, but then the code will be duplicated in insertNode(). So I just set it once there.

redmondbowang: Oh I didn't see it calls insertNode() eventually.

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
```diff
@@ -94,25 +102,70 @@ export default class PasteManager implements EditorPlugin {
...
     } else {
-        // let browser handle paste itself if it's not a known browser
-        return null;
+        pasteContainer.style.display = '';
     }
+    pasteContainer.focus();
+
+    window.requestAnimationFrame(() => {
+        // restore original selection range in editor
+        editor.updateSelection(originalSelectionRange);
```
redmondbowang: Why do you updateSelection before execute the callback? Should it be updated afterwards?

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
```diff
@@ -94,25 +102,70 @@ export default class PasteManager implements EditorPlugin {
...
+    // cache original selection range in editor
+    let originalSelectionRange = editor.getSelectionRange();
+
+    if (!pasteContainer) {
+        pasteContainer = fromHtml(CONTAINER_HTML)[0] as HTMLElement;
+        document.body.appendChild(pasteContainer);
     } else {
-        // let browser handle paste itself if it's not a known browser
-        return null;
+        pasteContainer.style.display = '';
```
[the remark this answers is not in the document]
JiuqingSong: It is null when first time call this function. Then it will be reused.
Setting display to '' makes the div to be shown, same effect with 'block' since 'block' is the default value for a DIV.

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
JiuqingSong: This need to be done before we call callback() to insert target nodes. Otherwise it will not be inserted to right position.

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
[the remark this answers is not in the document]
redmondbowang: I see. But I suppose this container is just for storing the paste content, we don't want to display that, why not always set it to hidden?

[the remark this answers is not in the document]
JiuqingSong: it is required to show the container so that we can put in focus and accept the pasted content. A hidden element can't be focused.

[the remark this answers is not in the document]
redmondbowang: Got it.

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
```diff
@@ -94,25 +102,70 @@ export default class PasteManager implements EditorPlugin {
...
+function getImage(dataTransfer: DataTransfer): File {
+    // Chrome, Firefox, Edge support dataTransfer.items
+    let fileCount = dataTransfer.items ? dataTransfer.items.length : 0;
+    for (let i = 0; i < fileCount; i++) {
+        let item = dataTransfer.items[i];
+        if (item.type && item.type.indexOf('image/') == 0) {
+            return item.getAsFile();
+        }
+    }
+    // IE, Safari suuport dataTransfer.files
```
MSJiaGuo: typo: suuport => support

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
```diff
@@ -94,25 +102,70 @@ export default class PasteManager implements EditorPlugin {
...
+    for (let i = 0; i < fileCount; i++) {
+        let file = dataTransfer.files.item(i);
+        if (file.type && file.type.indexOf('image/') == 0) {
+            return file;
+        }
+    }
+    return null;
+}
+
+function getHtml(editor: Editor, callback: (container: HTMLElement) => void) {
```
MSJiaGuo: The function name is a little confused. Usually a function starts with "get" will return something, but this method has no return value. Suggest to change the function name.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -234,7 +234,7 @@ export default class Editor {
...
         if (content) {
             let allNodes = fromHtml(content);
             // If it is to insert on new line, and there are more than one node in the collection, wrap all nodes with
```
MSJiaGuo: In line 243
if (option.insertOnNewLine && allNodes.length > 0) {

If the option is null, it will throw exception.

JiuqingSong: good catch!

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
JiuqingSong: fixed.

JiuqingSong: changed to "retrieveHtml"

PR #9, opened 2017-10-12
Watermark plugin

JiuqingSong:
Implement a watermark plugin

Other changes:
1. Fix bug "Editor.updateSelection()" always returns false
2. export two utils for undo snapshot: buildSnapshot, restoreSnapshot
3. optimize format

### DIFF
--- a/packages/roosterjs-plugin-watermark/lib/Watermark.ts
+++ b/packages/roosterjs-plugin-watermark/lib/Watermark.ts
@@ -23,7 +23,7 @@ class Watermark implements EditorPlugin {
     private visibleElementSelectors: string[] = ['table', 'img', 'li'];
 
     /**
-     * Create an instace of Watermark plugin
+     * Create an instance of Watermark plugin
      * @param watermark The watermark string
      * @param visibleElementSelectors Optional. Additional selectors for visible element
      */
--- a/packages/roosterjs-plugin-watermark/lib/Watermark.ts
+++ b/packages/roosterjs-plugin-watermark/lib/Watermark.ts
@@ -60,7 +60,8 @@ class Watermark implements EditorPlugin {
             event.eventType == PluginEventType.ContentChanged
         ) {
             this.showHideWatermark();
-        } else if (event.eventType == PluginEventType.ExtractContent) {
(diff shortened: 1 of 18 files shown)


packages/roosterjs-plugin-watermark/lib/Watermark.ts
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * A watermark plugin to manage watermark string for roosterjs
+ */
+class Watermark implements EditorPlugin {
+    private editor: Editor;
+    private isWatermarkShowing: boolean;
+    private visibleElementSelectors: string[] = ['table', 'img', 'li'];
+
+    /**
+     * Create an instace of Watermark plugin
```
redmondbowang: "instance"

packages/roosterjs-plugin-watermark/lib/Watermark.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    onPluginEvent(event: PluginEvent) {
+        if (
+            event.eventType == PluginEventType.Focus ||
+            event.eventType == PluginEventType.Blur ||
+            event.eventType == PluginEventType.ContentChanged
+        ) {
+            this.showHideWatermark();
+        } else if (event.eventType == PluginEventType.ExtractContent) {
+            this.removeWartermarkFromHtml(event as ExtractContentEvent);
```
redmondbowang: Should we check isShowingWatermark and then remove from html?

packages/roosterjs-plugin-watermark/lib/Watermark.ts
```diff
@@ -0,0 +1,120 @@
...
+            event.eventType == PluginEventType.ContentChanged
+        ) {
+            this.showHideWatermark();
+        } else if (event.eventType == PluginEventType.ExtractContent &&
+            this.isWatermarkShowing) {
+            this.removeWartermarkFromHtml(event as ExtractContentEvent);
+        }
+    }
+
+    private showHideWatermark() {
```
MSJiaGuo: Suggest to change name to toggleWatermark()

PR #19, opened 2017-11-21
Fix cursor position when paste

JiuqingSong:


packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
```diff
@@ -9,6 +9,7 @@ import {
 import { Editor, EditorPlugin } from 'roosterjs-editor-core';
 import { processImages } from './PasteUtility';
 import { fromHtml, unwrap } from 'roosterjs-editor-dom';
+import { browserData } from 'roosterjs-editor-core';
```
MSJiaGuo: You could merge the import with the first one since they're all imported from 'rootster-editor-core'

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
```diff
@@ -76,8 +77,16 @@ export default class PasteManager implements EditorPlugin {
                 let span = this.editor.getDocument().createElement('span');
                 span.innerHTML = normalizeContent(clipboardData.htmlData);
                 convertPastedContentFromWord(span);
-                this.editor.insertNode(span);
-                unwrap(span);
+
+                if (browserData.isChrome) {
+                    // In Chrome, if insert the span node first then do unwrap, the cursor position will be wrong
+                    while (span.firstChild) {
```
redmondbowang: Why do you use 'while', wouldn't it keep inserting the node?

JiuqingSong: insertNode will cause the node removed from original parent and append into new parent. So it will not get into infinite loop here.

PR #21, opened 2017-11-27
Fix cursor position issue after undo

JiuqingSong:
We should add undo snapshot add the beginning when delete something, even if selection range is collapsed. And we also need to set hasNewContent to true when delete, so that after delete if undo, we will take one more snapshot then undo to the previous one before deleting.

### DIFF
--- a/packages/roosterjs-editor-core/lib/undo/Undo.ts
+++ b/packages/roosterjs-editor-core/lib/undo/Undo.ts
@@ -151,7 +151,7 @@ export default class Undo implements UndoService {
 
             // Add snapshot when
             // 1. Something has been selected (not collapsed), or
-            // 2. This is not the same operation with previous one, or
+            // 2. It has a different key code from the last keyDown event (to prevent adding too many snapshot when keeping press the same key), or
             // 3. Ctrl/Meta key is pressed so that a whole word will be deleted
             if (
                 selectionRange &&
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-core/lib/undo/Undo.ts
```diff
@@ -147,24 +146,31 @@ export default class Undo implements UndoService {
         // Handle backspace/delete when there is a selection to take a snapshot
         // since we want the state prior to deletion restorable
         let evt = pluginEvent.rawEvent as KeyboardEvent;
-        let shouldTakeUndo = false;
         if (evt.which == KEY_BACKSPACE || evt.which == KEY_DELETE) {
             let selectionRange = this.editor.getSelectionRange();
+
+            // Add snapshot when
+            // 1. Something has been selected (not collapsed), or
+            // 2. This is not the same operation with previous one, or
```
redmondbowang: What does 'This is not the same operation with previous one' mean? Also why do we leave out image?

JiuqingSong: I updated the comment. This is to avoid adding to many snapshot when user keep pressing delete/backspace, we only add a snapshot before the first press.

For image, it can already be covered by "this.lastKeyPress != evt.which", so we don't need to check it separately.

Issue #23, opened 2017-12-04
Undo should allow empty snapshot


JiuqingSong:
There is a check in undo code:
if (snapshot) ...

This check will fail when snapshot is an empty string. We should allow empty string as snapshot.

PR #24, opened 2017-12-04
RoosterJs 6.4.0: Add APIs, some other fixes

JiuqingSong:
1. Add API: insertTable
2. Add API: insertImage, change pasteManager to use this API
3. Add an editor option param "doNotRestoreSelectionOnFocus", when set it to true, editor will not restore selection when focus
4. Fix a code bug in undo: we should allow empty undo snapshot ( #23 )
5. Fix insertNode API to allow insert a HTML fragment
6. Fix pasteManager to use HtmlDocumentFragment to have a better performance
7. Allow passing in extra Css Style node to convertInlineCss API
8. Add buttons for insertTable and insertImage in sample page

### DIFF
--- a/packages/roosterjs-editor-api/lib/test/format/clearFormatTest.ts
+++ b/packages/roosterjs-editor-api/lib/test/format/clearFormatTest.ts
@@ -37,6 +37,6 @@ describe('clearFormat()', () => {
         clearFormat(editor);
 
         // Assert
-        expect(editor.getContent()).toBe('<div id="text" style="">text</div>');
+        expect(editor.getContent()).toBe('<div id="text" style=""><span style="font-family: &quot;times new roman&quot;; font-size: 16px; color: black; background-color: rgba(0, 0, 0, 0);">text</span></div>');
     });
 });
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "roosterjs",
-  "version": "6.3.6",
+  "version": "6.4.0",
   "description": "Framework-independent javascript editor",
   "repository": {
     "type": "git",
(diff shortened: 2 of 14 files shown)


packages/roosterjs-editor-api/lib/test/format/clearFormatTest.ts
```diff
@@ -37,6 +37,6 @@ describe('clearFormat()', () => {
         clearFormat(editor);
 
         // Assert
-        expect(editor.getContent()).toBe('<div id="text" style="">text</div>');
+        expect(editor.getContent()).toBe('<div id="text" style=""><span style="font-family: &quot;times new roman&quot;; font-size: 16px; color: black; background-color: rgba(0, 0, 0, 0);">text</span></div>');
```
Lego6245: Based on this test, shouldn't all the formatting be removed? It looks like it partially shifted to the interior span element, but those styles don't match.

JiuqingSong: Ideally, yes, we should remove those styles.
But it is difficult to make it work correctly, especially when there are many nested elements.
So the dev code is to apply default format to the content. This results in a lot of styles are set into the elements.

I was thinking about having a parameter to clearFormat() to allow switching apply default format. But since in some browsers (IE, Edge), it will not work correctly without this action, it doesn't make sense to me to skip this action and leave a not well-cleared html content. So I decide to always clear these formats by applying default format, and allow it has these styles on.

Lego6245: That makes sense. Is there a reason for the addition of a span instead of applying these styles to the div? I suppose to handle the case of a partial selection being cleared, but should that be tested?

Just a thought.

PR #25, opened 2017-12-05
Bowang/apply style in table

redmondbowang:
Fixed issue: Table pasted from OneNote looks distorted after applying background color


packages/roosterjs-editor-dom/lib/inlineElements/NodeInlineElement.ts
```diff
@@ -164,7 +164,13 @@ class NodeInlineElement implements InlineElement {
             }
 
             // Apply the style
-            if (startNode.nodeType == NodeType.Text && startNode.nodeValue) {
+            // If a node has only white space and new line and is in table, we ignore it,
+            // otherwise the table will be distorted
+            if (
+                startNode.nodeType == NodeType.Text &&
+                startNode.nodeValue &&
+                !(startNode.nodeValue.trim() == '' && this.isNodeInTable(startNode))
```
JiuqingSong: 1. When find parent element with tag, we normally pass in containerNode as end node, otherwise it always need to find to very top.
2. Here I think you can just check 1 layer of parent see if it is TR. Text element under TD should still be styled. So  that you don't need to have isNodeInTable()

redmondbowang: Changed.

PR #27, opened 2017-12-07
New API toggleHeader

JiuqingSong:
Support toggleHeader() API.
Refactor some existing code (execFormatWithUndo, toggleBlockQuote) for better code reusing.

### DIFF
--- a/packages/roosterjs-editor-api/lib/format/cacheGetHeaderLevel.ts
+++ b/packages/roosterjs-editor-api/lib/format/cacheGetHeaderLevel.ts
@@ -1,0 +1,17 @@
+import queryNodesWithSelection from '../cursor/queryNodesWithSelection';
+import { Editor, cacheGetEventData } from 'roosterjs-editor-core';
+import { ListState, PluginEvent } from 'roosterjs-editor-types';
+import { getTagOfNode } from 'roosterjs-editor-dom';
+
+const EVENTDATACACHE_HEADER_LEVEL = 'HeaderLevel';
+
+export default function cacheGetHeaderLevel(editor: Editor, event?: PluginEvent): number {
+    return cacheGetEventData<number>(event, EVENTDATACACHE_HEADER_LEVEL, () => {
+        for (let i = 1; i <= 6; i++) {
+            if (queryNodesWithSelection(editor, 'H' + i).length > 0) {
+                return i;
+            }
+        }
+        return 0;
+    });
+}
--- a/packages/roosterjs-editor-api/lib/format/cacheGetHeaderLevel.ts
+++ b/packages/roosterjs-editor-api/lib/format/cacheGetHeaderLevel.ts
@@ -1,7 +1,6 @@
 import queryNodesWithSelection from '../cursor/queryNodesWithSelection';
 import { Editor, cacheGetEventData } from 'roosterjs-editor-core';
-import { ListState, PluginEvent } from 'roosterjs-editor-types';
-import { getTagOfNode } from 'roosterjs-editor-dom';
+import { PluginEvent } from 'roosterjs-editor-types';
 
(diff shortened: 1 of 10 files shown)


packages/roosterjs-editor-api/lib/format/toggleHeader.ts
```diff
@@ -0,0 +1,40 @@
...
+            let inlineElement = traverser ? traverser.currentInlineElement : null;
+            while (inlineElement) {
+                let node = inlineElement.getContainerNode();
+                if (node.nodeType == NodeType.Text) {
+                    node = node.parentNode
+                }
+                if (node.nodeType == NodeType.Element) {
+                    (<HTMLElement>node).style.fontSize = '';
+                }
+                inlineElement = traverser.getNextInlineElement();
```
redmondbowang: Should we getNextInlineElement() before adding the style? I saw we did that way in applyInlineStyle to avoid messing up with DOM.

JiuqingSong: In some cases, yes, we need to do that. But for this one, I just changed a css style and DOM structure is not changed. So I think either way will work.

packages/roosterjs-editor-api/lib/format/cacheGetHeaderLevel.ts
```diff
@@ -0,0 +1,16 @@
+import queryNodesWithSelection from '../cursor/queryNodesWithSelection';
+import { Editor, cacheGetEventData } from 'roosterjs-editor-core';
+import { PluginEvent } from 'roosterjs-editor-types';
+
+const EVENTDATACACHE_HEADER_LEVEL = 'HeaderLevel';
+
+export default function cacheGetHeaderLevel(editor: Editor, event?: PluginEvent): number {
+    return cacheGetEventData<number>(event, EVENTDATACACHE_HEADER_LEVEL, () => {
+        for (let i = 1; i <= 6; i++) {
```
redmondbowang: Why check 6 levels? Any special reasons?

JiuqingSong: It matches to HTML tag <h1> .. <h6>

packages/roosterjs-editor-api/lib/format/execFormatWithUndo.ts
```diff
@@ -1,13 +1,69 @@
...
+
+    editor.updateSelection(range);
+}
+
+function validateEditorPoint(editor: Editor, point: EditorPoint): boolean {
+    if (point.containerNode && editor.contains(point.containerNode)) {
+        if (point.containerNode.nodeType == NodeType.Text) {
+            point.offset = Math.min(point.offset, (<Text>point.containerNode).data.length);
+        } else if (point.containerNode.nodeType == NodeType.Element) {
+            point.offset = Math.min(point.offset, (<HTMLElement>point.containerNode).childNodes.length);
```
redmondbowang: If it is non textual node, shouldn't just be 0 or 1? Or does it actually go by the childNodes number?

packages/roosterjs-editor-api/lib/format/toggleHeader.ts
redmondbowang: I see.

packages/roosterjs-editor-api/lib/format/execFormatWithUndo.ts
JiuqingSong: From my testing, the value should be from 0 to the length of child node. So today the definition of NodeBoundary migth be wrong.

redmondbowang: I see.

PR #29, opened 2018-01-05
Show html and js sample code when initiate sample site

redmondbowang:


sample/scripts/initSampleCode.ts
```diff
@@ -0,0 +1,44 @@
+export default function initSampleCode() {
+    insertHtmlSampleCode();
+    insertJsSampleCode();
+
+    var fileref=document.createElement('script');
+    fileref.setAttribute("type","text/javascript");
+    fileref.setAttribute("src", "https://cdn.rawgit.com/google/code-prettify/master/loader/run_prettify.js");
```
JiuqingSong: This file is under Apache license. Can we use it here?

sample/scripts/initSampleCode.ts
```diff
@@ -0,0 +1,44 @@
...
+    document.getElementsByTagName("head")[0].appendChild(fileref);
+}
+
+function insertHtmlSampleCode() {
+    let htmlSampleCode = document.getElementsByClassName('htmlSample')[0];
```
JiuqingSong: Suggest getElementById()

sample/scripts/initSampleCode.ts
```diff
@@ -0,0 +1,44 @@
...
+        '</html>',
+    ].join("\n");
+    htmlSampleCode.appendChild(createElementFromContent('htmlSampleCode', codeString, 'prettyprint lang-html'));
+}
+
+function insertJsSampleCode() {
+    let jsSampleCode = document.getElementsByClassName('jsSample')[0];
+    let codeString: string = [
+        'var roosterjs = require(\'roosterjs\');',
+        'var editorDiv = document.getElementById(\'editorDiv\');',
```
JiuqingSong: Since we have limited possibilities of code, I think we can just hard-code the styles and no need to format them in runtime.

sample/scripts/sample.ts
```diff
@@ -10,6 +11,7 @@ window.onload = () => {
     initEditor();
     initFormatBar();
     initOptions();
+    initSampleCode();
     switchTab('quickstart');
 };
```
JiuqingSong: So this change only add one sample code. I'd like to see more functionalities to be completed then check in. If the sample code showing in the UI doesn't reflect any change of user choice, it may mislead users.

sample/scripts/initSampleCode.ts
redmondbowang: Looked up online and the Apache license documentation. It should be fine to use Apache licensed work and redistribute in any license as long as we retain any original Apache License notices in source files of the original project.

sample/scripts/initSampleCode.ts
redmondbowang: Changed.

sample/scripts/sample.ts
redmondbowang: Added logic to reflect user choice for plugins and format.

sample/scripts/initSampleCode.ts
redmondbowang: I found it harder than thought to hard code the styles. Basically we need to wrap each string segment with a span to set color. This makes the code hard to read. Currently I will just remove the dependency on code-prettify.

sample/scripts/initOptions.ts
```diff
@@ -12,12 +12,24 @@ import { ImageResizePlugin } from 'roosterjs-plugin-image-resize';
...
+    document.getElementById('pasteManagerCheckbox').addEventListener('change', initEditorForOptions);
+    document.getElementById('contentEditCheckbox').addEventListener('change', initEditorForOptions);
+    document.getElementById('watermarkCheckbox').addEventListener('change', initEditorForOptions);
+    document.getElementById('imageResizeCheckbox').addEventListener('change', initEditorForOptions);
+
+    document.getElementById('boldCheckbox').addEventListener('change', initEditorForOptions);
+    document.getElementById('italicCheckbox').addEventListener('change', initEditorForOptions);
+    document.getElementById('underlineCheckbox').addEventListener('change', initEditorForOptions);
+    document.getElementById('textColorDefaultFormat').addEventListener('change', initEditorForOptions);
+    document.getElementById('fontNameDefaultFormat').addEventListener('change', initEditorForOptions);
```
JiuqingSong: ['defaultShortcutCheckbox', '...'].forEach(id => document.getElementById(id).addEventListener(...);

sample/scripts/updateSampleCode.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    return node;
+}
+
+function assemblePluginsString(plugins: EditorPlugin[]): string {
+    if (plugins) {
+        let pluginsString = 'var plugins = [\n';
+        plugins.forEach(plugin => {
+            if (plugin instanceof Watermark) {
+                pluginsString = pluginsString.concat('  new roosterPlugins.Watermark(),\n');
```
JiuqingSong: Watermark plugin requires a parameter

sample/scripts/updateSampleCode.ts
```diff
@@ -0,0 +1,165 @@
...
+    if (defaultFormat.bold) {
+        defaultFormatString = defaultFormatString.concat('defaultFormat.bold = true;\n');
+    }
+    if (defaultFormat.italic) {
+        defaultFormatString = defaultFormatString.concat('defaultFormat.italic = true;\n');
+    }
+    if (defaultFormat.underline) {
+        defaultFormatString = defaultFormatString.concat('defaultFormat.underline = true;\n');
+    }
+    defaultFormatString = defaultFormatString.concat(`defaultFormat.textColor = ${defaultFormat.textColor};\n`);
```
JiuqingSong: I'd expect the defaultFormat code finally looks like:

var defaultFormat = {
  bold: true,
 ...
};

JiuqingSong: You need '' around textColor and fontFamily

sample/scripts/initOptions.ts
redmondbowang: Changed.

sample/scripts/updateSampleCode.ts
redmondbowang: Changed

sample/scripts/updateSampleCode.ts
redmondbowang: Changed

redmondbowang: Added

PR #31, opened 2018-01-08
Support direct pasting

JiuqingSong:
PasteManager class is using a temp DIV to accept pasted content, then retrieve it and finally add into editor. There are some issues with this solution:
1. It sometimes hangs Edge
2. Can't support HTML with global styles (e.g. Copy from Excel) in Chrome
3. Edge will remove some styles when copy from Word then some format (bullet, numbering) will lose.
4. Perf of pasting large content is bad

In this change I use getAsString() function to get HTML content directly from clipboard (and fallback to temp DIV when failed). Then use some regular expressions to remove unsafe tags.

This solution is not enabled by default. To enabled it, pass "true" to the 2nd parameter of PasteManager class.


packages/roosterjs-editor-api/lib/paste/buildClipBoardData.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+function getImage(dataTransfer: DataTransfer): File {
+    // Chrome, Firefox, Edge support dataTransfer.items
+    let fileCount = dataTransfer.items ? dataTransfer.items.length : 0;
+    for (let i = 0; i < fileCount; i++) {
+        let item = dataTransfer.items[i];
+        if (item.type && item.type.indexOf('image/') == 0) {
+            return item.getAsFile();
```
redmondbowang: What if there are multiple images? Or do they come together as one file?

JiuqingSong: It won't.

For example, if copy from OneNote when select two images, OneNote will generate a screenshot which contains both images and set it into clipboard.

In some other cases, say copy from browser with 2 images, there will be no image set into clipboard, but only some html text.

redmondbowang: Got it.

packages/roosterjs-editor-api/lib/paste/buildClipBoardData.ts
```diff
@@ -0,0 +1,149 @@
...
+        if (file.type && file.type.indexOf('image/') == 0) {
+            return file;
+        }
+    }
+    return null;
+}
+
+function directRetrieveHtml(
+    event: ClipboardEvent,
+    editor: Editor,
```
redmondbowang: editor is not used here?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -468,6 +475,28 @@ export default class Editor {
...
+                this.getDocument()
+            )[0] as HTMLElement;
+            this.insertNode(this.pasteDiv, {
+                position: ContentPosition.Outside,
+                updateCursor: false,
+                replaceSelection: false,
+                insertOnNewLine: false,
+            });
+        } else {
+            this.pasteDiv.style.display = '';
```
redmondbowang: Is it necessary to set display to '' ?

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
```diff
@@ -1,50 +1,64 @@
...
+     * When set to true, we retrieve HTML from clipboard directly rather than using a hidden pasting DIV,
+     * then filter out unsafe HTML tags and attributes. Although we removed some unsafe tags such as SCRIPT,
+     * OBJECT, ... But there is still risk to have other kinds of XSS scripts embeded. So please do NOT use
+     * this parameter if you don't have other XSS detecting logic outside the edtior.
+     * @param defaultPasteOption: Default option of pasting. Default value is PasteAsIs
      */
-    constructor(private readonly pasteHandler?: (clipboardData: ClipBoardData) => void) {}
+    constructor(
+        private readonly imageHandler?: (clipboardData: ClipBoardData) => void,
+        private useDirectPaste?: boolean,
```
redmondbowang: Shall we move 'useDirectPaste' first since we have defaultImageHandler, that way user doesn't need to explicitly pass 'null' for it.

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
```diff
@@ -59,129 +73,77 @@ export default class PasteManager implements EditorPlugin {
...
-                let nodes = fromHtml(clipboardData.htmlData, document);
-                for (let node of nodes) {
-                    fragment.appendChild(node);
-                }
-
-                convertPastedContentFromWord(fragment);
-                this.editor.insertNode(fragment);
-                this.onPasteComplete(clipboardData);
-            });
+            restoreSnapshot(this.editor, clipboardData.snapshotBeforePaste);
```
redmondbowang: Not quite understand this: why build snapshot inside clipboardData and in what kind of cases we restore to snapshot before paste?

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
```diff
@@ -59,129 +73,77 @@ export default class PasteManager implements EditorPlugin {
...
-        if (!this.pasteContainer || !this.pasteContainer.parentNode) {
-            this.pasteContainer = fromHtml(CONTAINER_HTML, editor.getDocument())[0] as HTMLElement;
-            editor.insertNode(this.pasteContainer, {
-                position: ContentPosition.Outside,
-                updateCursor: false,
-                replaceSelection: false,
-                insertOnNewLine: false,
-            });
+    private pasteHTML(clipboardData: ClipBoardData, fragmentOverride?: DocumentFragment) {
+        if (clipboardData.textData || !clipboardData.imageData.file) {
```
redmondbowang: Why check textData instead of htmlData?

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
```diff
@@ -59,129 +73,77 @@ export default class PasteManager implements EditorPlugin {
...
-        content = '<div>' + content.replace(/<br>/gi, '</div><div>') + '<br></div>';
+    private pasteWithCurrentFormat(clipboardData: ClipBoardData) {
+        throw new Error('Not implemented');
     }
 
-    return content;
+    private defaultImageHandler = (clipboardData: ClipBoardData) => {
+        let file = clipboardData.imageData ? clipboardData.imageData.file : null;
+        if (file) {
+            insertImage(this.editor, file);
```
redmondbowang: In processImages I saw images are categorized into clipboardData.imageData.noSrcImageIds, clipboardData.imageData.localFileImages and clipboardData.imageData.base64Images, are they treated the same way here?

redmondbowang:
Copy multiple images or image with text from Word seems not working.

packages/roosterjs-editor-api/lib/paste/buildClipBoardData.ts
JiuqingSong: good catch.

packages/roosterjs-editor-core/lib/editor/Editor.ts
JiuqingSong: Yes. This is necessary.
Each time after use, I set display to none. So I need to set it back to '' to use again.

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
JiuqingSong: This is to be compatible with old code. If I change original paramater's meaning, existing code will be broken.

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
JiuqingSong: This is for a new requirement. After pasted, user may want to change the way to paste, say we already pasted in HTML, but then user want to paste in plain text. So that we need to restore original content first, then paste again in plain text. Otherwise the content will be pasted twice.

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
JiuqingSong: Copy an image from OneNote, there HTML data will look like: <div></div>. In that case htmlData is not empty but textData is empty.

So if we only check htmlData, we will always paste in HTML instead of image.

packages/roosterjs-editor-plugins/lib/PasteManager/PasteManager.ts
JiuqingSong: That is to give imageHandler a chance to deal with those images (e.g., remove noSrcImage and localFileImages).

JiuqingSong:
It is by design when copy multiple image or image+text that only text is pasted because there's no base64 image data inside clipboard.

PR #38, opened 2018-01-20
Add comments for roosterjs-editor-api (part3)

redmondbowang:


packages/roosterjs-editor-api/lib/format/toggleSubscript.ts
```diff
@@ -2,7 +2,15 @@ import execFormatWithUndo from './execFormatWithUndo';
...
 
-export default function toggleSubscript(editor: Editor): void {
+/**
+ * Toggle subscript at selection
+ * If selection is collapsed, it will only affect the input after caret
+ * If selection contains only subscript text, the subscript style will be removed
+ * If selection contains only normal text, subscript style will be added to the whole selected text
+ * If selection contains both subscript and normal text, the subscript style will be removed from whole selected text
+ * @param editor The editor instance
+ */
```
JiuqingSong: What if the selection contains superscript?
Same question for subscript function

redmondbowang: Added description for each case

packages/roosterjs-editor-api/lib/format/toggleSubscript.ts
```diff
@@ -2,7 +2,16 @@ import execFormatWithUndo from './execFormatWithUndo';
...
 
-export default function toggleSubscript(editor: Editor): void {
+/**
+ * Toggle subscript at selection
+ * If selection is collapsed, it will only affect the input after caret
+ * If selection contains only subscript text, the subscript style will be removed
+ * If selection contains only normal text, subscript style will be added to the whole selected text
+ * If selection contains both subscript and normal text, the subscript style will be removed from whole selected text
+ * If selection contains any superscript text, the superscript style will be removed and then subscript style will be applied
+ * @param editor The editor instance
```
JiuqingSong: This is not true for some browsers, e.g. IE.

redmondbowang: Changed.

PR #43, opened 2018-04-18
Added option to Paste plugin to allow custom insert image handler

adriantran:
Added option to Paste plugin to allow custom insert image handler. Also improved webpack.config.js to start the sample page.


JiuqingSong:
Actually you can do customized handling when pasting by handling BeforePaste event, and check if the PasteOption is Image, then do any modification to the passed-in DOM fragment and paste option.

Please reference to https://github.com/Microsoft/roosterjs/wiki/BeforePasteEvent

PR #54, opened 2018-05-10
Add support for modifying title text when calling createLink

### LINKED ISSUE #52: When updating a pre-existing link, display text is not updated too.
We should probably do something similar here that we do for title/alt text.

https://github.com/Microsoft/roosterjs/blob/d9dba73e0ec753dfddcdf4e56bda499e48d623ac/packages/roosterjs-editor-api/lib/format/createLink.ts#L90-L95

Lego6245:
Resolves #52 

Updated createLink as well as the sample code to showcase the new behavior.


packages/roosterjs-editor-api/lib/format/createLink.ts
```diff
@@ -86,6 +90,9 @@ export default function createLink(
                 /* the selection is not collapsed, use browser execCommand */
                 editor.getDocument().execCommand('createLink', false, normalizedUrl);
                 anchor = getAnchorNodeAtCursor(editor);
+                if (displayText && anchor.textContent != displayText) {
```
JiuqingSong: Wrap to a function?

Lego6245: We do duplicate this... I'd be fine with that.

PR #55, opened 2018-05-13
Fix createLink issue

JiuqingSong:
The last change in createLink() causes an issue that the display text is always changed even if it is not specified. In this change I fix this issue and only change display text if it is specified.


Lego6245:
What scenario is being fixed here?

Is it CreateLink -> Call CreateLink again on the range -> Delete Display Text -> Get link with the URL as the display text?

The difference is now it wouldn't change. Is that what we want?

JiuqingSong:
1.  type "abc"
  2.  select "abc"
  3.  insert link with "http://www.bing.com"

Expect: <a href=http://www.bing.com>abc</a>
Actual: <a href=http://www.bing.com>www.bing.com</a>

Generally, I expect the display text doesn't change if displayText param is not passed, however, with the previous fix, it is always changed to the url because "displayText || originalUrl" is passed into the function, and here "originalUrl" is never null.

After this fix, if we still want to change the display text, we need to explicitly pass displayText param, otherwise the display text leaves no change.


Lego6245:
I see. I had assumed we'd always scan the link and provide the current display text for the link, but I think this makes sense for this specific case.

PR #63, opened 2018-05-24
Add Picker Plugin to RoosterJS

Lego6245:
This commit adds another plugin to the RoosterJS project: A Picker Plugin.

The Picker Plugin is designed to support scenarios where you'd want to type a trigger character, for instance, an @, and then mention someone. You can see this on Github, where typing an @ opens a drop down menu with a list of people that you can mention. It also works for the # symbol as well.

The Picker Plugin is generalized to accept configuration through it's PickerDataProvider, which manages the data and UX portions of the picker experience. It can be further configured through the PickerPluginOptions interface to see trigger characters and change sources.

### DIFF
--- a/packages/roosterjs-plugin-picker/README.md
+++ b/packages/roosterjs-plugin-picker/README.md
@@ -2,4 +2,4 @@
 
 The Picker Plugin is designed to support scenarios where you'd want to type a trigger character, for instance, an @, and then mention someone. You can see this on Github, where typing an @ opens a drop down menu with a list of people that you can mention. It also works for the # symbol as well, as well as for any self defined prefix.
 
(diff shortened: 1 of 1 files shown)


JiuqingSong:
Generally, please put this plugin to a new package, like what we did for ImageResize. Because I plan to move TableResize and Watermark to new packages as well, and finally remove the roosterjs-editor-plugin package.

packages/roosterjs-plugin-picker/README.md
```diff
@@ -0,0 +1,5 @@
+# roosterjs-editor-plugins
```
JiuqingSong: Correct the package name

PR #64, opened 2018-05-31
Fixed HyperLink plugin's TEMP_TITLE_REGEX

adriantran:
Fixed HyperLink plugin's TEMP_TITLE_REGEX where after replace anchor tag has an extra space (i.e. &lt;a href="..." &gt;foo&lt;/a&gt;). This creates dirty content when getContent is called.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "roosterjs",
-  "version": "6.10.8",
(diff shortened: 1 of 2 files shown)


adriantran:
I've just updated the description. GitHub didn't escape the i.e. code.

PR #65, opened 2018-06-07
Users/ricassan/pass editor reference to picker

RichardCassan:
6.11.0 add a method for setting the Editor for picker plugin data providers

### DIFF
--- a/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
+++ b/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
@@ -50,9 +50,12 @@ export default class EditorPickerPlugin implements EditorPickerPluginInterface {
             },
             (isSuggesting: boolean) => {
                 this.setIsSuggesting(isSuggesting);
-            },
-            this.editor
+            }
         );
+
+        if (this.dataProvider.setEditor) {
+            this.dataProvider.setEditor(editor);
+        }
     }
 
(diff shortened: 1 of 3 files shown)


packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -52,6 +52,10 @@ export default class EditorPickerPlugin implements EditorPickerPluginInterface {
                 this.setIsSuggesting(isSuggesting);
             }
         );
+
+        if (this.dataProvider.setEditor) {
```
Lego6245: Thoughts about this being part of `onInitialize` instead of it's own separate function?

RichardCassan: Yeah, I was going to do that. Just means we have to do a major version bump because it would be a breaking change.

Lego6245: adding to the api as an optional (non required) parameter shouldn't be, that's an additive API change. Unless you're arguing to make it required?

RichardCassan: Yeah, good point. I was thinking to make it required (since the plugin will always be providing it), but it's probably cleaner to include it as an optional argument instead of a different method.

PR #67, opened 2018-06-11
Fix bug in resolveInlineElement

JiuqingSong:
We should always use the node we are checking to build inline element

### DIFF
--- a/packages/roosterjs-editor-dom/lib/inlineElements/resolveInlineElement.ts
+++ b/packages/roosterjs-editor-dom/lib/inlineElements/resolveInlineElement.ts
@@ -22,19 +22,19 @@ export default function resolveInlineElement(
         parent = parent.parentNode
     ) {
         nodeChain.push(parent);
-        parent = parent.parentNode;
     }
 
     let inlineElement: InlineElement;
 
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-dom/lib/inlineElements/resolveInlineElement.ts
```diff
@@ -22,19 +22,19 @@ export default function resolveInlineElement(
...
     }
 
     let inlineElement: InlineElement;
 
     for (let i = nodeChain.length - 1; i >= 0 && !inlineElement; i--) {
-        let tag = getTagOfNode(nodeChain[i]);
+        let currentNode = nodeChain[i];
+        let tag = getTagOfNode(currentNode);
         if (tag == 'A') {
-            inlineElement = new LinkInlineElement(node, parentBlock);
```
Lego6245: Did this just not work as intended before?

JiuqingSong: It was just introduced from my refactoring, so before 6.10.9 it worked well.

PR #88, opened 2018-07-13
A few improvements for ImageResize plugin

adriantran:
Added option to allow image to be bypass ImageResize plugin. Also fixed extra bottom spacing between the image and resize box.

### DIFF
--- a/packages/roosterjs-plugin-image-resize/lib/ImageResize.ts
+++ b/packages/roosterjs-plugin-image-resize/lib/ImageResize.ts
@@ -39,14 +39,15 @@ export default class ImageResize implements EditorPlugin {
      * @param minHeight Minimum height of image when resize in pixel, default value is 10
      * @param selectionBorderColor Color of resize border and handles, default value is #DB626C
      * @param forcePreserveRatio Whether always preserve width/height ratio when resize, default value is false
-     * @param bypassClassName Class name of IMG tag which to bypass resize logic (e.g. for a placeholder image)
+     * @param resizableImageSelector Selector for picking which image is resizable (e.g. for all images not placeholders), note
+     * that the tag must be IMG regardless what the selector is
      */
     constructor(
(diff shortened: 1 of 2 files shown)


packages/roosterjs-plugin-image-resize/lib/ImageResize.ts
```diff
@@ -39,12 +39,14 @@ export default class ImageResize implements EditorPlugin {
...
      * @param forcePreserveRatio Whether always preserve width/height ratio when resize, default value is false
+     * @param bypassClassName Class name of IMG tag which to bypass resize logic (e.g. for a placeholder image)
      */
     constructor(
         private minWidth: number = 10,
         private minHeight: number = 10,
         private selectionBorderColor: string = '#DB626C',
-        private forcePreserveRatio: boolean = false
+        private forcePreserveRatio: boolean = false,
+        private bypassClassName: string = undefined
```
JiuqingSong: Instead of having a bypass class name, I'd like to make it more common. Suggest to pass in a "resizableImageSelector" with default value "img". So that people can use any selector to determine their image to resize.

packages/roosterjs-plugin-image-resize/lib/ImageResize.ts
```diff
@@ -223,7 +229,7 @@ export default class ImageResize implements EditorPlugin {
         parent.insertBefore(document.createComment(END_TAG), resizeDiv.nextSibling);
 
         resizeDiv.style.position = 'relative';
-        resizeDiv.style.display = 'inline-table';
+        resizeDiv.style.display = 'inline-flex';
```
JiuqingSong: Please make sure this also works for IE11.

adriantran: yes, this is working on IE11 for me

PR #93, opened 2018-08-03
Added toggleCodeBlock

adriantran:


packages/roosterjs-editor-api/lib/format/toggleCodeBlock.ts
```diff
@@ -0,0 +1,36 @@
+import { Editor } from 'roosterjs-editor-core';
+import { getTagOfNode, unwrap, wrap } from 'roosterjs-editor-dom';
+import toggleTagCore from './toggleTagCore';
+
+const PRE_TAG = 'pre';
+const CODE_TAG = 'code';
+const CODE_NODE_TAG = 'CODE';
+const defaultStyler = (_: HTMLElement) => {};
```
JiuqingSong: Better move this into toggleTagCore() so we don't need always pass it.

JiuqingSong: Or check null in toggleTagCore() and dothing for null case

packages/roosterjs-editor-api/lib/format/toggleTagCore.ts
```diff
@@ -0,0 +1,88 @@
...
+ * the elements will be untagge and other elements will take no affect
+ * @param editor The editor instance
+ * @param tag The tag name
+ * @param wrapFunction The wrap function
+ * @param unwrapFunction The unwrap function
+ * @param styler The styler for setting the style for the blockquote element
+ */
+export default function toggleTagCore<T extends keyof HTMLElementTagNameMap>(
+    editor: Editor,
+    wrapper: T,
```
JiuqingSong: "tag" or "wrapper"? Need to be consistent with comment

packages/roosterjs-editor-api/lib/format/toggleTagCore.ts
```diff
@@ -0,0 +1,88 @@
...
+    wrapFunction: (nodes: Node[]) => HTMLElement,
+    unwrapFunction: (node: Node) => Node,
```
JiuqingSong: Can we make wrapFunction and unwrapFunction optional and default behavior is to wrap/unwrap with the given tag since we already know the tag name?

And move styler param before these two, so that toggleBlockQuote doesn't need to pass in these params.

adriantran: sure

packages/roosterjs-editor-api/lib/format/toggleCodeBlock.ts
adriantran: I'll move it, generally prefer null functions over if checks

packages/roosterjs-editor-api/lib/format/toggleBlockQuote.ts
```diff
@@ -1,19 +1,9 @@
...
-    Position,
-    unwrap,
-    wrap,
-    fromHtml,
-    getTagOfNode,
-    splitBalancedNodeRange,
-} from 'roosterjs-editor-dom';
-import { ChangeSource, PositionType, QueryScope } from 'roosterjs-editor-types';
 import { Editor } from 'roosterjs-editor-core';
+import { unwrap, wrap } from 'roosterjs-editor-dom';
```
JiuqingSong: Remove these imports, otherwise it can't build.

packages/roosterjs-editor-api/lib/format/toggleTagCore.ts
```diff
@@ -0,0 +1,90 @@
...
+const DEFAULT_STYLER = (_: HTMLElement) => {};
+
+/**
+ * Toggle a tag at selection, if selection already contains elements of such tag,
+ * the elements will be untagge and other elements will take no affect
+ * @param editor The editor instance
+ * @param tag The tag name
+ * @param wrapFunction The wrap function
+ * @param unwrapFunction The unwrap function
+ * @param styler The styler for setting the style for the blockquote element
```
JiuqingSong: update the order of parameters in comment

packages/roosterjs-editor-api/lib/format/toggleBlockQuote.ts
adriantran: oops, had webpack on

PR #100, opened 2018-09-06
Improve get cursor performance for mobile safari

sebastienvillar:
1. Add support for iOS/Android user agent
2. Improve getRect performance. Adding a span in the DOM should be used as a last resort since it's really not efficient to change the dom for a rect computation.

Tested on Mobile Safari on iOS 11.4 and Chrome for Android on versions 4 and 8

### DIFF
--- a/packages/roosterjs-editor-dom/lib/selection/Position.ts
+++ b/packages/roosterjs-editor-dom/lib/selection/Position.ts
@@ -131,46 +131,51 @@ export default class Position {
      * Get bounding rect of this position
      */
     getRect(): Rect {
-        let document = this.node && this.node.ownerDocument;
+        const document = this.node && this.node.ownerDocument;
 
         if (!document) {
             return null;
         }
 
-        let range = document.createRange();
+        const range = document.createRange();
         range.setStart(this.node, this.offset);
 
         // 1) try to get rect using range.getBoundingClientRect()
-        let rect = normalizeRect(range.getBoundingClientRect());
+        let rect = range.getBoundingClientRect ? normalizeRect(range.getBoundingClientRect()) : null;
+        if (rect) {
+            return rect;
+        }
 
-        if (!rect) {
-            let { node, element, offset } = this.normalize();
+        // 2) try to get rect using range.getClientRects
+        const { node, element, offset } = this.normalize();
+        const rects = range.getClientRects ? range.getClientRects() : null;
+        rect = rects && rects.length == 1 ? normalizeRect(rects[0]) : null;
+        if (rect) {
+            return rect;
+        }
 
-            // 2) if current cursor is inside text node, use range.getClientRects() for safari
-            // or insert a SPAN and get the rect of SPAN for others
-            if (Browser.isSafari) {
-                let rects = range.getClientRects();
-                if (rects && rects.length == 1) {
-                    rect = normalizeRect(rects[0]);
-                }
-            } else {
-                if (node.nodeType == NodeType.Text) {
-                    let span = document.createElement('SPAN');
-                    range.setStart(node, offset);
-                    range.collapse(true /*toStart*/);
-                    range.insertNode(span);
-                    rect = normalizeRect(span.getBoundingClientRect());
-                    span.parentNode.removeChild(span);
-                }
-            }
-
-            // 3) fallback to element.getBoundingClientRect()
-            if (!rect && element) {
-                rect = normalizeRect(element.getBoundingClientRect());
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-dom/lib/selection/Position.ts
```diff
@@ -146,9 +146,9 @@ export default class Position {
         if (!rect) {
             let { node, element, offset } = this.normalize();
 
-            // 2) if current cursor is inside text node, use range.getClientRects() for safari
+            // 2) if current cursor is inside text node, use range.getClientRects() for safari/chrome
             // or insert a SPAN and get the rect of SPAN for others
-            if (Browser.isSafari) {
+            if (Browser.isSafari || Browser.isChrome) {
```
sebastienvillar: I'm not sure why this is the way this was implemented. I'm not super comfortable with this implementation. Is there any reason we don't test for getClientRects to exists and check that we end up with a unique correct normalized rect when called and fallback on the span method?

I would then remove what I implemented and just do this check so not to rely on user agents.

sebastienvillar:
Updated the PR @JiuqingSong

packages/roosterjs-editor-dom/lib/selection/Position.ts
```diff
@@ -131,46 +131,51 @@ export default class Position {
...
-            } else {
-                if (node.nodeType == NodeType.Text) {
-                    let span = document.createElement('SPAN');
-                    range.setStart(node, offset);
-                    range.collapse(true /*toStart*/);
-                    range.insertNode(span);
-                    rect = normalizeRect(span.getBoundingClientRect());
-                    span.parentNode.removeChild(span);
-                }
+        let rect = range.getBoundingClientRect ? normalizeRect(range.getBoundingClientRect()) : null;
```
sebastienvillar: Added extra checks to make sure the functions exist.

packages/roosterjs-editor-dom/lib/selection/Position.ts
```diff
@@ -131,46 +131,51 @@ export default class Position {
...
+        if (rect) {
```
sebastienvillar: Flatten the implementation by returning when we have the rect

packages/roosterjs-editor-dom/lib/selection/Position.ts
```diff
@@ -131,46 +131,51 @@ export default class Position {
...
             }
+        }
 
-            // 3) fallback to element.getBoundingClientRect()
-            if (!rect && element) {
-                rect = normalizeRect(element.getBoundingClientRect());
+        // 4) try getBoundingClientRect on element
+        if (element && element.getBoundingClientRect) {
+            rect = normalizeRect(element.getBoundingClientRect());
+            if (rect) {
```
sebastienvillar: Not needed since I could return rect at the end but makes the implementation more consistent and easier to read

packages/roosterjs-editor-dom/lib/selection/Position.ts
JiuqingSong: The function getBoundingClientRect() may not always work, especially when the position is in middle of a text node. So the original code is still required. The only change we need here is to remove the check of isSafari and always try getClientRects(), and then fallback to the code to insert a temp SPAN if it fails.

packages/roosterjs-editor-dom/lib/selection/Position.ts
```diff
@@ -131,46 +131,51 @@ export default class Position {
...
-                    span.parentNode.removeChild(span);
-                }
+        let rect = range.getBoundingClientRect ? normalizeRect(range.getBoundingClientRect()) : null;
+        if (rect) {
+            return rect;
+        }
+
+        // 2) try to get rect using range.getClientRects
+        const { node, element, offset } = this.normalize();
+        const rects = range.getClientRects ? Array.prototype.slice.call(range.getClientRects()) : [];
```
Lego6245: Why this construction of `Array.prototype.slice.call`? Why grab a shallow copy if all your'e going to do with it is check it's length and then grab the first element?

sebastienvillar: The goal was to satisfy type checking when creating the array to default to (type returned by getClientRects is not instantiable). But it's not needed I agree. Removed and defaulted to null. I already check if rects exists below.

Lego6245: You could type the rects returned as appropriate for a type safeguard. like `const rects: T = blah`. But, this should be fine too.

PR #105, opened 2018-09-19
Switch from .which to .key

Lego6245:
.which is deprecated, we should use .key instead, so this change will do that.

Open question: Does this break in IE / Edge which have spotty support for .key?


Adjective-Object:
Maybe it's worth generating an ie11 key table with autoHotKey and constraining our types with a typedef? I can't find any information on which version of the spec they were using when they implemented it :(

Lego6245:
A cursory look at IE11 shows that they definitely aren't using the same names here. We will have to conditionally use the correct one.

PR #108, opened 2018-09-24
Add alternative selection marker format

Adjective-Object:
The existing selection marker forces a reflow and invalidates the selection in webkit-based browsers, forcing another reflow to reselect the cursor. In large applications with many nodes in their DOM, this can lead to slowdown when typing, since selection is marked every time the user types a space after a non-space character by UndoPlugin.

This PR introduces an alternative experimental selection marker implementation that consumers can opt into consuming. This new selection marker API uses attributes on the DOM instead of spans to track where the selection was on the document.

### DIFF
--- a/packages/roosterjs-editor-dom/lib/selection/experimentalAttributeBasedSelectionMarker.ts
+++ b/packages/roosterjs-editor-dom/lib/selection/experimentalAttributeBasedSelectionMarker.ts
@@ -1,0 +1,149 @@
+import Position from "./Position";
+import createRange from "./createRange";
+import queryElements from "../utils/queryElements";
+
+const SELECTION_START_OFFSET_ATTRIBUTE = "data-offset1";
+const SELECTION_END_OFFSET_ATTRIBUTE = "data-offset2";
+const SELECTION_START_SELECTOR = `[${SELECTION_START_OFFSET_ATTRIBUTE}]`;
+const SELECTION_END_SELECTOR = `[${SELECTION_END_OFFSET_ATTRIBUTE}]`;
+
+type SerializablePosition = {
+  anchorElement: Element;
+  offsetCharacters: number;
+};
+
+/**
+ * Insert selection marker element into content, so that after doing some modification,
+ * we can still restore the selection as long as the selection marker is still there
+ * @param container Container HTML element to query selection markers from
+ * @param rawRange Current selection range
+ * @param useInlineMarker Inline marker will be inserted at the position where current selection is,
+ * so that even some content is changed, we can still still restore the selection. But this can cause
+ * adjacent text nodes to be created. If we are sure the content won't be changed and we don't want to
+ * create adjacent text nodes, set this parameter to false. This usually happens for undo/redo.
+ * @returns True if selection markers are added, otherwise false.
+ */
+export function markSelection(
+  container: HTMLElement,
+  range: Range,
+  useInlineMarker: boolean
+): boolean {
+  if (!range || queryElements(container, SELECTION_START_SELECTOR).length > 0) {
+    return false;
+  }
+  let start = Position.getStart(range).normalize();
+  let end = Position.getEnd(range).normalize();
+  markPosition(
+    positionToSerializablePosition(start),
+    SELECTION_START_OFFSET_ATTRIBUTE
+  );
+  markPosition(
+    positionToSerializablePosition(end),
+    SELECTION_END_OFFSET_ATTRIBUTE
+  );
+  return true;
+}
+
+/**
+ * If there is selection marker in content, convert into back to a selection range and remove the markers,
+ * otherwise no op.
+ * @param container Container HTML element to query selection markers from
+ * @param retrieveSelectionRange Whether retrieve selection range from the markers if any
+ * @returns The selection range converted from makers, or null if no valid marker found.
+ */
+export function removeMarker(
+  container: HTMLElement,
+  retrieveSelectionRange: boolean
+): Range {
+  const selectionStart = container.querySelector(SELECTION_START_SELECTOR);
+  const selectionEnd = container.querySelector(SELECTION_END_SELECTOR);
+  if (selectionStart == null || selectionEnd == null) {
+    return null;
+  }
+
+  const toReturn = retrieveSelectionRange
+    ? createRange(
+        serializablePositionToPosition({
+          anchorElement: selectionStart,
+          offsetCharacters: Number.parseInt(
+            selectionStart.getAttribute(SELECTION_START_OFFSET_ATTRIBUTE)
+          )
+        }),
+
+        serializablePositionToPosition({
+          anchorElement: selectionEnd,
+          offsetCharacters: Number.parseInt(
+            selectionEnd.getAttribute(SELECTION_END_OFFSET_ATTRIBUTE)
+          )
+        })
+      )
+    : null;
+
+  selectionStart.removeAttribute(SELECTION_START_OFFSET_ATTRIBUTE);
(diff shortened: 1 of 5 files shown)


packages/roosterjs-editor-dom/lib/selection/experimentalAttributeBasedSelectionMarker.ts
```diff
@@ -0,0 +1,137 @@
...
+        return {
+            anchorElement: parent,
+            offsetCharacters: offsetCharacters + p.offset,
+        };
+    } else {
+        throw new Error(`node of unknown type ${typeof node}`);
+    }
+}
+
+function serializablePositionToPosition(sp: SerializablePosition): Position {
```
Adjective-Object: RFC name suggestions

packages/roosterjs-editor-dom/lib/selection/experimentalAttributeBasedSelectionMarker.ts
```diff
@@ -0,0 +1,137 @@
...
+    selectionEnd.removeAttribute(SELECTION_END_OFFSET_ATTRIBUTE);
+
+    return toReturn;
+}
+
+function markPosition(sp: SerializablePosition, offsetAttribute: string) {
+    sp.anchorElement.setAttribute(offsetAttribute, sp.offsetCharacters + '');
+}
+
+function positionToSerializablePosition(p: Position): SerializablePosition {
```
Adjective-Object: RFC name suggestions

packages/roosterjs-editor-dom/lib/selection/experimentalAttributeBasedSelectionMarker.ts
```diff
@@ -0,0 +1,137 @@
+import Position from './Position';
+import createRange from './createRange';
+import queryElements from '../utils/queryElements';
+
+const SELECTION_START_OFFSET_ATTRIBUTE = 'data-offset1';
+const SELECTION_END_OFFSET_ATTRIBUTE = 'data-offset2';
+const SELECTION_START_SELECTOR = `[${SELECTION_START_OFFSET_ATTRIBUTE}]`;
+const SELECTION_END_SELECTOR = `[${SELECTION_END_OFFSET_ATTRIBUTE}]`;
+
+type SerializablePosition = {
```
Adjective-Object: RFC name suggestions

This type is different from a Selection in that 
- it guarantees it's anchorElement is an Element, not a Node. This means that the anchor will always have an HTML tag that we can assign attributes to.
- offsetCharacters can step into child text nodes.

packages/roosterjs-editor-dom/lib/selection/experimentalAttributeBasedSelectionMarker.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Insert selection marker element into content, so that after doing some modification,
+ * we can still restore the selection as long as the selection marker is still there
+ * @param container Container HTML element to query selection markers from
+ * @param rawRange Current selection range
+ * @param useInlineMarker Inline marker will be inserted at the position where current selection is,
+ * so that even some content is changed, we can still still restore the selection. But this can cause
+ * adjacent text nodes to be created. If we are sure the content won't be changed and we don't want to
+ * create adjacent text nodes, set this parameter to false. This usually happens for undo/redo.
```
Adjective-Object: I don't think this comment is actually true anymore.

packages/roosterjs-editor-dom/lib/selection/experimentalAttributeBasedSelectionMarker.ts
```diff
@@ -0,0 +1,137 @@
+import Position from './Position';
+import createRange from './createRange';
+import queryElements from '../utils/queryElements';
+
+const SELECTION_START_OFFSET_ATTRIBUTE = 'data-offset1';
+const SELECTION_END_OFFSET_ATTRIBUTE = 'data-offset2';
```
Adjective-Object: these could probably be made more complicated to avoid collisions with native DOM content

packages/roosterjs-editor-dom/lib/selection/experimentalAttributeBasedSelectionMarker.ts
```diff
@@ -0,0 +1,137 @@
...
+        return {
+            anchorElement: node,
+            offsetCharacters,
+        };
+    } else if (node instanceof Text || node instanceof Comment) {
+        let parent = node.parentElement;
+        let offsetCharacters = 0;
+        for (let child: Node = parent.childNodes[0]; child != node; child = child.nextSibling) {
+            offsetCharacters += child.textContent.length;
+        }
```
Adjective-Object: for loop required here because we're iterating a DOMNodeList

packages/roosterjs-editor-dom/lib/selection/experimentalAttributeBasedSelectionMarker.ts
```diff
@@ -0,0 +1,137 @@
...
+        return {
+            anchorElement: parent,
+            offsetCharacters: offsetCharacters + p.offset,
```
JiuqingSong: you can always use Position.element as anchorElement.

packages/roosterjs-editor-dom/lib/selection/experimentalAttributeBasedSelectionMarker.ts
```diff
@@ -0,0 +1,137 @@
...
+        let offsetCharacters = 0;
+        for (let i = 0; i < p.offset; i++) {
+            offsetCharacters += p.element.childNodes[i].textContent.length;
+        }
+        return {
+            anchorElement: node,
+            offsetCharacters,
+        };
+    } else if (node instanceof Text || node instanceof Comment) {
+        let parent = node.parentElement;
```
adriantran: what if you have:

`<div contenteditable="true" id="rooster-div">hello</div>
`
Also, are markers only used for Undo service? Can we return metadata to store along side the snapshot (i.e. non-DOM data).

Adjective-Object: This method doesn't handle root level text elements. For that I think we'd have to come up with an alternate solution.

The only other call I see is in `replaceWithNode`, which doesn't have a clear reason to be using it afaict. @JiuqingSong would know better than I would.

I was thinking about a selection metadata approach as well, but since you are traversing a new DOM tree you have to store the full path required to recover the selection. (e.g. the indexes of each node you need to step down from the root?). That runs into issues if, in `runWithSelectionMarker`, someone is editing the dom.

If we _are_ only using this for Undo, there's no reason we can't store the selection data separately, which seems like a cleaner overall approach to me.

Adjective-Object:
re: perf numbers

I've done puppetteer traces on the sample site. This avoid re-renders, but the actual numbers are pretty nominal. It's only when rooster is in a larger app that undos forcing style & layout calculations gets so expensive.

JiuqingSong:
> re: perf numbers
> 
> I've done puppetteer traces on the sample site. The actual numbers are pretty nominal. It's only when rooster is in a larger app that Undo forcing style & layout calculations gets so expensive.

How possible test the perf in a larger app (e.g. OWA) between original code and your change? I'm so interested with the perf improvement of the change.

Adjective-Object:
We can capture larger traces from owa using something like puppeteer or WebDriver and analyze the time spent in these callstacks in the perf trace.

That requires some investment in automation up front but is definitely worth it.

Adjective-Object:
Closing in favor of removing the need for selection markers at all.

PR #109, opened 2018-09-24
Provide alternate experimental async `select` implementation.

Adjective-Object:
**Problem:** On Master right now, multiple select calls within the same event handler leads to multiple forced layout recalculations in key handlers. This happens in picker plugin because it calls `focus` immediately followed by `replaceWithNode` on insert, but could conceivably happen in any plugin.

**Solution** Rather than write the selection back to the DOM immediately, we can take advantage of the selection caching mechanism in Editor by only updating the cache on `select` calls. Then, in a RAF after all the `select` calls for the event handler have run, we can update the selection.

This means that instead of forcing N layout / style recalculations, we only force 1.


packages/roosterjs-editor-core/lib/coreAPI/experimentalSelectAsync.ts
```diff
@@ -0,0 +1,89 @@
+import EditorCore, { Select } from '../editor/EditorCore';
+import { Browser, Position, contains, createRange } from 'roosterjs-editor-dom';
+import { PositionType } from 'roosterjs-editor-types';
```
Adjective-Object: This file is copied and modified from `select.ts`

JiuqingSong:
Since the implementation is overriding existing select() API, I think you can make this experiment in OWA code first, and if the result is good, you can just totally replace the select API directly, so that no need to introduce extra parameter to EditorOptions.

Lego6245:
I'd argue we should make this publicly available so that way, other consumers can try out the new benefits and we can share the gains as we do it. This will also help cover our blind spots as to other scenarios.

Adjective-Object:
also worth noting: async select fights with UndoPlugin, since UndoPlugin records the selection prior. This work is going to be unusable without a rework of UndoPlugin to not use inline spans (e.g. attributes or a path into cached document like we've been talking about).

JiuqingSong:
Let's chat about the change when I am back.

Adjective-Object:
Testing internally first

PR #110, opened 2018-09-28
6.16.0

Lego6245:
Bump to version 6.16.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "roosterjs",
-  "version": "6.15.1",
+  "version": "6.16.0",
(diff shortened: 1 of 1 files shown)


Adjective-Object:
> Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality. I would usually expect additional code for the latter instance, in which case it’s obviously a minor level increment.

I'd argue the regex fix and the picker-dismiss are bugfixes.

Lego6245:
Picker dismiss feels like an additional functionality that was introduced (a novel behavior we did not have before). I stand by this being a minor bump in this case.

Issue #118, opened 2018-10-17
AltGr is treated as CTRL in Polski in Edge


JiuqingSong:
**Describe the bug**
AltGr+Z should input ż, but not trigger Undo

**To Reproduce**
1. Switch language to Polski
2. Press AltGr (right Alt) + Z

**Expected behavior**
Type "ż"

**Device Information**
 - Browser Edge

PR #119, opened 2018-10-17
Fix issue 118

JiuqingSong:
In Edge, AltGr (right ALT in some languages, like Polski) key sets both e.altKey and e.ctrlKey to true. We need to check e.altKey not to be true before treat them as CTRL+* shortcut.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/ContentEdit/features/shortcutFeatures.ts
+++ b/packages/roosterjs-editor-plugins/lib/ContentEdit/features/shortcutFeatures.ts
@@ -64,7 +64,9 @@ function cacheGetCommand(event: PluginKeyboardEvent) {
     return cacheGetEventData(event, 'DEFAULT_SHORT_COMMAND', () => {
         let e = event.rawEvent;
         let key =
-            event.eventType == PluginEventType.KeyDown
+            // Need to check ALT key to be false since in some language (e.g. Polski) uses AltGr to input some special charactors
+            // In that case, ctrlKey and altKey are both true in Edge, but we should not trigger any shortcut function here
+            event.eventType == PluginEventType.KeyDown && !e.altKey
                 ? e.which |
                   (e.metaKey && Keys.Meta) |
                   (e.shiftKey && Keys.Shift) |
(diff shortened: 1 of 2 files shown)


philipzloh:
With this change, are all these cases identified differently?
1. `Ctrl + V`
2. `Alt + V`
3. `Shift + V`
4. `Ctrl + Alt + V`
5. `Ctrl + Shift + V`
6. `Alt + Shift + V`
7. `Ctrl + Alt + Shift + V`

JiuqingSong:
Yes, Ctrl/Shift/Meta is already handled, Alt is the only thing we missed.

packages/roosterjs-editor-plugins/lib/ContentEdit/features/shortcutFeatures.ts
```diff
@@ -64,7 +64,9 @@ function cacheGetCommand(event: PluginKeyboardEvent) {
     return cacheGetEventData(event, 'DEFAULT_SHORT_COMMAND', () => {
         let e = event.rawEvent;
         let key =
-            event.eventType == PluginEventType.KeyDown
+            // Need to check ALT key to be false since in some language (e.g. Polski) uses AltGr to input some special charactors
+            // In that case, ctrlKey and altKey are both true in Edge, but we should not trigger any shortcut function here
+            event.eventType == PluginEventType.KeyDown && !e.altKey
```
Lego6245: As aluded to in the comment, ctrlKey and altKey are true for AltGraph being pressed. Want to make sure we want to explicitly check this case? Or would we prefer to do a check like what I've got here. Is AltGraph part of any shortcuts in other languages? I know Alt + key is sometimes used which is why I'm hesitant.
```suggestion
            event.eventType == PluginEventType.KeyDown && e.key != AltGraph
```

Lego6245: Just remembered that when spoofing the keyboard, I'm not sure edge recognizes e.key as AltGraph. So maybe this change isn't warranted.

JiuqingSong: At this point e.key is letter key user pressed. In the case I fixed, it is "z", so (e.key != altGr) is always true here.

philipzloh:
Fixes #118

PR #122, opened 2018-10-25
 Encapsulate undo snapshots

Adjective-Object:
encapsulates taking / writing an undo snapshot into a new part of
EditorCore (UndoSnapshotTranslator). This is so we can continue to
test our new approach to serializing the undo state without merging
straight into rooster.

We ensure that undo snapshots can only be created / written to the dom
by using the closest thing typescript has to an opaque type alias (see
Microsoft/TypeScript#15807). Since other
classes can't reach into the UndoSnapshot type, it will be safe to
switch out the UndoSnapshotTranslator implementation

### DIFF
--- a/packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
+++ b/packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
@@ -27,7 +27,7 @@ export default function createEditorCore(
         defaultFormat: calcDefaultFormat(contentDiv, options.defaultFormat),
         corePlugin,
         undo,
-        undoSnapshotTranslator: new UndoSnapshotTranslator(contentDiv),
+        undoSnapshotTranslator: new UndoSnapshotTranslator(),
         currentUndoSnapshot: null,
         customData: {},
         cachedSelectionRange: null,
--- a/packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
+++ b/packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
@@ -13,6 +13,7 @@ import select from '../coreAPI/select';
 import triggerEvent from '../coreAPI/triggerEvent';
 import { DefaultFormat } from 'roosterjs-editor-types';
 import { getComputedStyles } from 'roosterjs-editor-dom';
+import { UndoSnapshotTranslator } from '../undo/UndoSnapshotTranslator';
 
 export default function createEditorCore(
     contentDiv: HTMLDivElement,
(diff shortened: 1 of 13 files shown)


JiuqingSong:
Please fix the build issue.

packages/roosterjs-editor-core/lib/undo/Undo.ts
```diff
@@ -13,7 +11,7 @@ const KEY_DOWN = 40;
 /**
  * Provides snapshot based undo service for Editor
  */
-export default class Undo implements UndoService {
```
JiuqingSong: When remove "default" here?

packages/roosterjs-editor-core/lib/undo/Undo.ts
```diff
@@ -128,22 +126,16 @@ export default class Undo implements UndoService {
...
+        const snapshot = this.editor.undoSnapshotTranslator.getUndoSnapshot();
+        this.getSnapshotsManager().addSnapshot(JSON.stringify(snapshot));
         this.hasNewContent = false;
         return snapshot;
     }
 
     protected getSnapshotsManager(): UndoSnapshotsService {
         if (!this.undoSnapshots) {
-            this.undoSnapshots = new UndoSnapshots(this.maxBufferSize);
+            this.undoSnapshots = new UndoSnapshotsService(this.maxBufferSize);
```
JiuqingSong: Isn't UndoSnapshotsService an interface? I didn't see you change it to a class in your change.

packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
```diff
@@ -26,6 +27,7 @@ export default function createEditorCore(
         defaultFormat: calcDefaultFormat(contentDiv, options.defaultFormat),
         corePlugin,
         undo,
+        undoSnapshotTranslator: new UndoSnapshotTranslator(),
```
JiuqingSong: How to override undoSnapshotTranslator?

Adjective-Object:
The webpack build works fine. There's some issue where the bespoke build tools aren't picking up the UndoSnapshotTranslatorService import properly.

image

@JiuqingSong you wrote these build tools. Do you recognize this issue? Is there a step in the documentation I'm missing to let the build system know about the new file?

JiuqingSong:
> The webpack build works fine. There's some issue where the bespoke build tools aren't picking up the UndoSnapshotTranslatorService import properly.
> 
> image
> 
> @JiuqingSong you wrote these build tools. Do you recognize this issue? Is there a step in the documentation I'm missing to let the build system know about the new file?

You added new type "UndoSnapshotTranslatorService" and use it with a public function, so you need to export it in index.ts

JiuqingSong:
Instead of having a "translator", I prefer to totally replace the current selectionMarker code using your approach, which means whenever we need to take a snapshot (e.g. via getContent() with includeSelectionMarker=true), call your code and replace innerHTML as well as the node index sequence.

PR #126, opened 2018-10-31
Build AMD version of the modules for production

adriantran:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "roosterjs",
-  "version": "6.16.3",
+  "version": "6.16.4",
```
JiuqingSong: Would you please upgrate to 6.17.0 since I also added some new APIs.

adriantran: sure

Issue #107, opened 2018-09-20
Undo Selection Marker logic prevents certain content from be replaced


adriantran:
**To Reproduce**
1. Initialize the editor with this content:

`<div>AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA</div><ol><li>BBBBBBB BBBBBBB BBB BB BBBBBBBBBB BB BBBBBBB BBB BB BBBBBBBBBBB (2d)<ul><li>a</li><li>CCCCCCCCC CCCC: CCCC CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCoCCCCCC</li></ul></li></ol><ol><li>AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA (5d)</li><li>UUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUU. (3d)</li><li>AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA</li></ol>`

2. Open the sample site in Edge and select the first line (e.g. double click it), or the last line
3. Press a key on the keyboard like 'a', text isn't replaced

**Expected behavior**
Text should be replaced

(sometimes it works, most times it doesn't)

**Investigation so far**
It looks like if you're trying add or remove elements (in this case the SPAN marker) under keypress event, there is a race condition which prevents your text from being replaced.

I've also made an attempt to fix the issue by using attribute only markers--adding them to the node's parent attribute list (this includes the node's index). There's at least one problem with this solution: the node's parent could be the contenteditable div itself, but the original issue is fixed.

A different attempt involved clearing the current selection and setting it back after calling remove markers, but that doesn't work (the selection range was created from scratch too).

[...]

PR #135, opened 2018-11-29
6.18.0 Introduce index-based snapshot

JiuqingSong:
This change fixes issue #107 

Before, we were using SPAN tag based snapshot, this need to insert/delete some DOM node into content when take snapshot. This causes some issue, both functional and perf.

Max just developed index-based snapshot, which only store the node index path into snapshot, this doesn't need to modify DOM tree.

In this change I integrate Max' solution into roosterjs with some refactoring (better cooperate with roosterjs code, avoid using recursive function) and remove some workaround code we added before since we don't need them now with the new solution.

e.g, original HTML is: (I use [ for &lt; to avoid auto HTML by github)
[DIV][SPAN]test1[/SPAN][SPAN]test2[/SPAN][/DIV]
and current selection starts before "1" and ends before "2", so the paths for this selection are:
start: [0,4],
end: [1, 4]
And final HTML snapshot will be:

[DIV][SPAN]test1[/SPAN][SPAN]test2[/SPAN][/DIV][!--{start:[0,4],end:[1,4]}--]


JiuqingSong:
This is because we don't add undo snapshot when you just start typing.

adriantran:
is there a negative effect if we do?

JiuqingSong:
No negative effect, but just need to figure out when to add.

PR #181, opened 2018-12-11
Rewrite Hyperlink plugin, fix #131

### LINKED ISSUE #131: Copy pasting content with links having istemptitle doesn't remove title
Hyperlink plugin creates temporary title for CTRL+Click message, but it doesn't work well with copy-paste scenario.

Main issue is when pasting one of these anchor elements, the temp title is still there and is now the true title. Perhaps the correct fix is to setData to the clipboard ourselves when user runs copy.

JiuqingSong:
I rewrote HyperLink plugin to change the way of implementing hyperlink title. The new approach doesn't need to add temporary attribute into A tag, so copy/paste won't have other side effect.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/HyperLink/HyperLink.ts
+++ b/packages/roosterjs-editor-plugins/lib/HyperLink/HyperLink.ts
@@ -37,7 +37,7 @@ export default class HyperLink implements EditorPlugin {
             : [];
     }
 
-    private onMouse = (e: MouseEvent) => {
+    protected onMouse = (e: MouseEvent) => {
         const a = this.editor.getElementAtCursor('a[href]', e.srcElement) as HTMLAnchorElement;
         const href = this.tryGetHref(a);
 
--- a/packages/roosterjs-editor-plugins/lib/HyperLink/HyperLink.ts
+++ b/packages/roosterjs-editor-plugins/lib/HyperLink/HyperLink.ts
@@ -1,19 +1,14 @@
 import { Browser } from 'roosterjs-editor-dom';
-import { ChangeSource, PluginEvent, PluginEventType } from 'roosterjs-editor-types';
 import { Editor, EditorPlugin } from 'roosterjs-editor-core';
-
-const TEMP_TITLE = 'istemptitle';
-const TEMP_TITLE_REGEX = new RegExp(
-    `<a\\s+([^>]*\\s+)?(title|${TEMP_TITLE})="[^"]*"\\s*([^>]*)\\s+(title|${TEMP_TITLE})="[^"]*"(\\s+[^>]*)?>`,
-    'gm'
-);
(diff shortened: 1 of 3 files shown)


packages/roosterjs-editor-plugins/lib/HyperLink/HyperLink.ts
```diff
@@ -33,13 +28,30 @@ export default class HyperLink implements EditorPlugin {
...
+                  editor.addDomEventHandler('mouseout', this.onMouse),
+              ]
+            : [];
     }
 
+    private onMouse = (e: MouseEvent) => {
+        let href = this.tryGetHref(e.srcElement);
+        if (href) {
+            this.editor.setEditorDomAttribute(
+                'titile',
```
adriantran: title*

packages/roosterjs-editor-plugins/lib/HyperLink/HyperLink.ts
```diff
@@ -33,13 +28,30 @@ export default class HyperLink implements EditorPlugin {
...
+            ? [
+                  editor.addDomEventHandler('mouseover', this.onMouse),
+                  editor.addDomEventHandler('mouseout', this.onMouse),
+              ]
+            : [];
     }
 
+    private onMouse = (e: MouseEvent) => {
+        let href = this.tryGetHref(e.srcElement);
+        if (href) {
```
adriantran: When there's no href (e.g. you're mouse over text element), should we remove the title?

JiuqingSong: It doesn't matter where the mouse is moving to, but where it is moving from, so once it is moved out of an A tag, we clear the title.

adriantran: ah, mouseout--but how about when you use your keyboard switch tab or minimize the browser, then move the keyboard elsewhere? Is it guaranteed that mouseout on anchor will always occur?

JiuqingSong: Tried, yes, it will always be triggered.

packages/roosterjs-editor-plugins/lib/HyperLink/HyperLink.ts
```diff
@@ -33,13 +28,30 @@ export default class HyperLink implements EditorPlugin {
...
+            this.editor.setEditorDomAttribute(
+                'titile',
+                e.type == 'mouseover' ? this.getTooltipCallback(href) : null
```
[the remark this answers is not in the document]
JiuqingSong: done

packages/roosterjs-editor-plugins/lib/HyperLink/HyperLink.ts
JiuqingSong: yes, it will be triggered.

PR #188, opened 2018-12-28
fix ctrl+a -> typing issue

Adjective-Object:
Fixes an issue where typing normalization did not run on ctrl+a -> typing 

The bug is because normalization runs before the keypress, and only on an empty div.

so
- have text content
- ctrl+a to select all
- press a button that would type text
- text format normalization does not run because you have text
- browser CED behaviour puts the text into a div at the root level of the document
- text is not formatted.

See #189 for tests.

### DIFF
--- a/packages/roosterjs-editor-core/lib/editor/CorePlugin.ts
+++ b/packages/roosterjs-editor-core/lib/editor/CorePlugin.ts
@@ -124,6 +124,7 @@ export default class CorePlugin implements EditorPlugin {
     /**
      * Ensure we are typing in an HTML Element inside editor, and apply default format if current block is empty
      * @param node Current node
+     * @param event (optional) The keyboard event that we are ensuring is typing in an element.
      * @returns A new position to select
      */
     public ensureTypeInElement(position: NodePosition, event?: PluginKeyboardEvent): NodePosition {
@@ -223,7 +224,10 @@ export default class CorePlugin implements EditorPlugin {
         //
         // This handles the case where the keyboard event that first inserts content happens when
         // there is already content under the selection (e.g. Ctrl+a -> type new content).
-        if (!this.tryNormalizeTyping(event)) {
+        //
+        // Only scheudle when the range is not collapsed to catch this edge case.
+        let range = this.editor.getSelectionRange();
+        if (!this.tryNormalizeTyping(event) && !range.collapsed) {
             requestAnimationFrame(() => {
                 this.tryNormalizeTyping(event);
             });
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-core/lib/editor/CorePlugin.ts
```diff
@@ -125,7 +126,7 @@ export default class CorePlugin implements EditorPlugin {
      * @param node Current node
      * @returns A new position to select
      */
-    public ensureTypeInElement(position: NodePosition): NodePosition {
+    public ensureTypeInElement(position: NodePosition, event?: PluginKeyboardEvent): NodePosition {
```
JiuqingSong: please add comment for the new param

packages/roosterjs-editor-core/lib/editor/CorePlugin.ts
```diff
@@ -210,22 +217,38 @@ export default class CorePlugin implements EditorPlugin {
...
     }
 
+    private onKeyPress(event: PluginKeyboardEvent) {
+        // If normalization was not possible before the keypress,
+        // check again after the keyboard event has been processed by browser native behaviour.
+        //
+        // This handles the case where the keyboard event that first inserts content happens when
+        // there is already content under the selection (e.g. Ctrl+a -> type new content).
+        if (!this.tryNormalizeTyping(event)) {
+            requestAnimationFrame(() => {
```
JiuqingSong: So in most case, "findClosestElementAncestor(range.startContainer) == this.contentDiv" will be false, which means it will request an animation frame. This will happen to almost all key input. I think here we are trying to fix the Ctrl+A case, but should not impart all the normal input.

JiuqingSong: You can check range.collapsed, and only do this when it is false.

Adjective-Object: :+1:

packages/roosterjs-editor-core/lib/editor/CorePlugin.ts
Adjective-Object: :+1:

packages/roosterjs-editor-core/lib/editor/CorePlugin.ts
Adjective-Object: also just want to note that though it would schedule a raf, it wouldn't act within it, so it wouldn't be _that_ much of a hit.

PR #189, opened 2018-12-28
Add instrumentation tests for ctrl+a editor typing issue

Adjective-Object:
Adds instrumentation testing & uses it to test #188 

Jest config setup is based on https://jestjs.io/docs/en/puppeteer. In order to spin up the devserver as part of the test run, and in order to test against firefox, I had to use the manual approach rather than `jest-puppeteer`'s approach.

## Test plan:
`yarn run instrumentation` to run instrumentation tests against the sample site.
`RUN_WITH_DISPLAY=true yarn run instrumentation` to see what the instrumentation tests are doing.


packages/roosterjs-editor-core/lib/editor/CorePlugin.ts
```diff
@@ -2,12 +2,13 @@ import Editor from './Editor';
 import EditorPlugin from './EditorPlugin';
```
Adjective-Object: This file is a dupe of changes in #188 and should disappear once that merges

package.json
```diff
@@ -23,41 +23,50 @@
...
-        "@microsoft/loader-load-themed-styles": "^1.7.118",
-        "@types/node": "^10.12.15",
-        "@types/object-assign": "4.0.30",
-        "@types/react": "^16.7.17",
-        "@types/react-dom": "^16.0.11",
-        "css-loader": "^2.0.1",
-        "node-sass": "^4.11.0",
-        "sass-loader": "^7.1.0",
-        "url-loader": "^1.1.2"
+        "webpack-dev-server": "^3.1.10"
```
Adjective-Object: ordering was normalized by `yarn`. I can manually reorder.

packages/roosterjs-editor-core/tsconfig.json
```diff
@@ -20,7 +20,10 @@
     "lib": [
       "es6",
       "dom"
-    ]
+    ],
+    "types": [
+      "@types/jasmine",
+    ],
```
Adjective-Object: these tsconfig edits are required because by default tsc includes _all_ @types in node_modules.

This lead to a conflict, because this pr introduces jest (as our test runner for the puppeteer test). This change disables the automatic ambient inclusion of types and makes the inclusion of jasmine explicit.

Lego6245:
Can you open an issue/PR for including jest tests as part of our CI checks? I believe it should be straightforward in the .yaml but if it's not we can figure it out when i'm back in the office.

package.json
```diff
@@ -22,7 +22,8 @@
         "test": "karma start",
         "test:debug": "karma start --no-single-run",
         "prettier": "node tools/prettier.js",
-        "publish": "node tools/build.js clean normalize tslint buildcommonjs buildamd dts pack packprod copysample builddemo publish"
+        "publish": "node tools/build.js clean normalize tslint buildcommonjs buildamd dts pack packprod copysample builddemo publish",
+        "instrumentation": "cd publish/samplesite/instrumentation-tests && npm run instrumentation"
```
JiuqingSong: Should we change back to original folder after the test is finished to run?

.travis.yml
```diff
@@ -6,10 +6,12 @@ node_js:
     - 9
 before_install:
     - yarn
+    - cd publish/samplesite/instrumentation-tests && yarn
```
JiuqingSong: same here

package.json
Adjective-Object: npm does not run tasks in the current shell

.travis.yml
Adjective-Object: 👍

package.json
```diff
@@ -22,7 +22,10 @@
         "test": "karma start",
         "test:debug": "karma start --no-single-run",
         "prettier": "node tools/prettier.js",
-        "publish": "node tools/build.js clean normalize tslint buildcommonjs buildamd dts pack packprod copysample builddemo publish"
+        "publish": "node tools/build.js clean normalize tslint buildcommonjs buildamd dts pack packprod copysample builddemo publish",
+        "instrumentation": "cd publish/samplesite/instrumentation-tests && npm run instrumentation",
+        "instrumentation:chrome": "cd publish/samplesite/instrumentation-tests && npm run instrumentation:chrome",
+        "instrumentation:firefox": "cd publish/samplesite/instrumentation-tests && npm run instrumentation:firefox"
```
JiuqingSong: change folder back?

PR #191, opened 2018-12-31
[PickerPlugin] Use performAutoComplete when flagged in PickerPluginOptions

titrongt:
This makes it so Undo happens when hitting Backspace. The emoji is re-autocompleted after because the KeyUp event triggers queryStringUpdated - so this should be fixed in another PR.

Code Changes
1. Add `handleAutoComplete` flag to PickerPluginOptions
2. When defined, use `performAutoComplete` instead of `addUndoSnapshot `

Deploy Link: https://outlook-sdf.office.com/mail/?branch=DESKTOP-ONVDVE4-u-titrongt-testRoosterJs

### DIFF
--- a/packages/roosterjs-plugin-picker/lib/PickerDataProvider.ts
+++ b/packages/roosterjs-plugin-picker/lib/PickerDataProvider.ts
@@ -26,7 +26,7 @@ export interface PickerDataProvider {
     // The first is called in order to "commit" a new element to the editor body that isn't handled automatically by the editor plugin.
     // The second sets the isSuggesting value for situations wherethe UX needs to manipulate the suggesting state that's otherwise plugin managed.
     onInitalize: (
-        insertNodeCallback: (nodeToAvoiddd: HTMLElement) => void,
+        insertNodeCallback: (nodeToAvoid: HTMLElement) => void,
         setIsSuggestingCallback: (isSuggesting: boolean) => void,
         editor?: Editor
     ) => void;
--- a/packages/roosterjs-plugin-picker/lib/PickerDataProvider.ts
+++ b/packages/roosterjs-plugin-picker/lib/PickerDataProvider.ts
@@ -16,6 +16,9 @@ export interface PickerPluginOptions {
     // Vertical (the default, when this is false), will call shiftHighlight with up (false) and down (true).
     // Horizontal (when this is true), will call shiftHighlight with left (false) and right (true).
     isHorizontal?: boolean;
+
+    // Option for letting the editor handle autocomplete on node insertion
+    handleAutoComplete?: boolean;
 }
 
 export interface PickerDataProvider {
@@ -23,7 +26,7 @@ export interface PickerDataProvider {
     // The first is called in order to "commit" a new element to the editor body that isn't handled automatically by the editor plugin.
     // The second sets the isSuggesting value for situations wherethe UX needs to manipulate the suggesting state that's otherwise plugin managed.
     onInitalize: (
-        commitMentionCallback: (nodeToAvoiddd: HTMLElement) => void,
+        insertNodeCallback: (nodeToAvoiddd: HTMLElement) => void,
         setIsSuggestingCallback: (isSuggesting: boolean) => void,
         editor?: Editor
     ) => void;
--- a/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
+++ b/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
@@ -36,7 +36,7 @@ export default class PickerPlugin implements EditorPickerPluginInterface {
     constructor(
         public readonly dataProvider: PickerDataProvider,
         private pickerOptions: PickerPluginOptions
-    ) {}
+    ) { }
 
     /**
      * Get a friendly name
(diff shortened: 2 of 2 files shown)


packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -53,20 +53,21 @@ export default class PickerPlugin implements EditorPickerPluginInterface {
         this.editor = editor;
         this.dataProvider.onInitalize(
             (htmlNode: Node) => {
-                let nodeInserted = false;
                 this.editor.focus();
                 this.editor.addUndoSnapshot();
```
JiuqingSong: You can remove this line now

JiuqingSong: btw, do you want to change the event handler for keyup to keypress instead?

packages/roosterjs-plugin-picker/lib/PickerDataProvider.ts
```diff
@@ -16,14 +16,17 @@ export interface PickerPluginOptions {
...
+    handleAutoComplete?: boolean;
 }
 
 export interface PickerDataProvider {
     // Function called when the plugin is intialized to register two callbacks with the data provider and a reference to the Editor.
     // The first is called in order to "commit" a new element to the editor body that isn't handled automatically by the editor plugin.
     // The second sets the isSuggesting value for situations wherethe UX needs to manipulate the suggesting state that's otherwise plugin managed.
     onInitalize: (
-        commitMentionCallback: (nodeToAvoiddd: HTMLElement) => void,
+        insertNodeCallback: (nodeToAvoiddd: HTMLElement) => void,
```
Lego6245: Is this parameter name typo'd?

Lego6245:
I'm not clear based on your description if this, if it goes in, will break the behavior of the PickerPlugin. It sounds like to me it does? Means you can't backspace with this flag active. Can we fix it so it does work as part of this checkin?

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
titrongt: >             btw, do you want to change the event handler for keyup to keypress instead?
I tested changing keyUp to keyPress, but this caused strange behavior because wordBeforeCursor was always a step behind. For example, if I typed ":)", the keypress event would register it just as ":".

packages/roosterjs-plugin-picker/lib/PickerDataProvider.ts
titrongt: You mean `nodeToAvoiddd`? Oh yeah I wondered about that too, I will remove the extra d's 😄

titrongt:
>           I'm not clear based on your description if this, if it goes in, will break the behavior of the PickerPlugin. It sounds like to me it does? Means you can't backspace with this flag active. Can we fix it so it does work as part of this checkin?

@Lego6245 , how will this break behavior? Backspace will only be affected when `autoCompleteInfo` is set in CorePlugin, but it gets set to null as soon as content is changed again. Is backspace not working as expected on the deploy link?

Lego6245:
“The emoji is re-autocompleted after because the KeyUp event triggers
queryStringUpdated - so this should be fixed in another PR.“

This makes it seem to me like backspace won’t work when that option is
true, which seems to defeat the whole point of this PR. Is this not an
accurate read?

titrongt:
@Lego6245 oh yeah, backspace will undo instead of removing the autocompleted emoji. I didn't handle the KeyUp issue in this PR because I wanted to discuss more how this should be done when everyone was back in office. Maybe it makes more sense to do it in all in one check in then.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
JiuqingSong: Ok. I remember in original code there is some special logic to handle this issue. We can reevaluate it in a separated PR.

Lego6245:
I’ll be back in tomorrow, we can talk then.

packages/roosterjs-plugin-picker/lib/PickerDataProvider.ts
Lego6245: Don't see it fixed on latest.

Lego6245: nvm github being dumb

titrongt:
> This makes it seem to me like backspace won’t work when that option is
> true, which seems to defeat the whole point of this PR. Is this not an
> accurate read?
@Lego6245 backspace breaking PickerPlugin shouldn't be a problem anymore with the latest commit and the use of the KeyPress event. 

> Ok. I remember in original code there is some special logic to handle this issue. We can reevaluate it in a separated PR.
@JiuqingSong sounds good to me, we can address KeyUp re-autocompleting emoji in a separate PR

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -110,6 +111,8 @@ export default class PickerPlugin implements EditorPickerPluginInterface {
         }
         if (event.eventType == PluginEventType.KeyUp && !this.eventHandledOnKeyDown) {
             this.onKeyUpDomEvent(event);
+        } else if (event.eventType == PluginEventType.KeyPress) {
+            this.isCharacterValue = true;
```
Lego6245: Why is this always true?

Lego6245: According to spec it's because modifier keys don't send this event. Mind adding a comment? Do we know of browsers that don't do this?

PR #195, opened 2019-01-14
Provide a fallback to PickerPlugin to use a last known range if focus can't be restored.

Lego6245:
Safari had a problem where the editor would blur and because of it, we couldn't reliably insert the node that the DataProvider wanted us to insert. This change will cache a "last known range" value whenever we're suggesting, and clear it when we're not. Then, if someone blurs the editor, we'll know where they last were so if the normal focus operation doesn't help us, we can fall back to using that piece of info.


Lego6245:
Nuked this by accident to fix my local fork.

PR #198, opened 2019-01-18
Dark mode editor support.

Lego6245:
This is part RFC, part code review, for dark mode support in rooster js. For sample site implementation, please see forthcoming PR which will have this as a dependent.

Philosophy:
Editor, on create, will accept an additional parameter for dark mode support. If provided (either a boolean or a DarkModeOptions object), the editor will be in dark mode.

Dark mode is enabled by having two data propertie, ogsc and ogsb. ogsc represents the color for light mode, ogsb represents the background color for light mode. Conversion is currently one way.

Editor will expose two new APIs:
* isDarkMode -> returns a boolean that will indicate if the editor is in dark mode or not
* getDarkModeOptions -> returns the DarkModeOptions object.

Editor's API will change:
* getContent will take an additional parameter. It will, by default, always return the "light" mode of content, unless passed a boolean to get the dark mode content, if it exists. Otherwise, it's a no-op if in light mode.
There's a new helper method to convert from dark mode to light mode.

Editor Core will have two new properties, isDarkMode and darkModeOptions to facilitate the above. getDefaultFormat will rectify the dark mode options instead of the other defaults if dark mode is enabled.

DefaultFormat has been expanded to provide the ability to set default ogsc and ogsb properties. Dependent files (applyFormat) have been updated to allow for this as well.

setTextColor and setBackgroundColor both take in ogsc and ogsb parameters now.

Paste.ts now applies a transform to pasted content if dark mode is enabled. We provide our own default (strip color) but allow a user to define a behavior through onExternalContentTransform.

Soliciting feedback on the entire project, but I'll ask questions inline.


packages/roosterjs-editor-api/lib/format/setTextColor.ts
```diff
@@ -1,14 +1,24 @@
 import applyInlineStyle from '../utils/applyInlineStyle';
 import { Editor } from 'roosterjs-editor-core';
 
+export interface ColorTuple {
```
Lego6245: Will remove, this was before I thought of my current approach.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -53,14 +54,26 @@ export default class Editor {
      * @param contentDiv The DIV HTML element which will be the container element of editor
      * @param options An optional options object to customize the editor
      */
-    constructor(contentDiv: HTMLDivElement, options: EditorOptions = {}) {
+    constructor(contentDiv: HTMLDivElement, options: EditorOptions = {}, darkMode: boolean | DarkModeOptions = false) {
```
Lego6245: Do you think this function signature is sensible?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -53,14 +54,26 @@ export default class Editor {
...
         // 1. Make sure all parameters are valid
         if (getTagOfNode(contentDiv) != 'DIV') {
             throw new Error('contentDiv must be an HTML DIV element');
         }
 
+        // Set up dark mode defaults if not provided.
+        if (darkMode && typeof darkMode === 'boolean') {
+            darkMode = <DarkModeOptions>{
+                defaultFormat: <DefaultFormat>{
```
Lego6245: Is there a better place to store this formatting? In a const, perhaps in a separate file?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -327,17 +340,28 @@ export default class Editor {
         return isNodeEmpty(this.core.contentDiv, trim);
     }
 
+    public isDarkMode(): boolean {
```
Lego6245: Function comments are missing, will be added.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -359,9 +383,33 @@ export default class Editor {
...
         }
 
+        if (this.core.inDarkMode && !getInDarkMode) {
+            content = this.getLightModeContent(content);
+        }
+
         return content;
     }
 
+    private getLightModeContent(content: string): string {
```
Lego6245: Should this be in its own file? Likely, as it seems Editor.ts is designed for just the public stuff, and a private helper isn't in the right place.

packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
```diff
@@ -52,10 +54,24 @@ export default function createEditorCore(
...
+        inDarkMode: darkModeOptions ? true : false,
+        darkModeOptions: darkModeOptions,
     };
 }
 
-function calcDefaultFormat(node: Node, baseFormat: DefaultFormat): DefaultFormat {
+function calcDefaultFormat(node: Node, options: EditorOptions, darkModeOptions?: DarkModeOptions): DefaultFormat {
+    let baseFormat = null;
+    if (darkModeOptions) {
+        baseFormat = darkModeOptions.defaultFormat ? darkModeOptions.defaultFormat : <DefaultFormat>{
```
Lego6245: This probably furthers the idea there should be a const file somewhere that can be used to store the default dark mode formatting.

packages/roosterjs-editor-plugins/lib/Paste/Paste.ts
```diff
@@ -131,7 +131,18 @@ export default class Paste implements EditorPlugin {
...
-                    this.applyTextFormat(node, clipboardData.originalFormat);
+                    this.applyToElements(node, this.applyFormatting(clipboardData.originalFormat));
+                }
+                if (this.editor.isDarkMode()) {
+                    // either use their paste handler or ours, but check it here.
+                    this.applyToElements(node, (element) => {
+                        if (this.editor.getDarkModeOptions().onExternalContentTransform) {
+                            this.editor.getDarkModeOptions().onExternalContentTransform(element);
+                        } else {
+                            element.style.color = null;
```
Lego6245: Split this into its own file? Or premature?

packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -46,10 +46,10 @@ export default function clearFormat(editor: Editor) {
                 setFontSize(editor, defaultFormat.fontSize);
             }
             if (defaultFormat.textColor) {
-                setTextColor(editor, defaultFormat.textColor);
+                setTextColor(editor, defaultFormat.textColor, defaultFormat.ogsc);
```
titrongt: I know ogsc represents the color for light mode, but why not just say that directly? e.g. defaultFormat.lightMode

packages/roosterjs-editor-api/lib/format/setTextColor.ts
```diff
@@ -1,14 +1,24 @@
...
 
+export interface ColorTuple {
+    lightMode: string;
+    darkMode: string;
+}
+
 /**
  * Set text color at selection
  * @param editor The editor instance
  * @param color The color string, can be any of the predefined color names (e.g, 'red')
```
titrongt: Add param description for ogsc

packages/roosterjs-editor-core/lib/editor/Editor.ts
titrongt: Do you need the two types? Can you just always require DarkModeOptions and initialize defaultFormat if that field is null?

titrongt: Also, do you expect there to be more Dark Mode themes in the future? So that the backgroundColor and textColor will be different than default?

packages/roosterjs-editor-core/lib/editor/Editor.ts
titrongt: +1 for separate file

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -359,9 +383,33 @@ export default class Editor {
             content = extractContentEvent.content;
         }
 
+        if (this.core.inDarkMode && !getInDarkMode) {
+            content = this.getLightModeContent(content);
```
titrongt: This makes it seem like otherwise you return the dark mode content (even if not in DarkMode). Maybe change name to something like `convertToLightModeContent`?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -359,9 +383,33 @@ export default class Editor {
...
+                    element.style.color = element.dataset.ogsc;
+                }
+
+                if (element.dataset.ogsb) {
+                    element.style.backgroundColor = element.dataset.ogsb;
+                }
+            }
+        });
+        const newContent = el.innerHTML;
+        //document.removeChild(el);
```
titrongt: Should this be removed?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -327,17 +340,28 @@ export default class Editor {
...
      * When restore this content, editor will set the selection to the position marked by these markers
+     * @param getInDarkMode Set to true if you want to get the content of the editor in dark mode
+     * This is a no-op when in light mode. If true, instead of converting to light, it will return the 'real' editor content.
      * @returns HTML string representing current editor content
      */
     public getContent(
         triggerExtractContentEvent: boolean = true,
-        includeSelectionMarker: boolean = false
+        includeSelectionMarker: boolean = false,
+        getInDarkMode: boolean = false,
```
Adjective-Object: getInDarkMode is a misleading name (seems like you're getting what the content will be in darkmode, and shouldn't relate to the state of the editor body on the DOM).

I'd suggest something like `normalizeColor = true`. That way you're communicating that by default a transform on the editor's body is being done here.



Also -- all these options are triggering my code smell. This seems like a good candidate for an options kwarg function.

packages/roosterjs-editor-core/lib/editor/Editor.ts
Adjective-Object: `getLightModeContent` is more accurate, since this does not mutate the state of the editor on the dom

packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
Adjective-Object: 👍 for adding a const for that

packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
```diff
@@ -70,6 +86,8 @@ function calcDefaultFormat(node: Node, baseFormat: DefaultFormat): DefaultFormat
         bold: baseFormat.bold,
         italic: baseFormat.italic,
         underline: baseFormat.underline,
+        ogsb: baseFormat.ogsb,
+        ogsc: baseFormat.ogsc,
```
Adjective-Object: ogsb/ogsc shouldn't be used in code -- it's extremely unfriendly to new readers.
Use descriptive names & we can uglify our bundles if we care about the bundle size impact

packages/roosterjs-editor-core/lib/interfaces/DarkModeOptions.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * RoosterJS provides an experemental "external content handler" that transforms text
+     * This is used when content is pasted or inserted via a method we can hook into.
+     * This transform is currently "lossy" and will eliminate color information.
+     * If you want change this behavior, you may define a different function here.
+     * It takes in the impacted HTMLElement
+     */
+    onExternalContentTransform?: (htmlIn: HTMLElement) => void;
+
```
Adjective-Object: nit extra trailing line

packages/roosterjs-editor-core/lib/undo/Undo.ts
```diff
@@ -134,7 +134,8 @@ export default class Undo implements UndoService {
     public addUndoSnapshot(): string {
         let snapshot = this.editor.getContent(
             false /*triggerExtractContentEvent*/,
-            true /*markSelection*/
+            true /*markSelection*/,
+            this.editor.isDarkMode()
```
Adjective-Object: nit: comment for the meaning of the parameter.

packages/roosterjs-editor-plugins/lib/Paste/Paste.ts
Adjective-Object: I think it'd be more elegant if the options argument had a this as a function by default rather than having to check at the callsite...

but we don't have strict null checks turned on so arguably that's less safe

packages/roosterjs-editor-plugins/lib/Paste/Paste.ts
```diff
@@ -180,7 +191,11 @@ export default class Paste implements EditorPlugin {
         }, ChangeSource.Paste);
     }
 
-    private applyTextFormat(node: Node, format: DefaultFormat) {
+    private applyFormatting = (format: DefaultFormat) => (element: HTMLElement) => {
+        applyFormat(element, format);
+    }
+
+    private applyToElements(node: Node, callback: (element: HTMLElement) => void) {
```
Adjective-Object: docstring pls (callback is not a descriptive arg name -- an explanation or a better name would be needed).

packages/roosterjs-editor-types/lib/interface/DefaultFormat.ts
```diff
@@ -36,6 +36,16 @@ interface DefaultFormat {
      * Whether has underline
      */
     underline?: boolean;
+
+    /**
+     * original source background, used for converting back to light mode
+     */
+    ogsb?: string;
```
Adjective-Object: like I said above these would be nicer as just `originalSourceBackground` or `lightmodeBackgroundColor` -- descriptive names onegai

packages/roosterjs-editor-api/lib/format/setTextColor.ts
Lego6245: Good point, especially re: your other comments. Will see about implementing.

packages/roosterjs-editor-plugins/lib/Paste/Paste.ts
Lego6245: That would be very nice, but again, since no strict null checking we're kinda hosed for that. Leaving as is unless you feel like there's a better response, and we can split it out in the future.

packages/roosterjs-editor-core/lib/editor/Editor.ts
Lego6245: gonna change it to `getColorNormalizedContent`

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -359,9 +383,33 @@ export default class Editor {
...
+        if (this.core.inDarkMode && !getInDarkMode) {
+            content = this.getLightModeContent(content);
+        }
+
         return content;
     }
 
+    private getLightModeContent(content: string): string {
+        let el = document.createElement('div');
+        el.innerHTML = content;
```
Lego6245: I'm not convinced this is the right way to do this. Is there a way to do like above, but without having to create an element in the dom?

packages/roosterjs-editor-core/lib/editor/Editor.ts
Lego6245: I'd like to make sure that any consumer of rooster will be able to to define their own dark mode color pallet if they want. They also should be able to override the paste handler if they choose. I think the right call is to wrap everything in the editor options, even though one of the parameters will essentially be a no-op currently (but may be later if we implement switching).

packages/roosterjs-editor-core/lib/editor/Editor.ts
JiuqingSong: 1. Consider having a DarkMode plugin, and handle ExtractContentEvent to replace the colors. So that we can add or not add this plugin to support or not support dark mode.
2. Try put all other configurations into this plugin (if we can), so that Editor and its core doesn't need to know too much about dark mode.
3. With this solution, you don't need to add new parameter to getContent()
4. For replacing color, you can leverage HtmlSanitizer from roosterjs-html-sanitizer package and passing in element callback to handle the colors, and the sanitizer code will handle the looping logic for you.

packages/roosterjs-editor-api/lib/format/setTextColor.ts
```diff
@@ -1,14 +1,30 @@
...
+ * The color string, can be any of the predefined color names (e.g, 'red')
  * or hexadecimal color string (e.g, '#FF0000') or rgb value (e.g, 'rgb(255, 0, 0)') supported by browser.
  * Currently there's no validation to the string, if the passed string is invalid, it won't take affect
+ * Alternatively, you can pass a @typedef ModeIndepenentColor. If in light mode, the lightModeColor property will be used.
+ * If in dark mode, the darkModeColor will be used and the lightModeColor will be set as the ogsc.
  */
-export default function setTextColor(editor: Editor, color: string) {
-    color = color.trim();
-    applyInlineStyle(editor, element => (element.style.color = color));
+export default function setTextColor(editor: Editor, color: string | ModeIndependentColor) {
```
titrongt: This looks very similar to `setBackgroundColor`. Could they be consolidated? Perhaps take a param to set ogsc or ogsb?

packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
```diff
@@ -52,10 +56,19 @@ export default function createEditorCore(
...
+        inDarkMode: options.inDarkMode,
+        darkModeOptions: options.darkModeOptions,
     };
 }
 
-function calcDefaultFormat(node: Node, baseFormat: DefaultFormat): DefaultFormat {
+function calcDefaultFormat(node: Node, options: EditorOptions): DefaultFormat {
+    let baseFormat = null;
+    if (options.inDarkMode) {
+        baseFormat = options.darkModeOptions.defaultFormat ? options.darkModeOptions.defaultFormat : DARK_MODE_DEFAULT_FORMAT;
```
titrongt: 👍🏼

packages/roosterjs-editor-api/lib/format/setTextColor.ts
Lego6245: Fair point, I'd be open to it, but that'd not be in scope for this PR.

packages/roosterjs-editor-types/lib/interface/DarkModeOptions.ts
```diff
@@ -0,0 +1,24 @@
+import { DefaultFormat } from 'roosterjs-editor-types';
+
+/**
+ * The options to specify dark mode editor behavior
+ */
+interface DarkModeOptions {
+    /**
+     * Default formating used for dark mode content. This will be applied to empty content.
+     * If there is already content inside editor, format of existing content will not be changed.
+     * Default value is a set of styles located in INSERT_FILE_HERE.
```
titrongt: INSERT_FILE_HERE --> consts/DarkMode.ts?

packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -46,10 +46,34 @@ export default function clearFormat(editor: Editor) {
...
-                setTextColor(editor, defaultFormat.textColor);
+                if (defaultFormat.originalSourceTextColor) {
+                    const colorPair = {
+                        lightModeColor: editor.isDarkMode() ?
+                        defaultFormat.originalSourceTextColor :
+                        defaultFormat.textColor,
+                        darkModeColor: !editor.isDarkMode() ?
+                        defaultFormat.originalSourceTextColor :
+                        defaultFormat.textColor
+                    }
```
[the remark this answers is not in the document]
Adjective-Object: The value of defaultFormat's colors based on if isDarkMode is set or not seems risky to me. It's the same scenario where callers need to be aware of the current state of the editor (see here).

In this setup, `originalSource<Foo>` also seems to actually refer to `otherMode<Foo>`.

Since we're worried about backwards compatibility, would something like this work:

```typescript
defaultFormat = {
  textColors: {
    lightModeColor: ...
    darkModeColor: ...
  }
  get textColor() {
    return editor.isDarkMode()
      ? this.textColors.darkModeColor
      : this.textColors.lightModeColor
  }
  set textColor(value: string) {
    this.textColors = editor.isDarkMode()
      ? colorPairFromDarkmodeColor(value)
      : colorPairFromLightmodeColor(value)
  }
}
```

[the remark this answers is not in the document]
Lego6245: I may change this entirely, as the more I look at it the more dis-satisfied I am with the result.

My original thinking was that DarkMode would use a seperate default format, and, in the future, if we enabled switching between the two modes, we would swap between one and the other.

OriginalSource as a concept has also been clobbered to mean "other mode" as Max highlighted.

A DefaultFormat should provide all color mappings applicable. If, for whatever reason, they aren't defined, we should provide our defaults, just like before. No more separate dark mode default format to prevent over complication.

I'm gonna take another pass at this, taking into consideration the proposed changes above, and try to arrive at something that makes sense.

packages/roosterjs-editor-api/lib/format/setBackgroundColor.ts
```diff
@@ -1,14 +1,30 @@
...
+        const darkMode = editor.isDarkMode();
+        const appliedColor =  darkMode ? color.darkModeColor : color.lightModeColor;
+        applyInlineStyle(editor, (element) => {
+            element.style.backgroundColor = appliedColor;
+            if (darkMode) {
+                element.dataset.ogsb = color.lightModeColor;
+            }
+        });
+    }
+
```
Adjective-Object: nit: empty line

packages/roosterjs-editor-api/lib/format/setTextColor.ts
```diff
@@ -1,14 +1,30 @@
...
+        const darkMode = editor.isDarkMode();
+        const appliedColor =  darkMode ? color.darkModeColor : color.lightModeColor;
+        applyInlineStyle(editor, (element) => {
+            element.style.color = appliedColor;
+            if (darkMode) {
+                element.dataset.ogsc = color.lightModeColor;
+            }
+        });
+    }
+
```
Adjective-Object: nit: empty line

packages/roosterjs-editor-api/lib/format/setTextColor.ts
```diff
@@ -1,14 +1,30 @@
...
 /**
  * Set text color at selection
  * @param editor The editor instance
- * @param color The color string, can be any of the predefined color names (e.g, 'red')
+ * @param color One of two options:
+ * The color string, can be any of the predefined color names (e.g, 'red')
  * or hexadecimal color string (e.g, '#FF0000') or rgb value (e.g, 'rgb(255, 0, 0)') supported by browser.
  * Currently there's no validation to the string, if the passed string is invalid, it won't take affect
+ * Alternatively, you can pass a @typedef ModeIndepenentColor. If in light mode, the lightModeColor property will be used.
+ * If in dark mode, the darkModeColor will be used and the lightModeColor will be set as the ogsc.
```
Adjective-Object: Is `ogsc` that something we're using in the externally facing API language?

packages/roosterjs-editor-types/lib/consts/DarkMode.ts
```diff
@@ -0,0 +1,13 @@
+import DarkModeOptions from '../interface/DarkModeOptions';
+import DefaultFormat from '../interface/DefaultFormat';
+
+export const DARK_MODE_DEFAULT_FORMAT = <DefaultFormat>{
```
Adjective-Object: cast hiding a bug here. 
`const DARK_MODE_DEFAULT_FORMAT: DefaultFormat = {
}`

packages/roosterjs-editor-types/lib/consts/DarkMode.ts
```diff
@@ -0,0 +1,13 @@
...
+    backgroundColor: 'rgb(51,51,51)',
+    textColor: 'rgb(255,255,255)',
+    originalSourceBackgroundColor: 'rgb(255,255,255)',
+    originalSourceTextColor: 'rgb(0,0,0)',
```
Adjective-Object: these need to go away.

packages/roosterjs-editor-dom/lib/utils/applyFormat.ts
```diff
@@ -8,7 +8,7 @@ import { DefaultFormat } from 'roosterjs-editor-types';
 export default function applyFormat(element: HTMLElement, format: DefaultFormat) {
     if (format) {
         let elementStyle = element.style;
-        let { fontFamily, fontSize, textColor, backgroundColor, bold, italic, underline } = format;
+        let { fontFamily, fontSize, textColor, backgroundColor, bold, italic, underline, originalSourceBackgroundColor, originalSourceTextColor } = format;
```
Adjective-Object: do originalSourceBackgroundColor / originalSourceTextColor still exist on DefaultFormat? should they? since we reworked it to be a getter.

packages/roosterjs-editor-dom/lib/utils/applyFormat.ts
```diff
@@ -31,5 +31,11 @@ export default function applyFormat(element: HTMLElement, format: DefaultFormat)
         if (underline) {
             elementStyle.textDecoration = 'underline';
         }
+        if (originalSourceBackgroundColor) {
+            element.dataset.ogsb = originalSourceBackgroundColor;
+        }
+        if (originalSourceTextColor) {
+            element.dataset.ogsc = originalSourceTextColor;
```
Adjective-Object: `isDarkMode() ? textColors.lightmodeColor : textColors.darkmodeColor`

similarly elsewhere here.

packages/roosterjs-editor-types/lib/consts/DarkMode.ts
```diff
@@ -0,0 +1,12 @@
+import DefaultFormat from '../interface/DefaultFormat';
+
+export const DARK_MODE_DEFAULT_FORMAT = <DefaultFormat>{
```
Adjective-Object: We still shouldn't be casting here. Why do you need to cast if you control the full value & full type?

Lego6245: you're right, the cast no longer applies

packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -46,10 +46,18 @@ export default function clearFormat(editor: Editor) {
                 setFontSize(editor, defaultFormat.fontSize);
             }
             if (defaultFormat.textColor) {
-                setTextColor(editor, defaultFormat.textColor);
+                if (defaultFormat.textColors) {
+                    setTextColor(editor, defaultFormat.textColors);
+                } else {
+                    setTextColor(editor, defaultFormat.textColor);
```
JiuqingSong: nit: setTextColor(editor, defaultFormat.textColors || defaultFormat.textColor);
(current way is also ok to me)

adriantran:
curious to know if this works with copy-paste scenarios

Lego6245:
Pasting content in will be supported (see the changes to Paste.ts)

Copying from editor into another surface currently isn't normalized. I'd do that outside this PR just because of its current size.

PR #205, opened 2019-01-23
Add autocomplete plugin

Adjective-Object:
This is to support moving emoji autocompletes (shortcuts?) out of emoji plugin and into a standalone plugin within OWA.

Adds a configurable autocomplete plugin which the user can opt into

## Test Plan:
`yarn instrumentation customReplacePlugin`

`yarn start` + check autocompletes manually.


packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,157 @@
...
+     * @param editor The editor instance
+     */
+    public initialize(editor: Editor): void {
+        this.editor = editor;
+    }
+
+    /**
+     * Dispose this plugin
+     */
+    public dispose(): void {
```
Lego6245: Practice here is to null out your editor ref on dispose.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,157 @@
...
+    replacementHTML: string,
+    matchSourceCaseSensitive: boolean
+}
+
+const defaultReplacements: Replacement[] = [
+    {
+        sourceString: ':)',
+        replacementHTML: '🙂',
+        matchSourceCaseSensitive: true,
+    },
```
JiuqingSong: suggest create a local util function to help create this object in order to save some bytes. e.g.
function createReplacements(sourceString: string, ... ) {
  return { ... }
}

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Create instance of CustomReplace plugin
+     * @param features An optional feature set to determine which features the plugin should provide
+     */
+    constructor(private replacements: Replacement[] = defaultReplacements) {
+        this.longestReplacementLength = getLongestReplacementSourceLength(this.replacements);
+    }
+
+    updateReplacements(newReplacements: Replacement[]) {
```
JiuqingSong: comment for this function?

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,157 @@
...
+    public dispose(): void {
+    }
+
+    public onPluginEvent(event: PluginEvent) {
+        if (this.editor.isInIME() || event.eventType != PluginEventType.KeyPress) {
+            return;
+        }
+
+        // This handler fires before the event's native behaviour -- defer until after it completes
+        setTimeout(() => this.handlePluginEvent(event), 0)
```
JiuqingSong: 1. try editor.runAsync().
2. Suggest pick all the possible last charactor first, then only run this for those last charactors.
For example, for the default replacements, only ")Oo3" can be the last charactor, so we don't need to do any other check if the typing key is not one of them.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    private handlePluginEvent(event: PluginKeyPressEvent) {
+        // Get the matching replacement
+        const range = document.getSelection().getRangeAt(0);
```
JiuqingSong: Use editor.getSelectionRange()

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,157 @@
...
+        if (range == null) return;
+        const searcher = this.editor.getContentSearcherOfCursor();
```
JiuqingSong: Suggest use cacheGetContentSearcher(), it will returns a cached version if any, of a new one otherwise.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,157 @@
...
+        const stringToSearch = searcher.getSubStringBefore(this.longestReplacementLength);
```
JiuqingSong: In very early implementation, it work like this. But then I found doing a string search here maybe a little low-perf. So I changed it by only search text in the nearest text node. It doesn't work in some corner case (e.g. text is splitted into several text nodes), but it works in most case with high efficency.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,157 @@
...
+        const matchingText = searcher.getSubStringBefore(replacement.sourceString.length)
+        const matchingRange = searcher.getRangeFromText(matchingText, true /* exactMatch */);
+
+        // parse the html string off the dom and inline the resulting element.
+        const parsingSpan = document.createElement('span');
+        parsingSpan.innerHTML = replacement.replacementHTML;
+        const nodeToInsert = (parsingSpan.childNodes.length == 1) ? parsingSpan.childNodes[0] : parsingSpan;
+
+        // Switch the node for the selection range
+        this.editor.addUndoSnapshot(() => {
```
JiuqingSong: editor.performAutoComplete() ?

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,157 @@
...
+            matchingRange.deleteContents();
+            matchingRange.insertNode(nodeToInsert);
+            if (nodeToInsert instanceof Text) {
+                // Fix backspace after text node insertion
+                // (when collapsing the range to the end, the selection goes _after_ the text
+                // node, which lets it delete half of the unicode multipart heart.)
+                const endOfTextNodeRange = new Range();
+                endOfTextNodeRange.setStart(nodeToInsert, nodeToInsert.length);
+                endOfTextNodeRange.setEnd(nodeToInsert, nodeToInsert.length);
+                this.editor.select(endOfTextNodeRange);
```
JiuqingSong: try this.editor.select(nodeToInsert, PositionType.End) rather than creating a range.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,157 @@
...
+        if (stringToSearch.length == 0) {
+            return null;
+        }
+        const lowerCaseStringToSearch = stringToSearch.toLocaleLowerCase();
+        for (const replacement of this.replacements) {
+            const [sourceMatch, replacementMatch] = replacement.matchSourceCaseSensitive
+                ? [stringToSearch, replacement.sourceString]
+                : [lowerCaseStringToSearch, replacement.sourceString.toLocaleLowerCase()];
+
+            if (sourceMatch.endsWith(replacementMatch)) {
```
JiuqingSong: endsWith doesn't work in IE

JiuqingSong:
I run "npm run build" but got several build time error for your branch:

ERROR: D:/roosterjs/2/packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts[95, 9]: if statements must be braced
ERROR: D:/roosterjs/2/packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts[105, 9]: Calls to 'console.log' are not allowed.
ERROR: D:/roosterjs/2/packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts[106, 9]: if statements must be braced
ERROR: D:/roosterjs/2/packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts[156, 9]: if statements must be braced
ERROR: D:/roosterjs/2/packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts[165, 5]: Calls to 'console.log' are not allowed.
ERROR: D:/roosterjs/2/publish/samplesite/scripts/controls/editor/Editor.tsx[57, 9]: if statements must be braced
ERROR: D:/roosterjs/2/publish/samplesite/scripts/controls/editor/Editor.tsx[61, 9]: if statements must be braced

Not sure why this can pass travis-ci build..

JiuqingSong:
I see, it actually failed to build, but not cause the whole check fail. We should fix this. Opened issue #207 

Saw this log in travis-ci:

Start building roosterjs version 7.0.2
ERROR: /home/travis/build/Microsoft/roosterjs/packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts[87, 9]: if statements must be braced
ERROR: /home/travis/build/Microsoft/roosterjs/packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts[97, 9]: Calls to 'console.log' are not allowed.
ERROR: /home/travis/build/Microsoft/roosterjs/packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts[98, 9]: if statements must be braced
ERROR: /home/travis/build/Microsoft/roosterjs/packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts[147, 9]: if statements must be braced
ERROR: /home/travis/build/Microsoft/roosterjs/packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts[156, 5]: Calls to 'console.log' are not allowed.
ERROR: /home/travis/build/Microsoft/roosterjs/publish/samplesite/scripts/controls/editor/Editor.tsx[57, 9]: if statements must be braced
ERROR: /home/travis/build/Microsoft/roosterjs/publish/samplesite/scripts/controls/editor/Editor.tsx[61, 9]: if statements must be braced
(node:6351) UnhandledPromiseRejectionWarning: Error: Command failed: node /home/travis/build/Microsoft/roosterjs/node_modules/tslint/bin/tslint --project /home/travis/build/Microsoft/roosterjs/tools/tsconfig.tslint.json
    at checkExecSyncError (child_process.js:575:11)
    at execSync (child_process.js:612:13)
    at runNode (/home/travis/build/Microsoft/roosterjs/tools/build.js:170:5)
    at tslint (/home/travis/build/Microsoft/roosterjs/tools/build.js:179:5)
    at Object.callback (/home/travis/build/Microsoft/roosterjs/tools/build.js:435:23)
    at /home/travis/build/Microsoft/roosterjs/tools/build.js:464:28
    at Runner.run (/home/travis/build/Microsoft/roosterjs/tools/build.js:475:11)
    at buildAll (/home/travis/build/Microsoft/roosterjs/tools/build.js:558:12)
    at Object.<anonymous> (/home/travis/build/Microsoft/roosterjs/tools/build.js:578:1)
    at Module._compile (internal/modules/cjs/loader.js:654:30)
(node:6351) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 3)
(node:6351) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
JiuqingSong: As I commented in the code, please use input event (for IE, use textinput event instead) so that you don't need this async operation.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
JiuqingSong: 1. If editor has focus at this time, there won't be cached range, each getSelectionRange() call will retrive a real range (just the same as you wrote).
2. If editor doesn't have focus, each time we call editor.select(), it will update the cached range. See https://github.com/Microsoft/roosterjs/blob/e49ff1a655dc6278269b65d027917a0633f94d49/packages/roosterjs-editor-core/lib/coreAPI/select.ts#L66

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
JiuqingSong: We have uglified build, and you can find the build result from dist\roosterjs\dist\rooster-min.js.
However, this won't minify the interface members. If you search in the minified file you will still see those member names there.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,158 @@
...
+
+    private handlePluginEvent(event: PluginKeyPressEvent): void {
+        // Get the matching replacement
+        const range = this.editor.getSelectionRange();
+        if (range == null) return;
+        const searcher = this.editor.getContentSearcherOfCursor();
+        const stringToSearch = searcher.getSubStringBefore(this.longestReplacementLength);
+
+        // Do nothing on control / navigation keys (e.g. keys that did not just type the last character)
+        if (!stringToSearch.endsWith(event.rawEvent.key)) {
```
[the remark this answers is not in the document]
Adjective-Object: mm, must not have added this to the commit. sec.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
Adjective-Object: What I saw while debugging using cacheGetContentSearcher was:

1. I start with the text `the quick brown fo` in my editor
2. I press the `x` key
3. rooster wraps the event and passes it through plugins
4. This plugin schedules an event for after the event is processed.
5. another plugin calls `cacheGetContentSearcher` and attaches it to the event. The PositionContentSearcher saves the current position of the selection(immediately after the 'o')
6. the event handler stack resolves & the browser native behaviour kicks in
7. the async handler fires, inserts an 'x' after the 'o', and moves the selection to the right 1 character
8. the async handler calls `cacheGetContentSearcher`, which fetches the same PositionContentSearcher off the event's cache. That position is no longer correct, since it is still saved to where the 'o' was.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,158 @@
...
+            return;
+        }
+
+        // Exit early on keyboard events that do not use a replacement's final character.
+        if (!this.replacementEndCharacters.has(event.rawEvent.key)) {
+            return
+        }
+
+        // This handler fires before the event's native behaviour -- defer until after it completes
+        setTimeout(() => this.editor.isDisposed() || this.handlePluginEvent(event), 0);
```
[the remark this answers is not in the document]
Adjective-Object: 👍

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -0,0 +1,131 @@
...
+        if (this.editor.isInIME() || event.eventType != PluginEventType.Input) {
+            return;
+        }
+
+        // Get the matching replacement
+        const range = this.editor.getSelectionRange();
+        if (range == null) {
+            return;
+        }
+        const searcher = cacheGetContentSearcher(event, this.editor);
```
JiuqingSong: Why removed the check of all possible last char? Doing a string search is kind of expensive. We should not do it if we already know the result must be empty.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
JiuqingSong: Ok, that makes sense. I also saw this happen to me before when develop auto bullet.

packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
Adjective-Object: Yes. I wasn't seeing InputEvent.data available in tsc or working in firefox, so I assumed it was not available for a reason.

It [looks like `@types/dom-inputevent` is in the ecosystem and has a definition for InputEvent](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/dom-inputevent/index.d.ts
), but there are still issues with using this optimization with inputevent data

- I don't see anywhere in the spec a restriction that InputEvent.data is the character that was inserted by a keyboard event.
- IE11 does not support InputEvent.inputType to distinguish the input events.
- Firefox does not support InputEvent.data _or_ InputEvent.inputType


Because the browser support is so spotty, I saw a risk here for introducing incorrect behavior for basing early exit on this.

Adjective-Object: https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/data
image

Adjective-Object: We could preserve the behaviour reliably by looking at both key and input events, but we'd have to record when the keyboard event happened and clear it when the current event stack ended, which would mean scheduling a raf (or possibly setTimout -- I'm unsure how timers interact with input event firing).

Via https://unixpapa.com/js/key.html

> To understand the difference between keydown and keypress, it is useful to distinguish between "characters" and "keys". A "key" is a physical button on the computer's keyboard. A "character" is a symbol typed by pressing a button. On a US keyboard, hitting the 4 key while holding down the Shift key typically produces a "dollar sign" character. This is not necessarily the case on every keyboard in the world. In theory, the keydown and keyup events represent keys being pressed or released, while the keypress event represents a character being typed. In practice, this is not always the way it is implemented.
>
>For a while, some browers fired an additional event, called textInput, immediately after keypress. Early versions of the DOM 3 standard intended this as a replacement for the keypress event, but the whole notion was later revoked. Webkit supported this between versions 525 and 533, and I'm told IE supported it, but I never detected that, possibly because Webkit required it to be called "textInput" while IE called it "textinput".
>
>There is also an event called input, supported by all browsers, which is fired just after a change is made to to a textarea or input field. Typically keypress will fire, then the typed charater will appear in the text area, then input will fire. **The input event doesn't actually give any information about what key was typed - you'd have to inspect the textbox to figure it out what changed** - so we don't really consider it a key event and don't really document it here. Though it was originally defined only for textareas and input boxes, I believe there is some movement toward generalizing it to fire on other types of objects as well. 

Which I think is the work being done with InputEvent.data now.

Adjective-Object: That leaves us in a situation where we can only reliably get the inserted character off of a keypress event. If we still make the edit happen on an Input event we're left with a more complicated version of what we had before, since now you'd have to be maintaining state between keypress events and clearing it at the end of the current block of events -- which would require multiple calculations of the cached values on events + setTimeout(0)

Adjective-Object:
Alright this implementation should short-circuit properly in Chrome / IE, but won't in Edge or Firefox.

Adjective-Object:
(and it should start short circuiting if the InputEvent spec gets accepted and implemented)

PR #212, opened 2019-01-28
Picker Autocomplete Interactions

Adjective-Object:
- Adds a sample picker to the sample site (pick from one of several colours)
- Adds interaction between autocomplete plugin & picker plugin: Autocomplete plugin should no longer complete text when a picker is open (e.g. the autocomplete `:o` or `:)` will not complete if the picker is open).
- Adds tests for the interaction between autocomplete plugin and picker plugin

Some older files were caught by the autoformatter in my editor.


packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -141,6 +142,17 @@ export default class PickerPlugin implements EditorPickerPluginInterface {
         if (!isSuggesting) {
             this.setLastKnownRange(null);
         }
+        const eventData: PickerPluginSetSuggestingData = {
+            isSuggesting,
+            pickerId: this.pickerOptions.elementIdPrefix
+        }
+        this.editor.triggerEvent(
+            {
+                eventType: PluginEventType.ContentChanged,
```
Lego6245: is `ContentChanged` the right event here?

packages/roosterjs-plugin-picker/lib/PickerDataProvider.ts
```diff
@@ -26,7 +26,7 @@ export interface PickerDataProvider {
     // The first is called in order to "commit" a new element to the editor body that isn't handled automatically by the editor plugin.
     // The second sets the isSuggesting value for situations wherethe UX needs to manipulate the suggesting state that's otherwise plugin managed.
     onInitalize: (
-        insertNodeCallback: (nodeToAvoid: HTMLElement) => void,
+        insertNodeCallback: (nodeToInsert: HTMLElement) => void,
```
Lego6245: 👍 💯 🥇

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
Adjective-Object: This is something I have open questions about -- there doesn't seem to be an existing way to do plugin-to-plugin messaging other than ContentChanged.

@JiuqingSong : what was your intent with this API design? Should I be introducing a different message type here for plugin events or is ContentChanged appropriate?

publish/samplesite/instrumentation-tests/__instrumentation-tests__/customReplacePlugin.test.ts
```diff
@@ -20,100 +20,133 @@ describe('Editor', () => {
...
 
         // Act
-        const resultContent: string = await page.evaluate(() => (window as any).globalRoosterEditor.getTextContent());
-        expect(resultContent.trim()).toEqual("👋");
+        const resultContent: string = await page.evaluate(() =>
+            (window as any).globalRoosterEditor.getTextContent()
+        );
+        expect(resultContent.trim()).toEqual('👋');
     });
 });
```
Adjective-Object: Not sure how this wasn't autoformatted before. The only actual change is renaming the `describe` block

packages/roosterjs-plugin-picker/lib/PickerPluginSetSuggestingEvent.ts
```diff
@@ -0,0 +1,5 @@
+export const PICKER_PLUGIN_SET_SUGGESTING_EVENT = 'PICKER_PLUGIN_SET_SUGGESTING_EVENT'
```
titrongt: Curious whether you think it's worthwhile to allow this to be provided in PickerOptions in the future? Is there value in knowing specifically which picker plugin the event source is?

titrongt:
Code looks good to me, thanks for adding these changes so promptly!

packages/roosterjs-plugin-picker/lib/PickerPluginSetSuggestingEvent.ts
Adjective-Object: that's what the intent behind including the idPrefix as part of the the event was to cover. That way autocomplete plugin can track each of the pickers independently to make sure no pickers are open.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
JiuqingSong: If you'd like some event should be handled in this plugin only (e.g., when picker is opened, other plugins should not handle the key inputs), you can use "willHandleEventExclusively".

PickerPlugin already did this, what it missed here is to return true for InputEvent.

publish/samplesite/scripts/controls/plugins.ts
```diff
@@ -16,6 +18,7 @@ export default interface Plugins {
     eventView: EventViewPlugin;
     api: ApiPlaygroundPlugin;
     customReplace: CustomReplacePlugin,
+    picker: PickerPlugin,
```
JiuqingSong: Please move customReplace and picker plugin declarations [here](https://github.com/Microsoft/roosterjs/blob/00553d8646c059502125f3b870566e9d2b1a1f2f/publish/samplesite/scripts/controls/editor/Editor.tsx#L76), so that they are configurable from side pane.

And suggest adding them into the "Editor Options" side pane, and add sample code for it under "Typescript Code" section.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
Adjective-Object: I was looking at willHandleEventExclusively, but it seemed like it wouldn't make sense here because it's possible to instantiate multiple autocomplete plugins in the same editor, isn't it?

PR #214, opened 2019-01-29
Insert node on empty wordToReplace in PickerPlugin

titrongt:
Currently, PickerPlugin's insertNode callback does nothing when the wordToReplace is empty 

In order for the ExpressionPane in client-web to use this callback, it needs to be able to insert nodes even when there is no text to replace


Adjective-Object:
This makes sense for the use case that we described in person (the picker plugin spawns a persistent UI that allows multiple insertions, so there can be a word to replace for the first insertion, but not for subsequent insertions).

This seems fine design wise to me unles @Lego6245 has objections.

PR #216, opened 2019-01-31
Modify Dark Mode transform order, and enable setContent to transform as well.

Lego6245:
1. Move the dark mode transform logic to be post insertion of the node. This means we can use getComputedStyles if desired in the future to compute dark mode transformations, which is necessary for proper and accessible color normalization.

2. Add a parameter to set content to declare you want the content inserted to be transformed for dark mode.

3. Add a parameter to DarkModeOptions to be used when initializing to transform the content when inserted.

### DIFF
--- a/packages/roosterjs-editor-core/lib/editor/Editor.ts
+++ b/packages/roosterjs-editor-core/lib/editor/Editor.ts
@@ -67,7 +67,11 @@ export default class Editor {
         this.core.plugins.forEach(plugin => plugin.initialize(this));
 
         // 4. Ensure initial content and its format
-        this.setContent(options.initialContent || contentDiv.innerHTML || '');
+        this.setContent(
+            options.initialContent || contentDiv.innerHTML || '',
+            true /* triggerContentChangedEvent */,
+            this.core.darkModeOptions.transformOnInitialize
+        );
 
         // 5. Create event handler to bind DOM events
         this.eventDisposers = [
@@ -170,7 +174,52 @@ export default class Editor {
      * @returns true if node is inserted. Otherwise false
      */
     public insertNode(node: Node, option?: InsertOption): boolean {
-        return node ? this.core.api.insertNode(this.core, node, option) : false;
+        let darkModeNormalize = null;
+        if (this.isDarkMode()) {
+            darkModeNormalize = this.convertContentToDarkMode(node);
+        }
+
+        const result = node ? this.core.api.insertNode(this.core, node, option) : false;
+
+        if (result && darkModeNormalize) {
+            darkModeNormalize();
+        }
+        return result;
+    }
+
+    private isElement(node: Node | HTMLElement): node is HTMLElement {
+        return (<HTMLElement>node).getElementsByTagName !== undefined && (<HTMLElement>node).style !== undefined;
+    }
+
+    private isParentNode(node: Node | ParentNode): node is ParentNode {
+        return (<ParentNode>node).querySelectorAll !== undefined;
+    }
+
+    private convertContentToDarkMode(node: Node): () => void {
+        let childElements: HTMLElement[];
+        // For dark mode, we must ammass a list of all the decendents of a node.
+        // To do this, we have to take mulitple tactics, as a Node could be literally anything.
+        // If a node is an element, getElementsByTagName will return a list of HTML elements.
+        // If it's not an element, it could be a DocumentFragment. querySelectorAll has broad support and achieves the same result.
+        // If it's neither of these things, just give up.
+        if (this.isElement(node)) {
+            childElements = Array.prototype.slice.call(node.getElementsByTagName('*'));
+            childElements.unshift(node);
+        } else if (this.isParentNode(node)) {
+            childElements = Array.prototype.slice.call(node.querySelectorAll('*'));
+        }
+
+        return () => {
+            childElements.forEach((element) => {
+                const darkModeOptions = this.getDarkModeOptions();
+                if (darkModeOptions && darkModeOptions.onExternalContentTransform) {
+                    darkModeOptions.onExternalContentTransform(element);
+                } else {
+                    element.style.color = null;
+                    element.style.backgroundColor = null;
+                }
+            });
+        }
     }
 
     /**
@@ -418,7 +467,7 @@ export default class Editor {
      * @param content HTML content to set in
      * @param triggerContentChangedEvent True to trigger a ContentChanged event. Default value is true
      */
-    public setContent(content: string, triggerContentChangedEvent: boolean = true) {
+    public setContent(content: string, triggerContentChangedEvent: boolean = true, convertToDarkMode?: boolean) {
         let contentDiv = this.core.contentDiv;
         if (contentDiv.innerHTML != content) {
             contentDiv.innerHTML = content || '';
@@ -432,6 +481,10 @@ export default class Editor {
                     let range = getRangeFromSelectionPath(contentDiv, path);
                     this.select(range);
                 } catch {}
+            }
+
+            if (convertToDarkMode) {
+                this.convertContentToDarkMode(contentDiv)();
             }
 
             if (triggerContentChangedEvent) {
--- a/packages/roosterjs-editor-core/lib/editor/Editor.ts
+++ b/packages/roosterjs-editor-core/lib/editor/Editor.ts
@@ -70,7 +70,7 @@ export default class Editor {
         this.setContent(
             options.initialContent || contentDiv.innerHTML || '',
             true /* triggerContentChangedEvent */,
-            this.core.darkModeOptions.transformOnInitialize
+            this.core.darkModeOptions && this.core.darkModeOptions.transformOnInitialize
         );
 
         // 5. Create event handler to bind DOM events
--- a/packages/roosterjs-editor-core/lib/editor/Editor.ts
+++ b/packages/roosterjs-editor-core/lib/editor/Editor.ts
@@ -174,42 +174,32 @@ export default class Editor {
      * @returns true if node is inserted. Otherwise false
      */
     public insertNode(node: Node, option?: InsertOption): boolean {
-        let darkModeNormalize = null;
-        if (this.isDarkMode()) {
-            darkModeNormalize = this.convertContentToDarkMode(node);
-        }
+        // DocumentFragment type nodes become empty after they're inserted.
+        // Therefore, we get the list of nodes to transform prior to their insertion.
+        const darkModeTransform = this.isDarkMode() ? this.convertContentToDarkMode(node) : null;
 
         const result = node ? this.core.api.insertNode(this.core, node, option) : false;
 
-        if (result && darkModeNormalize) {
-            darkModeNormalize();
+        if (result && darkModeTransform) {
+            darkModeTransform();
         }
         return result;
     }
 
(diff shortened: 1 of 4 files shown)


packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -170,7 +174,52 @@ export default class Editor {
      * @returns true if node is inserted. Otherwise false
      */
     public insertNode(node: Node, option?: InsertOption): boolean {
```
Adjective-Object: curious why we opted to implement this at the Editor vs the EditorCore level.

The distinction isn't clear to me, besides the fact that the consumer can override the core API.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -170,7 +174,52 @@ export default class Editor {
      * @returns true if node is inserted. Otherwise false
      */
     public insertNode(node: Node, option?: InsertOption): boolean {
-        return node ? this.core.api.insertNode(this.core, node, option) : false;
+        let darkModeNormalize = null;
```
Adjective-Object: nit: prefer const initialization with ternary -- reads cleaner.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -170,7 +174,52 @@ export default class Editor {
...
+        return (<HTMLElement>node).getElementsByTagName !== undefined && (<HTMLElement>node).style !== undefined;
+    }
+
+    private isParentNode(node: Node | ParentNode): node is ParentNode {
+        return (<ParentNode>node).querySelectorAll !== undefined;
+    }
+
+    private convertContentToDarkMode(node: Node): () => void {
+        let childElements: HTMLElement[];
+        // For dark mode, we must ammass a list of all the decendents of a node.
```
Adjective-Object: typo: amass

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -170,7 +174,52 @@ export default class Editor {
...
+        const result = node ? this.core.api.insertNode(this.core, node, option) : false;
+
+        if (result && darkModeNormalize) {
+            darkModeNormalize();
+        }
+        return result;
+    }
+
+    private isElement(node: Node | HTMLElement): node is HTMLElement {
+        return (<HTMLElement>node).getElementsByTagName !== undefined && (<HTMLElement>node).style !== undefined;
```
Adjective-Object: Does `instanceof HTMLElement`? not work?

[It appears to be a builtin class ](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement) so it should

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -170,7 +174,52 @@ export default class Editor {
...
+        // To do this, we have to take mulitple tactics, as a Node could be literally anything.
+        // If a node is an element, getElementsByTagName will return a list of HTML elements.
+        // If it's not an element, it could be a DocumentFragment. querySelectorAll has broad support and achieves the same result.
+        // If it's neither of these things, just give up.
+        if (this.isElement(node)) {
+            childElements = Array.prototype.slice.call(node.getElementsByTagName('*'));
+            childElements.unshift(node);
+        } else if (this.isParentNode(node)) {
+            childElements = Array.prototype.slice.call(node.querySelectorAll('*'));
+        }
```
Adjective-Object: log warning on unexpected condition.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -170,7 +174,52 @@ export default class Editor {
...
+
+        return () => {
+            childElements.forEach((element) => {
```
Adjective-Object: This will error on a node that has no children (e.g. a Text node) because childElements will not be initialized.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -170,7 +174,52 @@ export default class Editor {
...
+    private isParentNode(node: Node | ParentNode): node is ParentNode {
+        return (<ParentNode>node).querySelectorAll !== undefined;
+    }
+
+    private convertContentToDarkMode(node: Node): () => void {
+        let childElements: HTMLElement[];
+        // For dark mode, we must ammass a list of all the decendents of a node.
+        // To do this, we have to take mulitple tactics, as a Node could be literally anything.
+        // If a node is an element, getElementsByTagName will return a list of HTML elements.
+        // If it's not an element, it could be a DocumentFragment. querySelectorAll has broad support and achieves the same result.
```
Adjective-Object: This whole paragraph is written in a v. flowery way.

In previous jobs I've seen rules (and even linters) to avoid use of overly "poetic" comments / turns of phrase that are overcomplicated because they make maintaining code harder for an international team where English is not everyone's first language.

suggest something like
> Get a collection of the descendants of this node.
> If the node is not an Element or ParentNode, it has no descendants

Adjective-Object: Also, 
https://developer.mozilla.org/en-US/docs/Web/API/ParentNode
> The ParentNode mixin contains methods and properties that are common to all types of Node objects that can have children. It's implemented by Element, Document, and DocumentFragment objects.

Do you even need the Element check?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -170,7 +174,52 @@ export default class Editor {
...
-        return node ? this.core.api.insertNode(this.core, node, option) : false;
+        let darkModeNormalize = null;
+        if (this.isDarkMode()) {
+            darkModeNormalize = this.convertContentToDarkMode(node);
+        }
+
+        const result = node ? this.core.api.insertNode(this.core, node, option) : false;
+
+        if (result && darkModeNormalize) {
+            darkModeNormalize();
```
Adjective-Object: curious why you need to bind the darkModeNormalize node argument before insert.
Can you explain why + document with a comment?

Adjective-Object: And if you don't you could simplify `convertToDarkMode` by moving the ParentNode type check outside of it.

Lego6245: The reason I do it this way is a Node can be one of three things (at least, 3 useful things that we have to do this transform on)

1) An Element (usually an HTML Element)
2) A Document (unlikely, but possible)
3) A DocumentFragment (we do this sometimes)

The first 2, I can get live collections of and I can reliably guarantee that they will still exist even after insertions.

But, when a DocumentFragment is inserted into a document using DOM manipulation APIs, it becomes empty. It's a special cased sort of node that has its _contents_ put in to the document instead of the node itself.

This means that you can't get any of the nodes that exist on it after it's been inserted. You have to get a live list, and then clone the list into an array, and then you can manipulate it afterwords.

I agree it's not clear and I was meaning to put in a comment, but here's the context. If there's a good way to get around it, I'm open to it.

packages/roosterjs-editor-core/lib/editor/Editor.ts
Lego6245: Yeah, you're right, this could live at either level. SetContent needs to remain at the editor level, however (as there's no corresponding core api for it) Defer to Jason for more context about this.

packages/roosterjs-editor-core/lib/editor/Editor.ts
Lego6245: Yeah, you could use this, good catch.

packages/roosterjs-editor-core/lib/editor/Editor.ts
Lego6245: I'll initialize a default empty array so this won't occur.

packages/roosterjs-editor-core/lib/editor/Editor.ts
Lego6245: Yes. See my other long comment about why it's being done this way. I'll simplify the comment here and correct the logic slightly.

packages/roosterjs-editor-core/lib/editor/Editor.ts
Lego6245: You mentioned that there are nodes where there are no child elements, so this condition being unmet isn't unexpected.

packages/roosterjs-editor-core/lib/editor/Editor.ts
Adjective-Object: Cool.

We can still simplify this to a single `instanceof ParentNode` check, can't we? since it's the unifying parent type of Document / DocumentFragment / Node.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -170,7 +174,42 @@ export default class Editor {
...
+        }
+        return result;
+    }
+
+    private convertContentToDarkMode(node: Node): () => void {
+        let childElements: HTMLElement[] = [];
+
+        // Get a list of all the decendents of a node.
+        // querySelectorAll doesn't return a live list when called on an HTMLElement
+        // So we use getElementsByTagName instead for HTMLElement types.
```
Adjective-Object: good comments

Adjective-Object:
lgtm. Deferring accept to Jason for the EditorCore vs Editor design decision though.

packages/roosterjs-editor-core/lib/editor/Editor.ts
JiuqingSong: I don't want to block this check in, so for now I feel it is ok to put the code here.

But as a long term thinking, I'd like to keep everything about dark mode in a small scope, say, a plugin.

My basic idea is, for everything which needs taking care of color (such as set/getContent, setText/BkColor, Paste, ...), we can fire an event before/after the content is changed, and let the plugin handle everything about color change. So we don't need to put any dark specified code everywhere. And we already have ContentChangedEvent today, but if that event is not sufficient, we can add a new event.

And this can happen when we finally integrate the code back to master.

Lego6245: I know that's your position, but we run into problems with colors and guaranteeing execution order. I'm not pleased with the "this plugin HAS to be last" because it doesn't feel like that's enforceable, but that's the only way to make any guarantees about the content. Let's come up with a better strategy for this (core API override?) before going straight to plugin.

JiuqingSong: core API override is also an option.

About your concern of plugin order, there are two options:
1. Make it as a core plugin, so we can decide the order in editor code, and user won't see it.
2. Since we are adding a new event just for color handling, other plugins are not likely to handle it, so it doesn't matter in what order we add the plugin.

Lego6245: Core plugin, we still have to guarantee that the plugin list never changes, and then we end up special casing the plugin order, which still leaves a lot of dark mode specific code, just in the plugin init.

As for 2, just because it's not likely to be used doesn't mean that it _won't_ be, and obscurity isn't a good enough guarantee for this.

PR #219, opened 2019-02-01
7.1.2 Fix #218

### LINKED ISSUE #218: Paste a plaintext hyperlink should go through autolink flow and  trigger ContentChangedEvent with data=anchor node

JiuqingSong:
In an earlier change I fixed an issue when paste a link next to existing charactor, e.g.
Before paste: ()
After paste: (www.test.com)
And the fix allows do auto link to the pasted text. However, this won't trigger ContentChangedEvent with the new anchor element.

So the fix here is to defer this action after text is pasted, and go through the same code path of regular autolink.

### DIFF
--- a/packages/roosterjs-editor-api/lib/format/replaceWithNode.ts
+++ b/packages/roosterjs-editor-api/lib/format/replaceWithNode.ts
@@ -58,6 +58,15 @@ export default function replaceWithNode(
     if (range) {
         let backupRange = editor.getSelectionRange();
 
+        // If the range to replace is rgith before current cursor, it is actually an exact match
+        if (
+            backupRange.collapsed &&
+            range.endContainer == backupRange.startContainer &&
+            range.endOffset == backupRange.startOffset
+        ) {
+            exactMatch = true;
+        }
+
         range.deleteContents();
         range.insertNode(node);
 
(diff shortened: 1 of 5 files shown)


packages/roosterjs-editor-api/lib/format/replaceWithNode.ts
```diff
@@ -58,6 +58,15 @@ export default function replaceWithNode(
     if (range) {
         let backupRange = editor.getSelectionRange();
 
+        // If the range to replace is rgith before current cursor, it is actually an exact match
```
Lego6245: Typo

packages/roosterjs-editor-plugins/lib/ContentEdit/features/autoLinkFeatures.ts
```diff
@@ -43,6 +44,19 @@ function cacheGetLinkData(event: PluginEvent, editor: Editor): LinkData {
     return event.eventType == PluginEventType.KeyDown ||
         (event.eventType == PluginEventType.ContentChanged && event.source == ChangeSource.Paste)
         ? cacheGetEventData(event, 'LINK_DATA', () => {
+              // First try to match link from the whole paste string from the plain text in clipboard.
+              // This helps when we paste a link next to some existing charactor, and the text we got
```
Lego6245: Typo

Lego6245: Charactor -> Character

PR #222, opened 2019-02-01
Track content changed events and report up to PickerPluginDataProvider

Lego6245:
When we get a content changed event, and it's set content, we can assume that something drastic has changed about the state of the editor document. This is used for Undo for instance.

When we get one of these events, we can scan through the document and report any picker placed elements that still exist, and report it to the data provider. For instance, if the data provider is maintaining a list of elements for some reason (like if it needs to track how many are inserted for logging), they currently can't tell if elements are removed through that system. This adds a way for a DataProvider to be informed of the current element list's state after the change happens.

### DIFF
--- a/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
+++ b/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
@@ -8,6 +8,7 @@ import {
     PluginEvent,
     PluginEventType,
     PositionType,
+    ChangeSource,
 } from 'roosterjs-editor-types';
 
 // Character codes.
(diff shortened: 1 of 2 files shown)


packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -123,6 +124,18 @@ export default class PickerPlugin implements EditorPickerPluginInterface {
      * @param event PluginEvent object
      */
     public onPluginEvent(event: PluginEvent) {
+        if (event.eventType == PluginEventType.ContentChanged &&
+            event.source == ChangeSource.SetContent && this.dataProvider.onContentChanged) {
+                // Undos and other major changes to document content fire this type of event.
+                // Inform the data provider of the current picker placed elements in the body.
+                const elementsInDocument = this.editor.getDocument()
```
JiuqingSong: Use editor.queryElements(). Otherwise elements outside editor will also be quried.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -123,6 +124,18 @@ export default class PickerPlugin implements EditorPickerPluginInterface {
...
      */
     public onPluginEvent(event: PluginEvent) {
+        if (event.eventType == PluginEventType.ContentChanged &&
+            event.source == ChangeSource.SetContent && this.dataProvider.onContentChanged) {
+                // Undos and other major changes to document content fire this type of event.
+                // Inform the data provider of the current picker placed elements in the body.
+                const elementsInDocument = this.editor.getDocument()
+                                            .querySelectorAll("[id^='" + this.pickerOptions.elementIdPrefix + "']");
+                let elementIds: string[] = [];
+                elementsInDocument.forEach(element => {
```
JiuqingSong: and editor.queryElements() has a forEachCallback parameter, you can use it to do the loop.

PR #223, opened 2019-02-04
Refactor ReplaceWithNode to use the editor core API

Lego6245:
ReplaceWithNode currently manipulates the dom directly, when there exists an editor API for inserting nodes to begin with. If we extend the definition of the core API to allow for replacing an existing node (instead of relying on calculating what node to replace every time), we can use the editor API. This also will help us with dark mode, as replacing with node is a common operation, and centralizing dom manipulation calls helps us maintain good editor state.

### DIFF
--- a/packages/roosterjs-editor-api/lib/format/replaceWithNode.ts
+++ b/packages/roosterjs-editor-api/lib/format/replaceWithNode.ts
@@ -1,6 +1,6 @@
 import { Editor } from 'roosterjs-editor-core';
 import { PositionContentSearcher } from 'roosterjs-editor-dom';
-import { PositionType } from 'roosterjs-editor-types';
+import { ContentPosition } from 'roosterjs-editor-types';
 
 /**
  * Replace text before current selection with a node, current selection will be kept if possible
@@ -56,16 +56,13 @@ export default function replaceWithNode(
     }
 
     if (range) {
-        let backupRange = editor.getSelectionRange();
-
-        range.deleteContents();
-        range.insertNode(node);
-
-        if (exactMatch) {
-            editor.select(node, PositionType.After);
-        } else {
-            editor.select(backupRange);
-        }
+        editor.insertNode(node, {
+            position: ContentPosition.Range,
+            updateCursor: exactMatch,
+            replaceSelection: true,
+            insertOnNewLine: false,
+            range: range,
+        });
 
         return true;
     }
--- a/packages/roosterjs-editor-types/lib/interface/InsertOption.ts
+++ b/packages/roosterjs-editor-types/lib/interface/InsertOption.ts
@@ -22,12 +22,22 @@ export interface InsertOptionBase {
     replaceSelection?: boolean;
 }
 
+/**
+ * The "basic" insertNode related ContentPositions that require no additional parameters to use.
+ */
 export interface InsertOptionBasic extends InsertOptionBase {
     position: ContentPosition.Begin | ContentPosition.End | ContentPosition.Outside | ContentPosition.SelectionStart;
 }
 
+/**
+ * The Range varient where insertNode will opperate on a range disjointed from the current selection state.
+ */
 export interface InsertOptionRange extends InsertOptionBase {
     position: ContentPosition.Range;
+
+    /**
+     * The range to be targeted when insertion happens.
+     */
     range: Range;
 }
 
--- a/packages/roosterjs-editor-types/lib/interface/InsertOption.ts
+++ b/packages/roosterjs-editor-types/lib/interface/InsertOption.ts
@@ -1,28 +1,44 @@
 import { ContentPosition } from '../enum/ContentPosition';
 
 /**
- * Options for insertContent API
+ * Shared options for insertNode related APIs
  */
-export default interface InsertOption {
-    /**
-     * Target position
-     */
-    position: ContentPosition;
-
+export interface InsertOptionBase {
     /**
      * Whether need to update cursor
      */
-    updateCursor: boolean;
+    updateCursor?: boolean;
 
     /**
      * Whether need to replace current selection
      */
-    replaceSelection: boolean;
+    replaceSelection?: boolean;
 
     /**
      * Whether need to insert the content into a new line
      */
-    insertOnNewLine: boolean;
-
-    range?: Range;
+    insertOnNewLine?: boolean;
 }
+
+/**
+ * The "basic" insertNode related ContentPositions that require no additional parameters to use.
+ */
+export interface InsertOptionBasic extends InsertOptionBase {
+    position: ContentPosition.Begin | ContentPosition.End | ContentPosition.Outside | ContentPosition.SelectionStart;
+}
+
+/**
+ * The Range varient where insertNode will opperate on a range disjointed from the current selection state.
+ */
+export interface InsertOptionRange extends InsertOptionBase {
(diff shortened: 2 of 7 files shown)


packages/roosterjs-editor-types/lib/interface/InsertOption.ts
```diff
@@ -23,4 +23,6 @@ export default interface InsertOption {
      * Whether need to insert the content into a new line
      */
     insertOnNewLine: boolean;
+
+    range?: Range;
```
Lego6245: I know this is missing a descriptive comment.

I previously tried doing this with a union type, but ran into troubles. @Adjective-Object gave me some advice to try exporting more types, so I'll try that instead and see if that works. Any other advice to make this property conditional explicitly on having Range as the ContentPosition is appreciated.

packages/roosterjs-editor-api/lib/format/replaceWithNode.ts
```diff
@@ -56,9 +56,9 @@ export default function replaceWithNode(
     }
 
     if (range) {
-        let backupRange = editor.getSelectionRange();
+        const backupRange = editor.getSelectionRange();
```
Lego6245: Is there a way we can/should move this logic into the core API instead of having it live here?

packages/roosterjs-editor-types/lib/interface/InsertOption.ts
JiuqingSong: To be honest, I don't really like the existing "options" pattern of insertNode API. It makes the method very hard to use, and some of the combinations don't make sense. So I was thinking about changing it to separate parameters, however I didn't do it because the method is already widely used.

I think you may start to change it by having several method overrides, and make sure each override has a valid combination of the parameters, so that you only need to have one override need to take range parameter. And to keep competable with existing code, we can still keep the InsertOption version of override and we can mark it as deprecated.

packages/roosterjs-editor-api/lib/format/replaceWithNode.ts
JiuqingSong: If you want, you can call core.api.getSelectionRange() inside a core API.

Lego6245: The problem is the following "exact match" computation for determining if it's an exact match even if unspecified. Should that be in the core api?

packages/roosterjs-editor-types/lib/interface/InsertOption.ts
Lego6245: The way the core API is currently set up doesn't support this type of override. My advice would be to delay this refactor for a little bit and work on it out of band with this PR. We can discuss a good approach for this then.

JiuqingSong: You don't need to override core API, but only override the Editor method, and keep core API no change.

packages/roosterjs-editor-api/lib/format/replaceWithNode.ts
JiuqingSong: Hmm, that's a good question. In that case I'm ok to leave it here.

packages/roosterjs-editor-types/lib/interface/InsertOption.ts
Lego6245: Overloading there won't help, as overloading is strictly based on the number of parameters. This is more complicated and will refactor more than I would like in this PR. I am open to redesigning this API, but not as part of this PR.

JiuqingSong: My actual concern is we are adding new property to a public type (InsertOption), and once it is added into an official version of roosterjs, people may use it. So it will make it hard to remove it later.

JiuqingSong: I tried this and it seems work:

```typescript
export interface InternalInsertOption extends InsertOption {
    range?: Range;
}
```

and

```typescript

    public insertNode(
        node: Node,
        position: ContentPosition.Begin | ContentPosition.End,
        updateCursor: boolean,
        insertOnNewLine: boolean
    ): boolean;
    public insertNode(
        node: Node,
        position: ContentPosition.SelectionStart,
        updateCursor: boolean,
        insertOnNewLine: boolean,
        replaceSelection: boolean
    ): boolean;
    public insertNode(node: Node, position: ContentPosition.Outside): boolean;
    public insertNode(node: Node, rangeToReplace: Range, updateCursor: boolean): boolean;

    /**
     * @deprecated
     */
    public insertNode(node: Node, option?: InsertOption): boolean;

    public insertNode(
        node: Node,
        positionOrRangeOrOption: ContentPosition | Range | InsertOption,
        updateCursor?: boolean,
        insertOnNewLine?: boolean,
        replaceSelection?: boolean
    ): boolean {
        let options: InternalInsertOption;

        if (positionOrRangeOrOption instanceof Range) {
            options = {
                position: ContentPosition.SelectionStart,
                updateCursor: updateCursor,
                insertOnNewLine: false,
                replaceSelection: true,
                range: positionOrRangeOrOption,
            };
        } else if (typeof positionOrRangeOrOption == 'number') {
            options = {
                position: positionOrRangeOrOption,
                updateCursor: updateCursor,
                insertOnNewLine: insertOnNewLine,
                replaceSelection: replaceSelection,
            };
        } else {
            options = positionOrRangeOrOption;
        }

        return node ? this.core.api.insertNode(this.core, node, options) : false;
    }
```

Here InternalInsertOption may not be a good approach, I just use it to hold the range. But it is only used by the core API, so I feel it should be ok. Another way is to add an additional parameter "range" for the core API. For the core API since it won't be called by consumer code, it should be ok to have some unreasonable parameter combination.

Lego6245: The problem with the above is that if you're replacing with a range, `replaceSelection` is also a valid parameter. Same with `updateCursor` for `ContentPosition.Outside`. If you add in those parameters to those function signatures, then you run into the collision problem again. There's no clean way to conditionally require the range parameter.

Lego6245: An asynchronous plugin that's updating content behind or in front of the cursor as content comes in. Like, something that might replace a link with a "smart title", or automatically embed an image based on typed content. There may be no need to re-focus the editor in those cases, which is what updateCursor does (and is poorly named for what it's doing, I might add).

JiuqingSong: Any valid scenario that replaceSelection is false for a range?

Lego6245: An asynchronous plugin that's updating content behind or in front of the cursor as content comes in. Like, something that might replace a link with a "smart title", or automatically embed an image based on typed content. There may be no need to re-focus the editor in those cases, which is what updateCursor does in addition to restoring the selection (and is poorly named for what it's doing, I might add).

JiuqingSong: So in that case we should collapse the range first, or using a Position object instead of a range.

Then the API needs to support:
- insert at begin/end of the doc
- replace a range
- insert at a position
- insert at selection start (which can be implemented by position)
- put outside

In that case, I'll consider redesign the whole API to support all these cases better. And for now I think you can keep your current implementation.

PR #226, opened 2019-02-07
Add the ability to toggle DarkMode on the editor.

Lego6245:
PR is self describing. I also moved the DarkMode related functions into its own region in Editor.ts.

### DIFF
--- a/packages/roosterjs-editor-core/lib/editor/Editor.ts
+++ b/packages/roosterjs-editor-core/lib/editor/Editor.ts
@@ -106,8 +106,8 @@ export default class Editor {
             ));
             this.core.document.execCommand(DocumentCommand.EnableInlineTableEditing, false, <
                 string
-            >(<any>false));
-        } catch (e) {}
+                >(<any>false));
+        } catch (e) { }
 
         // 9. Let plugins know that we are ready
         this.triggerEvent(
@@ -432,7 +432,7 @@ export default class Editor {
                     this.deleteNode(pathComment);
                     let range = getRangeFromSelectionPath(contentDiv, path);
                     this.select(range);
-                } catch {}
+                } catch { }
             }
 
             if (convertToDarkMode) {
@@ -650,8 +650,8 @@ export default class Editor {
         nameOrMap:
             | string
             | {
-                  [eventName: string]: (event: UIEvent) => void;
-              },
+                [eventName: string]: (event: UIEvent) => void;
+            },
         handler?: (event: UIEvent) => void
     ): () => void {
         if (nameOrMap instanceof Object) {
@@ -884,19 +884,26 @@ export default class Editor {
 
     //#region Dark mode APIs
 
-        /**
-     * Set the dark mode state and converts the content.
-     * @param desiredDarkMode The desired status of dark mode. True if the editor should be in dark mode, false if not.
+    /**
+     * Set the dark mode state and transforms the content to match the new state.
+     * @param nextDarkMode The next status of dark mode. True if the editor should be in dark mode, false if not.
      */
-    public setDarkModeState(desiredDarkMode?: boolean) {
-        if (this.isDarkMode() != desiredDarkMode) {
-            const currentContent = this.getContent(true /* triggerExtractContentEvent */, true /* getSelectionMarker */);
-            this.core.inDarkMode = desiredDarkMode;
-            if (desiredDarkMode) {
-                this.setContent(currentContent, true /* triggetContentChangedEvent */, true /* convertToDarkMode */);
-            } else {
-                this.setContent(currentContent);
-            }
+    public setDarkModeState(nextDarkMode?: boolean) {
+        if (this.isDarkMode() == nextDarkMode) {
+            return;
+        }
+
+        const currentContent = this.getContent(
+            undefined /* triggerContentChangedEvent */,
+            true /* getSelectionMarker */);
+        this.core.inDarkMode = nextDarkMode;
+        if (nextDarkMode) {
+            this.setContent(
+                currentContent,
+                undefined /* triggerContentChangedEvent */,
+                true /* convertToDarkMode */);
+        } else {
+            this.setContent(currentContent);
         }
     }
 
@@ -933,9 +940,9 @@ export default class Editor {
             childElements = Array.prototype.slice.call(node.querySelectorAll('*'));
         }
 
+        const darkModeOptions = this.getDarkModeOptions();
         return childElements.length > 0 ? () => {
             childElements.forEach((element) => {
-                const darkModeOptions = this.getDarkModeOptions();
                 if (darkModeOptions && darkModeOptions.onExternalContentTransform) {
                     darkModeOptions.onExternalContentTransform(element);
                 } else {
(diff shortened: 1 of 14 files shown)


Lego6245:
I dunno why this is displaying like there's a bunch of nonsense changes that aren't related. Give me a moment to try to fix this.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -923,4 +881,70 @@ export default class Editor {
     }
 
     //#endregion
+
+    //#region Dark mode APIs
+
+        /**
```
Adjective-Object: copypasta comment indent problem

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -923,4 +881,70 @@ export default class Editor {
     }
 
     //#endregion
+
+    //#region Dark mode APIs
+
+        /**
+     * Set the dark mode state and converts the content.
+     * @param desiredDarkMode The desired status of dark mode. True if the editor should be in dark mode, false if not.
```
Adjective-Object: 'desired' is kinda a weird name for this. How do you feel about 'shouldBeDarkMode' or 'nextDarkMode' or something a little more standard?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -923,4 +881,70 @@ export default class Editor {
     }
 
     //#endregion
+
+    //#region Dark mode APIs
+
+        /**
+     * Set the dark mode state and converts the content.
```
Adjective-Object: converts the content is ambiguous. consider something like "converts the editor content to match the requested darkmode state"

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -923,4 +881,70 @@ export default class Editor {
...
     //#endregion
+
+    //#region Dark mode APIs
+
+        /**
+     * Set the dark mode state and converts the content.
+     * @param desiredDarkMode The desired status of dark mode. True if the editor should be in dark mode, false if not.
+     */
+    public setDarkModeState(desiredDarkMode?: boolean) {
+        if (this.isDarkMode() != desiredDarkMode) {
```
Adjective-Object: Personally I prefer early exits for methods that are 100% inside an if statement but up2u

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -923,4 +881,70 @@ export default class Editor {
...
+            const currentContent = this.getContent(true /* triggerExtractContentEvent */, true /* getSelectionMarker */);
```
Adjective-Object: Mind breaking up this line? It's kinda hard to read.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -923,4 +881,70 @@ export default class Editor {
...
+            this.core.inDarkMode = desiredDarkMode;
+            if (desiredDarkMode) {
+                this.setContent(currentContent, true /* triggetContentChangedEvent */, true /* convertToDarkMode */);
```
Adjective-Object: ditto on line style

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -923,4 +881,70 @@ export default class Editor {
...
+        if (node instanceof HTMLElement) {
+            childElements = Array.prototype.slice.call(node.getElementsByTagName('*'));
+            childElements.unshift(node);
+        } else if (node instanceof DocumentFragment) {
+            childElements = Array.prototype.slice.call(node.querySelectorAll('*'));
+        }
+
+        return childElements.length > 0 ? () => {
+            childElements.forEach((element) => {
+                const darkModeOptions = this.getDarkModeOptions();
```
Adjective-Object: nit: this can be pulled out of the forEach loop

packages/roosterjs-editor-core/lib/editor/Editor.ts
Adjective-Object: typo: trigget

Adjective-Object: We don't actually care about the value of triggerContentChanedEvent here, we just want to use the default.

In that case I think you can pass `undefined`

https://www.typescriptlang.org/docs/handbook/functions.html#optional-and-default-parameters
>In TypeScript, we can also set a value that a parameter will be assigned if the user does not provide one, or if the user passes undefined in its place. These are called default-initialized parameters. Let’s take the previous example and default the last name to "Smith".

packages/roosterjs-editor-core/lib/editor/Editor.ts
Lego6245: Yeah, it's a style thing. Save an indent, save a life.

packages/roosterjs-editor-core/lib/editor/Editor.ts
Lego6245: Oh, didn't know undefined had that behavior. Good call.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -692,8 +650,8 @@ export default class Editor {
         nameOrMap:
             | string
             | {
-                  [eventName: string]: (event: UIEvent) => void;
-              },
+                [eventName: string]: (event: UIEvent) => void;
+            },
```
Adjective-Object: dat autoformat. we really need prettier or smth on this repo.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -923,4 +881,77 @@ export default class Editor {
...
+        // querySelectorAll doesn't return a live list when called on an HTMLElement
+        // So we use getElementsByTagName instead for HTMLElement types.
+        if (node instanceof HTMLElement) {
+            childElements = Array.prototype.slice.call(node.getElementsByTagName('*'));
+            childElements.unshift(node);
+        } else if (node instanceof DocumentFragment) {
+            childElements = Array.prototype.slice.call(node.querySelectorAll('*'));
+        }
+
+        const darkModeOptions = this.getDarkModeOptions();
```
Adjective-Object: This can only be pulled out to this level if the darkmode options will never change, since the lifetime of the lambda might outlive the lifetime of the options we get here otherwise. is that the case?

Lego6245: You're right about that, but there's no mechanism to currently update the darkModeOptions. However, we might make those options mutable in the future (though I'm not sure that's good design). I think it's safer to put it inside the callback but outside the for-each

packages/roosterjs-editor-core/lib/editor/Editor.ts
Lego6245: Not as a pre-commit hook, but I agree. #228 opened.

PR #225, opened 2019-02-07
Merge in current master in preperation for additional changes prior to 7.1.3

Lego6245:
I want to make a few more changes prior to 7.1.3 going out on the dark mode branch, so this PR is to separate out the merge of the current state of master so the resulting PRs can be clean.

PR #227, opened 2019-02-07
Address comments raised in #225 but not addressed

Lego6245:
@Adjective-Object raised some concerns in #225, but that was literally right after I closed the PR. The actual changes went in with #223. I opened this PR to address those concerns.


Lego6245:
A comment was made about consistency between inline and explicit declarations of one-use options objects. The reason it was declared the way it was inside InsertContentPane is because TypeScript wasn't appropriately inferring the type of the object as InputTypeBasic when it was declared inline (likely because it wasn't inferring that the enum was limited by the preceeding if statement). I'm open to making this consistent but wasn't sure what other places you identified that we'd want to ensure were the same.

PR #229, opened 2019-02-09
Fix a js error in insertNode

JiuqingSong:
In previous check in there is a potential javascript error in insertNode() function when the "option" parameter. This change fix the error.

### DIFF
--- a/packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
+++ b/packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
@@ -17,7 +17,10 @@ import {
     createRange,
 } from 'roosterjs-editor-dom';
 
-function getInitialRange(core: EditorCore, option: InsertOption): { range: Range, rangeToRestore: Range } {
+function getInitialRange(
+    core: EditorCore,
+    option: InsertOption
+): { range: Range; rangeToRestore: Range } {
     // Selection start replaces based on the current selection.
     // Range inserts based on a provided range.
     // Both have the potential to use the current selection to restore cursor position
@@ -31,18 +34,16 @@ function getInitialRange(core: EditorCore, option: InsertOption): { range: Range
         rangeToRestore = range.cloneRange();
     }
 
-    return {range, rangeToRestore };
+    return { range, rangeToRestore };
 }
 
 const insertNode: InsertNode = (core: EditorCore, node: Node, option: InsertOption) => {
-    if (!option) {
-        option = {
-            position: ContentPosition.SelectionStart,
-            insertOnNewLine: option.insertOnNewLine != null ? option.insertOnNewLine : false,
-            updateCursor: option.updateCursor != null ? option.updateCursor : true,
-            replaceSelection: option.replaceSelection != null ? option.replaceSelection : true,
-        }
-    }
+    option = option || {
+        position: ContentPosition.SelectionStart,
+        insertOnNewLine: false,
+        updateCursor: true,
+        replaceSelection: true,
+    };
     let contentDiv = core.contentDiv;
 
     if (option.updateCursor) {


packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
```diff
@@ -31,18 +34,16 @@ function getInitialRange(core: EditorCore, option: InsertOption): { range: Range
...
 const insertNode: InsertNode = (core: EditorCore, node: Node, option: InsertOption) => {
-    if (!option) {
-        option = {
-            position: ContentPosition.SelectionStart,
-            insertOnNewLine: option.insertOnNewLine != null ? option.insertOnNewLine : false,
-            updateCursor: option.updateCursor != null ? option.updateCursor : true,
-            replaceSelection: option.replaceSelection != null ? option.replaceSelection : true,
-        }
-    }
+    option = option || {
```
Lego6245: @Adjective-Object mentioned in the last PR that this would not work, because if option is non-falsey but has nulled/undefined parameters, we wouldn't be providing the correct set of defaults. What JS error are we guarding against, and is there a way we can guarantee the intent of having a default set of a parameters?

JiuqingSong: old code is "if (!option) { option.***)“

here option is null. it repros 100% when auto bullet

JiuqingSong: If user passes option.something = null, we will treat it as false. This is by design.

Lego6245: Proposal to capture both of our intents:

```javascript
option = option ? {
        position: option.position,
        insertOnNewLine: option.insertOnNewLine != null ? option.insertOnNewLine : false,
        updateCursor: option.updateCursor != null ? option.updateCursor : true,	
        replaceSelection: option.replaceSelection != null ? option.replaceSelection : true,
} : {
        position: ContentPosition.SelectionStart,
        insertOnNewLine: false,
        updateCursor: true,
        replaceSelection: true,
}

Comment (JiuqingSong) [reply packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts]: So the question here is how should we treat null value here.

1. We marked the whole option param as optional, so people either pass a full object, or a null object.

2. People can still set some of the members in option as null, but they need to do this explicitly, which means

```
option = {
        position: ContentPosition.SelectionStart,
        insertOnNewLine: null,
        updateCursor: null,
        replaceSelection: null,
}
```

is valid, but 

```
option = {
        position: ContentPosition.SelectionStart,
}
```

is invalid. So we can say we don't support passing some of option as "default" value here, and passing a null means people want to pass a null value, and we should treat it as false.

3. This is already the existing behavior. People may already have null value in their code to be treated as false, although this is super uncommon, but we should still keep the meaning of parameter no change unless we deprecate the old API and have a new one, otherwise this will be a breaking change without build-time error.

4. We also have a lot of function takes boolean parameter, I don't think we should do null check for each of them and change to default value. The only case for default value is using undefined rather than null, and in that case it is already done by compiler as long as we have default value in function signature.

So my proposal is not having this check.

Comment (Lego6245) [reply packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts]: Here was the original code block, prior to my changes.

```javascript
    let position = option ? option.position : ContentPosition.SelectionStart;
    let updateCursor = option ? option.updateCursor : true;
    let replaceSelection = option ? option.replaceSelection : true;
    let insertOnNewLine = option ? option.insertOnNewLine : false;
    let contentDiv = core.contentDiv;
```

I have in recent changes made those above parameters optional on the object. So, the potential value space here could be `null`, `undefined`, `false`, or `true` (or maybe some other JS weirdness, IDK).

So, all of the following are valid objects to be passed in to insert node.
```javascript
let validOne = null;
let validTwo = { }
let validThree = {
    position: ContentPosition.Begin,
    insertOnNewLine: undefined,
    replaceSelection: false,
}
```

Without explicit null checks, we can't be sure of the intent of a null or undefined. So, ultimately, I think you're correct in that we should go with your solution. Thanks for working this out with me and explaining your reasoning.

PR #236, opened 2019-02-15
Fix #74

### LINKED ISSUE #74: RoosterJS: Editor only has list or table, content is not inserted properly for inserting at begin or end
RoosterJS base its insertion on object model.
For inserting at begin, it finds first block and insert content before it.
For inserting at end, it finds last block and insert content after it.

The current object model in RoosterJS has the block concept which essentially represents a parsing boundary, and is not equipped with Table or List. A List item or a table cell is a block, but beyond that (list or table), object model does not know.

So when you insert at end, and the end happens to be a table cell. The content will be inserted after the cell, but before the closing TR/Table.

A possible fix, as I discussed with Jia is to introduce a ContentBlock concept to representing block larger than block we have, i.e. TR, Table, List. BlockElement cannot be nested, but ContentBlock can be nested. For inserting at end and end is a table cell, we should go from a block element, and find topest Content Block and insert after.

Regards,
Nick Huang

JiuqingSong:
#74 RoosterJS: Editor only has list or table, content is not inserted properly for inserting at begin or end

Add a new content edit feature "InsertLineBeforeStructuredNodeFeature " (default set to false), it checks if the cursor is at the beginning of a TABLE or a LIST and there's no other blocks before, when user press ENTER, we insert a new line before the element. This is the same behavior of Word.


packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -93,15 +94,18 @@ export default class Editor {
             this.contenteditableChanged = true;
         }
 
-        // 8. Disable these operations for firefox since its behavior is usually wrong
+        // 8. Make settings change to make editor work better
```
titrongt: Clarify this comment - how will this make change make the editor "work better"?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -93,15 +94,18 @@ export default class Editor {
             this.contenteditableChanged = true;
         }
 
-        // 8. Disable these operations for firefox since its behavior is usually wrong
+        // 8. Make settings change to make editor work better
         // Catch any possible exception since this should not block the initialization of editor
         try {
+            // Disable these operations for firefox since its behavior is usually wrong
```
titrongt: Same as above - in what ways is the behavior wrong?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -93,15 +94,18 @@ export default class Editor {
...
         try {
+            // Disable these operations for firefox since its behavior is usually wrong
             this.core.document.execCommand(DocumentCommand.EnableObjectResizing, false, <string>(
                 (<any>false)
             ));
             this.core.document.execCommand(DocumentCommand.EnableInlineTableEditing, false, <
                 string
             >(<any>false));
+            // Change the default paragraph separater to DIV. This is mainly for IE since its default setting is P
+            this.core.document.execCommand(DocumentCommand.DefaultParagraphSeparator, false, 'div');
```
titrongt: If the purpose of these operations is for Firefox and IE, why isn't there a browser check?

packages/roosterjs-editor-plugins/lib/ContentEdit/features/insertLineBeforeStructuredNodeFeature.ts
```diff
@@ -0,0 +1,55 @@
...
+                return editor.getElementAtCursor(
+                    getTagOfNode(element) == 'LI' ? 'OL,UL' : 'TABLE',
+                    element
+                );
+            }
+        }
+
+        return null;
+    });
+}
```
titrongt: I see a lot of hard-coding in this file. Is there a consts file that contain these element names?

JiuqingSong: Move most into const. We don't have a global const file because we want to control the .js file size. Typescript doesn't inline the const value while compiling.

PR #239, opened 2019-02-19
Skip the root element when setContent is called w/ dark mode transform.

Lego6245:
When we call setContent, we get a reference to the core CED and set the content that way. However, when we do this, all we have left that's an HTML node is that root CED. We actually don't want to modify the CED, but we do want to modify the content coming into it. So, we add a parameter to the dark mode transform function to skip that root element, meaning the only thing that gets transformed is the actual editor content itself.

### DIFF
--- a/packages/roosterjs-editor-core/lib/editor/Editor.ts
+++ b/packages/roosterjs-editor-core/lib/editor/Editor.ts
@@ -418,6 +418,7 @@ export default class Editor {
      * Set HTML content to this editor. All existing content will be replaced. A ContentChanged event will be triggered
      * @param content HTML content to set in
      * @param triggerContentChangedEvent True to trigger a ContentChanged event. Default value is true
+     * @param convertToDarkMode True to conver the editor's new content to dark mode formatting. Default value is false.
      */
     public setContent(content: string, triggerContentChangedEvent: boolean = true, convertToDarkMode?: boolean) {
         let contentDiv = this.core.contentDiv;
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -926,16 +927,19 @@ export default class Editor {
...
+    private convertContentToDarkMode(node: Node, skipRootElement?: boolean): () => void {
         let childElements: HTMLElement[] = [];
 
         // Get a list of all the decendents of a node.
         // querySelectorAll doesn't return a live list when called on an HTMLElement
         // So we use getElementsByTagName instead for HTMLElement types.
         if (node instanceof HTMLElement) {
             childElements = Array.prototype.slice.call(node.getElementsByTagName('*'));
-            childElements.unshift(node);
+            if (!skipRootElement) {
```
Lego6245: This is a style question mostly, but would folks prefer "includeRootElement" instead?

JiuqingSong: In what case skipRootElement is false?

Lego6245: Above, on line 183:
`const darkModeTransform = this.isDarkMode() ? this.convertContentToDarkMode(node) : null;`

JiuqingSong: I see.

PR #247, opened 2019-02-27
Selection Management step 1

JiuqingSong:
I was planning to start the PR once I finished all steps, but it turns out that I still need some more time to finialize the other steps, and the first step can already fix the "paste hyperlink cause cursor pos wrong" issue, so let's start the first step PR.

The general idea of selection management is to cache a selection range every time a selection has been changed. In the first step, I split the core API select() into two steps:
1. create selection range from several combinations of parameters (this part is merged to existing API createRange()):
1.1 Single node => range around the node
1.2 Two nodes => range around the two nodes
1.3 Node+offset => collapsed range at the node+offset
1.4 Node1+offset1+node2+offset2 => range start at node1+offset1, end at node2+offset2
1.5 Node+startPath (number[]) + endPath (number[]) => range start at start path, end at end path
1.6 Single NodePosition => collapsed range at the position
1.7 Two NodePositions => range start at position 1, end at position 2
2. select the range
2.1 Rename the API to selectRange
2.2 Add a parameter ''skipSameRange", when set to true, do nothing if new range is the same with existing one
2.3 deprecate the old API select()

Then change the core API focus() to use selectRange() API with skipSameRange=true, to avoid lose current format state (B/I/U) when call focus().
Change all places where select() is called to new API.
The name and parameter list of Editor method Editor.select() are not changed,


packages/roosterjs-editor-core/lib/coreAPI/selectRange.ts
```diff
@@ -1,73 +1,55 @@
...
+};
+
+export default selectRange;
+
+/**
+ * @deprecated Only for compatibility with existing code, don't use ths function, use selectRange instead
+ */
+export const select: Select = (core: EditorCore, arg1: any, arg2?: any, arg3?: any, arg4?: any) => {
+    let range = arg1 instanceof Range ? arg1 : createRange(arg1, arg2, arg3, arg4);
+    return core.api.selectRange(core, range);
```
titrongt: Why do you no longer need the logic for when (arg1 instanceof Node)?

packages/roosterjs-editor-core/lib/coreAPI/selectRange.ts
```diff
@@ -1,73 +1,55 @@
...
+        let selection = core.document.defaultView.getSelection();
+        if (selection) {
+            let needAddRange = true;
+
+            if (selection.rangeCount > 0) {
+                // Workaround IE exception 800a025e
+                try {
+                    // Do not remove/add range if current selection is the same with target range
+                    // Without this check, execCommand() may fail in Edge since we changed the selection
+                    let currentRange = selection.rangeCount == 1 ? selection.getRangeAt(0) : null;
```
titrongt: nit: move currentRange assignment and checks to after we already know (skipSameRange || Browser.isEdge)

packages/roosterjs-editor-core/lib/coreAPI/selectRange.ts
```diff
@@ -1,73 +1,55 @@
...
+import { Browser, contains, createRange } from 'roosterjs-editor-dom';
+
+const selectRange: SelectRange = (core: EditorCore, range: Range, skipSameRange?: boolean) => {
+    if (contains(core.contentDiv, range)) {
+        let selection = core.document.defaultView.getSelection();
+        if (selection) {
+            let needAddRange = true;
+
+            if (selection.rangeCount > 0) {
+                // Workaround IE exception 800a025e
```
titrongt: Is the IE exception that selection has no rangeCount? Curious what then is the behavior in IE?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -517,7 +520,8 @@ export default class Editor {
     ): boolean;
 
     public select(arg1: any, arg2?: any, arg3?: any, arg4?: any): boolean {
```
titrongt: Is this a duplicate of the const defined in selectRange?

packages/roosterjs-editor-core/lib/coreAPI/selectRange.ts
JiuqingSong: It should be an IE bug.

packages/roosterjs-editor-core/lib/coreAPI/selectRange.ts
JiuqingSong: It is checking arg1 is Range, not Node.
The old code support passing range as parameter, in that case we already have the range and no need to createRange

packages/roosterjs-editor-core/lib/editor/Editor.ts
JiuqingSong: This is the public editor API, while selectRange is a CoreAPI and can't be called outside editor.

packages/roosterjs-editor-core/lib/coreAPI/selectRange.ts
JiuqingSong: make sense.

packages/roosterjs-editor-core/lib/coreAPI/selectRange.ts
titrongt: I meant that in the old code, there was another conditional that checked if arg1 was a Node. But since this is now selectRange instead of select, it probably no longer applies.

packages/roosterjs-editor-dom/lib/selection/createRange.ts
```diff
@@ -44,3 +103,43 @@ function getFocusablePosition(position: NodePosition) {
...
+        offset = path[i];
+        if (
+            i < path.length - 1 &&
+            node &&
+            node.nodeType == NodeType.Element &&
+            node.childNodes.length > offset
+        ) {
+            node = node.childNodes[offset];
+        } else {
+            break;
```
titrongt: If you always break out of the for loop once the conditional is not true, doesn't that mean you only ever check the index of 0? Why do you need a loop in this case?

JiuqingSong: This is existing code, moved from getSelectionPath.ts. It has been there for a while and run perfectly. So I believe we don't need to change it.

PR #248, opened 2019-02-27
 Better color normalization handling for Dark Mode

Lego6245:
When normalizing dark mode content, we should also take a step to handle any other data properties attached to the element. For instance, if we assign a color to an element when we normalize it into dark mode, but we don't have a corresponding value that it was originally (say the element was colored by a color attribute), we need to fix that to have a better normalization behavior.

I'm very open to moving this into a custom dark mode option just like externalContentTransform, but I'm soliciting feedback on if that's the right move.


packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -403,14 +403,26 @@ export default class Editor {
...
+
+                // Some elements might have set attribute colors. We need to reset these as well.
+                if (element.dataset.ogac) {
+                    element.setAttribute('color', element.dataset.ogac);
                 }
 
-                if (element.dataset.ogsb) {
-                    element.style.backgroundColor = element.dataset.ogsb;
+                if (element.dataset.ogab) {
+                    element.setAttribute('bgcolor', element.dataset.ogac);
```
JiuqingSong: typo? should be ogab?

Lego6245: Yup, thanks for that catch.

PR #250, opened 2019-03-04
Fix ordered lists and set content for dark mode

Lego6245:
1. Ordered lists were applying styles that were based on the current state of the element and did not use the data attributes we store the original colors on. Because of this, when covering content to light mode, it would keep its dark mode styles.

2. When toggling dark mode on content, if the content being set was identical to the content already existing, we'd skip the transform step. This forces the transform to run if required.


packages/roosterjs-editor-api/lib/utils/processList.ts
```diff
@@ -64,7 +66,10 @@ function applyListFormat(node: Node, formats: string[], isDarkMode: boolean) {
         {
             fontFamily: formats[0],
```
Lego6245: Looking at getComputedStyles and this code block, it seems like we're relying on hard coded positioning of returned attributes. We should likely refactor this to return a typed object that we can then reference, as opposed to this hacky workaround. Thoughts on undertaking this work now?

titrongt: Sounds like an improvement to me. But if it's too high cost, I'm fine with leaving as is for now and logging a Wi to return to later.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -465,14 +467,19 @@ export default class Editor {
...
                 } catch {}
             }
+        }
 
-            if (convertToDarkMode) {
-                this.convertContentToDarkMode(contentDiv, true /* skipRootElement */)();
+        // Convert content even if it hasn't changed.
+        if (convertToDarkMode) {
+            const convertFunction = this.convertContentToDarkMode(contentDiv, true /* skipRootElement */);
+            if (convertFunction) {
```
titrongt: When would this not be true? Is it when the content div has no children?

Lego6245: Correct. For instance, initial calls to set content with no child elements.

packages/roosterjs-editor-api/lib/utils/processList.ts
Lego6245: I'd rather do this in the main branch and port over. Don't want this fix to live only in dark mode.

Lego6245: Yeah, and it's complicated to do... will defer this.

packages/roosterjs-editor-api/lib/utils/processList.ts
```diff
@@ -64,7 +66,10 @@ function applyListFormat(node: Node, formats: string[], isDarkMode: boolean) {
         {
             fontFamily: formats[0],
             fontSize: formats[1],
-            textColor: formats[2],
+            textColors: {
+                lightModeColor: formats[4],
```
JiuqingSong: formats[4] can be null. Do you think we should skip textColors when formats[4] is null?

Lego6245: Actually, you highlighted a problem with this whole block here, this is definitely not the right way to go. I'm gonna refactor this real quick.

PR #251, opened 2019-03-05
Fix keyboard event issue

JiuqingSong:
In my previous check in I added stopPropagation() for keyboard event. But this should only be added for text input key event. Event such as ALT+, CTRL+, META+ should still be propagated.


Lego6245:
This all seems fine to me.

PR #264, opened 2019-03-08
7.2.3-darkmode.1

Lego6245:
Integrates a fix to dark mode that's medium priority.


JiuqingSong:
I was asked to publish a new version from master.
So let me do that then you can merge from master with latest build.

PR #272, opened 2019-03-12
Fix some perf issue and some other changes

JiuqingSong:
1. Split Format state into 4 parts:
 1.1 Pendable format  (B/I/U/Strikethrough/superscript/subscript)
 1.2 Element based format (Bullet, numbering. quote, link, image, header)
 1.3 Style based format (font name, font size, colors)
 1.4 (deprecated) Editor managed state (canundo, canredo)

2. Add functions to get format state of 1.1, 1.2, 1.3, so that user only need to get subset of the format state when need. This is to save the call to getComputedStyles()
3. Don't call getComputedStyles() when initialize editor if default format already contains everything
4. Fix dts tool to support multiple extends

### DIFF
--- a/packages/roosterjs-editor-api/lib/format/getFormatState.ts
+++ b/packages/roosterjs-editor-api/lib/format/getFormatState.ts
@@ -1,37 +1,59 @@
 import { cacheGetElementAtCursor, Editor } from 'roosterjs-editor-core';
-import { DocumentCommand, FormatState, PluginEvent, QueryScope } from 'roosterjs-editor-types';
 import { getComputedStyles, getTagOfNode, Position } from 'roosterjs-editor-dom';
+import {
+    DocumentCommand,
+    ElementBasedFormatState,
+    FormatState,
+    PendableFormatState,
+    PluginEvent,
+    QueryScope,
+    StyleBasedFormatState,
+} from 'roosterjs-editor-types';
+
+type PendableFormatNames = keyof PendableFormatState;
+
+const PendableFormatCommandMap: { [key in PendableFormatNames]: DocumentCommand } = {
+    isBold: DocumentCommand.Bold,
+    isItalic: DocumentCommand.Italic,
+    isUnderline: DocumentCommand.Underline,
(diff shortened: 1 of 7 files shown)


packages/roosterjs-editor-api/lib/format/getFormatState.ts
```diff
@@ -1,45 +1,101 @@
...
+ */
+export function getStyleBasedFormatState(editor: Editor): StyleBasedFormatState {
+    let range = editor.getSelectionRange();
+    let node = range && Position.getStart(range).normalize().node;
+    let styles = node ? getComputedStyles(node) : [];
+    return {
+        fontName: styles[0],
+        fontSize: styles[1],
+        textColor: styles[2],
+        backgroundColor: styles[3],
```
titrongt: What happens when styles is []?

tools/dts.js
```diff
@@ -8,7 +8,7 @@ var singleLineComment = /\/\/[^\n]*\n/g;
 var multiLineComment = /(^\/\*(\*(?!\/)|[^*])*\*\/\s*)/m;
 
 // 1. [export ][default |declare ](class|interface) <NAME>[ extends| implements <BASECLASS>] {...}
```
titrongt: Curious, why did the regex need to be changed?

JiuqingSong: in order to match multiple extending, such as
interface MyInterface extends Interface1, Interface2, Interface3 { ...

packages/roosterjs-editor-api/lib/format/getFormatState.ts
JiuqingSong: it can't be [], you can read the code of getComputedStyles()

titrongt: I meant if node is null (line 75)

JiuqingSong: Oh I see. In that case it will just return a bunch of "undefined" which should be expected, also the original behavior.

PR #274, opened 2019-03-20
[PickerPlugin] Remove KeyPress listener

titrongt:
Take out the KeyPress PluginEvent to improve perf

Deploy Link: https://outlook-sdf.office.com/mail/inbox?branch=TISHAPC-u-titrongt-testRoosterRemoveKeyPressFromPicker

### DIFF
--- a/packages/roosterjs-editor-core/lib/eventApi/isCharacterValue.ts
+++ b/packages/roosterjs-editor-core/lib/eventApi/isCharacterValue.ts
@@ -1,6 +1,6 @@
 // Returns true when the event was fired from a key that produces a character value, otherwise false
-// This detection is not 100% accurate. event.key is not fully supported by all brwosers, and in some browser (e.g. IE)
-// event.key is longer than 1 for num pad input. But here we just want to improve performance as mush as possible.
+// This detection is not 100% accurate. event.key is not fully supported by all browsers, and in some browsers (e.g. IE),
+// event.key is longer than 1 for num pad input. But here we just want to improve performance as much as possible.
 // So if we missed some case here it is still acceptable.
 export default function isCharacterValue(event: KeyboardEvent): boolean {
     return !event.ctrlKey && !event.altKey && !event.metaKey && event.key && event.key.length == 1;
--- a/packages/roosterjs-editor-core/lib/eventApi/isCharacterValue.ts
+++ b/packages/roosterjs-editor-core/lib/eventApi/isCharacterValue.ts
@@ -1,0 +1,7 @@
+// Returns true when the event was fired from a key that produces a character value, otherwise false
+// This detection is not 100% accurate. event.key is not fully supported by all brwosers, and in some browser (e.g. IE)
+// event.key is longer than 1 for num pad input. But here we just want to improve performance as mush as possible.
+// So if we missed some case here it is still acceptable.
+export default function isCharacterValue(event: KeyboardEvent): boolean {
+    return !event.ctrlKey && !event.altKey && !event.metaKey && event.key && event.key.length == 1;
+}
--- a/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
+++ b/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
@@ -122,11 +122,8 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
      * @param event PluginEvent object
      */
     public onPluginEvent(event: PluginEvent) {
-        if (
-            event.eventType == PluginEventType.ContentChanged &&
-        ) {
-            if (event.source == ChangeSource.SetContent &&
-                this.dataProvider.onContentChanged) {
+        if (event.eventType == PluginEventType.ContentChanged) {
+            if (event.source == ChangeSource.SetContent && this.dataProvider.onContentChanged) {
                 // Undos and other major changes to document content fire this type of event.
                 // Inform the data provider of the current picker placed elements in the body.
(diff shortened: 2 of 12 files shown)


packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -112,9 +111,7 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
     public willHandleEventExclusively(event: PluginEvent) {
         return (
             this.isSuggesting &&
-            (event.eventType == PluginEventType.KeyDown ||
-                event.eventType == PluginEventType.KeyUp ||
-                event.eventType == PluginEventType.Input)
+            (event.eventType == PluginEventType.KeyDown || event.eventType == PluginEventType.KeyUp)
```
titrongt: I removed `PluginEventType.Input` because this is the event CustomReplace needs to listen to. Looking at the old EmojiPlugin and AtMentionsPlugin, `KeyDown` and `KeyUp` seemed sufficient. @Lego6245 let me know if Input needs to be added back in.

JiuqingSong: A little concern about this.
But let me think about it.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -123,34 +120,34 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
+                            elementIds.push(element.id);
+                        }
                     }
+                );
+                this.dataProvider.onContentChanged(elementIds);
+            } else if (event.source == ChangeSource.CustomReplace) {
+                // Emoji shortcut autocomplete is handled by a separate plugin, CustomReplace
+                // When CustomReplace performs autocomplete, we can stop suggesting
+                if (this.isSuggesting) {
+                    this.setIsSuggesting(false);
```
JiuqingSong: I don't like this idea. When suggesting, all other plugins should not handle keyboard event, including input event. Here it makes this logic mess up. A plugin should not know this kind of knowledge from another plugin.

We should still let emoji plugin handle the shortcuts starts with ":" otherwise picker plugin, as a common control, has too much knowledge about emoji.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
JiuqingSong: Let's just focus on handle keypress event in this PR.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -248,8 +245,7 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
             } else {
                 this.setIsSuggesting(false);
             }
-        } else if (this.isCharacterValue) {
-            // Check for isCharacterValue to filter out modifiers like Ctrl+Z and Backspace
```
JiuqingSong: You can still check if this is a text input event. Please reference to https://github.com/Microsoft/roosterjs/blob/63a7bf45faf61c1c1a2202d7302d1fd5610cda59/packages/roosterjs-editor-core/lib/coreAPI/attachDomEvent.ts#L16

If you want you can extract it into a util and put under https://github.com/Microsoft/roosterjs/tree/master/packages/roosterjs-editor-core/lib/eventApi

JiuqingSong: Another option is to add a property to PluginKeyDownEvent and PluginKeyUpEvent since we have already checked it, no need to check again.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
titrongt: So you think CustomReplace should only handle emoji shortcuts that don't begin with ":"?

JiuqingSong: From roosterjs layer, yes.

If you want CustomReplace also handle ":" emojis, you should override it in OWA rather handling it in roosterjs. See my comment in the PR of OWA.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
titrongt: Sounds good

packages/roosterjs-editor-core/lib/eventApi/isCharacterValue.ts
```diff
@@ -0,0 +1,7 @@
+// Returns true when the event was fired from a key that produces a character value, otherwise false
+// This detection is not 100% accurate. event.key is not fully supported by all brwosers, and in some browser (e.g. IE)
```
Lego6245: nit: browsers

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -1,3 +1,4 @@
+import isCharacterValue from 'roosterjs-editor-core/lib/eventApi/isCharacterValue';
```
Lego6245: Can these be exported from the root? Does it make sense to export these from the root?

JiuqingSong: Agree

titrongt: Sounds good

PR #275, opened 2019-03-21
Re-implemented the processList API.

Lego6245:
The previous version of the process list API relied on a messy traversal of the document tree to re-apply style information.

After investigating the spec I noticed the following:

1. Edge/Trident has the best/model behavior. Elements are moved and properties are assigned to the new parent when outdenting/indenting into lists, and while in lists, the entire LI moves up and down the list tree.

2. Firefox is _mostly_ correct, but drops the attributes off the LI when outdenting out of a list.

3. Chromium is mostly correct when indenting, but when outdenting has some very bizzare behavior, including wrapping empty children in font tags (???), wrapping inline children content in spans, dropping attributes, etc.

I have opened a bug (https://bugs.chromium.org/p/chromium/issues/detail?id=944178#c_ts1553038349) against Chromium for their behavior, but in the interim we need to solve this problem. Previously, we solved it by traversing the document tree and re-assigning the style information to the LI element based on the node that Chrome created. This, while functional, ran into a number of issues, especially when we attempted to presever other style elements.

I propose a secondary API that should work better: By caching a reference to the original node, we can then replace the newly created LI when it is inserted by the browser's execComand. This will allow Chrome to better match the behaviors of Firefox and Edge. We'll end up in a state that's closer to Firefox's implementation when run against Chrome. The implementation is not perfect, it doesn't handle the outdent to the root scenario, but we didn't handle that before anyway, and the solution is harder to come to, but this should solve a large percentage of the currently reported problems with indent/outdent.

I additionally typed our input arguments to ProcessList, as I felt that allowing for any DocumentCommand was incorrect intent. This should not change any outside consumers, but may be considered breaking since the type contract is now more restrictive. RFC on if we should proceed with this, and if we do, does this require a major semver change.

### DIFF
--- a/packages/roosterjs-editor-api/lib/utils/processList.ts
+++ b/packages/roosterjs-editor-api/lib/utils/processList.ts
@@ -19,8 +19,6 @@ type ValidProcessListDocumentCommands = DocumentCommand.Outdent | DocumentComman
  * So we workaround it by always adding format to list element
  */
 export default function processList(editor: Editor, command: ValidProcessListDocumentCommands): Node {
-    let currentNode = editor.getElementAtCursor();
-
     // track the current cursor position with a dummy element.
     const currentRange = editor.getSelectionRange();
     if (currentRange) {
@@ -33,15 +31,15 @@ export default function processList(editor: Editor, command: ValidProcessListDoc
         // Chrome has some bad behavior when outdenting
         // in order to work around this, we need to take steps to deep clone the current node
         // after the outdent, we'll replace the new LI with the cloned content.
-        clonedNode = currentNode.cloneNode(true);
+        clonedNode =  editor.getElementAtCursor('LI').cloneNode(true);
         workaroundForChrome(editor);
     }
 
     let existingList = editor.getElementAtCursor('OL,UL');
     editor.getDocument().execCommand(command, false, null);
-    let newLIParent: HTMLElement;
+    let newParentNode: HTMLElement;
     editor.queryElements('.' + TEMP_NODE_CLASS, node => {
-        newLIParent = node.parentElement;
+        newParentNode = node.parentElement;
         editor.deleteNode(node);
     });
     let newList = editor.getElementAtCursor('OL,UL');
@@ -49,11 +47,11 @@ export default function processList(editor: Editor, command: ValidProcessListDoc
         newList = null;
     }
 
-    if (newList && clonedNode && newLIParent) {
+    if (newList && clonedNode && newParentNode) {
         // if the clonedNode and the newLIParent share the same tag name
         // we can 1:1 swap them
-        if ((clonedNode instanceof HTMLElement) && clonedNode.tagName == newLIParent.tagName) {
-            newList.replaceChild(clonedNode, newLIParent);
+        if ((clonedNode instanceof HTMLElement) && clonedNode.tagName == newParentNode.tagName) {
+            newList.replaceChild(clonedNode, newParentNode);
         }
         // The alternative case is harder to solve, but we didn't specifically handle this before either.
     }
--- a/packages/roosterjs-editor-api/lib/utils/processList.ts
+++ b/packages/roosterjs-editor-api/lib/utils/processList.ts
@@ -5,12 +5,14 @@ import {
     isVoidHtmlElement,
     isBlockElement,
     Browser,
+    getSelectionPath,
+    getRangeFromSelectionPath,
 } from 'roosterjs-editor-dom';
 
 const TEMP_NODE_CLASS = 'ROOSTERJS_TEMP_NODE_FOR_LIST';
 const TEMP_NODE_HTML = "<img class=\"" + TEMP_NODE_CLASS + "\">";
-const SELECTION_NODE_CLASS = 'ROOSTERJS_SELECTION_NODE';
-const SELECTION_NODE_HTML = "<img class=\"" + SELECTION_NODE_CLASS + "\">";
+//const SELECTION_NODE_CLASS = 'ROOSTERJS_SELECTION_NODE';
+//const SELECTION_NODE_HTML = "<img class=\"" + SELECTION_NODE_CLASS + "\">";
 
 type ValidProcessListDocumentCommands = DocumentCommand.Outdent | DocumentCommand.Indent | DocumentCommand.InsertOrderedList | DocumentCommand.InsertUnorderedList;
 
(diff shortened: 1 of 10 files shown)


packages/roosterjs-editor-api/lib/utils/processList.ts
```diff
@@ -5,65 +5,65 @@ import {
...
+type ValidProcessListDocumentCommands = DocumentCommand.Outdent | DocumentCommand.Indent | DocumentCommand.InsertOrderedList | DocumentCommand.InsertUnorderedList;
 
 /**
  * Browsers don't handle bullet/numbering list well, especially the formats when switching list statue
  * So we workaround it by always adding format to list element
  */
-export default function processList(editor: Editor, command: DocumentCommand): Node {
-    if (Browser.isChrome) {
+export default function processList(editor: Editor, command: ValidProcessListDocumentCommands): Node {
+    let currentNode = editor.getElementAtCursor();
```
JiuqingSong: Should we get current LI node instead?

packages/roosterjs-editor-api/lib/utils/processList.ts
```diff
@@ -5,65 +5,65 @@ import {
...
-                let formatNode = getFirstLeafNode(li);
-                if (formatNode) {
-                    applyListFormat(li, getComputedStyles(formatNode));
-                }
-            }
+    if (newList && clonedNode && newLIParent) {
+        // if the clonedNode and the newLIParent share the same tag name
+        // we can 1:1 swap them
+        if ((clonedNode instanceof HTMLElement) && clonedNode.tagName == newLIParent.tagName) {
+            newList.replaceChild(clonedNode, newLIParent);
```
JiuqingSong: Can we preserve all child nodes under the LI nodes rather than use cloned nodes? Because plugin may hold reference to node under LI.

Lego6245: Unfortunately, I'm not sure we can. Looking at how chrome behaves, it seems that the node references get broken when re-parenting, even on master. Maybe I'm missing something, but this scenario may be broken currently, not just by the new change I'm proposing.

packages/roosterjs-editor-api/lib/utils/processList.ts
Lego6245: Yeah, that makes sense especially for complex structures under LI nodes.

packages/roosterjs-editor-api/lib/utils/processList.ts
```diff
@@ -5,76 +5,78 @@ import {
...
-    getComputedStyles,
-    applyFormat,
-    getFirstLeafNode,
+    getSelectionPath,
+    getRangeFromSelectionPath,
 } from 'roosterjs-editor-dom';
 
 const TEMP_NODE_CLASS = 'ROOSTERJS_TEMP_NODE_FOR_LIST';
-const TEMP_NODE_HTML = `<img class="${TEMP_NODE_CLASS}">`;
+const TEMP_NODE_HTML = "<img class=\"" + TEMP_NODE_CLASS + "\">";
```
JiuqingSong: Just curious, what's the difference between `` and ""?

packages/roosterjs-editor-api/lib/utils/processList.ts
```diff
@@ -5,76 +5,78 @@ import {
...
+//const SELECTION_NODE_CLASS = 'ROOSTERJS_SELECTION_NODE';
+//const SELECTION_NODE_HTML = "<img class=\"" + SELECTION_NODE_CLASS + "\">";
+
+type ValidProcessListDocumentCommands = DocumentCommand.Outdent | DocumentCommand.Indent | DocumentCommand.InsertOrderedList | DocumentCommand.InsertUnorderedList;
```
JiuqingSong: Suggest run prettier to fix this long line

packages/roosterjs-editor-api/lib/utils/processList.ts
```diff
@@ -5,76 +5,78 @@ import {
...
-export default function processList(editor: Editor, command: DocumentCommand): Node {
-    if (Browser.isChrome) {
+export default function processList(editor: Editor, command: ValidProcessListDocumentCommands): Node {
+    let clonedNode: Node;
+    let relativeSelectionPath;
+    if (Browser.isChrome && command == DocumentCommand.Outdent) {
+        const parentLINode =  editor.getElementAtCursor('LI');
+        if (parentLINode) {
+            let currentRange = editor.getSelectionRange();
+            relativeSelectionPath = getSelectionPath(parentLINode, currentRange);
```
JiuqingSong: Please do verify if currentRange is not collapsed and end position is not under parenLINode. e.g.
[LI]test{SelectionStart}test[/LI]
[LI]test{SelectionEnd}test[/LI]

Because in that case end position is out of the relative node (parentLiNode), not sure if it still work in that case

packages/roosterjs-editor-api/lib/utils/processList.ts
```diff
@@ -5,76 +5,78 @@ import {
...
     }
 
-    let currentNode = editor.getElementAtCursor();
-    let currentFormat = getComputedStyles(currentNode);
     let existingList = editor.getElementAtCursor('OL,UL');
     editor.getDocument().execCommand(command, false, null);
-    editor.queryElements('.' + TEMP_NODE_CLASS, node => editor.deleteNode(node));
+    let newParentNode: HTMLElement;
+    editor.queryElements('.' + TEMP_NODE_CLASS, node => {
+        newParentNode = node.parentElement;
```
JiuqingSong: Suggest use parentNode.

packages/roosterjs-editor-api/lib/utils/processList.ts
Lego6245: Template strings present problems on IE11. They cause javascript exceptions, so I think it's safer we remove them (unless we no longer target IE11)

packages/roosterjs-editor-api/lib/utils/processList.ts
Lego6245: Good point. Let me test a few scenarios with that.

packages/roosterjs-editor-api/lib/utils/processList.ts
Lego6245: Perticular reason for this? Or just naming.

packages/roosterjs-editor-api/lib/utils/processList.ts
Lego6245: Yeah, multiple outdents are midly busted, but I don't think they're any more busted than before. So, I'll check to make sure the range is collapsed before continuing the process.

packages/roosterjs-editor-api/lib/utils/processList.ts
JiuqingSong: In IE11, parentElement only works when current node is not Text node. Here it should be ok, but I'd like to always use parentNode for safety.

Lego6245: I see. Since this code will only run for chrome I'd prefer keeping it here to prevent messy typecasting situations, and to be more confident in the behavior we'd be executing.

PR #280, opened 2019-04-03
[PickerPlugin] Set ariaOwns and ariaActiveDescendant attributes

titrongt:
Deploy Link: https://outlook-sdf.office.com/mail/?branch=TISHAPC-u-titrongt-testRoosterNarratePickerSuggestions

### DIFF
--- a/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
+++ b/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
@@ -169,6 +169,9 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
             this.setLastKnownRange(null);
         }
         this.dataProvider.onIsSuggestingChanged(isSuggesting);
+
+        this.setAriaOwns(isSuggesting);
+        this.setAriaActiveDescendant(isSuggesting ? 0 : undefined);
     }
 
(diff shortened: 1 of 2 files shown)


packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -398,4 +406,22 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
         }
         return false;
     }
+
+    private setAriaOwns(isSuggesting: boolean) {
```
Lego6245: Do we have confidence that there will only be one of these set at a time? (I assume so, else things will get super messy).

Do we need to clean this value (and others) after a picker dismisses?

JiuqingSong: Agree.

titrongt: The aria attributes get set when a picker is open, and since only one picker can be opened at a time, only one should be set at a time. 

When a picker is dismissed, setIsSuggesting is triggered with isSuggesting as false, so ariaOwns and ariaActiveDescendant are set as undefined.

Let me change from undefined --> null since setEditorDomAttribute will fully remove the attribute when the value is null  👌

PR #284, opened 2019-04-08
Add dark mode publish variant

Lego6245:
Previously publishing dark mode was a manual process. I've made it not manuall, so now you can call npm run publish:dark to publish to the dark mode tag.

### DIFF
--- a/tools/build.js
+++ b/tools/build.js
@@ -438,13 +438,13 @@ function err(message) {
     throw ex;
 }
 
-function publish(isDark) {
+function publish(tagname) {
     packages.forEach(package => {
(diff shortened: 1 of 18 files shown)


tools/build.js
```diff
@@ -578,6 +581,11 @@ function buildAll(options) {
             callback: publish,
             enabled: options.publish,
         },
+        {
+            message: 'Publishing dark mode variants to npm...',
+            callback: () => publish(true),
```
JiuqingSong: Instead of passing true here, how about pass the tag name to make the publish function more reusable?

PR #289, opened 2019-04-21
Support export code to codepen from demo site

JiuqingSong:
Add a button "export to codepen" in demo site, so user can change options and use this button to easily try the code in codepen.

### DIFF
--- a/publish/samplesite/scripts/controls/sidePane/editorOptions/codes/PickerPluginCode.ts
+++ b/publish/samplesite/scripts/controls/sidePane/editorOptions/codes/PickerPluginCode.ts
@@ -5,24 +5,15 @@ export default class PickerPluginCode extends CodeElement {
         super();
     }
 
-    getImports() {
-        return [
-            {
-                name: 'PickerPlugin',
-                path: 'roosterjs-picker-plugin',
-                isDefault: false,
-            },
-            {
-                name: 'YourPickerPluginDataProvider',
-                path: 'your-picker-plugin-data-provider',
-                isDefault: false,
-            },
-        ];
-    }
-
     getCode() {
         return (
-            'new PickerPlugin(new YourPickerPluginDataProvider(), {\n' +
+            'new roosterjs.PickerPlugin({\n' +
+            '    onInitalize: (insertNodeCallback, setIsSuggestingCallback) => {},\n' +
+            '    onDispose: () => {},\n' +
+            '    onIsSuggestingChanged: (isSuggesting) => {},\n' +
+            '    queryStringUpdated: (queryString) => {},\n' +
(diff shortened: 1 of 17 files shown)


titrongt:
Do you have a deploy link?

JiuqingSong:
> Do you have a deploy link?

No, we don't support deploy for PR now. You can pull the branch and build locally

publish/samplesite/scripts/controls/sidePane/editorOptions/codes/PickerPluginCode.ts
```diff
@@ -5,24 +5,15 @@ export default class PickerPluginCode extends CodeElement {
...
     getCode() {
         return (
-            'new PickerPlugin(new YourPickerPluginDataProvider(), {\n' +
+            'new roosterjs.PickerPlugin({\n' +
+            '    onInitalize: (insertNodeCallback, setIsSuggestingCallback) => {},\n' +
+            '    onDispose: () => {},\n' +
+            '    onIsSuggestingChanged: (isSuggesting) => {},\n' +
+            '    queryStringUpdated: (queryString) => {},\n' +
+            '    onRemove: (nodeRemoved, isBackwards) => null,\n' +
+            '    }, {\n' +
```
Adjective-Object: our import style is changing from separate packages to using a single global roosterjs object.

I'm assuming that's because we're getting roosterjs as a global in codepen.

Are we just saying that user that are working in a real (e.g. webpack'd) environment will have to go to the docs instead of using this global example?

I'm worried we might be encouraging people to use rooster as a single import instead of the separate micropackages.

JiuqingSong: When I wrote the code, I also had this concern. But then I was thinking about 2 things:
1. Some people will use our npm packages, and some will use rooster.js file directly.
2. Should we always encourage people to import from separate packages?

The reason why we have separate packages at beginning is to manage code size and dependencies. But forcing people always import from sub packages introduces extra complexity. So if people want to always import from the 'roosterjs' package, I think that's ok.

On the other hand, is it possible to make it work in codepen by importing from packages?

JiuqingSong: more comments?

Adjective-Object: I agree that using rooster as a single dependency is probably the easiest way to bootstrap a project. If we're okay with users having to migrate over their import structure if they start getting bundlesize problems, then I say go ahead and merge as-is.

Not sure if there's a reasonable way to include as separate dependencies in codepen (I'm actually not sure how the dependencies end up exposed in the codepens as they stand now).

PR #294, opened 2019-04-29
7.3.1

JiuqingSong:
Bump version to 7.3.1


Lego6245:
Do we want to hold for the double link issue, or does this fix include it?

JiuqingSong:
> Do we want to hold for the double link issue, or does this fix include it?

It is included in this PR: https://github.com/Microsoft/roosterjs/issues/286

PR #296, opened 2019-04-29
7.3.2-darkmode.0

JiuqingSong:


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "roosterjs",
-    "version": "7.3.0-darkmode.0",
+    "version": "7.3.2-darkmode.0",
```
Lego6245: 7.3.2 darkmode? just double checking.

JiuqingSong: Yes. I made an emoji fix after 7.3.1, so this is 7.3.2

PR #297, opened 2019-04-29
Fix an issue when call onKeyUp handler in PickerPlugin

JiuqingSong:
A recent change makes onKeyUp handler will be called when functional keys like Backspace, ... are pressed. This causes isSuggesting is set to true even when it was not suggesting.

In this change, I make it check isCharacterValue as well:

isSuggesting && isCharacterValue => call onKeyUp
isSuggesting && !isCharacterValue && !isModifierKey => call onKeyUp
isSuggesting && isModifierKey => NOT call onKeyUp

!isSuggesting && isCharacterValue => call onKeyUp
!isSuggesting && !isCharacterValue && !isModifierKey => NOT call onKeyUp
!isSuggesting && isModifierKey => NOT call onKeyUp

### DIFF
--- a/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
+++ b/packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
@@ -1,8 +1,13 @@
 import { Browser, createRange, PartialInlineElement } from 'roosterjs-editor-dom';
-import { cacheGetContentSearcher, Editor, EditorPlugin } from 'roosterjs-editor-core';
-import { isModifierKey } from 'roosterjs-editor-core';
 import { PickerDataProvider, PickerPluginOptions } from './PickerDataProvider';
 import { replaceWithNode } from 'roosterjs-editor-api';
+import {
+    cacheGetContentSearcher,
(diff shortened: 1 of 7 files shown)


packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -260,8 +275,12 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
                     wordBeforeCursor[0] == this.pickerOptions.triggerCharacter
                 ) {
                     this.setIsSuggesting(true);
-                    let shortWord = wordBeforeCursor.substring(1).trim();
-                    this.dataProvider.queryStringUpdated(shortWord);
+                    const wordBeforeCursorWithoutTriggerChar = wordBeforeCursor.substring(1);
+                    let trimmedWordBeforeCursor = wordBeforeCursorWithoutTriggerChar.trim();
```
titrongt: Wish these two lines didn't have to be repeated, but code looks good to me otherwise

PR #298, opened 2019-05-09
[Bullets] Tab should resolve AtMention instead of indenting bullet

titrongt:
Adjust the order of how the plugins are added so that the PickerPlugin's `willHandleExclusively` takes precedence over EditPlugin. 

Deploy Link: https://outlook-sdf.office.com/mail/?branch=TISHAPC-u-titrongt-testTabShouldResolveAtMention

### DIFF
--- a/packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
+++ b/packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
@@ -32,10 +32,10 @@ export default function createEditorCore(
         firefoxTypeAfterLink: Browser.isFirefox && new FirefoxTypeAfterLink(),
     };
     let allPlugins: EditorPlugin[] = [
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
```diff
@@ -32,10 +32,10 @@ export default function createEditorCore(
         firefoxTypeAfterLink: Browser.isFirefox && new FirefoxTypeAfterLink(),
     };
     let allPlugins: EditorPlugin[] = [
+        ...(options.plugins || []),
         corePlugins.typeInContainer,
         corePlugins.edit,
         corePlugins.mouseUp,
-        ...(options.plugins || []),
```
JiuqingSong: I suggest we only move the order of EditPlugin, and keep others.

PR #306, opened 2019-06-03
Add getTextContent

FrancisMengx:
**Cause:** We used to use ```this.core.contentDiv.innerText```.  
Last ```<br>``` will have a extra ```\n```.
and when you switch back the extra ```\n``` will add an extra ```<br>```

**Fix:** Get text content for each line and only add ```\n``` when it's not the last line.


packages/roosterjs-editor-dom/lib/utils/getBlockElementTextContent.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * get block element's text content.
+ * @param rootNode Root node that the get the textContent of.
+ * @param startNode Node the node start traverser to start traversing.
+ * @returns text content of given text content.
+ */
+export default function getBlockElementTextContent(rootNode: Node, startNode?: Node): string {
+    const traverser = ContentTraverser.createBodyTraverser(rootNode, startNode);
+    let block = traverser && traverser.currentBlockElement;
+    let blocks: BlockElement[] = [];
```
JiuqingSong: No need to save blocks, you just need the texts.

packages/roosterjs-editor-dom/lib/utils/getBlockElementTextContent.ts
```diff
@@ -0,0 +1,40 @@
...
+        }
+        if (index != blocks.length - 1) {
+            textContent = `${textContent}\n`
+        }
+    });
+
+    return textContent;
+}
+
+function getBlockTextContent(block: BlockElement) {
```
JiuqingSong: There used to be a method of BlockElement: getTextContent(). But then I removed it since it was not used. Now that we need it, you can add it to BlockElement (and its child classes) then we don't need to add it here

packages/roosterjs-editor-dom/lib/utils/getBlockElementTextContent.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param startNode Node the node start traverser to start traversing.
+ * @returns text content of given text content.
+ */
+export default function getBlockElementTextContent(rootNode: Node, startNode?: Node): string {
+    const traverser = ContentTraverser.createBodyTraverser(rootNode, startNode);
+    let block = traverser && traverser.currentBlockElement;
+    let blocks: BlockElement[] = [];
+    let textContent: string = '';
+
+    while (block) {
```
JiuqingSong: Suggest:
let texts: string[] = [];

while (...) {
  texts.push(block.getTextContent());
  ...
}

return texts.join('\n');

packages/roosterjs-editor-dom/lib/utils/getTextContent.ts
```diff
@@ -0,0 +1,20 @@
+import ContentTraverser from '../contentTraverser/ContentTraverser';
+
+/**
+ * get block element's text content.
+ * @param rootNode Root node that the get the textContent of.
+ * @param startNode Node the node start traverser to start traversing.
```
JiuqingSong: Remove the comment

PR #312, opened 2019-06-13
Remove initialize method from edit feature

JiuqingSong:
The regression in 7.4.0 is caused by ContentEditFeature.initialize(). I realized this function can potentially break editor since when call it, EditPlugin.initialize() may not been called yet so EditPlugin.editor is null.

So I decide  to remove this method. And actually it is only used in one place -- AutoLinkPlugin, to run a command for IE. So I also move all initial "execCommand" call to a new function "adjustBrowserBeahvior"

Also move the event mapping part to a new file to make Editor.ts a little shorter.


Lego6245:
All seems good to me :D

Issue #313, opened 2019-06-16
Pending format loses when type in empty editor


JiuqingSong:
1. Open editor,
2. Backspace, in order to clear any existing html
3. Turn on Bold
4. Type

Result: still type in non-bold

This is because html content is changed when type (TypeInContainerPlugin), so pending format state is lost.

PR #314, opened 2019-06-16
Support pending format management, fix #224, #313

### LINKED ISSUE #224: Save  temp BIU state when blur, restore when focus if position is not changed
when toggle B/I/U:
if (isCollapsed) {
  save current state, and position path
}

when keydown || mousedown || (contentChanged && source != Format) {
 clear state
}

when focus {
  if (has saved state && position is not changed && current state != saved) {
    restore current state;
  }
}

Then we don't need to workaround of BIU state in RIBBON

JiuqingSong:
styles like bold, italic, underline, ... are pending style if selection is empty (collapsed), which means if selection is changed before user type something (apply the change), the pending style will be lost. This causes user can't set pending style at the same time with other formats like font and color since those styles needs open a picker which will cause editor loses focus.

The fix here is:
1. Add a new event "PendingFormatStateChangedEvent", and it is fired when user is change such styles
2. Handle the event in DomPlugin and cache current position and format
3. When editor got focused, or we call editor.select() to change selection, check if the selection is still the same with the cached one. If so, apply the cached format

### DIFF
--- a/packages/roosterjs-editor-api/lib/index.ts
+++ b/packages/roosterjs-editor-api/lib/index.ts
@@ -9,7 +9,6 @@ export { default as clearFormat } from './format/clearFormat';
 export { default as createLink } from './format/createLink';
 export {
     default as getFormatState,
-    getPendableFormatState,
     getElementBasedFormatState,
     getStyleBasedFormatState,
 } from './format/getFormatState';
@@ -38,3 +37,7 @@ export { default as toggleSubscript } from './format/toggleSubscript';
 export { default as toggleSuperscript } from './format/toggleSuperscript';
 export { default as toggleUnderline } from './format/toggleUnderline';
 export { default as toggleHeader } from './format/toggleHeader';
+
+// @deprecated the function getPendableFormatState will still be available from
+// roosterjs-editor-dom package, keep export it here just for compatibility
+export { getPendableFormatState } from 'roosterjs-editor-dom';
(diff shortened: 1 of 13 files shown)


packages/roosterjs-editor-api/lib/index.ts
```diff
@@ -38,3 +37,7 @@ export { default as toggleSubscript } from './format/toggleSubscript';
 export { default as toggleSuperscript } from './format/toggleSuperscript';
 export { default as toggleUnderline } from './format/toggleUnderline';
 export { default as toggleHeader } from './format/toggleHeader';
+
+// @deprecated the function getPendableFormatState will still be available from
+// roosterjs-editor-dom package, keep export it here just for compatibility
+export { getPendableFormatState } from 'roosterjs-editor-dom';
```
Lego6245: Does this mean that roosterjs-editor-api now has a hard dependency on roosterjs-editor-dom?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -105,10 +105,10 @@ export default class Editor {
             if (Browser.isFirefox) {
                 this.core.document.execCommand(DocumentCommand.EnableObjectResizing, false, <
                     string
-                    >(<any>false));
+                >(<any>false));
```
Lego6245: Does this get clobbered with #312?

JiuqingSong: I need to resolve

PR #316, opened 2019-07-09
Merge latest master to dark mode

JiuqingSong:


packages/roosterjs-editor-core/lib/editor/adjustBrowserBehavior.ts
```diff
@@ -0,0 +1,38 @@
...
+          /**
+           * Disable these object resizing for firefox since other browsers don't have these behaviors
+           */
+          [DocumentCommand.EnableObjectResizing]: false,
+          [DocumentCommand.EnableInlineTableEditing]: false,
+      }
+    : Browser.isIE
+    ? {
+          /**
+           * Change the default paragraph separater to DIV. This is mainly for IE since its default setting is P
```
titrongt: Should this only be done in IE? The comment implies it can be done in other browsers as well

publish/samplesite/scripts/controls/sidePane/eventViewer/EventViewPane.tsx
```diff
@@ -147,6 +148,10 @@ export default class EventViewPane extends React.Component<
                         ))}
                     </span>
                 );
+            case PluginEventType.PendingFormatStateChanged:
+                const formatState = event.formatState;
+                const keys = Object.keys(formatState) as (keyof PendableFormatState)[];
+                return <span>{keys.map(key => `${key}=${event.formatState[key]}; `)}</span>;
```
titrongt: Maybe I'm confused by this syntax, but does this mean a single span could have multiple keys?

JiuqingSong: yes, it shows multiple keys in one span. This is for the sample site only.

packages/roosterjs-editor-core/lib/editor/adjustBrowserBehavior.ts
JiuqingSong: hmm, the comment was moved from editor.ts. It means this is only for IE.
Since this is just a code integration from master, I prefer change it. We can refine it later in master branch.

PR #318, opened 2019-07-12
Fix for copying from dark mode

Lego6245:
I created a new plugin that intercepts the copy process so when copying from dark editors, you get the normalized content.

### DIFF
--- a/packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
@@ -1,3 +1,4 @@
+import getColorNormalizedContent from '../darkMode/getColorNormalizedContent';
 import { Editor, EditorPlugin } from 'roosterjs-editor-core';
 
 /**
@@ -36,16 +37,23 @@ export default class CopyPlugin implements EditorPlugin {
         // if it's dark mode...
         if (this.editor && this.editor.isDarkMode()) {
             // get whatever the current selection range is
-            const copyFragment = this.editor.getSelection().getRangeAt(0).cloneContents();
+            const selectionRange = this.editor.getSelectionRange();
+            if (selectionRange && !selectionRange.collapsed) {
+                const clipboardEvent = (event as ClipboardEvent);
+                const copyFragment = this.editor.getSelectionRange().cloneContents();
 
-            // revert just this selected range to light mode colors
-            const normalizerDiv = document.createElement('div');
-            normalizerDiv.appendChild(copyFragment);
-            const normalizedFragment = this.editor.getColorNormalizedContent(normalizerDiv.innerHTML);
+                // revert just this selected range to light mode colors
+                const normalizedContent = getColorNormalizedContent(copyFragment);
+                const containerDiv = this.editor.getDocument().createElement('div');
+                containerDiv.setAttribute('contenteditable', 'true');
+                containerDiv.innerHTML = normalizedContent;
 
-            // put it on the clipboard
-            (event as ClipboardEvent).clipboardData.setData('text/html', normalizedFragment);
-            event.preventDefault();
+                // put it on the clipboard
+                clipboardEvent.clipboardData.setData('text/html', normalizedContent);
+                clipboardEvent.clipboardData.setData('text/plain', containerDiv.innerText)
+
+                event.preventDefault();
+            }
         }
     }
 }
--- a/packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
@@ -1,0 +1,51 @@
+import { Editor, EditorPlugin } from 'roosterjs-editor-core';
+
+/**
+ * Paste plugin, handles onPaste event and paste content into editor
+ */
+export default class CopyPlugin implements EditorPlugin {
+    private editor: Editor;
+    private copyDisposer: () => void;
+
+    /**
+     * Get a friendly name of  this plugin
+     */
+    getName() {
+        return 'Copy';
+    }
+
+    /**
+     * Initialize this plugin. This should only be called from Editor
+     * @param editor Editor instance
+     */
+    public initialize(editor: Editor) {
+        this.editor = editor;
+        this.copyDisposer = editor.addDomEventHandler('copy', this.onCopy);
+    }
+
+    /**
+     * Dispose this plugin
+     */
+    public dispose() {
+        this.copyDisposer();
+        this.copyDisposer = null;
+        this.editor = null;
+    }
+
+    private onCopy = (event: Event) => {
+        // if it's dark mode...
+        if (this.editor && this.editor.isDarkMode()) {
+            // get whatever the current selection range is
+            const copyFragment = this.editor.getSelection().getRangeAt(0).cloneContents();
+
+            // revert just this selected range to light mode colors
+            const normalizerDiv = document.createElement('div');
+            normalizerDiv.appendChild(copyFragment);
+            const normalizedFragment = this.editor.getColorNormalizedContent(normalizerDiv.innerHTML);
+
+            // put it on the clipboard
+            (event as ClipboardEvent).clipboardData.setData('text/html', normalizedFragment);
+            event.preventDefault();
+        }
+    }
+}
--- a/packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
@@ -1,7 +1,7 @@
 import { Editor, EditorPlugin } from 'roosterjs-editor-core';
 
 /**
- * Paste plugin, handles onPaste event and paste content into editor
+ * Copy plugin, hijacks copy events to normalize the content to the clipboard.
  */
 export default class CopyPlugin implements EditorPlugin {
     private editor: Editor;
--- a/packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
@@ -45,6 +45,8 @@ export default class CopyPlugin implements EditorPlugin {
                 // revert just this selected range to light mode colors
                 const normalizedContent = getColorNormalizedContent(copyFragment);
                 const containerDiv = this.editor.getDocument().createElement('div');
+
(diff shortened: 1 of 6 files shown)


packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
```diff
@@ -0,0 +1,51 @@
...
+        this.copyDisposer();
+        this.copyDisposer = null;
+        this.editor = null;
+    }
+
+    private onCopy = (event: Event) => {
+        // if it's dark mode...
+        if (this.editor && this.editor.isDarkMode()) {
+            // get whatever the current selection range is
+            const copyFragment = this.editor.getSelection().getRangeAt(0).cloneContents();
```
JiuqingSong: getSelection().getRangeAt(0)  ====> getSelectionRange()

JiuqingSong: null and collapsed check?

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
```diff
@@ -0,0 +1,51 @@
...
+
+            // revert just this selected range to light mode colors
+            const normalizerDiv = document.createElement('div');
```
JiuqingSong: document ===> this.editor.getDocument()

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
```diff
@@ -0,0 +1,51 @@
...
+            normalizerDiv.appendChild(copyFragment);
+            const normalizedFragment = this.editor.getColorNormalizedContent(normalizerDiv.innerHTML);
+
```
JiuqingSong: Suggest trigger a editor copy event so that other plugins can modify the copied content as well. (low pri for now)

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
```diff
@@ -0,0 +1,51 @@
...
+    private onCopy = (event: Event) => {
+        // if it's dark mode...
+        if (this.editor && this.editor.isDarkMode()) {
+            // get whatever the current selection range is
+            const copyFragment = this.editor.getSelection().getRangeAt(0).cloneContents();
+
+            // revert just this selected range to light mode colors
+            const normalizerDiv = document.createElement('div');
+            normalizerDiv.appendChild(copyFragment);
+            const normalizedFragment = this.editor.getColorNormalizedContent(normalizerDiv.innerHTML);
```
JiuqingSong: Should we only do  this when editor is in dark mode?

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    private onCopy = (event: Event) => {
+        // if it's dark mode...
+        if (this.editor && this.editor.isDarkMode()) {
+            // get whatever the current selection range is
+            const copyFragment = this.editor.getSelection().getRangeAt(0).cloneContents();
+
+            // revert just this selected range to light mode colors
+            const normalizerDiv = document.createElement('div');
+            normalizerDiv.appendChild(copyFragment);
```
JiuqingSong: a little concern about XSS, since editor content can contain SCRIPT tag but it won't execute in CED, when we append it to a regular DOM, I'm afraid it can be executed.

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
```diff
@@ -0,0 +1,51 @@
...
+            const normalizedFragment = this.editor.getColorNormalizedContent(normalizerDiv.innerHTML);
+
+            // put it on the clipboard
+            (event as ClipboardEvent).clipboardData.setData('text/html', normalizedFragment);
+            event.preventDefault();
```
JiuqingSong: will this prevent setting plain text clipboard data and/or image?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -374,7 +377,7 @@ export default class Editor {
         return content;
     }
 
-    private getColorNormalizedContent(content: string): string {
+    public getColorNormalizedContent(content: string): string {
```
JiuqingSong: move this function and isDataAttributeSettable() out to a util function since they don't rely on editor

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
Lego6245: See the check earlier in the file for the .isDarkMode check.

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
Lego6245: You're right, with prevent default the only thing on there would be the HTML fragment. We need to think more about how to support images and other clipboard data. You can't just change one of the data lines.

Lego6245: I did some inspecting and research which confirmed my suspicions:

1) Right click -> Copy image. Doesn't fire an event, at least in my limited testing. We don't hijack it here.
2) Select image, copy -> Will by hijacked, but doesn't matter since it's just a document fragment.
3) Select multiple elements, copy -> Will be hijacked, but doesn't matter since it's a document fragment.

I'll add something that sets plain text.

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
Lego6245: I thought about this a lot. I think we're fairly well protected overall, but any additional protection that can be leveraged, we should use. I'm appending contenteditable to the created divs to take advantage of the browser security policy around scripts in there.

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
```diff
@@ -0,0 +1,59 @@
...
+            // get whatever the current selection range is
+            const selectionRange = this.editor.getSelectionRange();
+            if (selectionRange && !selectionRange.collapsed) {
+                const clipboardEvent = (event as ClipboardEvent);
+                const copyFragment = this.editor.getSelectionRange().cloneContents();
+
+                // revert just this selected range to light mode colors
+                const normalizedContent = getColorNormalizedContent(copyFragment);
+                const containerDiv = this.editor.getDocument().createElement('div');
+                containerDiv.setAttribute('contenteditable', 'true');
```
JiuqingSong: add comment to explain why we need to set contenteditable to true here.

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
```diff
@@ -0,0 +1,59 @@
...
+        this.editor = null;
+    }
+
+    private onCopy = (event: Event) => {
+        // if it's dark mode...
+        if (this.editor && this.editor.isDarkMode()) {
+            // get whatever the current selection range is
+            const selectionRange = this.editor.getSelectionRange();
+            if (selectionRange && !selectionRange.collapsed) {
+                const clipboardEvent = (event as ClipboardEvent);
```
JiuqingSong: Suggest adding a check see if the copied content is just a single image, then skip everything below and let browser handle it.

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
```diff
@@ -0,0 +1,59 @@
...
+
+                // revert just this selected range to light mode colors
+                const normalizedContent = getColorNormalizedContent(copyFragment);
+                const containerDiv = this.editor.getDocument().createElement('div');
+                containerDiv.setAttribute('contenteditable', 'true');
+                containerDiv.innerHTML = normalizedContent;
+
+                // put it on the clipboard
+                clipboardEvent.clipboardData.setData('text/html', normalizedContent);
+                clipboardEvent.clipboardData.setData('text/plain', containerDiv.innerText)
```
JiuqingSong: question: does this work for IE? If not, need to add a check. (or we totally disable dark mode for IE)

packages/roosterjs-editor-core/lib/darkMode/getColorNormalizedContent.ts
```diff
@@ -0,0 +1,58 @@
+export default function getColorNormalizedContent(content: string | DocumentFragment): string {
```
JiuqingSong: comment, please.

packages/roosterjs-editor-core/lib/darkMode/getColorNormalizedContent.ts
```diff
@@ -0,0 +1,58 @@
...
+    let el = document.createElement('div');
+    el.setAttribute('contenteditable', 'true');
```
JiuqingSong: comment to explain contentEditable

JiuqingSong: Or use DOMParser instead which is just a sandbox and won't execute any script

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
Lego6245: We actually can't, unfortunately. copy events don't allow you access to what's on the clipboard, only paste. However, if the copied content is just the image (right click copy image) then it'll skip this block anyway.

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
Lego6245: No, this will not work in IE. I'll add a check to adding the plugin for IE.

packages/roosterjs-editor-core/lib/darkMode/getColorNormalizedContent.ts
Lego6245: DOMParser requires a string of HTML content, and the only way to get that from a DocumentFragment is to add it to an element, which doesn't help us.

JiuqingSong: ok, got it.

PR #319, opened 2019-07-12
Fix for dark mode undo

Lego6245:
I fixed dark mode undo so the undo buffer always has normalized content, and when inserted back into the editor, it is normalized appropriately.

### DIFF
--- a/packages/roosterjs-editor-core/lib/undo/Undo.ts
+++ b/packages/roosterjs-editor-core/lib/undo/Undo.ts
@@ -134,8 +134,7 @@ export default class Undo implements UndoService {
     public addUndoSnapshot(): string {
         let snapshot = this.editor.getContent(
             false /*triggerExtractContentEvent*/,
-            true /*markSelection*/,
-            !this.editor.isDarkMode() /*normalizeColor*/
+            true /* includeSelectionmarker */
         );
         this.getSnapshotsManager().addSnapshot(snapshot);
         this.hasNewContent = false;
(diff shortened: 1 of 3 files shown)


packages/roosterjs-editor-core/lib/undo/Undo.ts
```diff
@@ -155,7 +154,11 @@ export default class Undo implements UndoService {
         if (snapshot != null) {
             try {
                 this.isRestoring = true;
-                this.editor.setContent(snapshot);
+                this.editor.setContent(
+                    snapshot,
+                    undefined /* triggerContentChangedEvent */,
+                    this.editor.isDarkMode()
```
JiuqingSong: Is there any place where we get content without normalize the color? If not, I think we can remove the dark mode parameter for get/setContent since we always get normalized color and we know whether editor is in dark mode when set.

Lego6245: Currently, no. But I was thinking for public consumption, it'd be a nice to be able to get the un-normalized content.

JiuqingSong: For getContent(), we may provide the ability to get raw HTML. But I still feel we don't need to provide it for now, since it will be always easier adding a new parameter than removing an existing one which will cause a breaking change.

For setContent(), providing this parameter means we rely on its value to determine whether we should change the color. I suggest not rely on the parameter. And if we want to provide this ability, we can add meta data using html comment inside the content like we did for selection in undo snapshot.

Currently I can approve this PR, but suggest making this change before we integrate dark mode back to master.

Lego6245: I see your reasoning. This makes sense. Let me remove the properties for now.

packages/roosterjs-editor-core/lib/undo/Undo.ts
```diff
@@ -155,7 +154,10 @@ export default class Undo implements UndoService {
         if (snapshot != null) {
             try {
                 this.isRestoring = true;
-                this.editor.setContent(snapshot);
+                this.editor.setContent(
+                    snapshot,
+                    undefined /* triggerContentChangedEvent */,
```
JiuqingSong: This line can be removed

PR #320, opened 2019-07-18
Support adding custom data when creating editor

JiuqingSong:
1. We have editor.getCustomData() function to get/set custom data to editor so editor can help hold the objects. But if we want to add some custom data and allow plugins to access them at any time, including when initializing, we need to add them when create editor. This change add customData property to EditorOption to enable this functionality.

2. Fix a bug of undo plugin, when reuse this plugin across editor session, we should not add initial undo snapshot for the 2nd session since there are already snapshots.


JiuqingSong:
> Approach looks reasonable. Are there any public uses of the custom data pattern right now? I never knew about this, and am curious what folks are using it.

The idea comes from my compose code clean up work. I was trying to save some editor related stuff into editor object which has the same lifecycle with editor so we don't need to track them in separate places.

PR #326, opened 2019-07-30
Merge dark mode into master!!!!

Lego6245:
It's finally here!


adriantran:
curious to know if wiki documentation will be available for this?

Lego6245:
> curious to know if wiki documentation will be available for this?

There is a basic wiki page currently active, and once merged, we have plans to update the sample site.

https://github.com/microsoft/roosterjs/wiki/Dark-Mode

packages/roosterjs-editor-types/lib/consts/DarkMode.ts
```diff
@@ -0,0 +1,10 @@
+export const DARK_MODE_DEFAULT_FORMAT = {
```
JiuqingSong: The initial idea of roosterjs-editor-types package is not to generate any js code, everything here is just interfaces, types, const enums. However the const will generate some js code. I don't have strong objection, but still think we can move it to somewhere else, unless we believe it will be used somewhere else as well.

.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.npmjs.com/
```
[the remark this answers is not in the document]
Lego6245: For anyone with a global .npmrc that has their registry pointing somewhere other than npmjs, attempts to deploy the project will fail. I don't see any reason to not have it just to make things simpler.

packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -46,10 +46,18 @@ export default function clearFormat(editor: Editor) {
                 setFontSize(editor, defaultFormat.fontSize);
             }
             if (defaultFormat.textColor) {
-                setTextColor(editor, defaultFormat.textColor);
+                if (defaultFormat.textColors) {
+                    setTextColor(editor, defaultFormat.textColors);
+                } else {
```
imWildCat: Can this logic be improved in this way?

```typescript
  setTextColor(editor, defaultFormat.textColor);
}
else if (defaultFormat.textColors) {
  setTextColor(editor, defaultFormat.textColors);
} 
```

I think we don't need to set `textColor` if we are using `textColors`.

PR #332, opened 2019-08-05
Add Scroll event

JiuqingSong:
Add a scroll event, trigger from DOMEventPlugin.
This can help simplify OWA scroll handling logic and get rid of "ScrollAdapter"

### DIFF
--- a/packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
@@ -21,6 +21,7 @@ import {
  * 2. Selection management
  * 3. Cut and Drop management
  * 4. Pending format state management
+ * 5. Scroll container and scroll event management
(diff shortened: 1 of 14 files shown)


packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
```diff
@@ -29,7 +30,10 @@ export default class DOMEventPlugin implements EditorPlugin {
     private cachedPosition: NodePosition;
     private cachedFormatState: PendableFormatState;
 
-    constructor(private disableRestoreSelectionOnFocus: boolean) {}
+    constructor(
+        private disableRestoreSelectionOnFocus: boolean,
+        public readonly scrollContainer?: HTMLElement
```
titrongt: Should this param be required? Will DOMEventPlugin be created somewhere other than in `createEditorCore`?

PR #334, opened 2019-08-05
Refactor UndoSnapshot

JiuqingSong:
This is to move function implementations out of UndoSnapshot class so that developer can easily build their own UndoSnapshot and reuse our existing logic. This can help the OWA code clean up and be able to put undo snapshot on store so that the code can easily clear undo snapshots without knowing Undo plugin.

### DIFF
--- a/packages/roosterjs-editor-core/lib/undo/UndoSnapshots.ts
+++ b/packages/roosterjs-editor-core/lib/undo/UndoSnapshots.ts
@@ -5,6 +5,7 @@ import {
     canMoveCurrentSnapshot,
     moveCurrentSnapsnot,
     clearProceedingSnapshots,
+    createSnapshots,
 } from 'roosterjs-editor-dom';
 
 // Max stack size that cannot be exceeded. When exceeded, old undo history will be dropped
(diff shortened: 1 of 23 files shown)


packages/roosterjs-editor-core/lib/undo/UndoSnapshots.ts
```diff
@@ -7,25 +14,20 @@ const MAXSIZELIMIT = 1e7;
...
  * A class to help manage undo snapshots
  */
-export default class UndoSnapshots implements UndoSnapshotsService {
-    private snapshots: string[];
-    private totalSize: number;
-    private currentIndex: number;
+export default class UndoSnapshots implements UndoSnapshotsService, Snapshots {
+    snapshots: string[] = [];
+    totalSize = 0;
+    currentIndex = -1;
```
titrongt: nit: Types were removed for `totalSize `and `currentIndex`

packages/roosterjs-editor-dom/lib/snapshots/clearSnapshotsAfterCurrent.ts
```diff
@@ -0,0 +1,17 @@
+import canMoveCurrentSnapshot from './canMoveCurrentSnapshot';
+import { Snapshots } from 'roosterjs-editor-types';
+
+/**
+ * Clear all snapshots after the current one
+ * @param snapshots The snapshots data structure to clear
+ */
+export default function clearSnapshotsAfterCurrent(snapshots: Snapshots) {
```
titrongt: nit: the name `clearSnapshotsAfterCurrent` is a bit confusing to me, suggest change to something like `clearProceedingSnapshots`?

packages/roosterjs-editor-core/lib/undo/UndoSnapshots.ts
JiuqingSong: yes. type is not necessary here

packages/roosterjs-editor-dom/lib/snapshots/clearSnapshotsAfterCurrent.ts
JiuqingSong: sounds good.

PR #337, opened 2019-08-06
PickerPlugin for Android 

titrongt:
### Context
Outlook on Android has integrated the AtMentions plugin, but is missing some of the functionality we see on OWA. 

On Android, the code for the keyboard event is "Unidentified". which causes `isCharacterValue-->onKeyUp` and `onKeyDown `to fail. This makes it so the inline picker does not appear and backspace will not properly remove the AtMention node. This change works around this by looking at input events for Android.
 
### Deploy Link
https://outlook-sdf.office.com/mail?branch=titrongt-refs-heads-u-titrongt-testFixPickerPluginForMobile


packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -199,7 +227,7 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
         this.setAriaActiveDescendant(isSuggesting ? 0 : null);
     }
 
-    private handleKeyDownEvent(event: PluginKeyboardEvent) {
+    private handleKeyDownEvent(event: PluginDomEvent) {
```
JiuqingSong: why need to change the type?

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -57,6 +65,11 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
     private isSuggesting: boolean;
     private lastKnownRange: Range;
 
+    // For detecting backspace in Android
+    private isUnidentifiedKey: boolean = false;
```
JiuqingSong: Suggest name: pendingInputEventHandling

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -414,6 +438,46 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
+        ) {
+            this.removeNode(event);
+        }
+
+        this.currentInputLength = this.newInputLength;
+    }
+
+    private removeNode(event: PluginDomEvent) {
+        const searcher = cacheGetContentSearcher(event, this.editor);
+        const nodeBeforeCursor = searcher.getInlineElementBefore()
```
JiuqingSong: save the result of getInlineElementBefore() first

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -414,6 +438,46 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
+        if (
+            nodeId &&
+            nodeId.indexOf(this.pickerOptions.elementIdPrefix) == 0 &&
+            (nodeAfterCursor == null || !(nodeAfterCursor instanceof PartialInlineElement))
+        ) {
+            let replacementNode = this.dataProvider.onRemove(nodeBeforeCursor, true);
+            if (replacementNode) {
+                this.replaceNode(nodeBeforeCursor, replacementNode);
+                this.editor.runAsync(() => {
+                    this.editor.select(replacementNode, PositionType.After);
```
JiuqingSong: I think we should only do this for Android scenario.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
JiuqingSong: Actually this name is confusing. It is not "handle" the event, but "cancel" the event. Need to use a better name

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -414,6 +438,46 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
         }
     }
 
+    private onAndroidInputEvent(event: PluginInputEvent) {
+        const wordBeforCursor = this.getInlineElementBeforeCursor(event);
+        this.newInputLength = wordBeforCursor ? wordBeforCursor.length : 0;
+
+        if (
+            this.newInputLength < this.currentInputLength ||
```
JiuqingSong: if I type on a long node, then move cursor to a short one then continue type, it is possible this still get triggered. In that case I think it is wrong.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
titrongt: Sounds good

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
titrongt: This was changed because currently, the event can either be a PluginKeyboardEvent or PluginInputEvent (from Android). Though in the Android case, we probably don't need to keep track of `eventHandledOnKeyDown `.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -165,16 +180,29 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
             this.dataProvider.onContentChanged(elementIds);
         }
+
         if (event.eventType == PluginEventType.KeyDown) {
-            this.eventHandledOnKeyDown = false;
-            this.onKeyDownEvent(event);
+            if (event.rawEvent.key == UNIDENTIFIED_KEY) {
+                // On Android, the key for KeyboardEvent is "Unidentified",
+                // so ignore onKeyDownEvent and handle with InputEvent instead
+                this.isPendingInputEventHandling = true;
```
JiuqingSong: Instead of calculating currentInputLength when insertNode, I think it will be more accurate to calculate the content length here, because here means input event is about to happen, so we can get content length before the input. Please work with John to see if it works correctly.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -408,8 +433,53 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
+        if (
+            nodeId &&
+            nodeId.indexOf(this.pickerOptions.elementIdPrefix) == 0 &&
+            (inlineElementAfter == null || !(inlineElementAfter instanceof PartialInlineElement))
+        ) {
+            const replacementNode = this.dataProvider.onRemove(nodeBeforeCursor, true);
+            if (this.isPendingInputEventHandling) {
+                if (replacementNode) {
+                    this.replaceNode(nodeBeforeCursor, replacementNode);
+                    this.editor.runAsync(() => {
```
JiuqingSong: By saying "only need to do this for Android", I mean the runAsync() part. For others like check null for replacementNode, call replaceNode or deleteNode, they should be fine for both.

titrongt: Oh okay, I will simplify this function then!

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
titrongt: Does this mean the short node will be removed? Can we detect when the cursor has been moved (also, is this by mouse or keyboard)?

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
titrongt: Sounds good. I've updated accordingly and asked John to see if the new code still works correctly 👍🏻

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
titrongt: Should be fixed if the latest change works as expected

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -371,34 +411,20 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
-                    (searcher.getInlineElementAfter() == null ||
-                        !(searcher.getInlineElementAfter() instanceof PartialInlineElement))
-                ) {
-                    let replacementNode = this.dataProvider.onRemove(nodeBeforeCursor, true);
-                    this.replaceNode(nodeBeforeCursor, replacementNode);
-                    this.editor.select(replacementNode, PositionType.After);
-                    this.handleKeyDownEvent(event);
-                }
+                this.removeNode(event);
+                this.cancelDefaultKeyDownEvent(event);
```
JiuqingSong: In original code, handleKeyDownEvent is only called when "nodeId &&
                    nodeId.indexOf(this.pickerOptions.elementIdPrefix) == 0 &&
                    (searcher.getInlineElementAfter() == null ||
                        !(searcher.getInlineElementAfter() instanceof PartialInlineElement))", but now it is always called. Is that ok?

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -408,12 +434,60 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
     }
 
+    private onAndroidInputEvent(event: PluginInputEvent) {
+        this.newInputLength = this.calcInputLength(event);
+
+        if (
+            this.newInputLength < this.currentInputLength ||
+            (event.rawEvent as any).inputType === DELETE_CONTENT_BACKWARDS_INPUT_TYPE
+        ) {
+            this.removeNode(event);
```
JiuqingSong: Better use a name like "checkAndRemoveNode" or "tryRemoveNode" since it is not really always removing node.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -408,12 +434,60 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
+        }
+
+        this.currentInputLength = this.newInputLength;
```
JiuqingSong: I guess we don't need to assign the length here since we will always recalculate it when keydown.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
```diff
@@ -408,12 +434,60 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
                 }
             }
         }
     }
 
+    private onAndroidInputEvent(event: PluginInputEvent) {
+        this.newInputLength = this.calcInputLength(event);
+
+        if (
+            this.newInputLength < this.currentInputLength ||
```
JiuqingSong: If cursor is at beginning of a node, and backspace will actually move cursor to the previous node. In that case current node length will be greater than originl length (0). Can the code handle that case? This is a question to John.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
titrongt: Good catch, I think it should only be called if the node was successfully removed.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
tito0224: No sure I understand. How can I test this behavior you're describing?

tito0224:
I just did a round of testing on Android using the latest code here and everything seems to be working.

packages/roosterjs-plugin-picker/lib/PickerPlugin.ts
JiuqingSong: [NODE 1][[CURSOR] NODE 2]

When cursor is in front of node 2, press Backspace will still delete the last char of node 1. But in that case node length we measure before and after the event are for different node (firstly node 2, then node 1). In that case comparing the length doesn't seem make sense.

But it might be a corner case. I'll leave this to John, you can decide if this is an real issue or it can never happen.

tito0224: I test the scenario where you have two mentions next to each other with no space and you put the cursor between them and hit backspace. Everything works.

Issue #330, opened 2019-05-01
When replacing links using ctrl-v, link doesnt disappear


bochong:
If you add some text and create a link in the text, and then try to remove the link by highlighting it and replacing it with other copied text with ctrl-v then inspect the text, the link will still be there when it should have been cleared.

JiuqingSong:
I can repro it.

PR #339, opened 2019-08-08
Fix #277 and #330

### LINKED ISSUE #277: Insert <A> tag inside a <A> tag, should split outer <A> tag
e.g., paste a hyperlink instead (or override) existing hyperlink, need to split outer link to avoid having A  tag inside A tag.

JiuqingSong:
#277 Insert <A> tag inside a [A] tag, should split outer [A] tag
The fix is to follow Chrome CED behavior:
- If the node to insert doesn't contain any [A] tag, no change.
- If the node to insert contains [A] tag, but there isn't any [A] tag around current insert position, no change
- If the node to insert contains [A] tag, and there is [A] tag around current position within current block, split the existing [A] tag into two and insert the new nodes between them.

#330 When replacing links using ctrl-v, link doesnt disappear
Before insert, let's check if there is an empty [A] tag around current insert position. If yes, remove it.

### DIFF
--- a/packages/roosterjs-editor-dom/lib/utils/adjustNodeInsertPosition.ts
+++ b/packages/roosterjs-editor-dom/lib/utils/adjustNodeInsertPosition.ts
@@ -1,13 +1,26 @@
 import changeElementTag from './changeElementTag';
+import contains from './contains';
+import createRange from '../selection/createRange';
 import findClosestElementAncestor from './findClosestElementAncestor';
+import getBlockElementAtNode from '../blockElements/getBlockElementAtNode';
 import getTagOfNode from './getTagOfNode';
+import isNodeEmpty from './isNodeEmpty';
 import isPositionAtBeginningOf from '../selection/isPositionAtBeginningOf';
 import isVoidHtmlElement from './isVoidHtmlElement';
 import Position from '../selection/Position';
+import queryElements from './queryElements';
+import splitTextNode from './splitTextNode';
 import unwrap from './unwrap';
 import VTable from '../table/VTable';
 import wrap from './wrap';
-import { NodePosition, NodeType, PositionType } from 'roosterjs-editor-types';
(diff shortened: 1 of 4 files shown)


packages/roosterjs-editor-dom/lib/utils/adjustNodeInsertPosition.ts
```diff
@@ -17,7 +30,76 @@ import { NodePosition, NodeType, PositionType } from 'roosterjs-editor-types';
  * @param position The original position to insert the node
  */
 export default function adjustNodeInsertPosition(
-    root: Node,
+    root: HTMLElement,
```
titrongt: I noticed you changed the `root `type from `Node `to `HTMLElement`. Was this related to the fix for the bug?

packages/roosterjs-editor-dom/lib/utils/splitTextNode.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Split a text node into two parts by an offset number, and return one of them
+ * @param textNode The text node to split
+ * @param offset The offset number to split at
+ * @param returnFirstPart True to return the first part, then the passed in textNode will become the second part.
+ * Otherwise return the second part, and the passed in textNode will become the first part
+ */
+export default function splitTextNode(textNode: Text, offset: number, returnFirstPart: boolean) {
+    let firstPart = textNode.nodeValue.substr(0, offset);
+    let secondPart = textNode.nodeValue.substr(offset);
+    let newNode = textNode.ownerDocument.createTextNode(returnFirstPart ? firstPart : secondPart);
```
titrongt: nit: use consts

packages/roosterjs-editor-dom/lib/utils/adjustNodeInsertPosition.ts
JiuqingSong: Yes. Because API queryElements() requires HTMLElement. And here root actually must be an HTMLElement.

PR #342, opened 2019-08-14
Fix insertNode at end

premnirmal:
Fixing an issue with the `insertNode()` function, where if two elements are inserted with `ContentPosition.End` then the last one is inserted inside the last child instead of inside the container.

Added unit tests to demonstrate the issue, and fixed in `coreApi/insertNode.ts`

### DIFF
--- a/packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
+++ b/packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
@@ -77,7 +77,8 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
                             isBegin ? refNode : refNode.nextSibling
                         );
                     } else {
-                        contentDiv.appendChild(node);
+                        // Inserting at the end on a new line - use appendChild to insert the content in contentDiv
+                        insertedNode = contentDiv.appendChild(node);
                     }
                 } else {
                     // if the refNode can have child, use appendChild (which is like to insert as first/last child)
--- a/packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
+++ b/packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
@@ -71,10 +71,11 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
                     // For insert on new line, or refNode is text or void html element (HR, BR etc.)
                     // which cannot have children, i.e. <div>hello<br>world</div>. 'hello', 'world' are the
                     // first and last node. Insert before 'hello' or after 'world', but still inside DIV
-                    insertedNode = refNode.parentNode.insertBefore(
-                        node,
-                        isBegin ? refNode : refNode.nextSibling
-                    );
+                    if (isBegin) {
+                        insertedNode = refNode.parentNode.insertBefore(node, refNode);
+                    } else {
+                        contentDiv.appendChild(node);
+                    }
                 } else {
                     // if the refNode can have child, use appendChild (which is like to insert as first/last child)
                     // i.e. <div>hello</div>, the content will be inserted before/after hello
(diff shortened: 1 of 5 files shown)


premnirmal:
@JiuqingSong - the unit tests in this PR demonstrate the issue when calling `insertNode()` on the editor

JiuqingSong:
> @JiuqingSong - the unit tests in this PR demonstrate the issue when calling `insertNode()` on the editor

Got it. Thanks.

premnirmal:
I am actively trying to fix this, but if you have any suggestions let me know. I think the bug lies in this function: https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs-editor-dom/lib/blockElements/getBlockElementAtNode.ts#L34 invoked from this https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs-editor-dom/lib/blockElements/getFirstLastBlockElement.ts#L10

JiuqingSong:
The issue is not these functions themselves, but how we use them.
When get last block element, it indeed returns the last block element, but it is not at the to level, which means:
[div 1]
  some text
  [div 2]
  [/div]
[/div]

It will return div 2, although it not at the top level. 

As a fix, I have two options:
1. From the end node of the last block element, run a loop to find the top level container, then insert after it
2. Do not use getBlockElement() here, but directly get the last element of editor and insert after it.

I think 2 will be more straightforward.

premnirmal:
Thanks, I updated and went with 2) i.e. directly get the last element of editor and insert after it.

packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
```diff
@@ -71,10 +71,14 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
...
-                        node,
-                        isBegin ? refNode : refNode.nextSibling
-                    );
+                    if (isBegin || !option.insertOnNewLine) {
+                        insertedNode = refNode.parentNode.insertBefore(
+                            node,
+                            isBegin ? refNode : refNode.nextSibling
+                        );
+                    } else {
+                        contentDiv.appendChild(node);
```
tito0224: Do we need to ensure `insertedNode` is set to that it can be wrapped in a div if necessary?

premnirmal: Good point - updated

JiuqingSong:
I'm rethinking about this solution, a little worried about the risk since it changes the default behavior of a broad-used API. Give me some time to think about it.

JiuqingSong:
Ideally, insert at end and insert at begin should have consistent behavior. And this change only fixed insert at end case.

However, I'm afraid if change insert at begin case, it may break some OWA scenario, which contains some hidden content in front of other contents. And make it always insert as the first node will insert before the hidden content. That's why I haven't decided if this fix can be merged.

JiuqingSong:
I think we'd better keep current behavior. If we want to really insert at the end of DOM structure, we'd better add new API.

premnirmal:
Thanks for your feedback @JiuqingSong 

>  if change insert at begin case, it may break some OWA scenario, which contains some hidden content in front of other contents

Can we add unit tests to protect these scenarios from being broken?

> If we want to really insert at the end of DOM structure, we'd better add new API

We would need to make sure the api isn't confusing, so that the developer knows the difference between `editor.insertNode(node, { position: ContentPosition.End })` and `editor.insertNodeAtEnd(node)`. Any thoughts on this?

JiuqingSong:
> Can we add unit tests to protect these scenarios from being broken?

Yes, we can. But I'm afraid there may be other scenarios can be broken, given that this is a very widely used API. That's why I don't want to change its behavior.



> We would need to make sure the api isn't confusing, so that the developer knows the difference between `editor.insertNode(node, { position: ContentPosition.End })` and `editor.insertNodeAtEnd(node)`. Any thoughts on this?

Agree that we should not make it confusing. But since it is already widely used and this will change its original behavior which may potentially cause some regression, I think we can have two behaviors co-exist for some versions and mark the original one as deprecated, then in next major release we can remove the old one. So that people have time to test the new behavior if they want and report bugs, and people can still use the original one before we fix all the bugs. Every time when we make big change to existing behavior we do like this.

If you don't want to introduce a new API which may cause confusing, you can add new insert position for this API, say "End" and "EndOfDom", or some better name.

premnirmal:
I see. I will add a new `ContentPosition` enum that indicates end of the content div

PR #347, opened 2019-08-21
Cut support in the Copy Plugin

Lego6245:
The two events are different, but we want to do largely the same thing to them: Transform them before putting them on the clipboard.


package.json
```diff
@@ -1,75 +1,75 @@
...
-        "sass-loader": "7.1.0",
-        "url-loader": "2.0.1"
-    },
-    "husky": {
-        "hooks": {
-            "pre-commit": "pretty-quick --staged"
-        }
-    }
-}
+{
```
JiuqingSong: Don't change the line end to LF only.

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
```diff
@@ -22,6 +23,7 @@ export default class CopyPlugin implements EditorPlugin {
     public initialize(editor: Editor) {
         this.editor = editor;
         this.copyDisposer = editor.addDomEventHandler('copy', this.onCopy);
+        this.cutDisposer = editor.addDomEventHandler('cut', this.onCopy);
```
JiuqingSong: editor.addDomEventHandler() supports adding multiple events:
editor.addDomEventHandler({
copy: this.onCopy,
cut: this.onCut,
});

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
```diff
@@ -30,6 +32,8 @@ export default class CopyPlugin implements EditorPlugin {
     public dispose() {
         this.copyDisposer();
         this.copyDisposer = null;
+        this.cutDisposer();
+        this.cutDisposer = null;
         this.editor = null;
     }
```
JiuqingSong: Question:

If we do preventDefault() for cut event, will the selected content be removed?

Lego6245: Ah, yeah. It does stop that from happening.

package.json
Lego6245: The file is now in CRLF right now (with this check-in), so this should not occur in the future.

packages/roosterjs-editor-core/lib/corePlugins/CopyPlugin.ts
Lego6245: This will need to be handled (I published this while I was testing my changes).

package.json
JiuqingSong: I'm sure that it was CrLf but this change makes it to be Lf.

Two evidences: 
1. File sizes: 3140 (after change) v.s. 3215 (before change), the diff is 75 bytes which equals the count of lines.
2. From a HEX viewer, I can see 0D0A before the change and 0A only after change.

Lego6245: It was a bad config that was clobbering it post commit. I resolved it.

PR #348, opened 2019-08-23
Expose selection path related methods

JiuqingSong:
Expose two methods with selection path:
1. getHtmlWithSelectionPath(), return innerHTML of a root element with selection path if a valid range is passed
2. setHtmlWithSelectionPath(), restore the innerHTML with selection path, and return a range (or null if no valid selection path is included)
So that people can call them to generate HTML content with selection path, when pass it into Editor.setContent(), content will be auto selected.

Other fix:
1. When appy default format on initialize, use current cursor position rather than beginning position if current position exists
2. Editor.setContent() should not wrap a single node for "insertOnNewLine" scenario because it will call insertNode() which will wrap the node if it is not a block element.

### DIFF
--- a/packages/roosterjs-editor-core/lib/editor/Editor.ts
+++ b/packages/roosterjs-editor-core/lib/editor/Editor.ts
@@ -101,9 +101,8 @@ export default class Editor {
         this.triggerPluginEvent(PluginEventType.EditorReady, {}, true /*broadcast*/);
 
         // 10. Before give editor to user, make sure there is at least one DIV element to accept typing
-        const range = this.getSelectionRange();
         this.core.corePlugins.typeInContainer.ensureTypeInElement(
-            range ? Position.getStart(range) : new Position(contentDiv, PositionType.Begin)
+            this.getFocusedPosition() || new Position(contentDiv, PositionType.Begin)
         );
     }
 
--- a/packages/roosterjs-editor-core/lib/editor/Editor.ts
+++ b/packages/roosterjs-editor-core/lib/editor/Editor.ts
@@ -16,14 +16,12 @@ import {
     InlineElement,
     InsertOption,
     NodePosition,
-    NodeType,
     PluginEvent,
     PluginEventData,
     PluginEventFromType,
     PluginEventType,
     PositionType,
     QueryScope,
-    SelectionPath,
     Rect,
 } from 'roosterjs-editor-types';
 import {
@@ -34,16 +32,16 @@ import {
     findClosestElementAncestor,
     fromHtml,
     getBlockElementAtNode,
+    getHtmlWithSelectionPath,
     getTextContent,
     getInlineElementAtNode,
     getPositionRect,
-    getRangeFromSelectionPath,
-    getSelectionPath,
     getTagOfNode,
     isNodeEmpty,
     Position,
     PositionContentSearcher,
     queryElements,
+    setHtmlWithSelectionPath,
     wrap,
 } from 'roosterjs-editor-dom';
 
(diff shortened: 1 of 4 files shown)


packages/roosterjs-editor-dom/lib/selection/setHtmlWithSelectionPath.ts
```diff
@@ -0,0 +1,29 @@
...
+import { NodeType, SelectionPath } from 'roosterjs-editor-types';
+
+/**
+ * Restore inner Html of a root element from given html string. If the string contains selection path,
+ * remove the selection path and return a range represented by the path
+ * @param root The root element
+ * @param html The html to restore
+ * @returns A selection range if the html contains a valid selection path, otherwise null
+ */
+export default function getHtmlWithSelectionPath(rootNode: HTMLElement, html: string): Range {
```
titrongt: typo: setHtmlWithSelectionPath

JiuqingSong: good catch

titrongt: actually, if the function returns a range, the name is a bit misleading -- maybe something like `restoreHtmlAndGetSelectionPath`?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -345,16 +343,10 @@ export default class Editor {
...
-        let content = contentDiv.innerHTML;
-        let selectionPath: SelectionPath;
-
-        if (
-            includeSelectionMarker &&
-            (selectionPath = getSelectionPath(contentDiv, this.getSelectionRange()))
-        ) {
-            content += `<!--${JSON.stringify(selectionPath)}-->`;
-        }
+        let content = getHtmlWithSelectionPath(
```
titrongt: curious why you decided to factor this out? I think it looked more clear when it was all written here

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -436,10 +418,11 @@ export default class Editor {
     public insertContent(content: string, option?: InsertOption) {
         if (content) {
             let allNodes = fromHtml(content, this.core.document);
+
             // If it is to insert on new line, and there are more than one node in the collection, wrap all nodes with
             // a parent DIV before calling insertNode on each top level sub node. Otherwise, every sub node may get wrapped
             // separately to show up on its own line
-            if (option && option.insertOnNewLine && allNodes.length > 0) {
+            if (option && option.insertOnNewLine && allNodes.length > 1) {
```
titrongt: why was this number changed?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -388,19 +380,9 @@ export default class Editor {
         let contentDiv = this.core.contentDiv;
         let contentChanged = false;
         if (contentDiv.innerHTML != content) {
-            contentDiv.innerHTML = content || '';
+            let range = setHtmlWithSelectionPath(contentDiv, content);
+            this.select(range);
```
titrongt: in the old code, `this.select(range)` was called in the try block -- is it fine to call `this.select(null)`?

packages/roosterjs-editor-core/lib/editor/Editor.ts
JiuqingSong: See "Other fix 2" in the description:

Editor.insertContent() should not wrap a single node for "insertOnNewLine" scenario because it will call insertNode() which will wrap the node if it is not a block element.

packages/roosterjs-editor-core/lib/editor/Editor.ts
JiuqingSong: yes. It's ok to call with null, and it will be ignored.

Issue #336, opened 2019-08-06
Make `transformOnInitialize` default to true?


imWildCat:
**Describe the bug**
While using `initialContent` option, the `inDarkMode` does not work.

For example,
```ts
        let options: EditorOptions = {
            plugins: plugins,
            defaultFormat: defaultFormat,
            undo: this.props.undo,
            inDarkMode: true,
            initialContent: 'This will break dark mode',
        };
```

With colors:
```ts
    defaultFormat: {
        textColors: {
            lightModeColor: 'black',
            darkModeColor: 'white',
        },
        backgroundColors: {
            darkModeColor: 'black',
            lightModeColor: 'white',
        },
    },
```

**To Reproduce**
As specified above.

If we get rid of `initialContent` in the options, it can work.

**Expected behavior**
This will break dark mode' will show in dark mode.

**Screenshots**
The content shows in light mode.

**Device Information**
 - OS: macOS
 - Browser: Chrome 76 or Edge for Mac Version 77.0.230.2 (Official build) Dev (64-bit)

Lego6245:
Modify the editior options block as follows:

```
let options: EditorOptions = {
            plugins: plugins,
            defaultFormat: defaultFormat,
            undo: this.props.undo,
            inDarkMode: true,
            initialContent: 'This will break dark mode',
            darkModeOptions: {
                     transformOnInitialize: true
            },
        };
```

Then it should work as expected. Arguably this should be the default option.

PR #354, opened 2019-09-09
Make the Transform on Initialize option do something useful, and default to true.

Lego6245:
The current option is a no-op due to a bug where having it undefined would cause it to not transform anything (which was un-intuitive). Now we respect the option again, and this time include a check for if the function is in its first run.

I am open to alternatively passing in the transformOnInitialize value in directly, but I think making it an internal function and making it a consequence of first run makes more sense. Feel free to disagree, as I'm not wedded to this approach 💃

### DIFF
--- a/packages/roosterjs-editor-core/lib/editor/Editor.ts
+++ b/packages/roosterjs-editor-core/lib/editor/Editor.ts
@@ -73,9 +73,10 @@ export default class Editor {
         this.core.plugins.forEach(plugin => plugin.initialize(this));
 
         // 4. Ensure initial content and its format
-        this.setContent(
+        this.setContentInternal(
             options.initialContent || contentDiv.innerHTML || '',
-            false /*triggerContentChangedEvent*/
+            false /*triggerContentChangedEvent*/,
+            true /* firstRun */
         );
 
         // 5. Create event handler to bind DOM events
@@ -377,6 +378,14 @@ export default class Editor {
      * @param triggerContentChangedEvent True to trigger a ContentChanged event. Default value is true
      */
     public setContent(content: string, triggerContentChangedEvent: boolean = true) {
+        this.setContentInternal(content, triggerContentChangedEvent);
+    }
+
+    private setContentInternal(
+        content: string,
+        triggerContentChangedEvent: boolean,
+        firstRun?: boolean
+    ) {
         let contentDiv = this.core.contentDiv;
         let contentChanged = false;
         if (contentDiv.innerHTML != content) {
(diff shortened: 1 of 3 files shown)


Lego6245:
#336

packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
```diff
@@ -43,6 +43,10 @@ export default function createEditorCore(
     let eventHandlerPlugins = allPlugins.filter(
         plugin => plugin.onPluginEvent || plugin.willHandleEventExclusively
     );
+    if (options.darkModeOptions && options.darkModeOptions.transformOnInitialize === undefined) {
+        // Since you can't provide default values on interfaces, provide the default here.
+        options.darkModeOptions.transformOnInitialize = true;
```
JiuqingSong: I think we'd better not directly change the value from options since it is created by caller, and caller is able to change it after creating editor, which may cause unpredictable issue.

JiuqingSong: We'd better clone a copy of darkModeOptions then we are safe to do any modification

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -385,9 +394,13 @@ export default class Editor {
...
 
-        // Convert content even if it hasn't changed.
-        if (this.core.inDarkMode) {
-            const darkModeOptions = this.getDarkModeOptions();
+        const darkModeOptions = this.getDarkModeOptions();
+        // Convert when...
+        // It's in dark mode, AND it's either not the first run OR transformOnInitialze is true
+        if (
+            this.core.inDarkMode &&
+            (!firstRun || (darkModeOptions && darkModeOptions.transformOnInitialize))
```
JiuqingSong: In what case we don't to color transformation when first run?

packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
Lego6245: This setting technically only matters the first time the editor is initialized... Does that change anything?

packages/roosterjs-editor-core/lib/editor/Editor.ts
Lego6245: If it's first run AND the setting exists AND it's set to false.

I should refactor this conditional.

packages/roosterjs-editor-core/lib/editor/createEditorCore.ts
JiuqingSong: hmm, functionally yes, it should work. But still feel we should not change the option value directly. In an extreme case, say user save the option object on a mobx store, and pass it into editor, then we changed it, and this will cause a runtime exception.

PR #356, opened 2019-09-19
Fix #355:  when paste from Word and empty lines are dropped

### LINKED ISSUE #355: When paste from Word, <span>&nbsp;</span> shouldn't get modified into <span></span>
[bug discussed in email; contains non-zero PII]

JiuqingSong:
Previously I added code to remove <o:p> tag for Word to make it work with bullet list. Now let's preserve it if its inner html is &nbsp;.


philipzloh:
Fixes #355

PR #360, opened 2019-10-01
Support mounting react elements in editor

Adjective-Object:
Adds a plugin to support mounting custom react elements in-editor.

### Goals:
- components should be copy/pasteable
- A should be able to be drag/dropped around the Dom without losing state
- components should be able to be deleted & reinserted via undo/redo, preserving a reasonable amount of internal state
- consumers using this plugin should have to be minimally aware of the fact that they're not mounting in the same DOM that is being rendered to the user.

---

### Approach

In order for darkmode to interoperate with custom react components in the editor, their transforms need to go through some darkmode-aware content insertion pipeline.

This proposal works by tracking component mountpoints across the editor surface, and mirroring components mounted off the actual document root into the discovered mountpoints. This option was chosen over direct mounting or a custom react reconciler for the following reasons:

- direct mounting: rerenders of the component can't be tracked and be made darkmode aware, nor can their updates be tracked by our undo buffer (if that's something we care to do).
- react reconciler: this seems like the best way in theory, but the additional complexity & code doesn't seem worth the tradeoff, and the react-reconciler api isn't stable.

---
### Test plan:

`yarn test`
`yarn start`, and play with the experimental custom component (copy/paste the component into the editor from its section in the sidebar).

---

### Misc:
1. I had to remove a snippet from the lint task, since it crashes out with a message about circular dependencies if you don't. This is because it tracks `undefined` as a parent of `undefined`, cyclically in toposort.
2. there's no task for normalizing line-endings, nor is it configured in the repo. I tried forcing the whole repo to `cr/lf` and just `lf`, but both of those had huge footprints. Is there a prettier task that I can't find, or is everything just supposed to be formatted by vscode as we edit? if so we should have a .editorconfig.


JiuqingSong:
Is this for darkmode beautified link?

Adjective-Object:
@JiuqingSong yes, this is for darkmode beautiful link. Or other live+interactive components we might want to mount in-editor in the future.

Adjective-Object:
Closing in favor of https://github.com/microsoft/roosterjs-react/pull/101

PR #361, opened 2019-10-07
Fix OWA 57080

JiuqingSong:
Bug 57080: Backspace should not convert text back into emoji

In CustomReplace plugin, we should not perform the auto complete action if user input is "backspace".

So the fix is to check input event data should not be null


packages/roosterjs-editor-plugins/lib/CustomReplace/CustomReplace.ts
```diff
@@ -91,7 +91,7 @@ export default class CustomReplacePlugin implements EditorPlugin {
         }
 
         // Exit early on input events that do not insert a replacement's final character.
```
Lego6245: does this comment need to be updated? or still accurate?

JiuqingSong: I feel it is still accurate.

PR #371, opened 2019-10-28
One note bullet paste

FrancisMengx:
Before: 
one note before

After:
one note after

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/Paste/fragmentHandler.ts
+++ b/packages/roosterjs-editor-plugins/lib/Paste/fragmentHandler.ts
@@ -1,6 +1,7 @@
 import convertPastedContentFromExcel from './excelConverter/convertPastedContentFromExcel';
 import convertPastedContentFromWord from './wordConverter/convertPastedContentFromWord';
 import convertPastedContentFromWordOnline, { isWordOnlineWithList } from './officeOnlineConverter/convertPastedContentFromWordOnline';
+import { WAC_IDENTIFING_SELECTOR, LIST_SELECTOR } from './wacConverter/constants';
 import { getTagOfNode } from 'roosterjs-editor-dom';
 import { splitWithFragment } from 'roosterjs-html-sanitizer';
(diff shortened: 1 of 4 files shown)


packages/roosterjs-editor-plugins/lib/Paste/fragmentHandler.ts
```diff
@@ -22,10 +23,18 @@ export default function fragmentHandler(doc: HTMLDocument, source: string) {
             // Handle HTML copied from MS Word
             doc.body.innerHTML = html;
             convertPastedContentFromWord(doc);
-        } else if (isWordOnlineWithList(firstNode)) {
+        } else if (firstNode && firstNode.querySelector(WAC_IDENTIFING_SELECTOR)) {
```
JiuqingSong: no need to check null for firstNode. We already use it before this line

packages/roosterjs-editor-plugins/lib/Paste/fragmentHandler.ts
```diff
@@ -22,10 +23,18 @@ export default function fragmentHandler(doc: HTMLDocument, source: string) {
...
             doc.body.innerHTML = html;
             convertPastedContentFromWord(doc);
-        } else if (isWordOnlineWithList(firstNode)) {
+        } else if (firstNode && firstNode.querySelector(WAC_IDENTIFING_SELECTOR)) {
+            // Once it is known that the document is from WAC
+            // We need to remove the display property and margin from all the list item
+            doc.querySelectorAll(WAC_IDENTIFING_SELECTOR).forEach((el: HTMLElement) => {
+                el.style.display = null;
+                el.style.margin = null;
+            });
```
JiuqingSong: I think you can just query once. Here you called both querySelector and querySelectorAll.

FrancisMengx:
For one note we are simply tripping the styles. All the existing tests will run through the code path as well, so we are good on the tests guarding regression.

PR #381, opened 2020-01-09
Fix a watermark bug

JiuqingSong:
Watermark is working in an incorrect way. The first time evaluation is done too early before editor is fully initialized. So if there is some content like "[DIV][/DIV]" watermark will be replaced.

Fix this by delay the first evaluation to EditorReady event

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/Watermark/Watermark.ts
+++ b/packages/roosterjs-editor-plugins/lib/Watermark/Watermark.ts
@@ -5,7 +5,6 @@ import {
     PluginEvent,
     PluginEventType,
     ContentPosition,
-    ContentChangedEvent,
     ExtractContentEvent,
     DefaultFormat,
 } from 'roosterjs-editor-types';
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-plugins/lib/Watermark/Watermark.ts
```diff
@@ -70,10 +68,12 @@ export default class Watermark implements EditorPlugin {
      * @param event PluginEvent object
      */
     onPluginEvent(event: PluginEvent) {
-        if (event.eventType == PluginEventType.ContentChanged) {
+        if (event.eventType == PluginEventType.EditorReady) {
+            this.showHideWatermark(false /*ignoreCachedState*/);
+        } else if (event.eventType == PluginEventType.ContentChanged) {
             // When content is changed from setContent() API, current cached state
             // may not be accurate, so we ignore it
-            this.showHideWatermark((<ContentChangedEvent>event).source == ChangeSource.SetContent);
```
Lego6245: Nice cleanup of the cast here.

PR #386, opened 2020-02-05
Support Projection

JiuqingSong:
Editor Support Projection
1. add "roosterjs-cross-window" package and utils to support cross window type check
2. add related test cases
3. replace "instanceof [Native type]" with the type utils
4. replace "instanceof Array" with "Array.isArray()"
5. stop using global window and document
6. add popout feature in demo site

### DIFF
--- a/packages/roosterjs-cross-window/README.md
+++ b/packages/roosterjs-cross-window/README.md
@@ -1,0 +1,5 @@
+# roosterjs-plugin-picker
+
+The Picker Plugin is designed to support scenarios where you'd want to type a trigger character, for instance, an @, and then mention someone. You can see this on Github, where typing an @ opens a drop down menu with a list of people that you can mention. It also works for the # symbol as well, as well as for any self defined prefix.
+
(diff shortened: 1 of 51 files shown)


packages/roosterjs-cross-window/README.md
```diff
@@ -0,0 +1,5 @@
+# roosterjs-plugin-picker
```
Lego6245: Remove or update :)

JiuqingSong:
> This all up SGTM. Just a general question, why the change to make these helper type checkers? Were those required for popout, and if so, why?

Yes, it is required for popout projection.

Consider this:
var childWindow = window.open(...);
var node = childWindow.document.createElement('DIV');
var result1 = node instanceof Node;
var result2 = node instanceof childWindow.Node;

In Chrome, result1 will be false and result2 will be true. Because window.Node and childWindow.Node are different types. So I made these helper functions to support check like "node instance of childWindow.Node"

Firefox returns true for both, fyi.

Lego6245:
wow, gross! good plan to change the check. is the fact it doesn't type check intentional in chrome, AFAIK?

PR #387, opened 2020-02-09
Trigger content change after adding/removing resize div.

iendeavor:


JiuqingSong:
Why do we need to trigger event when resize handling added/removed? 
The resize handle is temporary content and will be removed when getContent(), so adding/removing resize handle is not a real "content change"

iendeavor:
It's my fault. I use innerHtml directly in my project.

PR #389, opened 2020-03-05
Fix cross window typing issue.

Lego6245:
1) Elements created in the parent window and inserted into the child window will fail our current type checks. Instead, let's check both the parent and the child window's type.
2) Make the check in `contains` a positive check instead of a negative check.


Lego6245:
I'll bump it in a separate PR for cleanliness.

Issue #380, opened 2020-01-09
Hyperlinks absorb all text/images pasted next to them


racbatch:
If text or images are pasted directly next to a hyperlink, the pasted material will become part of the hyperlink instead of remaining stand-alone material.

Can reproduce in Chrome, but not Edge, Firefox, or Internet Explorer.

Example: if I pasted "some text" just before www.google.com I'll have "some textwww.google.com" as a single hyperlink, while the link itself will still point to www.google.com

This can create confusing scenarios if you are pasting unrelated material next to a hyperlink, or in scenarios where pasted material / hyperlinks are supposed to have additional formatting.

There is a work-around of adding spaces every time you paste near a link, but it's rather tedious and can be easily forgotten by users.

Issue #349, opened 2019-08-23
ToggleHeader doesn't remove current font-size


kemiporto:
Once the font size is set, calling toggleHeader doesn't remove it, making the text keep the original size instead of header size

**To Reproduce**
1. Go to https://microsoft.github.io/roosterjs/
2. Set up font size and name
3. Write some text
4. select some Heading style

**Expected behavior**
The text should keep the font name but change the font to use the selected header size

PR #392, opened 2020-03-10
7.10.3: Fix #388, #380, #349

### LINKED ISSUE #388: HTML sanitizer bypass
**Describe the bug**
HTML sanitizer in RoosterJS can be bypassed.

**To Reproduce**
1. Go to https://microsoft.github.io/roosterjs/index.html
2. Click on show side pane
3. Click on API Playground
4. Select HTML sanitizer
5. Add following to the input
```
<a href="javascript
:alert(1)">test</a>
```
6. Click on Sanitize button
7. JavaScript URL is still there

**Expected behavior**
Empty anchor tag

**Device Information**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 80

JiuqingSong:
#349 ToggleHeader doesn't remove current font-size
Currently we only remove 'font-size' for selected inlineElement. But we actually should do this for all related BlockElements, and use HtmlSanitizer to help loop each elements

#380 Hyperlinks absorb all text/images pasted next to them
We already have a similar fix for inserting text for Firefox, we just need to also apply that fix to paste event in all browsers

#388 HTML sanitizer bypass
Add a XSS check to the link url when insert link using HtmlSanitizer

Bump version to 7.10.3

### DIFF
--- a/packages/roosterjs-editor-api/lib/format/createLink.ts
+++ b/packages/roosterjs-editor-api/lib/format/createLink.ts
@@ -1,5 +1,6 @@
 import { ChangeSource, DocumentCommand, QueryScope } from 'roosterjs-editor-types';
 import { Editor } from 'roosterjs-editor-core';
+import { HtmlSanitizer } from 'roosterjs-html-sanitizer';
 import { matchLink } from 'roosterjs-editor-dom';
 
 // Regex matching Uri scheme
@@ -54,7 +55,7 @@ export default function createLink(
     displayText?: string
 ) {
     editor.focus();
-    let url = link ? link.trim() : '';
+    let url = (checkXss(link) || '').trim();
     if (url) {
         let linkData = matchLink(url);
         // matchLink can match most links, but not all, i.e. if you pass link a link as "abc", it won't match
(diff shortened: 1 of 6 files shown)


packages/roosterjs-editor-api/lib/format/createLink.ts
```diff
@@ -54,7 +55,7 @@ export default function createLink(
     displayText?: string
 ) {
     editor.focus();
-    let url = link ? link.trim() : '';
+    let url = (checkXss(link) || '').trim();
```
titrongt: Add back the check to make sure link is defined? `checkXss `should work anyways, but would prevent a bit of unnecessary work

packages/roosterjs-editor-core/lib/corePlugins/FirefoxTypeAfterLink.ts
```diff
@@ -1,13 +1,15 @@
...
 import { PluginEvent, PluginEventType, PositionType } from 'roosterjs-editor-types';
 
 /**
  * FirefoxTypeAfterLink Component helps handle typing event when cursor is right after a link.
- * When typing after a link, Firefox will always put the new charactor inside link.
- * This plugin overrides this behavior to make it consistent with other browsers.
+ * When typing/pasting after a link, browser may put the new charactor inside link.
+ * This plugin overrides this behavior to always insert outside of link.
+ *
+ * TODO: Rename this file in next major release since it is not only applied to Firefox now
```
titrongt: Is this comment up to date? I see there is a check that makes sure this is Firefox

JiuqingSong: The keyPress event needs isFirefox check. PasteEvent should apply to all browsers

packages/roosterjs-editor-api/lib/format/createLink.ts
JiuqingSong: I think they both work, no difference.

titrongt:
A few minor comments but overall looks good to me 👍🏼

PR #395, opened 2020-03-12
7.0.6: Support more types for safeIntanceOf

JiuqingSong:
And export safeInstanceOf so it can be used externally.


JiuqingSong:
@Lego6245 , just added a tool to help generate the interface file.

PR #399, opened 2020-04-15
Fix bullet style loss when exiting a list

Lego6245:
Chromium has a bug (https://bugs.chromium.org/p/chromium/issues/detail?id=944178#c4) which nukes a bunch of stuff when you outdent. We added some logic in a prior PR to attempt to clone the node when we outdent in a list.

This PR further extends this work to
1) Apply when enter is pressed on the end of the list (in an attempt to preserve style information in these cases)
2) Append the image element to the end instead of to the beginning to prevent Chromium from thinking that the span element should be removed due to no content.

### DIFF
--- a/packages/roosterjs-editor-api/lib/utils/processList.ts
+++ b/packages/roosterjs-editor-api/lib/utils/processList.ts
@@ -29,7 +29,7 @@ export default function processList(
 ): Node {
     let clonedNode: Node;
     let relativeSelectionPath;
-    if (Browser.isChrome && command != DocumentCommand.Indent) {
+    if (Browser.isChrome && command == DocumentCommand.Outdent) {
         const parentLINode = editor.getElementAtCursor('LI');
         if (parentLINode) {
             let currentRange = editor.getSelectionRange();
--- a/packages/roosterjs-editor-api/lib/utils/processList.ts
+++ b/packages/roosterjs-editor-api/lib/utils/processList.ts
@@ -1,9 +1,9 @@
+import { Browser, getRangeFromSelectionPath, getSelectionPath } from 'roosterjs-editor-dom';
 import { DocumentCommand } from 'roosterjs-editor-types';
 import { Editor } from 'roosterjs-editor-core';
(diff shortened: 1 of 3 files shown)


packages/roosterjs-editor-api/lib/utils/processList.ts
```diff
@@ -39,65 +34,61 @@ export default function processList(
...
+        } else if (clonedCursorNode) {
+            // Rooster should never be creating FONT tags on it's own,
+            // and chromium's behavior is consistant with empty nodes outdenting to a non list block element root.
+            const chromeFontTag = editor.getElementAtCursor('FONT');
+            if (chromeFontTag) {
+                chromeFontTag.parentNode.replaceChild(clonedCursorNode, chromeFontTag);
+                if (
+                    cursorSelectionPath &&
+                    isHTMLElement(clonedCursorNode) &&
+                    editor.getDocument().body.contains(clonedCursorNode)
```
JiuqingSong, line 85: Suggest use editor.contains(...)

packages/roosterjs-editor-api/lib/utils/processList.ts
```diff
@@ -39,65 +34,61 @@ export default function processList(
...
-    if (newList && clonedNode && newParentNode) {
-        // if the clonedNode and the newLIParent share the same tag name
-        // we can 1:1 swap them
-        if (isHTMLElement(clonedNode)) {
-            if (
-                isHTMLElement(newParentNode) &&
-                clonedNode.tagName == (<HTMLElement>newParentNode).tagName
-            ) {
-                newList.replaceChild(clonedNode, newParentNode);
+    if (Browser.isChrome) {
```
JiuqingSong, line 58: So this line will cause this block only runs for Chrome. It was running for all browers before. Is that ok?

Lego6245: Technically, it would only run for chrome earlier, as `clonedNode` would never be set unless the browser was chromium.

PR #401, opened 2020-04-17
7.11.1 Fix bug in Editor.insertContent

JiuqingSong:
When call Editor.insertContent() with HTML represents multiple HTML nodes, insertContent() will insert them in a reversed order, because we insert nodes one by one but cursor is not changed.

The fix is to create a fragment and put all nodes under it first, then insert the fragment with a single insertNode() call.

### DIFF
--- a/packages/roosterjs-editor-core/lib/test/editor/EditorTest.ts
+++ b/packages/roosterjs-editor-core/lib/test/editor/EditorTest.ts
@@ -92,6 +92,19 @@ describe('Editor insertContent()', () => {
         expect(editor.getContent()).toBe('<div id="text">hellotext</div>');
     });
 
+    it('insert begin with multiple nodes', () => {
+        // Act
+        editor.insertContent('hello<br>world', {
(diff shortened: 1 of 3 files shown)


packages/roosterjs-editor-core/lib/test/editor/EditorTest.ts
```diff
@@ -92,6 +92,19 @@ describe('Editor insertContent()', () => {
...
+        // Act
+        editor.insertContent('hello<br>world', {
+            position: ContentPosition.Begin,
+            updateCursor: true,
+            replaceSelection: true,
+            insertOnNewLine: false,
+        });
+
+        // Assert
+        expect(editor.getContent()).toBe('<div id="text">hello<br>worldtext</div>');
```
Lego6245, line 105: is that `text` supposed to be there? The test passes with it?

JiuqingSong: Yes. See the test case above, it also has a "text"

PR #404, opened 2020-05-01
Fix potential null in tableFeatures

Lego6245:
`editor.getFocusedPosition()` can return a null value. Without checking for it, we throw an exception inside of `cacheGetTableCell`. Adding an explicit null check should help.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/ContentEdit/features/tableFeatures.ts
+++ b/packages/roosterjs-editor-plugins/lib/ContentEdit/features/tableFeatures.ts
@@ -1,11 +1,5 @@
 import { cacheGetEventData, ContentEditFeature, Editor, Keys } from 'roosterjs-editor-core';
-import {
-    contains,
-    getTagOfNode,
-    isVoidHtmlElement,
-    Position,
-    VTable
-    } from 'roosterjs-editor-dom';
+import { contains, getTagOfNode, isVoidHtmlElement, Position, VTable } from 'roosterjs-editor-dom';
 import { NodeType, PluginEvent, PositionType } from 'roosterjs-editor-types';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-plugins/lib/ContentEdit/features/tableFeatures.ts
```diff
@@ -98,7 +92,9 @@ export const UpDownInTable: ContentEditFeature = {
 function cacheGetTableCell(event: PluginEvent, editor: Editor): HTMLTableCellElement {
     return cacheGetEventData(event, 'TABLECELL_FOR_TABLE_FEATURES', () => {
         let pos = editor.getFocusedPosition();
-        let firstTd = editor.getElementAtCursor('TD,TH,LI', pos.node);
-        return getTagOfNode(firstTd) == 'LI' ? null : (firstTd as HTMLTableCellElement);
+        let firstTd = pos && editor.getElementAtCursor('TD,TH,LI', pos.node);
+        return (
+            firstTd && (getTagOfNode(firstTd) == 'LI' ? null : (firstTd as HTMLTableCellElement))
```
JiuqingSong, line 97: When firstTd is null, getTagOfNode(firstTd) will return null. So the function will return null correctly.

Lego6245: You're right, but I was thinking about strict mode, and the current signature of `getTagOfNode` takes in a non null value. Granted, could expand that check to also take in a null value, or remove this check entirely. Up to you.

JiuqingSong: Ok. I can approve the change since it can make the code more clear.

PR #406, opened 2020-05-05
Upgrade dependencies

JiuqingSong:


JiuqingSong:
> Overall SGTM, but I'm noticing that in a few cases we're removing `^` from our semver requirements. Any particular reason why?

I just added them back.

PR #413, opened 2020-05-14
fix pasting issue in table

FrancisMengx:
When copy bullet in tables from word online. 
The issue is caused by when splitting for listblock. The root node is used as body, instead it should be the parent of the list block.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/Paste/officeOnlineConverter/convertPastedContentFromWordOnline.ts
+++ b/packages/roosterjs-editor-plugins/lib/Paste/officeOnlineConverter/convertPastedContentFromWordOnline.ts
@@ -4,7 +4,7 @@ import {
     WORD_ONLINE_IDENTIFYING_SELECTOR,
     LIST_CONTAINER_ELEMENT_CLASS_NAME,
     ORDERED_LIST_TAG_NAME,
-    UNORDERED_LIST_TAG_NAME
+    UNORDERED_LIST_TAG_NAME,
 } from './constants';
 
 import {
@@ -13,11 +13,10 @@ import {
     getFirstLeafNode,
     getTagOfNode,
     collapseNodes,
-    unwrap
+    unwrap,
 } from 'roosterjs-editor-dom';
 
 import ListItemBlock, { createListItemBlock } from './ListItemBlock';
-
 
 export function isWordOnlineWithList(node: HTMLElement): boolean {
     return !!(node && node.querySelector(WORD_ONLINE_IDENTIFYING_SELECTOR));
@@ -62,7 +61,6 @@ export default function convertPastedContentFromWordOnline(doc: HTMLDocument) {
     const listItemBlocks: ListItemBlock[] = getListItemBlocks(doc);
 
     listItemBlocks.forEach((itemBlock) => {
-
         // There are cases where consecutive List Elements are seperated into different divs:
         // <div>
         //   <div>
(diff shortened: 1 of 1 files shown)


JiuqingSong:
I only see format changes. Where is your real change?

JiuqingSong:
Please fix the build error

JiuqingSong:
Can you also add a test case about this issue?

packages/roosterjs-editor-plugins/lib/test/fragmentHandlerTest.ts
```diff
@@ -118,6 +118,26 @@ describe('fragmentHandler', () => {
...
+                    true,
+                    '  <div class="OutlineElement Ltr BCX0 SCXW244795937"> <div class="TableContainer SCXW244795937 BCX0"> <table> <tbody> <tr> <td> <div> <div class="OutlineElement Ltr BCX0 SCXW32709461"> <p><span><span>asdfasdf</span></span><span></span></p> </div> </div> </td> <td> <div> <div class="OutlineElement Ltr BCX0 SCXW32709461"> <p><span><span>asdfasdf222</span></span><span></span></p> </div> </div> </td> </tr> <tr> <td> <div> <div class="ListContainerWrapper SCXW244795937 BCX0"> <ul> <li role="listitem" data-aria-level="1" class="OutlineElement Ltr BCX0 SCXW244795937" style="margin: 0px 0px 0px 24px;"> A </li> </ul> </div> </div> </td> <td> <div> <div class="ListContainerWrapper SCXW244795937 BCX0"> <ul> <li role="listitem" data-aria-level="1" class="OutlineElement Ltr BCX0 SCXW244795937" style="margin: 0px 0px 0px 24px;"> A </li> </ul> </div> </div> </td> </tr> </tbody> </table> </div> </div> '
```
JiuqingSong, line 138: Is this the minimized repro? We should keep the test case as small as possible. So that another day when we make some change breaks the case, we can understand why it is broken and how to fix.

FrancisMengx: Just simplified the html removed some unecessary classnames and styles.

packages/roosterjs-editor-plugins/lib/test/fragmentHandlerTest.ts
```diff
@@ -118,6 +118,26 @@ describe('fragmentHandler', () => {
...
+            //   .d
+            //text text
+            // -------------- --------------
+            //| text text    | text text    |
+            // -------------- --------------
+            //| .a           | .a           |
+            // -------------- --------------
+            it('fragments contains text, list and table that consist of list', () => {
+                runTest(
+                    '<html><body><div class="OutlineElement"><div class="TableContainer"><table><tbody><tr><td><div><div class="OutlineElement"><p><span><span>asdfasdf</span></span><span></span></p></div></div></td><td><div><div class="OutlineElement"><p><span><span>asdfasdf222</span></span><span></span></p></div></div></td></tr><tr><td><div><div class="ListContainerWrapper"><ul><li role="listitem" data-aria-level="1" class="OutlineElement">A</li></ul></div></div></td><td><div><div class="ListContainerWrapper"><ul><li role="listitem" data-aria-level="1" class="OutlineElement">A</li></ul></div></div></td></tr></tbody></table></div></div></body></html>',
```
JiuqingSong, line 136: I believe you can simplify it more. Such as 
```html
<p><span><span>asdfasdf</span></span><span></span></p>
```
should be simplified to
```html
<p>asdfasdf</p>
```
Same for others

PR #420, opened 2020-05-26
VList preparation Step 3: Region

JiuqingSong:
This change introduces a new concept of editor: Region.

Consider a DOM structure:
```html
<div id="outerDiv">
<table>
  <tr><td>TD1.1</td><td>TD1.2</td></tr>
  <tr><td>TD1.3</td><td>TD1.4</td></tr>
</table>
<div>Line 1</div>
<div>Line 2</div>
<table>
  <tr><td>TD2.1</td><td>TD2.2</td></tr>
  <tr><td>TD2.3</td><td>TD2.4</td></tr>
</table>
</div>
```
And user selects from TD1.3 to TD2.2, it will contain 6 elements:

- TD1.3
- TD1.4
- Line 1
- Line 2
- TD2.1
- TD2.2

And user wants to toggle to bullet list. In that case, we should not break the table structure, so the operation should be splitted into 5 regions:

1. TD1.3
2. TD1.4
3. Line 1 and Line 2
4. TD2.1
5. TD2.2

A region is defined by:

- a root node (outerDiv in the example above)
- a boundary node before the region content
- a boundary node after the region content
- a full selection range used for determine the selection range in this region

The two boundary nodes can be null, which means the region starts/ends at the beginning/ending of root node.

This change adds the following new interfaces and APIs:

1. Region: The interface of region
2. RegionType: Types of region. Currently only Table region is supported.
3. getRegionsFromRange: An API to create regions from a selection range under a root node
4. getSelectedBlockElementsInRegion: An API to get selected block elements from a region which is selected by a full selection
5. isNodeInRegion: An API to check if a given node is in the given region
6. collapseNodesInRegion: An API to collapse nodes under a region
7. Editor.getSelectedRegions: A method of Editor to get selected regions.

Related test cases added.

### DIFF
--- a/packages/roosterjs-editor-dom/lib/region/getRegionsFromRange.ts
+++ b/packages/roosterjs-editor-dom/lib/region/getRegionsFromRange.ts
@@ -1,0 +1,209 @@
+import contains from '../utils/contains';
+import findClosestElementAncestor from '../utils/findClosestElementAncestor';
+import Position from '../selection/Position';
+import queryElements from '../utils/queryElements';
+import regionTypeData from './regionTypeData';
+import { getNextLeafSibling, getPreviousLeafSibling } from '../utils/getLeafSibling';
+import { QueryScope, Region, RegionType } from 'roosterjs-editor-types';
+
+/**
+ * Get regions impacted by the given range under the root node
+ * @param root Root node to get regions from
+ * @param range A selection range. Regions will be created acording to this range. Each region will be
+ * fully or partially covered by this range.
+ * @param type Type of region. Currently we only support TABLE region.
+ */
+export default function getRegionsFromRange(
+    root: HTMLElement,
+    range: Range,
+    type: RegionType
+): Region[] {
+    let regions: Region[] = [];
+    if (root && range) {
+        const { innerSelector, skipTags } = regionTypeData[type];
+        const boundaryTree = buildBoundaryTree(root, range, type);
+        const start = findClosestElementAncestor(range.startContainer, root, innerSelector) || root;
+        const end = findClosestElementAncestor(range.endContainer, root, innerSelector) || root;
+        const creator = getRegionCreator(range, skipTags);
+        [regions] = iterateNodes(creator, boundaryTree, start, end);
+    }
+
+    return regions.filter(r => !!r);
+}
+
+/**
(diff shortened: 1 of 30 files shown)


Lego6245:
It's not 100% clear to me what the end result will be. Will it be 5 distinct lists, one for each region? I.E. operations now take place region by region?

packages/roosterjs-editor-dom/lib/region/getRegionsFromRange.ts
```diff
@@ -0,0 +1,209 @@
...
+        return false;
+    } else {
+        const firstNodeOfRegion = nodeBefore && getNextLeafSibling(root, nodeBefore, skipTags);
+        const lastNodeOfRegion = nodeAfter && getPreviousLeafSibling(root, nodeAfter, skipTags);
+        const firstNodeValid =
+            !nodeBefore || (contains(root, nodeBefore) && contains(root, firstNodeOfRegion));
+        const lastNodeValid =
+            !nodeAfter || (contains(root, nodeAfter) && contains(root, lastNodeOfRegion));
+        const bothValid =
+            !nodeBefore ||
```
Lego6245, line 201: Does the bothValid check need to re-check if `nodeBefore` and `nodeAfter` are falsy? Aren't those covered by the previous two checks already, and since all the values are `&&`, you already know whether or not those values are or are not falsy?

JiuqingSong:
> It's not 100% clear to me what the end result will be. Will it be 5 distinct lists, one for each region? I.E. operations now take place region by region?

Yes, it will resut in 5 distinct lists. That is also the behavior of execCommand().

packages/roosterjs-editor-dom/lib/region/getRegionsFromRange.ts
JiuqingSong: Without checking falsy for nodeBefore and nodeAfter, if one of them is null, contains() will return false, then bothValid will be false, so the final result is also false.

Lego6245: But if one of them is null, then `!null` will be true, and the whole boolean check will short circuit and `bothValid` will be true, at least as written.

PR #429, opened 2020-06-12
Initial commit of P0 Markdown features for Rooster

ctuck04:
The Markdown features include:
-- Bold
-- Underline
-- Italic
-- Strikethrough


packages/roosterjs-editor-core/lib/interfaces/ContentEditFeature.ts
```diff
@@ -14,14 +14,18 @@ export const enum Keys {
...
     I = 73,
     U = 85,
     Y = 89,
     Z = 90,
+    PLUS = 187,
     COMMA = 188,
+    UNDERSCORE = 189,
     PERIOD = 190,
     FORWARDSLASH = 191,
+    SQUIGGLY = 192,
```
Lego6245, line 28: nit: should be tildehttps://en.wikipedia.org/wiki/Tilde

packages/roosterjs-editor-plugins/lib/ContentEdit/ContentEditFeatures.ts
```diff
@@ -131,5 +139,9 @@ export function getDefaultContentEditFeatures(): ContentEditFeatures {
         noCycleCursorMove: Browser.isChrome,
         smartOrderedList: false,
         smartOrderedListStyles: ['lower-alpha', 'lower-roman', 'decimal'],
+        bold: true,
```
Lego6245, line 142: we probably would not include these in the default enabled set, and instead pass them in to explicitly enable them

packages/roosterjs-editor-plugins/lib/ContentEdit/features/autoLinkFeatures.ts
```diff
@@ -12,7 +12,7 @@ import {
     Editor,
     cacheGetEventData,
     cacheGetContentSearcher,
-    clearContentSearcherCache,
```
Lego6245, line 15: nit: clean this up

packages/roosterjs-editor-plugins/lib/ContentEdit/features/autoLinkFeatures.ts
```diff
@@ -114,7 +114,7 @@ function autoLink(event: PluginEvent, editor: Editor) {
             replaceWithNode(editor, linkData.originalUrl, anchor, false /* exactMatch */, searcher);
 
             // The content at cursor has changed. Should also clear the cursor data cache
-            clearContentSearcherCache(event);
```
Lego6245, line 117: nit: revert?

packages/roosterjs-editor-plugins/lib/ContentEdit/features/markdownFeatures.ts
```diff
@@ -0,0 +1,127 @@
...
+        anchorNode.textContent.lastIndexOf(triggerChar) == 0 ||
+        anchorNode.textContent.lastIndexOf(triggerChar) == 1
+    ) {
+        genericTriggerCharacterIndex = anchorNode.textContent.lastIndexOf(triggerChar);
+    } else if (anchorNode.textContent.lastIndexOf(' ' + triggerChar) >= 0) {
+        genericTriggerCharacterIndex = anchorNode.textContent.lastIndexOf(' ' + triggerChar);
+    }
+
+    let firstWhiteSpaceChecker = anchorNode.textContent.charAt(
+        genericTriggerCharacterIndex + stringAdjust
```
Lego6245, line 31: why is stringAdjust a variable here when it's being used as a const

packages/roosterjs-editor-plugins/lib/ContentEdit/features/markdownFeatures.ts
```diff
@@ -0,0 +1,127 @@
...
+    event: PluginKeyboardEvent,
+    editor: Editor
+) => {
+    let selObj = window.getSelection();
+    let anchorNode = selObj.anchorNode;
+    let genericTriggerCharacterIndex = anchorNode.textContent.lastIndexOf(' ' + triggerChar);
+    let alterationRequired = false;
+    let stringAdjust = 2;
+
+    if (
```
Lego6245, line 21: compute this value once, and check it each time. goes for both of these conditionals

packages/roosterjs-editor-plugins/lib/ContentEdit/features/markdownFeatures.ts
```diff
@@ -0,0 +1,127 @@
...
+) => {
+    let selObj = window.getSelection();
+    let anchorNode = selObj.anchorNode;
+    let stringToBeAltered = new Range();
+    let genericTriggerCharacterRange = new Range();
+    let stringAdjust = 1;
+    let scenario = 'N/A';
+    let genericTriggerCharacterIndex;
+
+    if (
```
Lego6245, line 53: could probably pull this out into its own function

packages/roosterjs-editor-plugins/lib/ContentEdit/features/markdownFeatures.ts
```diff
@@ -0,0 +1,127 @@
...
+    } else if (anchorNode.textContent.lastIndexOf(' ' + triggerChar) >= 0) {
+        genericTriggerCharacterIndex = anchorNode.textContent.lastIndexOf(' ' + triggerChar);
+        scenario = 'mid';
+    }
+
+    if (scenario == 'begin') {
+        stringToBeAltered.setStart(anchorNode, genericTriggerCharacterIndex + stringAdjust);
+        stringToBeAltered.setEnd(anchorNode, anchorNode.textContent.length);
+
+        genericTriggerCharacterRange.setStart(anchorNode, genericTriggerCharacterIndex);
```
Lego6245, line 68: `anchorNode.textContent.replaceAll` might be simpler here and below

packages/roosterjs-editor-plugins/lib/ContentEdit/features/markdownFeatures.ts
```diff
@@ -0,0 +1,127 @@
...
+        anchorNode.textContent.lastIndexOf(triggerChar) == 1
+    ) {
+        genericTriggerCharacterIndex = anchorNode.textContent.lastIndexOf(triggerChar);
+        scenario = 'begin';
+    } else if (anchorNode.textContent.lastIndexOf(' ' + triggerChar) >= 0) {
+        genericTriggerCharacterIndex = anchorNode.textContent.lastIndexOf(' ' + triggerChar);
+        scenario = 'mid';
+    }
+
+    if (scenario == 'begin') {
```
Lego6245, line 64: could move this if block into the above if block

packages/roosterjs-editor-plugins/lib/ContentEdit/features/markdownFeatures.ts
```diff
@@ -0,0 +1,127 @@
...
+    handleEvent: applyStyle('+', 'u'),
+};
+
+export const Strikethrough: ContentEditFeature = {
+    keys: [Keys.SQUIGGLY],
+    shouldHandleEvent: findTriggerCharactersAndCheckBetween('~'),
+    handleEvent: applyStyle('~', 'strike'),
+};
+
+// 6/4/2020 1:16 PM PDT
```
Lego6245, line 125: nit: clean this up

packages/roosterjs-editor-plugins/lib/ContentEdit/features/markdownFeatures.ts
ctuck04: I'm getting "Property 'replaceAll' does not exist on type 'string'."

Lego6245: whoops, `replaceAll` isn't in all browsers: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll

Lego6245: `replace` should work just as well (replace and replace all both do the same thing pretty much)

Lego6245: you could use `slice` here alternatively, or leave as is. not a big deal either way, just think it's cleaner than making a range and deleting its contents

PR #435, opened 2020-06-22
Preparation step for Entity API

JiuqingSong:
This is a preparation step for upcoming Entity API change.

1. Add ExtractContentWithDomEvent, and mark ExtractContentEvent as deprecated
2. Editor.getContent() will also trigger ExtractContentWithDomEvent
3. Refactor some dark mode code to better adopt ExtractContentWithDomEvent
4. Add an option to HtmlSanitizer to allow specify preserved CSS class names


JiuqingSong:
> All SGTM, good changes overall. Concerned slightly that now since normalizing dark mode content is removed from the Editor.ts and moved to CopyPlugin, that dark mode now has an implicit dependency on CopyPlugin to "work" as intended.

CopyPlugin (later it will be renamed to DarkModePlugin) is a core plugin which means it is always added and can't be replaced.

PR #437, opened 2020-06-22
7.13.0 Entity API and demo site change

JiuqingSong:
1. Add Entity API page for demo site
2. Add a catch block in build script to ignore publish failure
3. Set the final version number for deprecated packages

### DIFF
--- a/packages/roosterjs-cross-window/package.json
+++ b/packages/roosterjs-cross-window/package.json
@@ -6,5 +6,5 @@
         "roosterjs-editor-dom": ""
     },
     "main": "./lib/index.ts",
-    "version": "1.2.0"
+    "version": "1.4.0"
 }
--- a/packages/roosterjs-html-sanitizer/package.json
+++ b/packages/roosterjs-html-sanitizer/package.json
@@ -6,5 +6,5 @@
         "roosterjs-editor-types": ""
     },
     "main": "./lib/index.ts",
-    "version": "1.3.0"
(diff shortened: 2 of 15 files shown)


packages/roosterjs-cross-window/package.json
```diff
@@ -6,5 +6,5 @@
         "roosterjs-editor-dom": ""
     },
     "main": "./lib/index.ts",
-    "version": "1.2.0"
+    "version": "1.4.0"
```
JiuqingSong, line 9: I published a 1.3.0 before but I realized it doesn't work since new version for main packages are not published yet. So I deprecated it. Now we need to use 1.4.0 instead

packages/roosterjs-html-sanitizer/package.json
```diff
@@ -6,5 +6,5 @@
         "roosterjs-editor-types": ""
     },
     "main": "./lib/index.ts",
-    "version": "1.3.0"
+    "version": "1.5.0"
 }
```
JiuqingSong, line 10: same here

tools/build.js
```diff
@@ -436,6 +436,9 @@ function publish() {
                     stdio: 'inherit',
                     cwd: path.join(distPath, package),
                 });
+            } catch (e) {
```
Lego6245, line 439: This was the only functional change in this file, yes?

JiuqingSong: Yes.

PR #444, opened 2020-07-06
7.13.2 Fix some entity bugs

JiuqingSong:
Fix two null-check issues
Don't set display to "inline-block" for editable entity since it will cause a resize border in IE
Modify the build script to respect the explicitly specified dependency version
Modify the dependency versions to make a final version of deprecated packages

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/Entity/EntityFeatures.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/Entity/EntityFeatures.ts
@@ -156,7 +156,7 @@ function cacheGetNeighborEntityElement(
 
                 if (checkForSameLine) {
                     const block = editor.getBlockElementAtNode(pos.node);
-                    if (!block.contains(node)) {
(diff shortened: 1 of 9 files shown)


packages/roosterjs-editor-plugins/lib/plugins/Entity/EntityFeatures.ts
```diff
@@ -156,7 +156,7 @@ function cacheGetNeighborEntityElement(
 
                 if (checkForSameLine) {
                     const block = editor.getBlockElementAtNode(pos.node);
-                    if (!block.contains(node)) {
+                    if (!block || !block.contains(node)) {
```
Lego6245, line 159: Just want to sanity check that you intend to fail this check if block is falsy.

JiuqingSong: Yes. If editor content is totally empty, not even an empty DIV there, block will be null. In that case we should return null since there's no entity at all.

PR #446, opened 2020-07-07
PickerPlugin for low version Android Webview

Rain-Zheng:
### Context

The mention functionality in Outlook Android does not work properly in WebView version below 51. We found the root cause is _[PickerPlugin](https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts)_ uses `event.rawEvent.key == UNIDENTIFIED_KEY` to identify keyboard events on Android. In low version WebView, 'KeyboardEvent.key' is not supported and always returns 'undefined'. 

### Solution

Besides 'KeyboardEvent.key', we can use 'KeyboardEvent.charCode' property for low version WebView. This property is 0 or 229 on Android devices. So we check this property in `onPluginEvent()` and `shouldHandleKeyUpEvent()`.


packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -180,7 +182,12 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
 
             case PluginEventType.KeyDown:
                 this.eventHandledOnKeyDown = false;
-                if (event.rawEvent.key == UNIDENTIFIED_KEY) {
+                // On Android Webview below 51, the key for KeyboardEvent is not supported and always returns undefined,
+                // so using the charCode property, which is 0 or 229.
+                const isLowVersionAndroid =
+                    event.rawEvent.key == undefined &&
+                    UNIDENTIFIED_CODE.indexOf(event.rawEvent.charCode) > -1;
+                if (event.rawEvent.key == UNIDENTIFIED_KEY || isLowVersionAndroid) {
```
tito0224, line 190: Since this check is used in two places should move it to a private method with a good name and comment?

tito0224:
@Rain-Zheng Can you point to the ADO ticket for this work? I'd like to understand why we need to support such an old version and how many users we have on it. Thanks

Rain-Zheng:
> @Rain-Zheng Can you point to the ADO ticket for this work? I'd like to understand why we need to support such an old version and how many users we have on it. Thanks

The ticket number is 4221282.

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
Rain-Zheng: `isAndroidKeyboardEvent()` method is added to check Android keyboard events.

tito0224:
This is ok with me. I don't have the ability to approve PRs here but you have my :+1:

Rain-Zheng:
> This is ok with me. I don't have the ability to approve PRs here but you have my 👍

Thanks for your review!

@JiuqingSong Could you help approve this PR? Thank you!

JiuqingSong:
Please change version number in package.json in root folder to "7.14.1" so that once you check in, new package will be published automatically

Issue #450, opened 2020-07-08
With enableExperimentFeatures=true, Enter in a bullet/number list doesn't start new list item in Safari


JiuqingSong:
1. Load demo site (https://microsoft.github.io/roosterjs/index.html) in Safari
2. Make sure "enableExperimentFeatures" is checked from side pane, editor option tab.
3. type "1. " to start a new list, and type some text
4. press Enter

Expect: 
see "2." in next line

Actual:
Still in "1." but in next line

JiuqingSong:
Fixed in 8.0.0-beta.1

PR #454, opened 2020-07-21
U/melvsparks/safari enter on list bug

melvsparks:
### Problem
**Note:** Some tests are failing - in the process of updating the tests
Creating new rows with "Enter" key in numbered or bulleted list in Safari doesn't work.
Fixes Issue #450 

In Safari, the problem with how it handles how our lists are created. Chrome is able to recognize the div wrapper as a list element and can handle the "enter" input correctly. In Safari, the extra div causes problems.

```
<ol>
  <li>
      <div>test</div>
  </li>
</ol>
```
 
### What was done
The solution I chose, was to restructure the VList so that we aren't adding an unnecessary div when appending items to the list.

**After changes**
gif2

### What other solutions were considered/attempted
The first solution I tried was to manually handle the new line creation when an ENTER event is triggered. I chose this approach to start because it, initially, made sense to be able to have full control of that scenario. The once I got into the nitty gritty implementation details of this is became extremely complex. There were a lot of different cases that would need to be handled and tested (cursor in different places, text being selected before the ENTER, inline tables, nested lists within tables, sublist row items, etc). Once the real complexity of this surfaced, it didn't make sense to re-write all that logic that's already built into the browser that we should be getting for free.


packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -207,8 +207,20 @@ export default class VList {
...
+        let nodeToAppend: Node;
+        if (getTagOfNode(node) != 'LI') {
+            nodeToAppend = document.createElement('LI');
+
+            while (node.hasChildNodes()) {
+                nodeToAppend.appendChild(node.firstChild);
+            }
+        } else {
+            nodeToAppend = node;
+        }
```
JiuqingSong, line 219: Consider use util function "changeElementTag" https://github.com/microsoft/roosterjs/blob/d246c0c/packages/roosterjs-editor-dom/lib/utils/changeElementTag.ts#L10

melvsparks: Great suggestion, I'll do that

JiuqingSong: One more concern here. What if the original tag is a structured node, such as a TABLE, in that case we should not change its tag name. What do you think?

melvsparks: Yea, I agree

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -207,8 +208,16 @@ export default class VList {
      * @param type Type of this list item, can be ListType.None
      */
     appendItem(node: Node, type: ListType) {
-        node = getTagOfNode(node) == 'LI' ? node : wrap(node, 'li');
-        this.items.push(type == ListType.None ? new VListItem(node) : new VListItem(node, type));
+        let newListNode = node;
+        const nodeTag = getTagOfNode(node);
+
+        if (nodeTag != 'LI' && nodeTag != 'TABLE') {
```
JiuqingSong, line 214: For table, we still need to wrap it with LI, right?
Please add a test case for table.

melvsparks: Yes, I forgot to add the else clause. I've resolved it

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -207,8 +208,18 @@ export default class VList {
...
+            newListNode = changeElementTag(<HTMLElement>node, 'LI');
+        } else if (nodeTag == 'TABLE') {
```
JiuqingSong, line 216: nit:
you can check in a better order:
if (table) { .. }
else if (not li) {
...}

packages/roosterjs-editor-dom/lib/test/list/createVListFromRegionTest.ts
```diff
@@ -61,7 +61,7 @@ describe('createVListFromRegion from selection, no sibling list', () => {
         runTest(`<div id="${FocusNode}"><br></div>`, [
```
JiuqingSong, line 61: would you please also add a test case for table?

Thanks.

PR #461, opened 2020-07-28
Add basic "teams style" markdown features

Lego6245:
Adds the simple teams style markdown features and behaviors to RoosterJS, and enables them in the samplesite.


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+function handleMarkdownEvent(
+    event: PluginKeyboardEvent,
+    editor: Editor,
+    triggerCharacter: string,
+    elementTag: string
+) {
+    editor.performAutoComplete(() => {
+        const range = getRangeForMarkdownOperation(event, editor, triggerCharacter);
```
JiuqingSong, line 71: Same here, use the cached event api (cacheGetEventData) to avoid run this twice for the same event

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
```diff
@@ -0,0 +1,127 @@
...
+        if (!!range) {
+            const elementToWrap = editor.getDocument().createElement(elementTag);
+            elementToWrap.appendChild(range.extractContents());
+            elementToWrap.innerText = elementToWrap.innerText.slice(
+                1,
+                elementToWrap.innerText.length
+            );
+            const nonPrintedSpaceTextNode = editor.getDocument().createTextNode(ZERO_WIDTH_SPACE);
```
JiuqingSong, line 79: why do we need this?

packages/roosterjs-editor-core/lib/interfaces/ContentEditFeature.ts
```diff
@@ -16,14 +16,17 @@ export const enum Keys {
     RIGHT = 39,
```
JiuqingSong, line 16: When type two "*", both of them are gone.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
```diff
@@ -0,0 +1,127 @@
...
+    return !!startPosition && !!endPosition && createRange(startPosition, endPosition);
+}
+
+function handleMarkdownEvent(
+    event: PluginKeyboardEvent,
+    editor: Editor,
+    triggerCharacter: string,
+    elementTag: string
+) {
+    editor.performAutoComplete(() => {
```
JiuqingSong, line 70: editor.runAsync() might be necessary here for now, otherwise the "backspace to undo auto complete" feature won't work correctly because the snapshot before auto complete will be immediately after it is taken.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
Lego6245: We need to put the cursor outside of the created wrapping element. By adding a ZWS we clearly separate the newly created bold/italic/strikethrough block and are able to put the cursor somewhere where they can keep typing normally.

Lego6245: Is there a cleaner way to do this, in your opinion? I tried fiddling with the range objects but wasn't able to get it to work.

JiuqingSong: We can just select after the new node, then call toggleBold/Italic/... to force turn off current state

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
```diff
@@ -0,0 +1,150 @@
...
+        const range = cacheGetRangeForMarkdownOperation(event, editor, triggerCharacter);
+        if (!!range) {
+            // modify the range here to include the new * character
+            range.setEnd(range.endContainer, range.endOffset + 1);
+            const elementToWrap = editor.getDocument().createElement(elementTag);
+            elementToWrap.appendChild(range.extractContents());
+            elementToWrap.innerText = elementToWrap.innerText.slice(
+                1,
+                elementToWrap.innerText.length - 1
```
JiuqingSong, line 102: A cornor case: if I type
[Star]text1[Ctrl+I]text2[Ctrl+I]text3[Star]
That only "text1" got bolded, and cursor is stopped between text1 and text2.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
Lego6245: Doesn't work for `code` as a feature.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
Lego6245: I can't even get that case to work. Debugging...

Lego6245: Figured it out. Some mixup between last and first. Also was able to clean it up a bit more, as we were, as a side effect nuking any currently applied styles.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
JiuqingSong: Ok. We can go with ZWS for now, since font and color formats are also using it.
I'll think about a way to apply format without using ZWS later.

Lego6245:
CLA bot seems to be having an outage.

Issue #233, opened 2019-02-12
TableResize not working well when table contains some long content


JiuqingSong:

JiuqingSong:
This has been fixed in 8.2.0

PR #465, opened 2020-08-11
U/melvsparks/tableresize2

melvsparks:
### Features Added
- Insert new column and row
- Resize vertically and horizontally

### What was done
I built off of Jason's existing resizing and row/column inserting prototype. 
- UI updates
- Code clean up
- Resizing improvements (dynamically resizing in real-time instead of only resizing after releasing the mouse) fixing #233 

**After changes**
ezgif-6-b383e94cd446

Screen Shot 2020-08-11 at 1 53 46 PM

Screen Shot 2020-08-11 at 1 54 12 PM

Screen Shot 2020-08-11 at 1 51 23 PM

Screen Shot 2020-08-11 at 1 50 39 PM


packages/roosterjs-editor-dom/lib/utils/normalizeRect.ts
```diff
@@ -0,0 +1,17 @@
+import { Rect } from 'roosterjs-editor-types';
+
+/**
+ * A ClientRect of all 0 is possible. i.e. chrome returns a ClientRect of 0 when the cursor is on an empty p
+ * We validate that and only return a rect when the passed in ClientRect is valid
+ */
+export default function normalizeRect(clientRect: ClientRect): Rect {
+    let { left, right, top, bottom } = clientRect || <ClientRect>{};
```
zorauf, line 8: suggestion/nit: I think the `<ClientRect>{}` fallback will make the left, right, top, bottom all as undefined and which will become `NaN` when added. Maybe it should be `<ClientRect>{ left: 0, right: 0, top: 0, bottom: 0 }`

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -1,219 +1,393 @@
-import { contains, fromHtml, isNode, isRtl, VTable } from 'roosterjs-editor-dom';
```
Lego6245, line 1: Can you double check your whitespace/lineendings on this (and other files) in your PR? We shouldn't be seeing deltas like this.

package.json
```diff
@@ -40,6 +40,7 @@
         "karma-phantomjs-launcher": "1.0.4",
         "karma-sourcemap-loader": "0.3.7",
         "karma-webpack": "4.0.2",
+        "lodash": "^4.17.10",
```
zorauf, line 43: Any particular reason for this added dev dependency?

packages/roosterjs-editor-dom/lib/utils/normalizeRect.ts
Lego6245: Technically this is "OK" as `nan > x` will return false.

package.json
Lego6245: `lodash.throttle` it seems, but i'm not sure if it's worth the whole dependency. Request animation frame would also work ehre.

zorauf: Ahh, i totally missed the file `TableResize.ts` as it didn't auto load the diff. Yep, I agree that we shouldn't take that dependency for throttle. And it is listed as dev dependency which is not the case I believe as its being used in the core code and not tests.

melvsparks: Yea, based on what I was looking at it seems as though lodash is a pretty widely used and quality package. That was one of the reasons why I went with it for throttling. We also have dependencies that use it (looking in our yarn.lock). I could refactor it to remove the dependency and use requestAnimationFrame instead

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
melvsparks: I have prettier enabled in VSCode by default, so depending on how that's setup it's probably doing this

package.json
melvsparks: In general, what is your team philosophy around taking dependencies vs writing custom in-repo solutions. What are the main things you tend to focus on when deciding to take a new dependency or not (maintainability , quality, flexibility, etc)?

packages/roosterjs-editor-dom/lib/utils/normalizeRect.ts
melvsparks: I'll change it to create more clarity around what's happening here

package.json
zorauf: Good question! I'll let @JiuqingSong chime in for that as he has more thoughts on it.

P.S: Jason, I think we should document somewhere on whats the philosophy of adding (non-dev) dependencies.

JiuqingSong:
In original implementation, there will be double lines while resizng, but with this change it seems not visible to me:
resizetable

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -1,219 +1,396 @@
...
+    };
+
+    private startResizeTable(e: MouseEvent) {
+        const doc = this.editor.getDocument();
+        doc.addEventListener('mousemove', this.frameAnimateResizeTable, true);
+        doc.addEventListener('mouseup', this.endResizeTable, true);
+    }
+
+    private frameAnimateResizeTable = (e: MouseEvent) => {
+        this.resizeTableAnimationFrameID = requestAnimationFrame(() => this.resizeTable(e));
```
JiuqingSong, line 331: What if we just handle the event directly? Since user is dragging, no other event can be triggered by user, so I think it should be fine not to run this in async way

JiuqingSong: Just took a quick look and generally seems fine to me. I'll review into more details tomorrow.

melvsparks: Okay, let me know if there is any other action that you think I need to take.I previously had throttling set up to deal with the performance of the dragging, but Will suggested using requestAnimationFrame, which made sense because this is a resizing animation.

melvsparks:
> In original implementation, there will be double lines while resizng, but with this change it seems not visible to me:
> resizetable

@JiuqingSong I ended up removing that because both word online and word desktop didn't use the double lines and I wanted to keep it consistent. I thought it was a temporary solution to the table not resizing dynamically (previously it was only showing the final resize after mousedown).

packages/roosterjs-editor-dom/lib/utils/normalizeRect.ts
M-050: Might

JiuqingSong:
> > In original implementation, there will be double lines while resizng, but with this change it seems not visible to me:
> > resizetable
> 
> @JiuqingSong I ended up removing that because both word online and word desktop didn't use the double lines and I wanted to keep it consistent. I thought it was a temporary solution to the table not resizing dynamically (previously it was only showing the final resize after mouseup).

Ok, that sounds make sense. Thanks for the clarify.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: To run with requestAnimationFrame(), suggest call editor.runAsync() instead. It is a wrapper of requestAnimationFrame and it also do null check for editor since in theory editor may be disposed after any async call.

PR #468, opened 2020-08-14
Adding function to calculate which cell to select after inserting row…

aniagonzalez:
… or column

This new function will calculate which cell to place the cursor in after insertion, either by tabbing or using the menu.
* If column was inserted, cursor will be placed in first row of the new column
* If row was inserted, cursor will be place in first column of the new row

Note: It does change the current Rooster behavior when deleting. Before this changes, whenever you delete a row or column, the content of the cell in the next row or column is selected:

image

image

With these changes, the cursor would be placed at the beginning of that cell, instead of selecting it.

### DIFF
--- a/packages/roosterjs-editor-api/lib/table/editTable.ts
+++ b/packages/roosterjs-editor-api/lib/table/editTable.ts
@@ -16,7 +16,7 @@ export default function editTable(editor: Editor, operation: TableOperation) {
             vtable.writeBack();
             editor.focus();
 
-            let cellToSelect = CalculateCellToSelect(operation, vtable.row, vtable.col);
+            let cellToSelect = calculateCellToSelect(operation, vtable.row, vtable.col);
             editor.select(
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-api/lib/table/editTable.ts
```diff
@@ -16,9 +16,37 @@ export default function editTable(editor: Editor, operation: TableOperation) {
...
+            let cellToSelect = CalculateCellToSelect(operation, vtable.row, vtable.col);
+            editor.select(
+                vtable.getCell(cellToSelect.newRow, cellToSelect.newCol).td,
+                PositionType.Begin
+            );
         }, ChangeSource.Format);
     }
 }
+
+function CalculateCellToSelect(operation: TableOperation, currentRow: number, currentCol: number) {
```
zorauf, line 28: Nit: Rename to `calculateCellToSelect`. In typescript/javascript we use camel casing for function names.

PR #469, opened 2020-08-17
Fix insert DocumentFragment node

Rain-Zheng:
When call Editor.insertContent() to insert some text on new line, the inserted text is still on the same line.

CUDYN1P4vF

This issue is related to [PR](https://github.com/microsoft/roosterjs/pull/401). The root cause is Node.insertBefore() returns an empty DocumentFragment if the node to be inserted is DocumentFragment, so the inserted node will not be wrapped with DIV.

The solution is use childNodes of DocumentFragment as insertedNode and pass them to wrap(), so that they can be wrapped.

@JiuqingSong Could you help review it? Thanks!

### DIFF
--- a/packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
+++ b/packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
@@ -101,14 +101,15 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
             // add a DIV wrapping
             if (insertedNode && option.insertOnNewLine) {
                 let start = Array.isArray(insertedNode) ? insertedNode[0] : insertedNode;
-                let end = Array.isArray(insertedNode)
-                    ? insertedNode[insertNode.length - 1]
-                    : insertedNode;
+                // Check the last node if there're more than 1 node in list
+                let end =
+                    Array.isArray(insertedNode) && insertedNode.length > 1
+                        ? insertedNode[insertedNode.length - 1]
+                        : null;
                 if (start && !isBlockElement(start)) {
                     wrap(start);
                 }
-                // Check the last node if there're more than 1 node in list
-                if (insertNode.length > 1 && end && !isBlockElement(end)) {
+                if (end && !isBlockElement(end)) {
                     wrap(end);
                 }
             }
--- a/packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
+++ b/packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
@@ -71,10 +71,20 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
                     // For insert on new line, or refNode is text or void html element (HR, BR etc.)
                     // which cannot have children, i.e. <div>hello<br>world</div>. 'hello', 'world' are the
                     // first and last node. Insert before 'hello' or after 'world', but still inside DIV
-                    insertedNode = refNode.parentNode.insertBefore(
-                        node,
-                        isBegin ? refNode : refNode.nextSibling
-                    );
+                    if (node instanceof DocumentFragment) {
+                        // if the node to be inserted is DocumentFragment, use its childNodes as insertedNode
+                        // because insertBefore() returns an empty DocumentFragment
+                        insertedNode = Array.prototype.slice.call(node.childNodes);
+                        refNode.parentNode.insertBefore(
+                            node,
+                            isBegin ? refNode : refNode.nextSibling
+                        );
+                    } else {
+                        insertedNode = refNode.parentNode.insertBefore(
+                            node,
+                            isBegin ? refNode : refNode.nextSibling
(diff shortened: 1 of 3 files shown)


packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
```diff
@@ -71,10 +71,20 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
...
                     // which cannot have children, i.e. <div>hello<br>world</div>. 'hello', 'world' are the
                     // first and last node. Insert before 'hello' or after 'world', but still inside DIV
-                    insertedNode = refNode.parentNode.insertBefore(
-                        node,
-                        isBegin ? refNode : refNode.nextSibling
-                    );
+                    if (node instanceof DocumentFragment) {
+                        // if the node to be inserted is DocumentFragment, use its childNodes as insertedNode
+                        // because insertBefore() returns an empty DocumentFragment
+                        insertedNode = Array.prototype.slice.call(node.childNodes);
```
JiuqingSong, line 77: use toArray() for type safe, and you need to change the type of insertedNode to Node|Node[]

packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
```diff
@@ -71,10 +71,20 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
                     // For insert on new line, or refNode is text or void html element (HR, BR etc.)
                     // which cannot have children, i.e. <div>hello<br>world</div>. 'hello', 'world' are the
                     // first and last node. Insert before 'hello' or after 'world', but still inside DIV
-                    insertedNode = refNode.parentNode.insertBefore(
-                        node,
-                        isBegin ? refNode : refNode.nextSibling
-                    );
+                    if (node instanceof DocumentFragment) {
```
JiuqingSong, line 74: use safeInstanceOf(node, 'DocumentFragment') for cross-window safe.

packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
Rain-Zheng: Type updated. Thanks!

packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
```diff
@@ -87,8 +99,13 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
 
             // Final check to see if the inserted node is a block. If not block and the ask is to insert on new line,
             // add a DIV wrapping
-            if (insertedNode && option.insertOnNewLine && !isBlockElement(insertedNode)) {
-                wrap(insertedNode);
+            if (insertedNode && option.insertOnNewLine) {
+                // When insert on new line, all the nodes to be inserted have been wrapped with a DIV in insertContent(),
+                // so just check the first node of insertedNode.
+                let wrappedNode = Array.isArray(insertedNode) ? insertedNode[0] : insertedNode;
+                if (!isBlockElement(wrappedNode)) {
```
JiuqingSong, line 106: Need to check if the last node is block element as well. Otherwise, when the first one is block element but last one is not, it is possible the last node and the next one are in the same block

Rain-Zheng: The check for the last node is added.

packages/roosterjs-editor-core/lib/coreAPI/insertNode.ts
```diff
@@ -87,8 +99,19 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
...
+                // Check the last node if there're more than 1 node in list
+                let end =
+                    Array.isArray(insertedNode) && insertedNode.length > 1
+                        ? insertedNode[insertedNode.length - 1]
+                        : null;
+                if (start && !isBlockElement(start)) {
+                    wrap(start);
+                }
+                if (end && !isBlockElement(end)) {
+                    wrap(end);
```
JiuqingSong, line 113: Actually, this may cause the inserted content looks different than expected.
For example:
```html
<span>Span1</span><span>Span2</span><span>Span3</span>
```

I expect the inserted content looks like:
Span1Span2Span3

But the actual result will be:
Span1
Span2
Span3

You don't need to wrap start and end node, but just need to wrap them together. And you can also make the check simpler:

```ts
if (insertNode) { // Need this check in case of empty document fragment
  const nodes = Array.isArray(insertedNode) ? insertedNode : [insertNode];
  if (!isBlockElement(nodes[0] || !isBlockElement(nodes[nodes.length - 1]) {
     wrap(nodes);
  }
}
```

Rain-Zheng: Got it. I misunderstood the logic of `wrap()`. Thanks for your clarification!

PR #471, opened 2020-08-18
Melvsparks/rotate image api

melvsparks:
### Add a new API for rotating an image
What was done
- Added the new rotateImage function
- Added tests for the rotate image function
- Added a rotate button to the ribbon

RotateImage

### DIFF
--- a/packages/roosterjs-editor-api/lib/format/rotateImage.ts
+++ b/packages/roosterjs-editor-api/lib/format/rotateImage.ts
@@ -5,23 +5,12 @@ import { Editor } from 'roosterjs-editor-core';
  * Rotate an image
  * @param editor The editor instance
  * @param image The selected image file
- * @param degreeIncrements The drgree increments at which to rotate the image (90 would rotate 90 degrees clockwise)
+ * @param angle The degree at which to rotate the image from it's center
  */
-export default function rotateImage(editor: Editor, image: HTMLImageElement, degreeIncrements?: number): void {
+export default function rotateImage(editor: Editor, image: HTMLImageElement, angle: number): void {
     if (image) {
         editor.addUndoSnapshot(() => {
-            const rotationalValue = degreeIncrements || 90;
-
-            if (image.style.transform) {
-                const currentRotation = getCurrentRotationFromTransform(image.style.transform);
-                let newRotation = currentRotation + rotationalValue;
-                newRotation = newRotation > 360 ? newRotation - 360 : newRotation;
-                image.style.transform = `rotate(${newRotation}deg)`;
-            } else {
-                image.style.transform = `rotate(${rotationalValue}deg)`;
-            }
+            image.style.transform = `rotate(${angle}deg)`;
         }, ChangeSource.Format);
     }
 }
-
-const getCurrentRotationFromTransform = (transform: string): number => +transform.match(/\d+/);
--- a/packages/roosterjs-editor-api/lib/format/rotateImage.ts
+++ b/packages/roosterjs-editor-api/lib/format/rotateImage.ts
@@ -1,0 +1,27 @@
+import { ChangeSource } from 'roosterjs-editor-types';
+import { Editor } from 'roosterjs-editor-core';
+
+/**
+ * Rotate an image
+ * @param editor The editor instance
+ * @param image The selected image file
+ * @param degreeIncrements The drgree increments at which to rotate the image (90 would rotate 90 degrees clockwise)
+ */
+export default function rotateImage(editor: Editor, image: HTMLImageElement, degreeIncrements?: number): void {
+    if (image) {
+        editor.addUndoSnapshot(() => {
+            const rotationalValue = degreeIncrements || 90;
+
+            if (image.style.transform) {
+                const currentRotation = getCurrentRotationFromTransform(image.style.transform);
+                let newRotation = currentRotation + rotationalValue;
+                newRotation = newRotation > 360 ? newRotation - 360 : newRotation;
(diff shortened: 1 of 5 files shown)


packages/roosterjs-editor-api/lib/format/rotateImage.ts
```diff
@@ -0,0 +1,16 @@
+import { ChangeSource } from 'roosterjs-editor-types';
+import { Editor } from 'roosterjs-editor-core';
+
+/**
+ * Rotate an image
+ * @param editor The editor instance
+ * @param image The selected image file
+ * @param angle The degree at which to rotate the image from it's center
+ */
+export default function rotateImage(editor: Editor, image: HTMLImageElement, angle: number): void {
```
JiuqingSong, line 10: question: for an image already rotated, when we rotate again, should we get a combined a result or a overwritten one?

publish/samplesite/scripts/controls/ribbon/Ribbon.tsx
```diff
@@ -30,6 +31,9 @@ export default class Ribbon extends React.Component<RibbonProps, {}> {
                         onClicked={this.onButtonClicked}
                     />
                 ))}
+                <button onClick={this.onRotateImage} className={styles.textButton}>
+                    RotateImage
```
JiuqingSong, line 35: (Low pri)
if you have got a SVG file from Philip, let's use it in the demo site, and you can add a drop down menu to provide more rotation choices.

melvsparks: Philip opened a work item for get an svg for this in ADO

packages/roosterjs-editor-api/lib/format/rotateImage.ts
melvsparks: I previously had it rotating based on the combined result, but I realize that would be very restricting depending on how our consumers want to rotate the image. 

For example, in OWA, we rotate dynamically based on where the mouse is. In this case we would want to update the rotation on the fly at whatever point the mouse is at. Other consumers may have different scenarios for how they want to handle this and doing it this way would allow them the flexibility to do that

publish/samplesite/scripts/controls/ribbon/Ribbon.tsx
```diff
@@ -46,6 +50,25 @@ export default class Ribbon extends React.Component<RibbonProps, {}> {
...
+        blockFormat(editor, (region) => {
+            const blocks = getSelectedBlockElementsInRegion(region);
+            blocks.forEach(block => {
+                const blockElement = block.collapseToSingleElement();
+                for (let j = 0; j < blockElement.children.length; j++) {
+                    {
+                        const child = blockElement.children[j];
+                        if (child.tagName == "IMG") {
+                            rotateImage(editor, child as HTMLImageElement, 90);
+                            break; // Only rotate the first image
```
aniagonzalez, line 64: Why do we want to rotate only the first image if we have multiple images selected?

packages/roosterjs-editor-api/lib/format/rotateImage.ts
```diff
@@ -0,0 +1,16 @@
...
+    if (image) {
+        editor.addUndoSnapshot(() => {
+            image.style.transform = `rotate(${angle}deg)`;
```
melvsparks, line 13: I need to handle image resize once it's rotated. it doesn't appear right

publish/samplesite/scripts/controls/ribbon/Ribbon.tsx
melvsparks: I just figured that since we are just trying to display how the behavior works that we didn't need to do all the selected images

packages/roosterjs-editor-api/lib/format/rotateImage.ts
melvsparks: completed

JiuqingSong:
When I try resize after rotate, I got a wrong border:

image

publish/samplesite/scripts/controls/ribbon/Ribbon.tsx
JiuqingSong: Yes, you just need to get current selected image, no need to get all.
You may need to do some change to ImageResize plugin to make the current selected image work for both places.

packages/roosterjs-editor-api/lib/format/rotateImage.ts
JiuqingSong: In OWA, if we do rotation on the fly, does it mean the function addUndoSnapshot() will be called multiple times when rotate? Does it mean user need to Ctrl+Z multiple times when they want to undo the rotation?

packages/roosterjs-editor-api/lib/index.ts
```diff
@@ -18,6 +18,7 @@ export { default as editTable } from './table/editTable';
 export { default as formatTable } from './table/formatTable';
 export { default as removeLink } from './format/removeLink';
 export { default as replaceWithNode } from './format/replaceWithNode';
+export { default as rotateImage } from './format/rotateImage';
```
aniagonzalez, line 21: Question: How do we decide if a feature should be a a plugin vs editor api? Specifically, I am wondering for image crop... I see ImageResize is a plugin.

publish/samplesite/scripts/controls/ribbon/Ribbon.tsx
melvsparks: Ive fixed this

PR #474, opened 2020-08-24
Add doc for dependencies

JiuqingSong:
Add doc for adding new dependencies.


README.md
```diff
@@ -250,28 +250,66 @@ To build the sample site code yourself, follow these instructions:
...
+   themselves in their code, and we should not add such dependency to people who don't really need it.
+
+2. What's the dependency tree of the dependency?
+   If we introduce a new dependency which has a deep dependency tree, we need to be careful since it
+   means we are actually adding a lot of new dependencies and our code size may be increased a lot.
+
+3. How much functionalities do we need from the dependency?
+   If the dependency provides a lot of functionalities but we actually only need a small piece of them,
+   we may need to consider other solutions, such as find another smaller one, or do it ourselves.
+
```
zorauf, line 306: Does it make sense to also add information on when a dependency should be a custom plugin (using that dependency) separate from the core Rooster JS?

JiuqingSong: Do you mean whether a new plugin should be part of roosterjs or a plugin in customer's projection? I think that is a separate topic.

zorauf: Make sense. Along with that should we also mention support and license for dependency?

JiuqingSong: Good question. Let me add license part.

PR #476, opened 2020-09-01
Add dark mode transformation in 'Editor.replaceNode()' 

Rain-Zheng:
When replacing an existing node with a new node in dark mode, `Editor.replaceNode()` doesn't transform the new node to dark mode. This PR transforms it like 'Editor.insertNode()' does.


Rain-Zheng:
@JiuqingSong Could you take a look at it? Thanks!

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -209,9 +209,23 @@ export default class Editor {
      * @returns true if node is replaced. Otherwise false
      */
     public replaceNode(existingNode: Node, toNode: Node): boolean {
+        // Transform the new node to replace to
+        const darkModeOptions = this.getDarkModeOptions();
+        const darkModeTransform = this.isDarkMode()
```
JiuqingSong, line 214: This changes the existing behavior. Some customer may rely on the existing behavior to not convert replaced node to dark mode. 
Consider add a parameter to control this behavior

Rain-Zheng: Got it, a param has been added.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "roosterjs",
-    "version": "7.15.2",
+    "version": "7.15.3",
```
JiuqingSong, line 3: New parameter added, use 7.16.0 instead

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -206,12 +206,30 @@ export default class Editor {
      * Replace a node in editor content with another node
      * @param existingNode The existing node to be replaced
      * @param new node to replace to
+     * @param darkMode (optional) Whether to transform new node to dark mode. Default is false
      * @returns true if node is replaced. Otherwise false
      */
-    public replaceNode(existingNode: Node, toNode: Node): boolean {
+    public replaceNode(existingNode: Node, toNode: Node, darkMode?: boolean): boolean {
```
JiuqingSong, line 212: suggest, rename "darkMode" to "transformColorForDarkMode"

Rain-Zheng: Updated. Thanks!

JiuqingSong (approved):

PR #479, opened 2020-09-14
HtmlSanitizer: Handle unknown tags

JiuqingSong:
Provide an option to decide how to handle unknown HTML tags when do sanitization.
- A valid string: replace unknown tags with this tag name
- An empty string: keep as it is
- null/undefined: Drop unknown tags and all its children

### DIFF
--- a/packages/roosterjs-editor-dom/lib/htmlSanitizer/HtmlSanitizer.ts
+++ b/packages/roosterjs-editor-dom/lib/htmlSanitizer/HtmlSanitizer.ts
@@ -1,3 +1,4 @@
+import changeElementTag from '../utils/changeElementTag';
 import getInheritableStyles from './getInheritableStyles';
 import getTagOfNode from '../utils/getTagOfNode';
 import htmlToDom from './htmlToDom';
@@ -5,12 +6,13 @@ import safeInstanceOf from '../utils/safeInstanceOf';
 import toArray from '../utils/toArray';
 import { cloneObject } from './cloneObject';
 import {
-    getAllowedTags,
     getAllowedAttributes,
-    getDefaultStyleValues,
-    getStyleCallbacks,
     getAllowedCssClassesRegex,
+    getAllowedTags,
+    getDefaultStyleValues,
+    getDisallowedTags,
     getPredefinedCssForElement,
+    getStyleCallbacks,
 } from './getAllowedValues';
 import {
     HtmlSanitizerOptions,
@@ -64,11 +66,13 @@ export default class HtmlSanitizer {
     private styleCallbacks: StyleCallbackMap;
     private attributeCallbacks: AttributeCallbackMap;
     private allowedTags: string[];
+    private disallowedTags: string[];
     private allowedAttributes: string[];
     private allowedCssClassesRegex: RegExp;
     private defaultStyleValues: StringMap;
     private predefinedCssForElement: PredefinedCssMap;
     private additionalGlobalStyleNodes: HTMLStyleElement[];
+    private unknownTagReplacement: string;
 
     /**
      * Construct a new instance of HtmlSanitizer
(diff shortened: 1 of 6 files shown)


packages/roosterjs-editor-dom/lib/htmlSanitizer/getAllowedValues.ts
```diff
@@ -1,14 +1,134 @@
...
+    section: true,
+    select: true,
+    small: true,
+    span: true,
+    strike: true,
+    strong: true,
+    sub: true,
+    summary: true,
+    sup: true,
+    table: true,
```
aniagonzalez, line 93: What about svg tag?

packages/roosterjs-editor-dom/lib/htmlSanitizer/HtmlSanitizer.ts
```diff
@@ -181,8 +187,39 @@ export default class HtmlSanitizer {
...
+            const tag = getTagOfNode(node);
+            const callback = this.elementCallbacks[tag];
+            if (callback) {
+                shouldKeep = callback(node as HTMLElement, context);
+            } else if (this.allowedTags.indexOf(tag) >= 0 || tag.indexOf(':') > 0) {
+                shouldKeep = true;
+            } else if (this.disallowedTags.indexOf(tag) >= 0) {
+                shouldKeep = false;
+            } else if (this.unknownTagReplacement === '') {
+                shouldKeep = true;
```
aniagonzalez, line 202: Having an unknownTagReplacement with an empty string, wouldn't the expected outcome be that the tag is stripped? Maybe it could be safer to have a specific string that said something like "keep" be less confusing? Maybe it is a common pattern that I am just not familiar with, but just a suggestion.

packages/roosterjs-editor-dom/test/htmlSanitizer/sanitizeHtmlTest.ts
```diff
@@ -365,3 +365,98 @@ describe('sanitizeHtml with white-space style', () => {
...
+describe('sanitizeHtml with unknown/disabled tags, keep unknown tags', () => {
+    function runTest(source: string, exp: string) {
+        let sanitizer: HtmlSanitizer = new HtmlSanitizer({
+            unknownTagReplacement: '',
+        });
+        let result = sanitizer.exec(source, false, false, { color: '' });
+        expect(result).toBe(exp);
+    }
+
+    it('Unknown tags, convert to SPAN', () => {
```
aniagonzalez, line 378: 'Unknown tags, should keep'?

packages/roosterjs-editor-dom/lib/htmlSanitizer/HtmlSanitizer.ts
zabirauf: +1, I also thought that empty string meant drop the unknown tags as I didn’t think there will be something different between null/undefined and empty string. Then I read the description and realized they are different.

zabirauf: Maybe one alternative can be

```
type UnknownTagHandlingStrategy = 
  | { strategy: ‘keep’ }
  | { strategy: ‘drop’ }
  | { strategy: ‘replace’, replacement: string }
```

packages/roosterjs-editor-dom/lib/htmlSanitizer/getAllowedValues.ts
JiuqingSong: SVG tags will be treated as unknown tags and replaced with SPAN.
Before this change, all SVG tags will be removed.
Let's see if users are satisfied with the new behavior.

packages/roosterjs-editor-dom/lib/htmlSanitizer/HtmlSanitizer.ts
JiuqingSong: How about use "*" to keep tags, and other valid string to replace, and falsy (null, undefined, empty string) to drop?

packages/roosterjs-editor-dom/test/htmlSanitizer/sanitizeHtmlTest.ts
JiuqingSong: fixed

aniagonzalez (approved):

PR #480, opened 2020-09-21
Partial code review for ImageCrop plugin.

aniagonzalez:
This is not ready for check in, but would appreciate feedback while I work on the next steps!

The things that I will add/fix on the next iteration: 
- Create tests for image crop
- Fix: "Uncaught TypeError: Failed to execute 'insertBefore' on 'Node': parameter 1 is not of type 'Node' at ImageResize.removeResizeDiv (ImageResize.ts:312)" when resizing after crop.
- Fix: need to undo twice to undo image crop
- Look into img switch from jpeg to png when using canvas for crop
- Get UI into OWA to be able to test there (does it work with the image resize OWA does?)

Some things that I would like feedback on:
- ImageCrop and ImageResize classes have very similar structure and code duplication. Is it advisable to create a base class?
- Was creating a new PluginEvent type the right thing to do?

### DIFF
--- a/packages/roosterjs-editor-types/lib/index.ts
+++ b/packages/roosterjs-editor-types/lib/index.ts
@@ -33,6 +33,7 @@ export { default as EditorReadyEvent } from './event/EditorReadyEvent';
 export { default as EntityOperationEvent } from './event/EntityOperationEvent';
 export { default as ExtractContentWithDomEvent } from './event/ExtractContentWithDomEvent';
 export { default as PendingFormatStateChangedEvent } from './event/PendingFormatStateChangedEvent';
+export { default as StartCropEvent } from './event/StartCropEvent';
 export {
     PluginDomEvent,
     PluginCompositionEvent,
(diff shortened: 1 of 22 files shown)


packages/roosterjs-editor-types/lib/index.ts
```diff
@@ -33,6 +33,7 @@ export { default as EditorReadyEvent } from './event/EditorReadyEvent';
 export { default as EntityOperationEvent } from './event/EntityOperationEvent';
 export { default as ExtractContentWithDomEvent } from './event/ExtractContentWithDomEvent';
 export { default as PendingFormatStateChangedEvent } from './event/PendingFormatStateChangedEvent';
+export { default as StartCropEvent } from './event/StartCropEvent';
```
JiuqingSong, line 36: We should try to make an event as common as possible, but not just for one scenario.

Here this is used for plugin to know that we want to start crop. But normally we don't use event to just talk to one special plugin, because all other plugins will also receive it but they never need to handle this event. So we need to rethink about how should we notify a plugin that user is about to do something.

JiuqingSong:
Generally, I suggest we do this feature in OWA first. Then later when everything stable we can consider move it back to roosterjs.

aniagonzalez:
Closing without merging. Will create a new one when moving the image crop to Rooster.

PR #481, opened 2020-09-21
Fix cut issue

JiuqingSong:
User reported that after cut, some extra content might be deleted.

This happens in this case ([ ] means the selection)
```
Li[ne 1
]Line 2
Line3
```
Expected result should be:
```
liLine2
Line3
```

But actually it becomes:
```
Li
Line3
```

In this change I rewrote the whole logic of deleting selected content so that all cases can be handled.

### DIFF
--- a/packages/roosterjs-editor-dom/test/selections/deleteSelectedContentTest.ts
+++ b/packages/roosterjs-editor-dom/test/selections/deleteSelectedContentTest.ts
@@ -1,0 +1,141 @@
+import createRange from '../../lib/selection/createRange';
+import deleteSelectedContent from '../../lib/selection/deleteSelectedContent';
+import getHtmlWithSelectionPath from '../../lib/selection/getHtmlWithSelectionPath';
+import setHtmlWithSelectionPath from '../../lib/selection/setHtmlWithSelectionPath';
+
+describe('deleteSelectedContent', () => {
+    let div: HTMLDivElement;
+    beforeEach(() => {
+        div = document.createElement('div');
+        div.contentEditable = 'true';
+        document.body.appendChild(div);
+    });
+
+    afterEach(() => {
+        document.body.removeChild(div);
+        div = null;
+    });
+
+    function runTest(html: string, expected: string) {
+        const range = setHtmlWithSelectionPath(div, html);
+        const position = deleteSelectedContent(div, range);
+        const result = getHtmlWithSelectionPath(div, createRange(position));
+
+        expect(result).toBe(expected);
+    }
+
+    it('Empty input', () => {
+        runTest('<!--{"start":[0],"end":[0]}-->', '');
+    });
+
+    it('Text input', () => {
+        runTest('test<!--{"start":[0,1],"end":[0,3]}-->', 'tt<!--{"start":[0,1],"end":[0,1]}-->');
+    });
+
+    it('Two text node', () => {
+        runTest(
+            'test1<br>test2<!--{"start":[0,2],"end":[2,2]}-->',
+            'test2<!--{"start":[0,2],"end":[0,2]}-->'
+        );
+    });
+
+    it('Two DIV node', () => {
+        runTest(
+            '<div>test1</div><div>test2</div><!--{"start":[0,0,2],"end":[1,0,2]}-->',
+            '<div>test2</div><!--{"start":[0,0,2],"end":[0,0,2]}-->'
+        );
+    });
+
+    it('Two DIV node, select to next line', () => {
+        runTest(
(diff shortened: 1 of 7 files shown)


packages/roosterjs-editor-dom/test/selections/deleteSelectedContentTest.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    afterEach(() => {
+        document.body.removeChild(div);
+        div = null;
+    });
+
+    function runTest(html: string, expected: string) {
+        const range = setHtmlWithSelectionPath(div, html);
+        const position = deleteSelectedContent(div, range);
+        const result = getHtmlWithSelectionPath(div, createRange(position));
```
aniagonzalez, line 22: @JiuqingSong , could you confirm that the loop in getHtmlWithSelectionPath is working appropriately?

I could not get the test file attached to the debugger, but I added some logs:
image

and when I insert a table like you have in the Whole Table 2 example:
aa<table><tbody><tr><td>line1</td><td>line2</td></tr><tr><td>line3</td><td>line4</td></tr></tbody></table>bb

I am getting that after the TBODY child the next sibling is NULL. Not sure why that is happening but it might be related to your test below where you don't get the expected result in the selection path...

image

aniagonzalez: Oh I get it now :) The body does not have siblings. I am sleepy... But then the loop is not working as intended because it never gets to check the row, right?

JiuqingSong: @aniagonzalez 
The code is to make sure there is TBODY under TABLE

In most cases, when we insert a table, there is no TBODY between TABLE and TR because it is optional. But this causes a problem when we get/set html with selection path. Because when we set HTML directly into container and the HTML doesn't have TBODY, broder will add TBODY for us, this causes the DOM structure changed, and selection path becomes invalid. To mitigate, we need to always add TBODY when we get HTML with selection. So that when set HTML, no extra tags will be added. In your case, there is already TBODY, so that it won't get TR and nothing need to be added, which is an expected behavior.

packages/roosterjs-editor-dom/test/selections/deleteSelectedContentTest.ts
```diff
@@ -0,0 +1,141 @@
...
+            'aabb<!--{"start":[0,2],"end":[0,2]}-->'
+        );
+    });
+
+    it('Whole talbe 2', () => {
+        // TODO: the result contains separated continuous text object at root
+        // Selection path gives wrong result. Need to revisit here
+        runTest(
+            'aa<table><tbody><tr><td>line1</td><td>line2</td></tr><tr><td>line3</td><td>line4</td></tr></tbody></table>bb<!--{"start":[1],"end":[2]}-->',
+            'aabb<!--{"start":[],"end":[]}-->'
```
aniagonzalez, line 96: So from what I can tell here... the issue seems to be that on deleteSelectedContent the startContainer is the text node "line1" which is text but has the parent "td". Then when ensureBeforeAndAfter node, it creates a new node with "" but which also has parentElement td. So when creating the new Position to select, the nodeBefore (which is the empty text with parent td) will try to findClosestElementAncestor, but the parent node is td.

But I wonder if this specific selection path scenario can happen outside of a test. When doing it manually I can cut the table and the selection after works.

packages/roosterjs-editor-dom/test/selections/deleteSelectedContentTest.ts
aniagonzalez: Thank you, that clears things up a lot! I am learning so much just by going through your CR :)

packages/roosterjs-editor-dom/test/selections/deleteSelectedContentTest.ts
JiuqingSong: Not quite understand what you mean here. You can try the selection path here:

image

You can put some text, make a selection and click "Take snapshot" to create a snapshot with selection path, modify it then click "restore snapshot" to apply it to editor.

aniagonzalez: Got it, thanks. I was just wondering if we needed to worry about fixing that specific scenario because I didn't know if our users could get into that state but I see now that there are ways to set a very specific selectionPath.

JiuqingSong: This has been there for long time and it should not be easy to hit.

aniagonzalez (approved):

PR #490, opened 2020-11-17
Adding single direction resize

aniagonzalez:
Adding ability to resize in single direction.
- Created new handles for n, s, e and w in the resizeDiv.
- When the resize is happening, reposition the handles in reference to the changing width and height of the image.
- When dragging the single direction handles, only allow the change in either width or height.
- Note: The forcePreserveRatio setting only applies to the corner handles and not the single direction ones.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResize.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResize.ts
@@ -21,6 +21,7 @@ const ENTITY_TYPE = 'IMAGE_RESIZE_WRAPPER';
 
 const HANDLE_SIZE = 7;
 const HANDLE_MARGIN = 3;
+const HANDLE_POSITIONS = ['nw', 'n', 'ne', 'e', 'se', 's', 'sw', 'w'];
 
 /**
  * ImageResize plugin provides the ability to resize an inline image in editor
@@ -235,8 +236,6 @@ export default class ImageResize implements EditorPlugin {
                     img.style.height = newHeight + 'px';
                 }
             }
-
-            this.positionSingleDirectionHandles(img);
         }
         this.stopEvent(e);
     };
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResize.ts
```diff
@@ -328,7 +371,20 @@ export default class ImageResize implements EditorPlugin {
...
 
     private isWest(direction: string): boolean {
-        return direction && direction.substr(1, 1) == 'w';
+        return direction && (direction.substr(0, 1) == 'w' || direction.substr(1, 1) == 'w');
+    }
+
+    private isSingleDirectionNS(direction: string): boolean {
+        return (
+            direction &&
+            (direction.substr(0, 1) == 'n' || direction.substr(0, 1) == 's') &&
```
melvsparks, line 380: can't this just be written direction == "n" || direction == "s"?

private isSingleDirectionNS = (direction: string): boolean => direction == "n" || direction == "s";

melvsparks (changes requested):

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResize.ts
```diff
@@ -237,10 +235,34 @@ export default class ImageResize implements EditorPlugin {
                     img.style.height = newHeight + 'px';
                 }
             }
+
+            this.positionSingleDirectionHandles(img);
```
JiuqingSong, line 239: How about this, then you never need to reposition them:

```html
<div style="background-color: gray; width: 50%; height: 50%; position: absolute; border: solid 1px black">
  <!--top left-->
  <div style="position: absolute;left: 0; top: 0">
    <div style="top: -4px; left: -4px; width: 7px; height: 7px; background-color: red; cursor: nw-resize; position: relative"></div>
  </div>

  <!--top right-->
  <div style="position: absolute;right: 0; top: 0">
    <div style="top: -4px; right: -4px; width: 7px; height: 7px; background-color: red; cursor: ne-resize; position: relative"></div>
  </div>

  <!--bottom left-->
  <div style="position: absolute;left: 0; bottom: 0">
    <div style="bottom: -4px; left: -4px; width: 7px; height: 7px; background-color: red; cursor: sw-resize; position: relative"></div>
  </div>

  <!--bottom right-->
  <div style="position: absolute;right: 0; bottom: 0">
    <div style="bottom: -4px; right: -4px; width: 7px; height: 7px; background-color: red; cursor: se-resize; position: relative"></div>
  </div>

  <!--top-->
  <div style="position: absolute;left: 50%; top: 0">
    <div style="top: -4px; left: -4px; width: 7px; height: 7px; background-color: red; cursor: n-resize; position: relative"></div>
  </div>

  <!--bottom-->
  <div style="position: absolute;left: 50%; bottom: 0">
    <div style="bottom: -4px; left: -4px; width: 7px; height: 7px; background-color: red; cursor: s-resize; position: relative"></div>
  </div>

  <!--left-->
  <div style="position: absolute;top: 50%; left: 0">
    <div style="top: -4px; left: -4px; width: 7px; height: 7px; background-color: red; cursor: w-resize; position: relative"></div>
  </div>

  <!--right-->
  <div style="position: absolute;top: 50%; right: 0">
    <div style="top: -4px; right: -4px; width: 7px; height: 7px; background-color: red; cursor: e-resize; position: relative"></div>
  </div>
</div>
```

JiuqingSong: You just need to make a loop to generate this HTML from the 8 positions.

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResize.ts
```diff
@@ -328,7 +371,20 @@ export default class ImageResize implements EditorPlugin {
...
+        return (
+            direction &&
+            (direction.substr(0, 1) == 'n' || direction.substr(0, 1) == 's') &&
+            direction.substr(1, 1) != 'e' &&
+            direction.substr(1, 1) != 'w'
+        );
+    }
+
+    private isSingleDirectionWE(direction: string): boolean {
+        return direction && (direction.substr(0, 1) == 'w' || direction.substr(0, 1) == 'e');
```
JiuqingSong, line 387: just check direction == 'w' || direction == 'e'

JiuqingSong (changes requested):

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResize.ts
aniagonzalez: The problem with this approach is that it would work for the html generation (line 308) where the handle position is 'n' but this.direction (used in line 251, 262) is set to the element style cursor (line 192) which would be "n-resize". So the method of getting the substring works for either pos ('n') or cursor property ('n-resize'), but using == 'n' doesn't.

If we really wanted to make direction and pos consistent, we could change the HANDLE_POSITIONS array to "n-resize" format, but then we need to use substrings here anyway. Or we could make this.direction be only the 'n' or 'ne' substring from the 'n-resize' or 'ne-resize' style cursor but that is just as complex.

Another option would be to use regex here.

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResize.ts
aniagonzalez: The problem with this approach is that it would work for the html generation (line 308) where the handle position is 'n' but this.direction (used in line 251, 262) is set to the element style cursor (line 192) which would be "n-resize". So the method of getting the substring works for either pos ('n') or cursor property ('n-resize'), but using == 'n' doesn't.

If we really wanted to make direction and pos consistent, we could change the HANDLE_POSITIONS array to "n-resize" format, but then we need to use substrings here anyway. Or we could make this.direction be only the 'n' or 'ne' substring from the 'n-resize' or 'ne-resize' style cursor but that is just as complex.

Another option would be to use regex here.

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResize.ts
JiuqingSong: I think in that case it may not be a good idea to reuse the CSS cursor value to check the direction.

How about add a new attribute to the DIV data-direction="..", then we always check this value instead of the CSS cursor. And the value will only contain "n", "w", ...

aniagonzalez:
This also made me realize that when we rotate an element (for example 90 deg), the handles are rotated too so the UX is weird (the handle that appears to be West is actually South and changes the height not width).

JiuqingSong:
> This also made me realize that when we rotate an element (for example 90 deg), the handles are rotated too so the UX is weird (the handle that appears to be West is actually South and changes the height not width).

Rotated image can be handle separately.

JiuqingSong (approved):

Issue #254, opened 2019-03-05
Editing a url does not update the link


dmhabib:
When pasting a url into the editor, it correctly turns it into a hyperlink.  Later, if I go back and edit the link text it does not update the link address.  For example, if I paste 'http://www.github.net' into the editor, realize it's wrong and update it to 'http://www.github.com' it will still link to the old url.

**To Reproduce**
1. Add a url to the editor
2. Using the keyboard, go back and edit the text of the url
3. Notice:  The text is updated but the link is not.

**Expected behavior**
I would expect that if the text of the link originally matches the target address, editing the link text would update the link url.

JiuqingSong:
This is a valid issue. Thanks for reporting. We'll prioritize the fix.

JiuqingSong:
This will be fixed in 8.0.0

PR #491, opened 2020-11-18
Allow editing of link URLs through display text

rurikoaraki:
This is an attempted fix for #254 where editing a link's URL does not update the href attribute.

The fix has the following behavior:

In the Hyperlink plugin, if we have either mouse up or key up events and the keycode is a directional key (pageup, pagedown, home, end, left/up/right/down), the code checks if the link's display text matches the href value of the link. If it does, it stores the link element away so that we can know if the cursor has moved away from the link. When we receive another event of the same kind, we will check if the current element at the cursor is different from the one that was stored away. If it is different, and the display text is a valid URL, then we will clone the element, update the href attribute on the cloned element to match the display text URL, and then replace the old element with the new one in the editor.

Some details:
- The reason that the keys are restricted to directional keys is so that we don't run a bunch of work if the user is typing in the editor.
- A match is defined as: the display text can be missing the http(s) portion or a backslash at the end of the URL, but otherwise must exactly match the href value. This means that if the display text says “www.bing.com” and the href value is “http://www.bing.com/” then it will match and allow the editing behavior, but it will not kick in if the “www.” is missing from the display text.
- The behavior kicks in if the link has extra formatting like changed text color or italicization. It will keep previous formatting that was applied to it.
- This behavior is done with the auto-correct behavior so that the user can undo it with backspace or ctrl-z. This can have some weird behavior in cases like: you have two links in the editor, you edit one link, use the mouse to put the cursor into another link (behavior kicks in), and then press backspace to edit the second link, which seems to do nothing. However what it does is undo the href editing that just happened because that's how the auto correct works, but it's not very obvious because the href editing is not a visible change. If people have suggestions on how I can fix this behavior, I would love to hear them.

I have tested:
- Editing a link which matches the href attribute exactly
- Editing a link which has the http(s) and backslash missing
- Editing a link which has the "www." portion missing, the behavior did not kick in
- Editing a link with red text, the red text was kept
- Behavior is undo-able in two stages (the href update and the display text edits are different undo steps)

If there are other test cases that I should run, let me know and I'd be happy to try them out.
I also just noticed that I should update the wiki page on the Hyperlink plugin to account for this behavior. I'd also be open to moving this into its own plugin - let me know what would be best.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
@@ -1,5 +1,10 @@
 import { Browser, matchLink } from 'roosterjs-editor-dom';
-import { Editor, EditorPlugin, isCtrlOrMetaPressed } from 'roosterjs-editor-core';
+import {
+    Editor,
+    EditorPlugin,
+    isCtrlOrMetaPressed,
+    cacheGetElementAtCursor,
+} from 'roosterjs-editor-core';
 import { PluginEvent, PluginEventType, ChangeSource } from 'roosterjs-editor-types';
 
 /**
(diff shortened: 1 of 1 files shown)


rurikoaraki:
Closing this out and moving this to be against the vnext branch.

PR #492, opened 2020-11-20
Allow editing of link URLs through display text

rurikoaraki:
This is an attempted fix for #254 where editing a link's URL does not update the href attribute.

The fix has the following behavior:

In the Hyperlink plugin, if we have either mouse up or key up events and the keycode is a directional key (pageup, pagedown, home, end, left/up/right/down), the code checks if the link's display text matches the href value of the link. If it does, it stores the link element away so that we can know if the cursor has moved away from the link. When we receive another event of the same kind, we will check if the current element at the cursor is different from the one that was stored away. If it is different, and the display text is a valid URL, then we will clone the element, update the href attribute on the cloned element to match the display text URL, and then replace the old element with the new one in the editor.

Some details:
- The reason that the keys are restricted to directional keys is so that we don't run a bunch of work if the user is typing in the editor. I think I got all of the ways the cursor can move via keyboard, but there may be cases I'm missing. If there's an easier way to listen to cursor position changes I'd be open to using that instead.
- A match is defined as: the display text can be missing the http(s) portion or a backslash at the end of the URL, but otherwise must exactly match the href value. This means that if the display text says “www.bing.com” and the href value is “http://www.bing.com/” then it will match and allow the editing behavior, but it will not kick in if the “www.” is missing from the display text.
- The behavior kicks in if the link has extra formatting like changed text color or italicization. It will keep previous formatting that was applied to it.
- This behavior is done with the auto-correct behavior so that the user can undo it with ctrl-z but not backspace.

I have tested:
- Editing a link which matches the href attribute exactly
- Editing a link which has the http(s) and backslash missing
- Editing a link which has the "www." portion missing, the behavior did not kick in
- Editing a link with red text, the red text was kept
- Behavior is undo-able in two stages (the href update and the display text edits are different undo steps)
- If you edit the href field of the link using the link dialog and then change the display text, the link is not updated.
- Link is updated after bringing focus out of the editor
- Link is updated after arrowing arround
- Link is updated if space is pressed at the end of a link
- Link is updated if content is pasted at the end of a link and middle of a link

If there are other test cases that I should run, let me know and I'd be happy to try them out.
I also just noticed that I should update the wiki page on the Hyperlink plugin to account for this behavior. I'd also be open to moving this into its own plugin - let me know what would be best.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
@@ -58,12 +58,11 @@ export default class HyperLink implements EditorPlugin {
     };
 
     protected onBlur = (e: FocusEvent) => {
-        if (this.trackedLink && !this.doesLinkDisplayMatchHref(this.trackedLink)) {
-            this.updateLinkHref();
+        if (this.trackedLink) {
+            this.updateLinkHrefIfShouldUpdate();
         }
 
-        this.trackedLink = null;
-        this.originalHref = '';
+        this.resetLinkTracking();
     };
 
     /**
@@ -104,15 +103,12 @@ export default class HyperLink implements EditorPlugin {
                     this.trackedLink &&
                     (anchor !== this.trackedLink || event.eventType == PluginEventType.KeyUp)
                 ) {
-                    if (!this.doesLinkDisplayMatchHref(this.trackedLink)) {
-                        this.updateLinkHref();
-                    }
+                    this.updateLinkHrefIfShouldUpdate();
                 }
 
                 // If the link's href value was edited, or the cursor has moved out of the previously tracked link,
                 // stop tracking the link.
-                this.trackedLink = null;
-                this.originalHref = '';
+                this.resetLinkTracking();
             }
 
             // Cache link and href value if its href attribute currently matches its display text
@@ -168,6 +164,23 @@ export default class HyperLink implements EditorPlugin {
         return (
             isCharacterValue(event) || event.which == Keys.BACKSPACE || event.which == Keys.DELETE
         );
+    }
+
+    /**
+     * Updates the href of the tracked link if the display text doesn't match href anymore
+     */
+    private updateLinkHrefIfShouldUpdate() {
+        if (!this.doesLinkDisplayMatchHref(this.trackedLink)) {
+            this.updateLinkHref();
+        }
+    }
+
+    /**
+     * Clears the tracked link and its original href value so that it's back to default state
+     */
+    private resetLinkTracking() {
+        this.trackedLink = null;
+        this.originalHref = '';
     }
 
     /**
--- a/packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
@@ -68,10 +68,14 @@ export default class HyperLink implements EditorPlugin {
      * @param event PluginEvent object
      */
     public onPluginEvent(event: PluginEvent): void {
-        if (event.eventType == PluginEventType.MouseUp) {
+        if (
+            event.eventType == PluginEventType.MouseUp ||
+            event.eventType == PluginEventType.KeyUp
+        ) {
             const anchor = this.editor.getElementAtCursor(
-                'A',
-                <Node>event.rawEvent.srcElement
+                'A[href]',
+                null /*startFrom*/,
+                event
             ) as HTMLAnchorElement;
 
             // If cursor has moved out of previously tracked link
@@ -84,12 +88,19 @@ export default class HyperLink implements EditorPlugin {
                 this.trackedLink = null;
             }
 
-            if (anchor) {
-                // Cache link if its href attribute currently matches its display text
-                if (!this.trackedLink && this.doesLinkDisplayMatchHref(anchor)) {
-                    this.trackedLink = anchor;
-                }
+            // Cache link if its href attribute currently matches its display text
+            if (!this.trackedLink && this.doesLinkDisplayMatchHref(anchor)) {
+                this.trackedLink = anchor;
+            }
+        }
 
+        if (event.eventType == PluginEventType.MouseUp) {
+            const anchor = this.editor.getElementAtCursor(
+                'A',
+                <Node>event.rawEvent.srcElement
+            ) as HTMLAnchorElement;
+
+            if (anchor) {
                 if (this.onLinkClick && this.onLinkClick(anchor, event.rawEvent) !== false) {
                     return;
                 }
--- a/packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
@@ -1,5 +1,5 @@
 import { Browser, isCtrlOrMetaPressed, matchLink } from 'roosterjs-editor-dom';
-import { EditorPlugin, IEditor, PluginEvent, PluginEventType } from 'roosterjs-editor-types';
+import { EditorPlugin, IEditor, Keys, PluginEvent, PluginEventType } from 'roosterjs-editor-types';
 
 /**
  * An editor plugin that show a tooltip for existing link
@@ -70,7 +70,9 @@ export default class HyperLink implements EditorPlugin {
     public onPluginEvent(event: PluginEvent): void {
         if (
             event.eventType == PluginEventType.MouseUp ||
-            event.eventType == PluginEventType.KeyUp
+            (event.eventType == PluginEventType.KeyUp &&
+                event.rawEvent.which >= Keys.PAGEUP &&
+                event.rawEvent.which <= Keys.DOWN)
         ) {
             const anchor = this.editor.getElementAtCursor(
                 'A[href]',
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
```diff
@@ -105,4 +134,38 @@ export default class HyperLink implements EditorPlugin {
             return anchor ? anchor.href : null;
         } catch {}
     }
+
+    /**
+     * Compares the normalized URL of inner text of element to its href to see if they match
+     */
+    private doesLinkDisplayMatchHref(element: HTMLAnchorElement): boolean {
+        let display = element.innerText.trim();
+        let rule = new RegExp(`^(?:https?:\\/\\/)?${display}\\/?`, 'i');
```
JiuqingSong, line 143: display may contain some special charactor for regular expression

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
```diff
@@ -105,4 +134,38 @@ export default class HyperLink implements EditorPlugin {
...
+     */
+    private updateLinkHref(event: PluginEvent) {
+        let originalLink = this.trackedLink; // keep the element available for the auto complete function
+
+        let linkData = matchLink(this.trackedLink.innerText.trim());
+        if (linkData !== null) {
+            let anchor = originalLink.cloneNode(true /*deep*/) as HTMLAnchorElement;
+            anchor.href = linkData.normalizedUrl;
+
+            this.editor.runAsync(() => {
```
JiuqingSong, line 163: why need run async and clone node? Can we change the original link in place?

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
```diff
@@ -67,6 +68,34 @@ export default class HyperLink implements EditorPlugin {
      * @param event PluginEvent object
      */
     public onPluginEvent(event: PluginEvent): void {
+        if (
+            event.eventType == PluginEventType.MouseUp ||
+            (event.eventType == PluginEventType.KeyUp &&
+                event.rawEvent.which >= Keys.PAGEUP &&
+                event.rawEvent.which <= Keys.DOWN)
```
JiuqingSong, line 75: 1. change a link href
2. without click or move cursor, save the result directly

Then the original link without changed will be saved

rurikoaraki: Ah yeah I see the issue, thanks for catching this. Working on fixing it.

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
rurikoaraki: I've added a line which will find all special characters in the display text and escape them so that matching works as expected.

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
rurikoaraki: Should be fixed with latest commit - let me know if you think of any issues with it

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
```diff
@@ -67,6 +76,43 @@ export default class HyperLink implements EditorPlugin {
...
+                this.trackedLink = null;
+                this.originalHref = '';
+            }
+
+            // Cache link and href value if its href attribute currently matches its display text
+            if (!this.trackedLink && this.doesLinkDisplayMatchHref(anchor)) {
+                this.trackedLink = anchor;
+                this.originalHref = this.tryGetHref(anchor);
+            }
+        }
```
JiuqingSong, line 114: The issue that link can't be updated  before click away can still happen.

Suggest always update the link for any key input or content changed event as long as this.trackedLink is not null.

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
```diff
@@ -105,4 +151,35 @@ export default class HyperLink implements EditorPlugin {
...
         } catch {}
     }
+
+    /**
+     * Compares the normalized URL of inner text of element to its href to see if they match
+     */
+    private doesLinkDisplayMatchHref(element: HTMLAnchorElement): boolean {
+        if (element) {
+            let display = element.innerText.trim();
+            let escapedDisplay = display.replace(/[-\/\\^$*+?.()|[\]{}]/g, '\\$&');
```
JiuqingSong, line 161: Add comment to explain why we need to do this

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
```diff
@@ -67,6 +76,43 @@ export default class HyperLink implements EditorPlugin {
...
+
+                this.trackedLink = null;
+                this.originalHref = '';
+            }
+
+            // If the link's href value was edited, stop tracking the link.
+            if (this.trackedLink && this.tryGetHref(this.trackedLink) !== this.originalHref) {
+                this.trackedLink = null;
+                this.originalHref = '';
+            }
```
JiuqingSong, line 107: These two if blocks (L94-L101 and L104-L107) seems to be able to merge.

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
rurikoaraki: Chatted offline to resolve this. Summary is that one of the issues with changing behavior to update the link on any key input is that it makes it harder for the changes to be undo-able. @JiuqingSong suggested that the behavior kicks in on clicking away, handling any non-content input, and on blur. The latest commit has these updates.

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
rurikoaraki: I did some refactoring to combine logic with the blur event as well. I could combine the logic further so that the link tracking and updating is all done in one function and both the onBlur event and the onPluginEvents could call into it, but I feel that it would make the logic more complicated since it would mean having many optional fields for that function.

Let me know what you think.

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
```diff
@@ -67,6 +81,43 @@ export default class HyperLink implements EditorPlugin {
...
+            if (
+                this.trackedLink &&
+                (anchor !== this.trackedLink ||
+                    event.eventType == PluginEventType.KeyUp ||
+                    this.tryGetHref(this.trackedLink) !== this.originalHref)
+            ) {
+                // If cursor has moved out of previously tracked link or non-content key event
+                // detected, update link href if display text doesn't match href anymore
+                if (
+                    this.trackedLink &&
```
JiuqingSong, line 103: No need to check this.trackedLink again since you already checked in L95

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
```diff
@@ -67,6 +81,43 @@ export default class HyperLink implements EditorPlugin {
      * @param event PluginEvent object
      */
     public onPluginEvent(event: PluginEvent): void {
+        if (
+            event.eventType == PluginEventType.MouseUp ||
+            (event.eventType == PluginEventType.KeyUp && !this.isContentEditValue(event.rawEvent))
```
JiuqingSong, line 86: I think you also need to handle ContentChanged event

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
JiuqingSong: Current code seems good to me.

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
rurikoaraki: I also added handling for the space key as you could edit a link at the end and press space to leave the link element

packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
rurikoaraki: Thanks, I tested that the behavior is triggered when pasting content at the end of an edited link

JiuqingSong: ContentChanged event can be triggered not only from paste, but also a lot of reasons, even from other plugins.

rurikoaraki: That's fair. Should we be selective on when it should trigger? I think it would make sense to trigger the behavior for Paste, Cut, and Drop (and maybe SetContent? I didn't see any weird behavior with undo logic when testing).

JiuqingSong: I think it should be fine to handle everything.

JiuqingSong (approved):

PR #495, opened 2020-12-03
Add context menu handler in image resize

FrancisMengx:
Add context menu handler in image resize


JiuqingSong:
Please check in to vnext branch instead, and make the fix with the new ContextMenuProvider interface so that it can benefit all plugins.

You can fix in DomEventPlugin where it invokes getContextMenuItems() and pass in the node. You can get the correct node before the function call so that no change to other plugins is needed.

PR #497, opened 2020-12-11
Fix image context menu

FrancisMengx:
Fix image context menu

### DIFF
--- a/packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
@@ -1,4 +1,4 @@
-import { arrayPush, Browser, isCharacterValue, safeInstanceOf } from 'roosterjs-editor-dom';
+import { arrayPush, Browser, isCharacterValue } from 'roosterjs-editor-dom';
 import {
     ChangeSource,
     ContextMenuProvider,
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
```diff
@@ -152,8 +152,18 @@ export default class DOMEventPlugin implements PluginWithState<DOMEventPluginSta
 
     private onContextMenuEvent = (event: MouseEvent) => {
         const allItems: any[] = [];
+        const position = this.editor.getFocusedPosition();
+        let eventTargetNode = event.target as Node;
+        if (
+            position &&
+            event.button != 2 &&
+            safeInstanceOf(position.node, 'HTMLDivElement') &&
```
JiuqingSong, line 160: why check DIV?

I think you can get InlineElement before cursor and use its container node as the node. 

Use editor.getContentSearcherOfCursor(event), then call searcher.getInlineElementBefore().

It doesn't need to be image, but for whatever inline element we can use it.

packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
```diff
@@ -152,8 +152,15 @@ export default class DOMEventPlugin implements PluginWithState<DOMEventPluginSta
 
     private onContextMenuEvent = (event: MouseEvent) => {
         const allItems: any[] = [];
+        const searcher = this.editor.getContentSearcherOfCursor();
+        const elementBeforeCursor = searcher.getInlineElementBefore();
```
JiuqingSong, line 156: need null check for searcher.

packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
```diff
@@ -152,8 +152,15 @@ export default class DOMEventPlugin implements PluginWithState<DOMEventPluginSta
...
+
+        let eventTargetNode = event.target as Node;
+        if (event.button != 2) {
+            eventTargetNode = elementBeforeCursor.getContainerNode();
```
JiuqingSong, line 160: need null check for elementBeforeCursor

JiuqingSong (approved):

PR #503, opened 2021-01-11
U/angon/capitalize

aniagonzalez:
This change adds a feature to change case of selection to UPPERCASE, lowercase or Capitalize Each Word options.

image

It is implemented using applyInlineStyle to traverse the elements in the selection. 
To get the capitalized test it uses text-transform which handles the language-specific mapping rules.

TO REVIEW:
- For correct capitalization to happen in all languages, we would need the editor to know of the language (through xml-lang with current implementation). See this example, the first test passes but the second fails:
image

- I am not sure the added SPAN elements are acceptable. Please see file changeCapitalizationTest.ts

- I grabbed an open source icon because we didn't have an appropriate one in our codebase. Should I change it to just a text button?


packages/roosterjs-editor-api/lib/format/changeCapitalization.ts
```diff
@@ -0,0 +1,47 @@
...
+            }
+        }
+    });
+
+    function preprocessText(textToPreprocess: string) {
+        const lowercaseText = getCapitalizedText(textToPreprocess, Capitalization.Lowercase);
+        return lowercaseText;
+    }
+
+    function getCapitalizedText(originalText: string, capitalization: Capitalization): string {
```
aniagonzalez, line 28: Per our discussion, I will be changing this method to accept an optional LANG.

packages/roosterjs-editor-api/lib/format/changeCapitalization.ts
```diff
@@ -0,0 +1,56 @@
...
+import { Capitalization, NodeType } from 'roosterjs-editor-types';
+import applyInlineStyle from '../utils/applyInlineStyle';
+import { getFirstLeafNode, getNextLeafSibling } from 'roosterjs-editor-dom';
+
+/**
+ * Change the capitalization of text in the selection
+ * @param editor The editor instance
+ * @param capitalization The case option
+ */
+export default function g(editor: IEditor, capitalization: Capitalization, language?: string) {
```
JiuqingSong, line 11: Use a meaningful function name

packages/roosterjs-editor-api/lib/format/changeCapitalization.ts
```diff
@@ -0,0 +1,56 @@
...
+        //     tempNode.style.textTransform = capitalization;
+        //     transformedText = tempNode.innerText;
+
+        //     document.body.removeChild(tempNode);
+        // }
+        // return transformedText;
+
+        switch (capitalization) {
+            case Capitalization.Lowercase:
+                return originalText.toLocaleLowerCase(language);
```
JiuqingSong, line 44: If language is undefined, will it respect browser's language setting?

packages/roosterjs-editor-api/lib/format/changeCapitalization.ts
```diff
@@ -0,0 +1,56 @@
...
+        //     if (language) {
+        //         tempNode.lang = language;
+        //     }
+        //     tempNode.innerText = originalText;
+        //     tempNode.style.textTransform = capitalization;
+        //     transformedText = tempNode.innerText;
+
+        //     document.body.removeChild(tempNode);
+        // }
+        // return transformedText;
```
JiuqingSong, line 40: Remove unused code /comment

aniagonzalez: Sorry, I forgot I had already started this PR when I pushed my last changes!

packages/roosterjs-editor-api/lib/format/changeCapitalization.ts
aniagonzalez: Yes, it is an optional parameter, the default locale is the host environment’s current locale.

packages/roosterjs-editor-api/lib/format/changeCapitalization.ts
```diff
@@ -0,0 +1,49 @@
+import { IEditor } from 'roosterjs-editor-types';
+import { Capitalization, NodeType } from 'roosterjs-editor-types';
+import applyInlineStyle from '../utils/applyInlineStyle';
+import { getFirstLeafNode, getNextLeafSibling } from 'roosterjs-editor-dom';
+
+/**
+ * Change the capitalization of text in the selection
+ * @param editor The editor instance
+ * @param capitalization The case option
+ */
```
JiuqingSong, line 10: Add comment for language, and explain the default behavior

packages/roosterjs-editor-api/lib/format/changeCapitalization.ts
```diff
@@ -0,0 +1,49 @@
...
+                return wordArray.join(' ');
+            case Capitalization.Sentence:
+                // TODO: Add rules on punctuation for internationalization - TASK 104769
+                const punctuationMarks = '[\\.\\!\\?]';
+                // Find a match of a word character either:
+                // at the beginning of a string with or without preceding whitespace
+                // or preceded by a punctuation mark and at least one whitespace.
+                const regex = new RegExp('^\\s*\\w|' + punctuationMarks + '\\s+\\w', 'g');
+                return originalText.toLocaleLowerCase(language).replace(regex, match => {
+                    return match.toUpperCase();
```
JiuqingSong, line 45: toUpperCase or toLocaleUpperCase ?

aniagonzalez: Thanks for catching this!

JiuqingSong (approved):

PR #517, opened 2021-03-09
Fix some table resize issues

flyingbee2012:
* Fixed the bug where a long word in a TD doesn't auto-wrap during resizing
* Fixed the bug where it would do vertical resizing after user resized the cell horizontally
* Enlarged the region of detecting the resizer so it would be easier to be triggered when the table is on top or on the most left in the container

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
@@ -13,6 +13,7 @@ import {
 const INSERTER_COLOR = '#4A4A4A';
 const INSERTER_SIDE_LENGTH = 12;
 const INSERTER_BORDER_SIZE = 1;
+const INSERTER_HOVER_OFFSET = 5;
 
 const CELL_RESIZER_WIDTH = 4;
 const HORIZONTAL_RESIZER_HTML =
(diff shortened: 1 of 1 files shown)


JiuqingSong:
Tried with the branch, I found one issue when there is a super long word in a cell and then do resize, somehow it can suddenly change the width then follow my mouse. I haven't found a stable repro step, but I realized that when this happen, the cell that I'm resizing doesn't have "width" in its CSS attribute.

flyingbee2012:
> Tried with the branch, I found one issue when there is a super long word in a cell and then do resize, somehow it can suddenly change the width then follow my mouse. I haven't found a stable repro step, but I realized that when this happen, the cell that I'm resizing doesn't have "width" in its CSS attribute.

Can you send a gif about this issue? I cannot get repro on my end. I can see the width is set while resizing.

JiuqingSong:
> > Tried with the branch, I found one issue when there is a super long word in a cell and then do resize, somehow it can suddenly change the width then follow my mouse. I haven't found a stable repro step, but I realized that when this happen, the cell that I'm resizing doesn't have "width" in its CSS attribute.
> 
> Can you send a gif about this issue? I cannot get repro on my end. I can see the width is set while resizing.

Hmm I can't repro now

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -190,6 +191,9 @@ export default class TableResize implements EditorPlugin {
     }
 
     private createInserter(tableRect: Rect) {
+        if (this.insertingState == ResizeState.None) {
+            return;
```
JiuqingSong, line 195: What does this fix for?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -342,7 +346,7 @@ export default class TableResize implements EditorPlugin {
     };
 
     private resizeTable = (e: MouseEvent) => {
-        if (this.currentTd) {
+        if (this.currentTd && this.resizingState !== ResizeState.None) {
```
JiuqingSong, line 349: In what case resizingState is not None?

flyingbee2012: This is to fix the issue where it would do vertical resizing after user resized the cell horizontally. Since in endResizeTable, we are setting resizingState as None, but in resizeTable it was only checking if Horizontal or not (any state that is not horizonal is considered as vertical), which is the root cause of this issue.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: Since we ignored the None state in resizeTable method and caused a resizing issue. I think we'd better consider the None state in this method as well. Since in this method, the the logic is below and is bug prone.

if (state == horizontal) {
// do horizontal logic
}
else {
// do vertical logic, even for None state.
}

JiuqingSong (approved):

PR #518, opened 2021-03-13
Allow move image

JiuqingSong:
Moving image is natively supported by browser, but our ImageResize plugin has event handler for MouseDown plugin and make change to the image, which causes the image can't be moved.
The fix is to split the action of mouse down event into mouse down and mouse up. For mouse down, we deselect image if any, then reselect when mouse up. So that for drag event, the image can be detected by browser.


flyingbee2012 (approved):
looks good!

PR #522, opened 2021-03-24
Graduate list features

JiuqingSong:
Graduate the following experimental features:

- NewIndentation
- NewBullet
- NewNumbering
- ListChain
- MergePastedLine


packages/roosterjs-editor-api/lib/format/setIndentation.ts
```diff
@@ -18,32 +24,53 @@ import {
...
+function outdent(region: RegionBase, blocks: BlockElement[]) {
+    blocks.forEach(blockElement => {
+        let node = blockElement.collapseToSingleElement();
+        const quote = findClosestElementAncestor(node, region.rootNode, 'blockquote');
+        if (quote) {
+            if (node == quote) {
+                node = wrap(toArray(node.childNodes));
+            }
+
+            while (isNodeInRegion(region, node) && getTagOfNode(node) != 'BLOCKQUOTE') {
```
flyingbee2012, line 67: Not sure if we can unify 'blockquote' and 'BLOCKQUOTE' to either upper case or lower case here?

packages/roosterjs-editor-types/lib/enum/ExperimentalFeatures.ts
```diff
@@ -3,27 +3,27 @@
  */
 export const enum ExperimentalFeatures {
     /**
-     * Toggle indentation using VList
+     * @deprecated This feature is always enabled
```
flyingbee2012, line 6: We don't remove the graduated feature flights?

packages/roosterjs-editor-api/lib/format/setIndentation.ts
JiuqingSong: Good question. Ideally we should use the same case. However, for history reason, getTagOfNode() always returns upper case. And when search or create node, we prefer lower case so that typescript can infer the node type.

packages/roosterjs-editor-types/lib/enum/ExperimentalFeatures.ts
JiuqingSong: No we don't. This is for backward compability.

We need to bump up major version if we remove element of public interface since it may break the build when others are still using it in their code.

packages/roosterjs-editor-api/lib/format/setIndentation.ts
flyingbee2012: okay, good to know that.

packages/roosterjs-editor-types/lib/enum/ExperimentalFeatures.ts
flyingbee2012: okay

flyingbee2012 (approved):

PR #523, opened 2021-03-30
Don't resize the whole table when resizing individual cells

flyingbee2012:
Today, when we resize each individual cell in a table, the total width of the table is changed as well when resizing. This change is to fix the table size while resizing individual cells so that we can add a resizer that control that whole table's size later. This is to simulate the similar behavior in Word.


packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -1,5 +1,6 @@
 import safeInstanceOf from '../utils/safeInstanceOf';
 import toArray from '../utils/toArray';
+import { normalizeRect } from 'roosterjs-editor-dom';
```
JiuqingSong, line 3: This is already in roosterjs-editor-dom package, use relative path to import

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -279,6 +280,51 @@ export default class VTable {
...
+                    }
+                }
+            }
+        }
+
+        if (cells.length > 0) {
+            const temp = cells[0];
+            cells[0] = cells[pos];
+            cells[pos] = temp;
+        }
```
JiuqingSong, line 324: What does this do?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -279,6 +280,51 @@ export default class VTable {
         this.forEachCellOfColumn(this.col, callback);
     }
 
+    /**
+     * Loop each table cell and get all the cells that share the same border.
+     * The result is an array of table cell elements where the first element is the narrowest td
+     * @param borderPos The position of the border
+     * @param isRightBorder Indicate whether it is a right border or left border
+     */
+    getCellsWithBorder(borderPos: number, isRightBorder: boolean): HTMLTableCellElement[] {
```
JiuqingSong, line 289: Better add more comments inside code, and add an example

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -335,6 +337,14 @@ export default class TableResize implements EditorPlugin {
 
     private startVerticalResizeTable = (e: MouseEvent) => {
         this.resizingState = ResizeState.Vertical;
+
+        const vtable = new VTable(this.currentTd);
+        if (vtable) {
+            const rect = normalizeRect(this.currentTd.getBoundingClientRect());
+            this.currentCellsToResize = vtable.getCellsWithBorder(rect.right, true);
```
JiuqingSong, line 344: add inline comment for constant parameters

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -335,6 +337,14 @@ export default class TableResize implements EditorPlugin {
...
+            this.nextCellsToResize = vtable.getCellsWithBorder(rect.right, false);
```
JiuqingSong, line 345: Any special handling for RTL?

packages/roosterjs-editor-dom/lib/table/VTable.ts
flyingbee2012: cells[pos] is the cell whose border is closest to the border we are dragging. This is to swap it with the first element in the array so that the caller can just use it without doing a loop to search for it.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: I don't think we need special handling for RTL as it is working regardless of LTR or RTL

JiuqingSong: I found that in RTL the whole resizing functionality not working correctly. This also repro without this change. I remember it worked before, not sure which change broke it.

I suggest we fix the RTL issue first then do this change.

image

packages/roosterjs-editor-dom/lib/table/VTable.ts
flyingbee2012: added comments

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: done

JiuqingSong (approved):

PR #526, opened 2021-04-08
Fix paste two lines plain text bug

JiuqingSong:
The code for pasting didn't handle the two lines of plain text scenario. So rewrite the part to make sure it works correctly for all cases. Also update test case.

### DIFF
--- a/packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
+++ b/packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
@@ -113,10 +113,23 @@ export const createPasteFragment: CreatePasteFragment = (
                 .replace(/^ /g, NBSP_HTML)
                 .replace(/\r/g, '')
                 .replace(/ {2}/g, ' ' + NBSP_HTML);
+            const textNode = document.createTextNode(line);
 
-            const node = line == '' ? document.createElement('br') : document.createTextNode(line);
-
-            fragment.appendChild(index == 0 || index == lines.length - 1 ? node : wrap(node));
+            // There are 3 scenarios:
+            // 1. Single line: Paste as it is
+            // 2. Two lines: Add <br> between the lines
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
```diff
@@ -113,10 +113,23 @@ export const createPasteFragment: CreatePasteFragment = (
...
-
-            const node = line == '' ? document.createElement('br') : document.createTextNode(line);
-
-            fragment.appendChild(index == 0 || index == lines.length - 1 ? node : wrap(node));
+            const textNode = document.createTextNode(line);
+
+            // There are 3 scenarios:
+            // 1. Single line: Paste as it is
+            // 2. Two lines: Add <br> between the lines
+            // 3. 3 or More lines, For first and last line, paste as it is. For middle lines, wrap with DIV, and add BR if it is empty line
```
flyingbee2012, line 121: Just curious...If we ignore the # of lines, just always do "wrap with DIV, and add BR if it is empty line" for each line, would it still work?

JiuqingSong:
Consider you only copy 1 line, say "aaa", and the target content is "bbbccc", and you want to paste it between "bbb" and "ccc".
The expected paste result should be "bbbaaaccc".
But if we always wrap the content with DIV, when paste, it will become "bbb
aaa
ccc"

packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
JiuqingSong: Consider you only copy 1 line, say "aaa", and the target content is "bbbccc", and you want to paste it between "bbb" and "ccc".
The expected paste result should be "bbbaaaccc".
But if we always wrap the content with DIV, when paste, it will become "bbb
aaa
ccc"

flyingbee2012: I see.

flyingbee2012 (approved):

PR #528, opened 2021-04-08
Fix table inserter for culture in RTL 

flyingbee2012:
fix_rtl2

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
@@ -1,4 +1,4 @@
-import { fromHtml, normalizeRect, VTable } from 'roosterjs-editor-dom';
+import { fromHtml, getComputedStyle, normalizeRect, VTable } from 'roosterjs-editor-dom';
 import {
     EditorPlugin,
     IEditor,
@@ -45,6 +45,7 @@ export default class TableResize implements EditorPlugin {
     private currentInsertTd: HTMLTableCellElement;
     private insertingState: ResizeState = ResizeState.None;
     private inserter: HTMLDivElement;
+    private isRTL: boolean;
 
     /**
      * Get a friendly name of  this plugin
@@ -112,18 +113,32 @@ export default class TableResize implements EditorPlugin {
             this.cacheRects();
         }
 
+        this.isRTL = getComputedStyle(this.editor.getDocument().body, 'direction') == 'rtl';
+
         if (this.tableRectMap) {
             let i = this.tableRectMap.length - 1;
             for (; i >= 0; i--) {
                 const { table, rect } = this.tableRectMap[i];
-                if (
-                    e.pageX >= rect.left - INSERTER_SIDE_LENGTH &&
-                    e.pageX <= rect.right &&
-                    e.pageY >= rect.top - INSERTER_SIDE_LENGTH &&
-                    e.pageY <= rect.bottom
-                ) {
-                    this.setCurrentTable(table, rect);
-                    break;
+                if (this.isRTL) {
+                    if (
+                        e.pageX <= rect.right + INSERTER_SIDE_LENGTH &&
+                        e.pageX >= rect.left &&
+                        e.pageY >= rect.top - INSERTER_SIDE_LENGTH &&
+                        e.pageY <= rect.bottom
+                    ) {
+                        this.setCurrentTable(table, rect);
+                        break;
+                    }
+                } else {
+                    if (
+                        e.pageX >= rect.left - INSERTER_SIDE_LENGTH &&
+                        e.pageX <= rect.right &&
+                        e.pageY >= rect.top - INSERTER_SIDE_LENGTH &&
+                        e.pageY <= rect.bottom
+                    ) {
+                        this.setCurrentTable(table, rect);
+                        break;
+                    }
                 }
             }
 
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -112,18 +113,32 @@ export default class TableResize implements EditorPlugin {
             this.cacheRects();
         }
 
+        this.isRTL = getComputedStyle(this.editor.getDocument().body, 'direction') == 'rtl';
```
JiuqingSong, line 116: Since you calculate this value everytime when mouse move, why need to cache it?

Actually you can cache it, but only calculate it when a new resize session is starting, e.g., move this line up to L114.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -112,18 +113,32 @@ export default class TableResize implements EditorPlugin {
...
+                    if (
+                        e.pageX <= rect.right + INSERTER_SIDE_LENGTH &&
+                        e.pageX >= rect.left &&
+                        e.pageY >= rect.top - INSERTER_SIDE_LENGTH &&
+                        e.pageY <= rect.bottom
+                    ) {
+                        this.setCurrentTable(table, rect);
+                        break;
+                    }
+                } else {
```
JiuqingSong, line 132: The "if" block and "else" block has exactly same action, so we should merge them.

You can write some util function to help you get real "right" of rect, e.g.

function getLeftWithDir(rect: Rect, margin: number, isRTL: boolean) {
  return isRTL ? rect.right + margin : rect.left - margin;
}

Then you can call it in a single if block

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -142,54 +157,125 @@ export default class TableResize implements EditorPlugin {
...
-                                let verticalInserterTd: HTMLTableCellElement = null;
-                                // set inserter at current td
-                                if (e.pageX >= tdRect.left + (tdRect.right - tdRect.left) / 2.0) {
-                                    verticalInserterTd = td;
-                                } else if (e.pageX >= tdRect.left) {
-                                    // set inserter at previous td if it exists
-                                    const preTd = td.previousElementSibling as HTMLTableCellElement;
-                                    if (preTd) {
-                                        verticalInserterTd = preTd;
+                        if (this.isRTL) {
```
JiuqingSong, line 160: similar here, try to merge them.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: Good question! I was about to ask if we can add this logic into the **initialize** method instead, that way, we only calculate this value once, but I did not make that change since we cannot test this feature that way. What do you think?
```
 
initialize(editor: IEditor) {
        this.editor = editor;
       this.isRTL = getComputedStyle(this.editor.getDocument().body, 'direction') == 'rtl';
        ....
    }
```

JiuqingSong: Calculate during initialize may not be good since RTL can be changed during editing. You can calculate it each time the inserter is added, then during the mouse move no need to calculate again.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: merged

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: merged

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -230,12 +251,15 @@ export default class TableResize implements EditorPlugin {
         if (this.insertingState == ResizeState.None) {
             return;
         }
+
+        this.isRTL = getComputedStyle(this.editor.getDocument().body, 'direction') == 'rtl';
```
JiuqingSong, line 255: this.isRTL is used before this line. Need to do this earlier. And no need to calculate everytime when mouse move.

JiuqingSong, line 255: this.isRTL is ued before inserter is created. Need a better place to do this.

flyingbee2012: I moved it after "const inserter = fromHtml..."

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -230,12 +251,14 @@ export default class TableResize implements EditorPlugin {
         if (this.insertingState == ResizeState.None) {
             return;
         }
+
         const rect = normalizeRect(this.currentInsertTd.getBoundingClientRect());
         const editorBackgroundColor = this.editor.getDefaultFormat().backgroundColor;
         const inserterBackgroundColor = editorBackgroundColor || 'white';
         const inserterColor = this.editor.isDarkMode() ? INSERTER_COLOR_DARK_MODE : INSERTER_COLOR;
+        const leftOrRight = this.isRTL ? 'right' : 'left';
```
JiuqingSong, line 259: Even in this same function, isRtl is used here before it is assigned.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -245,17 +268,27 @@ export default class TableResize implements EditorPlugin {
             this.editor.getDocument()
         )[0] as HTMLDivElement;
 
+        this.isRTL = getComputedStyle(this.editor.getDocument().body, 'direction') == 'rtl';
```
JiuqingSong, line 271: You can move this into funciton cacheRects

JiuqingSong (approved):

PR #529, opened 2021-04-09
All custom paste type

JiuqingSong:
Add a new editor option: allowedCustomPasteType. Use this option to allow retrieving more content type when paste besides default types.

To use it, pass your custom values when init editor.

```ts
const editor = new Editor(div, {
  ...
  allowedCustomPasteType: [ 'type1', 'type2' ],
};
```

Note that only **text types** as supported, and no need to add the "text/" prefix.

To get the custom value, use ClipboardData.customValues:

```ts
onPluginEvent(e) {
  if (e.eventType == PluginEventType.BeforePaste) {
     const valueOfType1 = e.clipboardData.customValues['type1'];
     ...
  }
}
```


elisabethcvs (approved):
So when putting the custom type on the clipboard, it needs to have a text/ prefix in front of it, but not when setting the editor settings or retrieving it. Probably worth documenting

PR #531, opened 2021-04-13
fix backspace in list issue

JiuqingSong:
Repro:
1. Enable content edit feature "Merge in new line when Backspace on first char in list"
2. create a list
3. locate to the beginning of the second list item, shift+right to select some text
4. BackSpace

Expect: selected text is deleted, no other change
Actual: a new line is created

Reason: the content edit feature "Merge in new line when Backspace on first char in list" is run here which should not
Fix: check if the selection is collapsed.


flyingbee2012 (approved):
looks good to me

PR #535, opened 2021-04-22
New resizer for the whole table

flyingbee2012:
Added a new resizer to control the size of the whole table for both vertical and horizontal resizing. It also works for RTL language culture. 

tableresizer

### DIFF
--- a/packages/roosterjs-editor-api/lib/table/insertTable.ts
+++ b/packages/roosterjs-editor-api/lib/table/insertTable.ts
@@ -30,6 +30,7 @@ export default function insertTable(
             tr.appendChild(td);
             td.appendChild(document.createElement('br'));
             td.style.width = getTableCellWidth(columns);
+            td.style.height = '20px';
         }
     }
 
--- a/packages/roosterjs-editor-api/lib/table/insertTable.ts
+++ b/packages/roosterjs-editor-api/lib/table/insertTable.ts
@@ -30,7 +30,6 @@ export default function insertTable(
             tr.appendChild(td);
             td.appendChild(document.createElement('br'));
             td.style.width = getTableCellWidth(columns);
-            td.style.height = '20px';
         }
     }
 
--- a/packages/roosterjs-editor-dom/lib/table/VTable.ts
+++ b/packages/roosterjs-editor-dom/lib/table/VTable.ts
@@ -273,17 +273,6 @@ export default class VTable {
     }
 
     /**
-     * Loop each cell of the table and invoke a callback function
-     * @param callback The callback function to invoke
-     */
-    forEachCell(callback: (cell: VCell) => any) {
-        for (let i = 0; i < this.cells.length; i++) {
-            for (let j = 0; j < this.cells[i].length; j++) {
-                callback(this.getCell(i, j));
-            }
-        }
-    }
-    /**
      * Loop each cell of current column and invoke a callback function
      * @param callback The callback function to invoke
      */
--- a/packages/roosterjs-editor-dom/lib/table/VTable.ts
+++ b/packages/roosterjs-editor-dom/lib/table/VTable.ts
@@ -273,6 +273,17 @@ export default class VTable {
     }
 
     /**
+     * Loop each cell of the table and invoke a callback function
+     * @param callback The callback function to invoke
+     */
+    forEachCell(callback: (cell: VCell) => any) {
+        for (let i = 0; i < this.cells.length; i++) {
+            for (let j = 0; j < this.cells[i].length; j++) {
+                callback(this.getCell(i, j));
+            }
+        }
+    }
+    /**
      * Loop each cell of current column and invoke a callback function
      * @param callback The callback function to invoke
      */
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
@@ -526,37 +526,47 @@ export default class TableResize implements EditorPlugin {
                     : (e.pageX - currentBorder) / (currentBorder - rect.left));
             const ratioY = 1.0 + (e.pageY - tableBottomBorder) / (tableBottomBorder - rect.top);
 
-            for (let i = 0; i < vtable.cells.length; i++) {
-                for (let j = 0; j < vtable.cells[i].length; j++) {
-                    const cell = vtable.cells[i][j];
-                    if (cell.td) {
-                        if (ratioX != 1.0) {
-                            const originalWidth: number = cell.td.style.width
-                                ? parseFloat(
-                                      cell.td.style.width.substr(0, cell.td.style.width.length - 2)
-                                  )
-                                : cell.td.getBoundingClientRect().right -
-                                  cell.td.getBoundingClientRect().left;
-                            const newWidth = originalWidth * ratioX;
-                            cell.td.style.boxSizing = 'border-box';
-                            if (newWidth >= MIN_CELL_WIDTH) {
-                                cell.td.style.wordBreak = 'break-word';
-                                cell.td.style.width = `${newWidth}px`;
+            const shouldResizeX = Math.abs(ratioX - 1.0) > 1e-3;
+            const shouldResizeY = Math.abs(ratioY - 1.0) > 1e-3;
+            if (shouldResizeX || shouldResizeY)
+                for (let i = 0; i < vtable.cells.length; i++) {
+                    for (let j = 0; j < vtable.cells[i].length; j++) {
+                        const cell = vtable.cells[i][j];
+                        if (cell.td) {
+                            if (shouldResizeX) {
+                                const originalWidth: number = cell.td.style.width
+                                    ? parseFloat(
+                                          cell.td.style.width.substr(
+                                              0,
+                                              cell.td.style.width.length - 2
+                                          )
+                                      )
+                                    : cell.td.getBoundingClientRect().right -
+                                      cell.td.getBoundingClientRect().left;
+                                const newWidth = originalWidth * ratioX;
+                                cell.td.style.boxSizing = 'border-box';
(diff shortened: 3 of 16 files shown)


packages/roosterjs-editor-api/lib/table/insertTable.ts
```diff
@@ -30,6 +30,7 @@ export default function insertTable(
             tr.appendChild(td);
             td.appendChild(document.createElement('br'));
             td.style.width = getTableCellWidth(columns);
+            td.style.height = '20px';
```
JiuqingSong, line 33: We should not rely on this value to do resize. It is possible that user paste some table which doesn't have a height style. We can calcuate the height at runtime, but not read its style value.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -272,6 +272,17 @@ export default class VTable {
         }
     }
 
+    /**
+     * Loop each cell of the table and invoke a callback function
+     * @param callback The callback function to invoke
+     */
+    forEachCell(callback: (cell: VCell) => any) {
```
JiuqingSong, line 279: I prefer not adding this public method unless it will be used by multiple files. You can move it into TableResize.ts. So that we don't need to bump minor version number.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -15,17 +15,23 @@ const INSERTER_COLOR_DARK_MODE = 'white';
...
-
+const CELL_MINIMUM_LENGTH = 20;
 const CELL_RESIZER_WIDTH = 4;
+const TABLE_RESIZER_LENGTH = 12;
 const HORIZONTAL_RESIZER_HTML =
     '<div style="position: fixed; cursor: row-resize; user-select: none"></div>';
 const VERTICAL_RESIZER_HTML =
     '<div style="position: fixed; cursor: col-resize; user-select: none"></div>';
+const TABLE_RESIZER_HTML_LTR =
+    '<div style="position: fixed; cursor: nw-resize; user-select: none; border: 1px solid #808080"></div>';
```
JiuqingSong, line 26: Can the color #808080 work correctly in dark mode?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -99,11 +107,21 @@ export default class TableResize implements EditorPlugin {
             replaceSelection: false,
             position: ContentPosition.Outside,
         });
+
+        this.tableResizerContainer = this.editor.getDocument().createElement('div');
```
JiuqingSong, line 111: document is used twice in this funciton, put into a local variable

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -99,11 +107,21 @@ export default class TableResize implements EditorPlugin {
...
+            insertOnNewLine: false,
+            replaceSelection: false,
+            position: ContentPosition.Outside,
+        });
     }
 
     private removeResizerContainer() {
         this.resizerContainer.parentNode.removeChild(this.resizerContainer);
         this.resizerContainer = null;
+        this.tableResizerContainer.parentNode.removeChild(this.tableResizerContainer);
```
JiuqingSong, line 123: Where there need to have a separate container for table resizer? can they share the same one?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -179,7 +196,7 @@ export default class TableResize implements EditorPlugin {
                                         verticalInserterTd,
                                         map.rect
                                     );
-                                    break;
+                                    return;
```
JiuqingSong, line 199: I prefer to use break.

In the future if we want to add some other code after this loop, we may not notice that the code can return here and the new added code won't be executed.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -207,7 +224,7 @@ export default class TableResize implements EditorPlugin {
                                         horizontalInserterTd,
                                         map.rect
                                     );
-                                    break;
+                                    return;
```
JiuqingSong, line 227: same here

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -217,14 +234,13 @@ export default class TableResize implements EditorPlugin {
                                     tdRect.bottom
                                 );
                                 this.setCurrentInsertTd(ResizeState.None);
-                                break;
+                                return;
```
JiuqingSong, line 237: same

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -322,9 +338,9 @@ export default class TableResize implements EditorPlugin {
         }, ChangeSource.Format);
     };
 
-    private setCurrentTable(table: HTMLTableElement, rect: Rect): void;
+    private setCurrentTable(table: HTMLTableElement): void;
     private setCurrentTable(table: null): void;
-    private setCurrentTable(table: HTMLTableElement, rect?: Rect) {
+    private setCurrentTable(table: HTMLTableElement) {
```
JiuqingSong, line 343: We can remove the function override now.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -377,13 +393,46 @@ export default class TableResize implements EditorPlugin {
         }
     }
 
-    private createResizer(
+    private setTableResizer(rect: null): void;
+    private setTableResizer(rect: Rect): void;
```
JiuqingSong, line 397: No need to have function override here.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -377,13 +393,46 @@ export default class TableResize implements EditorPlugin {
...
+    private setTableResizer(rect: Rect | null): void {
+        // remove old one if exists
+        while (this.tableResizerContainer.hasChildNodes()) {
```
JiuqingSong, line 400: null check for tableResizerContainer?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -377,13 +393,46 @@ export default class TableResize implements EditorPlugin {
...
+            this.tableResizerContainer.removeChild(this.tableResizerContainer.lastChild);
+        }
+        this.tableResizer = null;
```
JiuqingSong, line 403: remove event handler?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -515,29 +638,24 @@ export default class TableResize implements EditorPlugin {
...
+            this.frameAnimateResizeCells(e);
             this.editor.select(start, end);
         }, ChangeSource.Format);
 
         this.setCurrentTd(null);
+        this.setTableResizer(null);
         this.resizingState = ResizeState.None;
     };
 
-    private destoryRectMap() {
```
JiuqingSong, line 534: why do we need to removethis method?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: tableResizerContainer is created in initialize() and removed in dispose(), so it should exist when we call this method.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: Yes, the event handler is still onMouseMove(mousemove) and we already removed it in endResizeCells code.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: It could happen that we have both inserter and table resizer shown at the same time (same behavior in Word). If we share the same container, it would make thing a little bit complicated, let's say, it is showing both inserter and table resizer, now the user moves mouse to the middle of the table, so we want to remove the inserter from the container but still keep the table resizer. I feel like keeping them in separate containers will make it less bug prone.

packages/roosterjs-editor-dom/lib/table/VTable.ts
flyingbee2012: fixed

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: done

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: done

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: yes

Capture_dark

JiuqingSong (approved):

PR #540, opened 2021-04-30
Paste HTML when copy from PowerPoint

JiuqingSong:


packages/roosterjs-editor-plugins/lib/plugins/Paste/pptConverter/convertPastedContentFromPowerPoint.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    if (clipboardData.html && !clipboardData.text && clipboardData.image) {
+        // It is possible that PowerPoint copied both image and HTML but not plain text.
+        // We always prefer HTML if any.
+        const doc = new DOMParser().parseFromString(clipboardData.html, 'text/html');
+
+        while (fragment.firstChild) {
+            fragment.removeChild(fragment.firstChild);
+        }
+        while (doc?.body?.firstChild) {
```
flyingbee2012, line 19: Should we use `if `there, would it cause an endless loop?

JiuqingSong: no. the node will be removed from original tree in the next line

flyingbee2012 (approved):

PR #542, opened 2021-05-04
Do not show inserter while left mouse button is pressed

flyingbee2012:
A behavior in parity with Word

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
@@ -643,53 +643,30 @@ export default class TableResize implements EditorPlugin {
                                 : Number.MAX_SAFE_INTEGER;
                     }
 
-                    if (newPos <= leftBoundary + 50) {
-                        /*if (
+                    if (
                         newPos <= leftBoundary + MIN_CELL_WIDTH ||
                         newPos >= rightBoundary - MIN_CELL_WIDTH
-                    ) {*/
-
+                    ) {
                         return;
                     }
 
-                    /*this.currentCellsToResize.forEach(td => {
-                        console.log('***** current td: ' + td.textContent);
+                    this.currentCellsToResize.forEach(td => {
                         const rect = normalizeRect(td.getBoundingClientRect());
                         td.style.wordBreak = 'break-word';
-                        const offset = this.getTdOffsetWidth(td);
-                        const newWidth = newPos - rect.left - offset;
+                        td.style.boxSizing = 'border-box';
                         td.style.width = this.isRTL
-                            ? `${rect.right - newPos - offset}px`
-                            : `${newPos - rect.left - offset}px`;
-                    });*/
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -73,13 +74,18 @@ export default class TableResize implements EditorPlugin {
         this.editor = editor;
         this.setupResizerContainer();
         this.onMouseMoveDisposer = this.editor.addDomEventHandler('mousemove', this.onMouseMove);
+        this.editor.addDomEventHandler('mousedown', this.onMouseDown);
+        this.editor.addDomEventHandler('mouseup', this.onMouseUp);
```
JiuqingSong, line 78: no need to do this, you can directly get button info from MouseEvent.which from mousemove event

flyingbee2012: Good to know that, I was trying to use e.button the MouseEvent inside onMouseMove but it didn't work.

flyingbee2012: Used e.buttons instead as e.which seems deprecated.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -194,11 +194,14 @@ export default class TableResize implements EditorPlugin {
                                 }
                                 if (verticalInserterTd) {
                                     this.setCurrentTd(null);
-                                    this.setCurrentInsertTd(
-                                        ResizeState.Vertical,
-                                        verticalInserterTd,
-                                        map.rect
-                                    );
+                                    // we hide the inserter if left mouse button is pressed
+                                    if (e.buttons !== 1) {
```
JiuqingSong, line 198: what if I press both left and right button? Will buttons be 2? I prefer check buttons == 0 here

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -194,11 +194,14 @@ export default class TableResize implements EditorPlugin {
                                 }
                                 if (verticalInserterTd) {
                                     this.setCurrentTd(null);
-                                    this.setCurrentInsertTd(
-                                        ResizeState.Vertical,
-                                        verticalInserterTd,
-                                        map.rect
-                                    );
+                                    // we hide the inserter if left mouse button is pressed
+                                    if (e.buttons == 0) {
```
JiuqingSong, line 198: Actually, why not just do this check at very beginning of this function (line 130), and return directly if buttons is not 0? Not need to do all further calculation.

JiuqingSong: Oh I see, so in the dragging scenario, onmousemove still need to work.

JiuqingSong (approved):

PR #544, opened 2021-05-08
Add null check for resizerContainer

flyingbee2012:


packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -373,8 +379,10 @@ export default class TableResize implements EditorPlugin {
     ) {
         if (this.currentTd != td) {
             if (this.currentTd) {
-                this.resizerContainer.removeChild(this.horizontalResizer);
-                this.resizerContainer.removeChild(this.verticalResizer);
+                if (this.resizerContainer) {
+                    this.resizerContainer.removeChild(this.horizontalResizer);
```
JiuqingSong, line 383: Since we don't have a stable repro, we don't even know if this.horizontalResizer exists, or if this.resizerContainer is the parent of this.horizontalResizer.

So I suggest to do
```
this.horizontalResizer?.parentNode?.removeChild(this.horizontalResizer);
```

JiuqingSong: same for other places

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -120,10 +120,14 @@ export default class TableResize implements EditorPlugin {
     }
```
JiuqingSong, line 121: please bump version in package.json to 8.2.4

JiuqingSong (approved):

PR #550, opened 2021-05-14
Preprocess table to make table resize plugin more compatible

flyingbee2012:
In this change, we implement a preprocessor to transform the HTML content of the table so that we can resize it properly:
1. Remove useless width/height HTML attributes (presentational attributes) as they will be overriden by css properties
2. Convert pt to px if there is any
3. If unit used is neither px nor pt, or, no unit specified, caculate it based on bounding rectangle

After this change, the table resize plugin can handle tables pasted from most office apps: Word/Excel/PPT/OneNote, and some tables from external websites.


packages/roosterjs-editor-dom/lib/utils/preprocessTable.ts
```diff
@@ -0,0 +1,47 @@
+import normalizeRect from './normalizeRect';
```
JiuqingSong, line 1: If this function is only used by TableResize, put it under the table resize folder and no need to export from index.

packages/roosterjs-editor-dom/lib/utils/preprocessTable.ts
```diff
@@ -0,0 +1,47 @@
...
+ * Preprocess the HTML content of the table so that we can resize it properly:
+ *     1. Remove useless width/height HTML attributes (presentational attributes) as they will be overriden by css properties
+ *     2. Convert pt to px if there is any
+ *     3. If unit used is neither px nor pt, or, no unit specified, caculate it based on bounding rectangle
+ */
+
+const Pt2PxCoefficient = 1.333;
+
+export default function preProcessTable(table: HTMLTableElement) {
+    if (table.hasAttribute('width')) {
```
JiuqingSong, line 13: what if the table only has "width" and no cells has "width"? remove this attribute will make the table looks smaller (or bigger)

JiuqingSong: And when you want to remove an attribute, no need to check if "hasAttribute", just remove it directly.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -506,6 +512,10 @@ export default class TableResize implements EditorPlugin {
     }
 
     private frameAnimateResizeCells = (e: MouseEvent) => {
+        if (!this.currentTable.hasAttribute('isPreProcessed')) {
```
JiuqingSong, line 515: I think we can just preprocess it every time, no need to have "isPreProcessed".

flyingbee2012: good point

flyingbee2012: I tried with preprocess every time with the current change, it would set table width each time, causing some jumpy effect when we resize each column, let's keep it as it is.

JiuqingSong:
Do you think we should put this change under an experimentalFeature?

flyingbee2012:
> experimentalFeature

I think it should be fine as this change is to fix existing issues with external tables. Is there any concern with this change?

JiuqingSong:
table

When start resizing, the table becomes shorter.

I feel we should not remove all height and weight, but we should keep each cell's size no change, then remove unnecessary styles and attributes.

flyingbee2012:
> table
> 
> When start resizing, the table becomes shorter.
> 
> I feel we should not remove all height and weight, but we should keep each cell's size no change, then remove unnecessary styles and attributes.

This is caused by setting cell width where it will strangely adjust the height as well, but we do need to do some calculation/conversion and set each cell's style.width with uniformed unit. Otherwise, it won't resize properly later. It is fixed by giving the table a fixed height.

JiuqingSong:
I see weird behavior when try resize a table pasted from Excel:

tablebug

JiuqingSong:
I suggest we rethink about the implementation of table resize, and leverage VTable ability.
When start to resize, create VTable and calculate size of each cells. Then each time mouse move, we apply the new width/height to VTable, then let VTable write back to reflect the change, but not directly change size of TDs. If necessary, we can even recreate TABLE/TR/TD nodes.

JiuqingSong:
I still see the size shifting happen. To repro, just start Excel, copy some empty cells and paste into roosterjs, then start resize.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
```diff
@@ -0,0 +1,98 @@
...
+        if (dimension === Dimension.WIDTH || dimension === Dimension.BOTH) {
+            element.removeAttribute('width');
+            element.removeAttribute('height');
+            const unit = element.style?.width?.substr(element.style.width.length - 2);
+            if (!unit || unit !== 'px' /*&& unit !== 'pt'*/) {
+                const rect = normalizeRect(element.getBoundingClientRect());
+                element.style.boxSizing = 'border-box';
+                element.style.width = `${rect.right - rect.left}px`;
+            }
+        }
```
JiuqingSong, line 27: I still think we should not remove width and height that early.

Can we just always get the size using getBoundingClientRect() ? Then after that we apply the calcualted size then we can remove original width/height attribute.

Removing width/height before getBoundingClientRect() may cause the table size changed.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
```diff
@@ -0,0 +1,98 @@
...
+
+        if (dimension === Dimension.HEIGHT || dimension === Dimension.BOTH) {
+            const unit2 = element.style?.height?.substr(element.style.height.length - 2);
+            if (!unit2 || unit2 !== 'px' /*&& unit2 !== 'pt' */) {
```
JiuqingSong, line 31: remove unused comment

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
JiuqingSong: Then we don't ned to care the difference of unit (pt, px, ...)

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
```diff
@@ -0,0 +1,98 @@
...
+                const rect = normalizeRect(element.getBoundingClientRect());
+                element.style.boxSizing = 'border-box';
+                element.style.height = `${rect.bottom - rect.top}px`;
```
JiuqingSong, line 34: And set size of one cell may impact the size of other cells. It is also not good for perf.
The best practice is to get all sizes together, then write back together. So 1 read, 1 write, only cause 1 relayout. It is the worst to have read/write/read/write/... loop because each time browser need to relayout.

That is why I think relying on VTable may be good here. You can calculate size for all cells using getBoundingClientRect at once, write the calcualted size to VTable. Then call writeBack() to let VTable help you flush all the calculated sizes, and remove other attributes.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
flyingbee2012: can you try again, I set the height to ensure the height doesn't change. There is still minor shifting horizontally but it is only for the first time for Excel. It would be difficult to make it perfect and the change would be expensive. Even in Word, I am seeing similar shifting.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
flyingbee2012: It is already one-time calculate during one resizing operation. We have the variables such as isReadyForResizingWidth to ensure all the cell size is already calculated to avoid unnecessary calculation.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
flyingbee2012: Can we just always get the size using getBoundingClientRect() ?

That's the original approach I tried and it would cause the resizing not proportionally, We need to set style.width before reiszing.

JiuqingSong: I believe the size shifit is caused by "box-sizing: border-box". When set this style, border width will be counted as part of TD width. so that when we add this style if it was not there, we also need to calculate the width/size with border width.

By saying using getBoundingClientRect(), I mean:

1. When start to resize, calculate size of each cell using getBoundingClientRect.
2. After get size of all cells, write them back in CSS width/height, and remove HTML attribute width/height
3. Then we can really start to resize.

So that is still "before resizing"

The difference is, we never read the original HTML width/height and CSS width/height. Instead, we read the value from getBoundingClientRect(). Because the real size can be different than width/height in HTML or CSS when there is size conflict between TABLE/TR/TD width/height. Browser will calculate a size for us which may be different than the value in HTML.

JiuqingSong:
tablenew

As you can see, when I start resize by moving mouse a little bit, it change the size, but if I move mouse again, the size changes to normal. That means you can already calculate the correct size, although there will be a step to have wrong size. Why not just go with both steps together when the first time moving mouse? In that way I think it will look perfect.

flyingbee2012:
> tablenew
> 
> As you can see, when I start resize by moving mouse a little bit, it change the size, but if I move mouse again, the size changes to normal. That means you can already calculate the correct size, although there will be a step to have wrong size. Why not just go with both steps together when the first time moving mouse? In that way I think it will look perfect.

@JiuqingSong I think you are referring to 2 steps right and you suggest jumping directly from state 1 to state 3, but as long as we are starting from state 1, we will always experience the visual inconsistency. Can we talk about this in the meeting I just scheduled?
1. raw state (state 1)=> our state (state 2)
2. our state (state )2 => our new state (state 3)

PR #554, opened 2021-05-21
Moving to 3001 to make it easier to run at same time

aniagonzalez:
We sometimes need to run Rooster at the same time as we have something else running on port 3000.

Would it be ok to change Rooster to 3001?


aniagonzalez:
Closing because Lingling is adding a detect-port functionality which would be better

PR #555, opened 2021-05-21
8.2.8: Fix a bug when copy/cut with empty text node

JiuqingSong:
There is a bug when copy/cut content if there is empty text node insde or before the selection. Please see the new test case for the scenario.

Normally an empty text node means nothing, we should remove them. And when we get selection path, continuous text node will be treated as single text node, empty text node will be ignored. However, if we apply a selection path on the original DOM tree or a cloned DOM tree, this becomes a problem because those text node still exist.

e.g. DOM tree:
"text1"
"" <-- empty node
"text2"

It will be renderred as 
"text1text2"

But the offset inside text2 will be calculated together with "text1" and "". So if the original DOM tree or cloned DOM tree is not normalized (merge those text nodes), selection path becomes wrong.

To fix it, we can normalize the DOM tree when we know we will apply such selection path to the DOM tree directly. For those cases that selection path is applied to DOM tree generated from innerHTML, we don't need to fix it since when we get innerHTML, it is already normalized.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "roosterjs",
-    "version": "8.2.7",
+    "version": "8.2.8",
     "description": "Framework-independent javascript editor",
     "repository": {
         "type": "git",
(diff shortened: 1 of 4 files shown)


flyingbee2012 (approved):

JiuqingSong:
#1145

PR #556, opened 2021-05-24
start demo project using first available port from 3000

litong2:
Prevent conflicts with projects that use http://localhost:3000 for webpack-dev-server.  OWA gulp and fluentui can run on http://localhost:3000.  Port availability is found using https://github.com/node-modules/detect-port which is what fluentui storybook uses.  

'yarn start' will launch rooster demo on http://localhost:3000 if free otherwise http://localhost:3001 or next available.


JiuqingSong:
Generally looks good. Please also update readme.md, it mentioned port 3000.

JiuqingSong (approved):

PR #561, opened 2021-06-09
preprocess table to make table resize plugin more compatible

flyingbee2012:


packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -58,6 +51,10 @@ export default class TableResize implements EditorPlugin {
     private inserter: HTMLDivElement;
     private isRTL: boolean;
 
+    private isReadyForResizingWidth: boolean = false;
+    private isReadyForResizingHeight: boolean = false;
+    private isReadyForResizingTable: boolean = false;
+
```
JiuqingSong, line 57: I don't like the idea to have these variables. It makes us always need to manage the flow that we need to reset it to false. If any place we missed to reset, there will be wrong result after that.

I prefer always call preprocessTable, and inside the function we can do some check to determine if we need to preprocess, such as if the table has width/height attribute, if each cell wdith/height are in "px", if the table as border box, ...

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
-                        if (cell.td) {
-                            if (shouldResizeX) {
-                                const originalWidth: number = cell.td.style.width
+        if (!this.isReadyForResizingTable) {
+            preProcessTable(vtable.table, ResizeState.Both);
+            this.isReadyForResizingTable = true;
+        }
+        let rect = normalizeRect(vtable.table.getBoundingClientRect());
+        let currentBorder: number = parseFloat(
+            vtable.table.getAttribute(this.isRTL ? 'currentLeftBorder' : 'currentRightBorder')
```
JiuqingSong, line 523: Where is "currentLeftBorder" removed after resize?

JiuqingSong: And since this is only used internally, I prefer make it simpler to "currentVerticalBorder", so no need to do RTL check everywhere.

JiuqingSong: But actually why we need to write this result to table? Can't we just remember it as a local variable or member field?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+        );
+        const tableBottomBorder: number = parseFloat(
+            vtable.table.getAttribute('currentBottomBorder')
+        );
+        const ratioX =
+            1.0 +
+            (this.isRTL
+                ? (currentBorder - mouseX) / (rect.right - currentBorder)
```
JiuqingSong, line 531: Can rect.right - currentBorder be 0?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+                : (mouseX - currentBorder) / (currentBorder - rect.left));
+        const ratioY = 1.0 + (mouseY - tableBottomBorder) / (tableBottomBorder - rect.top);
```
JiuqingSong, line 533: can tableBottomBorder - rect.top be 0?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+        const shouldResizeX = Math.abs(ratioX - 1.0) > 1e-3;
+        const shouldResizeY = Math.abs(ratioY - 1.0) > 1e-3;
+        if (shouldResizeX || shouldResizeY) {
+            for (let i = 0; i < vtable.cells.length; i++) {
+                for (let j = 0; j < vtable.cells[i].length; j++) {
+                    const cell = vtable.cells[i][j];
+                    if (cell.td) {
+                        if (shouldResizeX) {
+                            // the width of some external table is fixed, we need to make it resizeable
+                            vtable.table.removeAttribute('width');
```
JiuqingSong, line 544: I think we already removed "width" when preprocess, right?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+                            vtable.table.style.width = null;
+                            const originalWidth: number = cell.td.style.width
+                                ? parseFloat(
+                                      cell.td.style.width.substr(0, cell.td.style.width.length - 2)
```
JiuqingSong, line 548: what does " - 2" mean here?
no magic number.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+                                  )
+                                : cell.td.getBoundingClientRect().right -
```
JiuqingSong, line 550: Save cell.td.getBoundingClientRect() to a local var since it is used twice

JiuqingSong: Or use cell.td.getBoundingClientRect().width directly

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+                            if (newWidth >= MIN_CELL_WIDTH) {
+                                cell.td.style.wordBreak = 'break-word';
+                                cell.td.style.whiteSpace = 'normal';
+                                cell.td.style.width = `${newWidth}px`;
+                            }
+                        }
+
+                        if (shouldResizeY) {
+                            // the height of some external table is fixed, we need to make it resizeable
+                            vtable.table.removeAttribute('height');
```
JiuqingSong, line 563: same

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+                            vtable.table.style.height = null;
+                            if (j == 0) {
+                                const originalHeight = cell.td.style.height
                                     ? parseFloat(
-                                          cell.td.style.width.substr(
+                                          cell.td.style.height.substr(
                                               0,
-                                              cell.td.style.width.length - 2
+                                              cell.td.style.height.length - 2
```
JiuqingSong, line 570: same

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
-                                    const originalHeight =
-                                        cell.td.getBoundingClientRect().bottom -
-                                        cell.td.getBoundingClientRect().top;
-                                    const newHeight = originalHeight * ratioY;
-                                    if (newHeight >= MIN_CELL_HEIGHT) {
-                                        cell.td.style.height = `${newHeight}px`;
-                                    }
-                                } else {
-                                    cell.td.style.height = '';
+                                    : cell.td.getBoundingClientRect().bottom -
```
JiuqingSong, line 573: same

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+                                      cell.td.getBoundingClientRect().top;
+                                const newHeight = originalHeight * ratioY;
+                                if (newHeight >= MIN_CELL_HEIGHT) {
+                                    cell.td.style.height = `${newHeight}px`;
                                 }
+                            } else {
+                                cell.td.style.height = null;
```
JiuqingSong, line 580: What if there is merged cells in table?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
-            currentBorder = this.isRTL ? rect.left : rect.right;
-            this.currentTable.setAttribute(
-                this.isRTL ? 'currentLeftBorder' : 'currentRightBorder',
-                currentBorder.toString()
-            );
+        }
+        rect = normalizeRect(vtable.table.getBoundingClientRect());
+        currentBorder = this.isRTL ? rect.left : rect.right;
+        vtable.table.setAttribute(
+            this.isRTL ? 'currentLeftBorder' : 'currentRightBorder',
```
JiuqingSong, line 590: again, why need to remember this value?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+
+    private resizeColumns = (newPos: number, vtable: VTable, isShiftPressed: boolean) => {
+        if (!this.isReadyForResizingWidth) {
+            preProcessTable(vtable.table, ResizeState.Vertical);
+            this.isReadyForResizingWidth = true;
+        }
+        // Since we allow the user to resize the table width on adjusting the border of the last cell,
+        // we need to make the table width resizeable by setting it as null;
+        // We also allow the user to resize the table width if Shift key is pressed
+        if (!this.currentCellsToResize[0].nextElementSibling || isShiftPressed) {
```
JiuqingSong, line 620: Better check if current column number is the last one in table, but not checking nextElementSibling

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+            vtable.table.removeAttribute('width');
```
JiuqingSong, line 621: We already removed width when preprocess.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+            vtable.table.style.width = null;
+        }
+
+        this.currentCellsToResize.forEach(td => {
+            const rect = normalizeRect(td.getBoundingClientRect());
+            const nextTd = td.nextElementSibling as HTMLTableElement;
```
JiuqingSong, line 627: Use VTable to help you get next TD. Because next Element sibling may not be TD.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -496,177 +496,187 @@ export default class TableResize implements EditorPlugin {
...
+            td.style.wordBreak = 'break-word';
+            td.style.whiteSpace = 'normal';
+            td.style.boxSizing = 'border-box';
+
+            const nextTdWidth = nextTd
+                ? this.isRTL
+                    ? newPos - normalizeRect(nextTd.getBoundingClientRect()).left
+                    : normalizeRect(nextTd.getBoundingClientRect()).right - newPos
```
JiuqingSong, line 635: This pattern occurs multiple times. Suggest extract to a function:

function getDistance(rect: Rect, pos: number, isRTL: boolean, toLeft: boolean) {
  return !!isRTL == !!toLeft ? newPos - rect.left : rect.right - newPos
}

JiuqingSong (changes requested):

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: It is used to convert string length to a number e.g. "19px" => 19, then we can calculate the new table width based on the ratio. I can add a constant for this.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: fixed

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: In that case I prefer use clientWidth instead of the value from CSS.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: What if the first column/row has merged cells?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: the table size is being changed during resizing so we need to remember this original value

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
```diff
@@ -0,0 +1,118 @@
...
+    None,
+    Horizontal,
+    Vertical,
+    Both, // when resizing the whole table
+}
+
+function setHTMLElementSizeInPx(element: HTMLElement, dimension: Dimension, storeOnly: boolean) {
+    if (!!element) {
+        if (dimension === Dimension.WIDTH || dimension === Dimension.BOTH) {
+            const rect = normalizeRect(element.getBoundingClientRect());
```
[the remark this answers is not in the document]
flyingbee2012: yes, we can :-)

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
```diff
@@ -0,0 +1,118 @@
...
+            }
+        }
+    }
+}
+
+function setTableCells(table: HTMLTableElement, dimension: Dimension) {
+    // measure and store the width/height into attribute
+    for (let i = 0, row; (row = table.rows[i]); i++) {
+        for (let j = 0, cell; (cell = row.cells[j]); j++) {
+            setHTMLElementSizeInPx(cell, dimension, true);
```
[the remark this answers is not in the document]
flyingbee2012: Yes, I thought the thing as well, but I am wondering what data structure we should use for this local variable, considering it is a 2-D array where the length of each row could be different (for merged cells)? Let's keep temp attribute here for simplicity?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: Ok, I agree we need to remember something during resize. So my next question is, is it the only solution to remember it in DOM? Can we remember it somewhere else, such as local var, or in VTable?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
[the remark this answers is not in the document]
JiuqingSong: VTable should be the best place. Event for the whole preprocess function, we can put it into VTable.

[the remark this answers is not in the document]
JiuqingSong: interface VCell {
 ...
 width: number;   // new 
 height: number;
}

Preprocess can get the size and store into VCell. Then writeback() can apply them.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
```diff
@@ -0,0 +1,118 @@
...
+        table.removeAttribute('height');
+        table.style.height = null;
+    } else if (resizeState === ResizeState.Vertical) {
+        setHTMLElementSizeInPx(table, Dimension.BOTH, false); // Make sure table width/height is fixed to avoid shifting effect
+        setTableCells(table, Dimension.WIDTH); // Make sure every cell has 'width' instead of null in order to be resized properly
+        /*
+        Since dragging the last border would cause table width to change, we need to remove width properties
+        if it is not dragging the last border, we should set table width to make sure the table width won't be changed during resizing
+        */
+        if (byAddingNewColumn) {
```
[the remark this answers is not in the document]
flyingbee2012: no, if we always remove table width, it will cause the table width to shrink when resizing, so we should fix the table width while resizing columns to avoid that shrink/shift effect.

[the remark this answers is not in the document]
JiuqingSong: I think it should not shrink as long as we preprocess everything and set width of each cells.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: And if we only set width/height for first column/row, what if user then delete the first column/row?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: we set a minimum width so the table cell cannot be 0

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: good catch!

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: we set a minimum height so the table cell's height cannot be 0

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: we don't remove it after resize, instead, we update it with the new border position based on the new resized position so that we can use set the correct value for the next resizing. I don't use as a local variable, considering the fact that we might have multiple tables, one single local variable won't be enough, so I save it as an attribute.

flyingbee2012: used the 'currentVerticalBorder'

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: fixed

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: fixed by using a new strategy.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: fixed

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: Fixed..Interesting that you are able to see my comment before submitting the review...

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: done

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: fixed

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: done

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: fistcell

Is this what you mean?

flyingbee2012: if use deleted the first column, we will recalculate the first column and it's still working. For rows, we recalculate every cell.  Please free feel to test

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: fixed

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: hmm...I can see there is only one pattern here, where are other ones?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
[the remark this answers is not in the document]
flyingbee2012: preProcessTable(vtable.table, ResizeState.Both);

I am already using the vtable here in this method, right? Interestingly it seems directly modifying the actual table. If we remove vtable.writeBack(), it is directly changing the actual table.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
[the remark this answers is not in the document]
flyingbee2012: fixed

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: Looks like we don't have existing method in VTable to get the next TD, we have Virtual cells there, do you mean we return the cells for the method getCellsWithBorder, and also create a method to get the next cells?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/preprocessTable.ts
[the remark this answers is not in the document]
flyingbee2012: okay. got your point

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -338,6 +344,13 @@ export default class TableResize implements EditorPlugin {
...
     private insertTd = () => {
+        if (this.insertingState === ResizeState.Vertical) {
+            let vtable = new VTable(this.currentTable);
+            vtable.setTableCells();
+            // Since adding new column will cause table width to change, we need to remove width properties
+            this.currentTable.removeAttribute('width');
+            this.currentTable.style.width = null;
+        }
         this.editor.addUndoSnapshot((start, end) => {
             const vtable = new VTable(this.currentInsertTd);
```
JiuqingSong, line 355: VTable is created twice

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -465,47 +478,43 @@ export default class TableResize implements EditorPlugin {
...
-                this.isRTL ? rect.left : rect.right,
-                this.isRTL
-            );
         }
-
         this.startResizeCells(e);
     };
 
     private startResizeCells(e: MouseEvent) {
+        this.resizingVtable.setEmptyTableCells();
```
JiuqingSong, line 517: Let's do this in ctor

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -515,159 +524,169 @@ export default class TableResize implements EditorPlugin {
...
+            this.currentTableHorizontalBorder = rect.bottom;
+
+            for (let i = 0; i < this.resizingVtable.cells.length; i++) {
+                this.cellsBeforeResize.push([]);
+                for (let j = 0; j < this.resizingVtable.cells[i].length; j++) {
+                    this.cellsBeforeResize[i].push(
+                        this.resizingVtable.cells[i][j].td?.cloneNode(true) as HTMLTableCellElement
+                    );
+                }
+            }
```
JiuqingSong, line 541: Actually what you need is just original width and height, why do you need to clone the whole TD?

And since we never preprocess a table again during resizing, you can just use vtable.cells[][].width/height as original size

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -515,159 +524,169 @@ export default class TableResize implements EditorPlugin {
...
+                    const cell = this.resizingVtable.cells[i][j];
+                    const originalTd = this.cellsBeforeResize[i][j];
+                    if (cell.td && originalTd) {
+                        if (shouldResizeX) {
+                            // the width of some external table is fixed, we need to make it resizeable
+                            this.resizingVtable.table.style.width = null;
+                            const originalWidth: number = parseFloat(
+                                originalTd.style.width.substr(
+                                    0,
+                                    originalTd.style.width.indexOf('px')
```
JiuqingSong, line 573: I suggest we never read the value of CSS width/height. That is for browser to read. We can always use either clientWidth/Height or getBoundingClientRect()

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: Can you point me the code to recalculate?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: For example, if you want to know if a cell is the last cell of this row, you can do:

const isLastTd = td == vTable.getTd(rowIndex, vTable.cells[rowIndex].length - 1)  // Need to amke getTd public

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: Line 550, Line 649, Line 656, Line 663

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: And to get nextTd for resize column, you can do 
vTable.getCell(rowIndex, colIndex + 1)?.td

If it return null, it means that is a merged cell so no need to resize.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: And no need to have isReadyForResizingTable.

JiuqingSong (changes requested):

JiuqingSong:
And found a bug, not related to this change, also repro in aka.ms/roosterjs

1. Create a table
2. Do whole table resize
3. Delete first row
Expect:
First row is deleted, table height is reduced by the height of the deleted row

Actual:
First row is deleted, table height is not changed.

flyingbee2012:
> And found a bug, not related to this change, also repro in aka.ms/roosterjs
> 
> 1. Create a table
> 2. Do whole table resize
> 3. Delete first row
>    Expect:
>    First row is deleted, table height is reduced by the height of the deleted row
> 
> Actual:
> First row is deleted, table height is not changed.

Hmm...I don't get the repro.

cellnorepro2

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -450,6 +435,69 @@ export default class VTable {
         }
         return result;
     }
+
+    private setHTMLElementSizeInPx(element: HTMLElement, newWidth?: number, newHeight?: number) {
```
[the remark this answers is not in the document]
flyingbee2012: fixed

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: fixed!

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -450,6 +435,69 @@ export default class VTable {
...
+        for (let i = 0, row; (row = this.table.rows[i]); i++) {
+            for (let j = 0, cell; (cell = row.cells[j]); j++) {
+                if (!cell.innerHTML) {
+                    cell.appendChild(document.createElement('br'));
+                }
+            }
+        }
+    }
+
+    public setTableCells() {
```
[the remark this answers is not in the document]
flyingbee2012: this is used in TableResize in insertTd method.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -450,6 +435,69 @@ export default class VTable {
...
+            element.style.width = `${
+                newWidth !== undefined ? newWidth : element.getBoundingClientRect().width
+            }px`;
+            element.style.height = `${
+                newHeight !== undefined ? newHeight : element.getBoundingClientRect().height
+            }px`;
+        }
+    }
+
+    public setEmptyTableCells() {
```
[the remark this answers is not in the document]
flyingbee2012: fixed

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: Fixed the first one, but I think we need isReadyForResizingTable to calculate the ratioX/ratioY based on the original table size (not the original td size.), right? The table size is being changed during resizing, so we need isReadyForResizingTable  so that we calculate only once.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -450,6 +435,69 @@ export default class VTable {
...
+        // measure and store the width/height into VCell
+        for (let i = 0, row; (row = this.table.rows[i]); i++) {
+            for (let j = 0, cell; (cell = row.cells[j]); j++) {
+                if (this.cells[i][j]) {
+                    const rect = cell.getBoundingClientRect();
+                    this.cells[i][j].width = rect.width;
+                    this.cells[i][j].height = rect.height;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
flyingbee2012: Done. I did this during resizing was to delay the shifting effect (mouse move), since now the shifting issue was resolved, I can just move this logic in ctor (mouse down).

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: I feel like the current pattern looks more straightforward to me, having a function with 4 parameter feels more obscure to understand, let's keep it as it is?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: fixed

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -450,6 +437,66 @@ export default class VTable {
         }
         return result;
     }
+
+    private setEmptyTableCells() {
+        for (let i = 0, row; (row = this.table.rows[i]); i++) {
+            for (let j = 0, cell; (cell = row.cells[j]); j++) {
+                if (cell) {
```
JiuqingSong, line 444: cell will be undefined when j == 0. Is that expected?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -450,6 +437,66 @@ export default class VTable {
...
+                    if (!cell.innerHTML || !cell.innerHTML.trim()) {
+                        cell.appendChild(document.createElement('br'));
+                    }
+                }
+            }
+        }
+    }
+
+    public setTableCells() {
```
JiuqingSong, line 453: What does this function do? Add comments.

All public functions (especially those will public to customers) need comments.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -450,6 +437,66 @@ export default class VTable {
         }
         return result;
     }
+
+    private setEmptyTableCells() {
```
JiuqingSong, line 441: Normally a function named "set***" should take some parameter to set. But here it doesn't require parameter. So use a better name which can reflect its functionality.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -450,6 +437,66 @@ export default class VTable {
...
+        for (let i = 0; i < this.cells.length; i++) {
+            for (let j = 0; j < this.cells[i].length; j++) {
+                const cell = this.cells[i][j];
+                if (cell.td) {
+                    const rect = cell.td.getBoundingClientRect();
+                    cell.width = rect.width;
+                    cell.height = rect.height;
+                }
+            }
+        }
```
JiuqingSong, line 464: This part can be moved into constructor of VTable since it just read the size and do no change to the table. So that even normalizeSize is false, we can use the width/height value of VCell.

packages/roosterjs-editor-dom/lib/table/VTable.ts
JiuqingSong: Same here, use a better name. e.g. normalizeTableCellSize.

And I still think this function should be private. When you need to call it from outside, consider do the full normalization of the table instead.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -465,43 +475,48 @@ export default class TableResize implements EditorPlugin {
     }
 
     private startResizingTable = (e: MouseEvent) => {
-        if (this.currentTable == null) {
+        if (!this.currentTable) {
             return;
         }
+        this.resizingVtable = new VTable(this.currentTable, true);
```
JiuqingSong, line 481: Add inline comments for "true"

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -515,159 +530,133 @@ export default class TableResize implements EditorPlugin {
...
+                    if (cell.td) {
+                        if (shouldResizeX) {
+                            // the width of some external table is fixed, we need to make it resizeable
+                            this.resizingVtable.table.style.width = null;
+                            const newWidth = cell.width * ratioX;
+                            cell.td.style.boxSizing = 'border-box';
+                            if (newWidth >= MIN_CELL_WIDTH) {
+                                cell.td.style.wordBreak = 'break-word';
+                                cell.td.style.whiteSpace = 'normal';
+                                cell.td.style.width = `${newWidth}px`;
```
JiuqingSong, line 562: Shouldn't this be done when normalized the table?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: I didn't see issue here, just want to see how it is done in the code.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: The goal is to reduce duplicated code. In your code here, normalizeRect(nextTd.getBoundingClientRect()) appears twice.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: cell.td.style.width = `${newWidth}px` here is to set new width for each cell when we resize the whole table based on the resizing ratio, why should we do it when normalizing the table in the beginning?

packages/roosterjs-editor-dom/lib/table/VTable.ts
flyingbee2012: fixed the name, the full normalization (normalizeSize) will make the table width/height fixed whereas we don't want fixed width in this scenario, so using only setTableCells instead;

packages/roosterjs-editor-dom/lib/table/VTable.ts
flyingbee2012: It will do cell = row-cells[j] before evaluating the cell when j == 0, so it won't be undefined.

packages/roosterjs-editor-dom/lib/table/VTable.ts
flyingbee2012: I think we should do this logic only when we need to resize the table, we don't need to set/use width/height of VCell when we are not resizing the table.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts

packages/roosterjs-editor-types/lib/interface/VCell.ts
```diff
@@ -16,4 +16,7 @@ export default interface VCell {
      * Whether this cell is spanned from above
      */
     spanAbove?: boolean;
+
+    width?: number;
+    height?: number;
```
JiuqingSong, line 21: need comment for members of public interface

packages/roosterjs-editor-dom/lib/table/VTable.ts
JiuqingSong: For table resize, yes it is only used when we need to do resize. But as a public interface, other people may also want to use it in other cases. So we should make sure it always have correct value. Otherwise, need to comment it to explain when the value will be available.

RoosterJs is open source project, each line of public code is not only for ourselves, but also for people who may use it. Keep that in mind.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: ok

packages/roosterjs-editor-types/lib/interface/VCell.ts
flyingbee2012: added

packages/roosterjs-editor-dom/lib/table/VTable.ts
flyingbee2012: discussed offline and fixed

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -62,6 +62,11 @@ export default class VTable {
                     }
                 }
             });
+
+            this.measureAndStoreCellSize();
```
JiuqingSong, line 66: Minor: I believe you can merge this function into line 56-60.

packages/roosterjs-editor-types/lib/interface/VCell.ts
```diff
@@ -16,4 +16,7 @@ export default interface VCell {
      * Whether this cell is spanned from above
      */
     spanAbove?: boolean;
+
+    width?: number /* the width in pixel of the actual td (including border and padding) */;
```
JiuqingSong, line 20: Please put the comment before the properties so that typedoc can generate document for these attributes, like what we have for other properties .

JiuqingSong:
I can't resize the last column.

flyingbee2012:
> I can't resize the last column.

Fixed!

JiuqingSong (approved):

PR #566, opened 2021-06-16
Image Resize refactor

JiuqingSong:
This is a refactor of ImageResize plugin, to get prepared for next change of ImageCrop and ImageRotate.

In this change I added two new classes to help do resize:

- ImageResizeHandler: Manage the resize handles, handle image operations
- ResizeHelper: Handle mouse event

Later these classes can be reused by other image operations.

Also add a new property "isClick" to MouseUp event so that if this event also triggers a "click" event. We should not show image resize handles when a mouse up event not triggers a click event. This can happens when we select some text and image.


packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResize.ts
```diff
@@ -69,17 +53,22 @@ export default class ImageResize implements EditorPlugin {
      */
     initialize(editor: IEditor) {
         this.editor = editor;
-        this.disposer = editor.addDomEventHandler({
-            dragstart: this.onDragStart,
-            blur: this.onBlur,
-        });
+        this.disposer = editor.addDomEventHandler('blur', this.onBlur);
+        this.resizeHandler = new ImageResizeHandler(
```
flyingbee2012, line 57: can we put this into ctor?

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResize.ts
```diff
@@ -69,17 +53,22 @@ export default class ImageResize implements EditorPlugin {
...
+            this.forcePreserveRatio
+        );
     }
 
     /**
      * Dispose this plugin
      */
     dispose() {
-        this.hideResizeHandle();
+        this.resizeHandler.setCurrentImage(null);
```
flyingbee2012, line 70: better check null for this.resizeHandler?

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResizeHandler.ts
```diff
@@ -0,0 +1,266 @@
...
+            let newHeight = verticalOnly
+                ? this.initHeight
+                : Math.max(
+                      this.initHeight + deltaY * (y == YCoordinate.Top ? -1 : 1),
+                      this.minHeight
+                  );
+
+            if (shouldPreserveRatio) {
+                newHeight = Math.min(newHeight, (newWidth * this.initHeight) / this.initWidth);
+                newWidth = Math.min(newWidth, (newHeight * this.initWidth) / this.initHeight);
```
flyingbee2012, line 221: since this.ratio is calculated already, should this work:

newHeight = Math.min(newHeight, newWidth / this.ratio);
newWidth = Math.min(newWidth, newHeight * this.ratio);

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResizeHandler.ts
```diff
@@ -0,0 +1,266 @@
...
+                    } else {
+                        newHeight = newWidth / this.ratio;
+                    }
+                }
+            }
+
+            this.setSize(newWidth, newHeight);
+
+            // double check
+            if (shouldPreserveRatio) {
```
flyingbee2012, line 235: Interesting...I am wondering why we double check and set size again. Any chance to get it right for once?

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResizeHandler.ts
```diff
@@ -0,0 +1,266 @@
...
+        if (this.image) {
+            this.editor.addUndoSnapshot();
+            this.editor.triggerContentChangedEvent(ChangeSource.ImageResize, this.image);
+        }
+    };
+
+    private setSize(width: number, height: number) {
+        this.image.style.width = width + 'px';
+        this.image.style.height = height + 'px';
+        this.image.width = width;
```
flyingbee2012, line 263: I think style.width/height should be enough

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResize.ts
JiuqingSong: No, in ctor we don't know editor yet.

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResizeHandler.ts
JiuqingSong: See here: https://github.com/microsoft/roosterjs/pull/384/files

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResizeHandler.ts
JiuqingSong: Outlook Desktop doesn't respect style.width/height, which causes resized image can't be shown in new size in Outlook Desktop. So that is necessary.

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResizeHandler.ts
JiuqingSong: I think you are right

packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResizeHandler.ts
flyingbee2012: okay. good to know that!

flyingbee2012 (approved):

PR #568, opened 2021-06-16
A new way to do boundary check

flyingbee2012:
Fix an issue where resizing is not working properly for vertically merged cells. The fix is to resize current column and next column separately.


packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -601,7 +601,44 @@ export default class TableResize implements EditorPlugin {
         });
     };
 
+    private canResizeColums = (newPos: number): boolean => {
```
JiuqingSong, line 604: typo: columns

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -601,7 +601,44 @@ export default class TableResize implements EditorPlugin {
...
+        for (let i = 0; i < this.currentCellsToResize.length; i++) {
+            const td = this.currentCellsToResize[i];
+            const rect = normalizeRect(td.getBoundingClientRect());
+            if (this.isRTL && rect.right - newPos < MIN_CELL_WIDTH) {
```
JiuqingSong, line 608: We still see this pattern to get width for LTR and RTL, I still think we should extract it to be a function.
Same for line 621

JiuqingSong: and line 626, 630

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
flyingbee2012: good catch!

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
JiuqingSong: Was this resolved? I still see it in the code.

I expect something like:
if (getDistance(rect, newPos, true /*toLeft*/, this.isRTL) <MIN_CELL_WIDTH) ...

If you think 4 parameters are too many, we can split it into two function:
function getDistanceToLeft(rect, pos, isRtl) {}
and
function getDistanceToRight(rect, pos, isRtl {}

flyingbee2012: fixed now

JiuqingSong (approved):

PR #571, opened 2021-06-23
Add HtmlSanitizer preserveHtmlComments option

adrienWeiss:
Hello,
this is a proposal to add an option to keep HTML comments when using `sanitizeHtml`.
I don't think HTML comments are a risk, but if I'm wrong I'd love to know why.


JiuqingSong (approved):

adrienWeiss:
Thank you for the review 🙏

PR #574, opened 2021-07-02
Improve paste pure LI behavior

JiuqingSong:
We have a fix before for the case when copy from Teams, there may be LI nodes directly under DIV without UL/OL. However, it is also possible that LI nodes are directly under the fragment, and no DIV at all. This change also fix that case.


packages/roosterjs-editor-plugins/lib/plugins/Paste/commonConverter/convertPastedContentForLI.ts
```diff
@@ -0,0 +1,35 @@
...
+    } else if (isPureLiNode(fragment.firstChild)) {
+        changeElementTag(fragment.firstChild as HTMLElement, 'UL');
+    }
+}
+
+function isPureLiNode(node: Node) {
+    return (
+        node &&
+        !node.nextSibling &&
+        ['OL', 'UL', 'MENU'].indexOf(getTagOfNode(node)) < 0 &&
```
flyingbee2012, line 28: can we just use getTagOfNode(childNode) == 'LI' here?

JiuqingSong: what we want here is to check LI is not under OL/UL/MENU, so only check LI si not enough

flyingbee2012 (approved):

juliaroldi (approved):

PR #572, opened 2021-07-01
Add ImageEdit plugin

JiuqingSong:
1. Add ImageEdit plugin to support image resize/rotate/crop
2. Add related types and event: EditImageEvent
3. Update demo site to demo the new plugin

PR #575, opened 2021-07-02
Fix JS error if !ExperimentalFeatures.SingleDirectionResize

adrienWeiss:
Hello,

If `ExperimentalFeatures.SingleDirectionResize` is disabled, there is a JS error when disposing of the resize handlers here
https://github.com/microsoft/roosterjs/blob/a79ba502b6a53a8a625b538eed6428a3dbf9f3d3/packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResizeHandler.ts#L118-L121

because we are skipping an index in here (step = 2)

https://github.com/microsoft/roosterjs/blob/a79ba502b6a53a8a625b538eed6428a3dbf9f3d3/packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResizeHandler.ts#L149-L150

resulting in `this.resizeHelpers` having undefined on every odd index

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResizeHandler.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ImageResize/ImageResizeHandler.ts
@@ -116,6 +116,9 @@ export default class ImageResizeHandler {
         const wrapper = this.getResizeWrapper();
 
         for (let i = 0; i < this.resizeHelpers.length; i++) {
+            if (!this.resizeHelpers[i]) {
+                continue;
+            }
             for (let j = 0; j < this.resizeHelpers[i].length; j++) {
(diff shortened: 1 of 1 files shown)


JiuqingSong:
Thanks for find the bug. I believe this can also be fixed by https://github.com/microsoft/roosterjs/pull/572

adrienWeiss:
> Thanks for find the bug. I believe this can also be fixed by #572

if ImageEdit is supposed to replace ImageResize yes I guess. Do you plan on keeping ImageResize plugin though ?

JiuqingSong (approved):

JiuqingSong:
Given that the new image edit plugin may need some time to review, I'll merge this change first to unblock people.

PR #578, opened 2021-07-06
Hide inserter when moving mouse to another component

Lufedi:
**Description**

When we move the mouse out of the inserter and land on other components, we need to hide the inserter element.

**Bug description:** 
When moving the mouse to other components the inserter bar was not being removed from the table and this was causing issues when resizing the screen.

**Solution** 

This bug is an edge case when the table is touching a border of the editor.

My approach is to add add a "mouseout" event handler in the [inserterTd ](https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts#L288) element to avoid the bug when the "+" button is rendered passing the border of the editing zone 

I also modified the [current](https://github.com/microsoft/roosterjs/blob/cfe4f3515833480b66f4c0214f93bc337410bddb/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts#L146)  validation to hide the inserter button using the box approach

image

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
@@ -135,6 +135,10 @@ export default class TableResize implements EditorPlugin {
         this.tableResizerContainer = null;
     }
 
+    private onMouseOutInserter = () => {
+        this.setCurrentInsertTd(ResizeState.None);
+    };
+
     private onMouseMove = (e: MouseEvent) => {
         if (this.resizingState != ResizeState.None) {
             return;
@@ -146,8 +150,8 @@ export default class TableResize implements EditorPlugin {
 
         if (this.tableRectMap) {
             this.setCurrentTable(null);
-            let i = this.tableRectMap.length - 1;
-            while (i >= 0) {
+
+            for (let i = this.tableRectMap.length - 1; i >= 0; i--) {
                 const { table, rect } = this.tableRectMap[i];
 
                 if (
@@ -161,8 +165,6 @@ export default class TableResize implements EditorPlugin {
                     this.setCurrentTable(table);
                     break;
                 }
-
-                i--;
             }
 
             if (this.currentTable) {
@@ -181,7 +183,11 @@ export default class TableResize implements EditorPlugin {
                             e.pageY <= tdRect.bottom
                         ) {
                             // check vertical inserter
-                            if (i == 0 && e.pageY <= tdRect.top + INSERTER_HOVER_OFFSET) {
+                            if (
+                                i == 0 &&
+                                e.pageY >= tdRect.top - INSERTER_HOVER_OFFSET &&
+                                e.pageY <= tdRect.top + INSERTER_HOVER_OFFSET
+                            ) {
                                 let verticalInserterTd: HTMLTableCellElement = null;
                                 // set inserter at current td
                                 if (
@@ -286,6 +292,7 @@ export default class TableResize implements EditorPlugin {
             this.currentInsertTd = td;
             if (this.currentInsertTd) {
                 this.inserter = this.createInserter(tableRect);
+                this.inserter.addEventListener('mouseout', this.onMouseOutInserter);
                 this.resizerContainer.appendChild(this.inserter);
             }
         }
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
@@ -292,7 +292,6 @@ export default class TableResize implements EditorPlugin {
             this.currentInsertTd = td;
             if (this.currentInsertTd) {
                 this.inserter = this.createInserter(tableRect);
-                this.inserter.addEventListener('mouseout', this.onMouseOutInserter);
                 this.resizerContainer.appendChild(this.inserter);
             }
         }
@@ -348,6 +347,7 @@ export default class TableResize implements EditorPlugin {
         }
 
         inserter.addEventListener('click', this.insertTd);
+        inserter.addEventListener('mouseout', this.onMouseOutInserter);
 
         return inserter;
     }
--- a/packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
+++ b/packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
@@ -100,7 +100,7 @@ describe('TableResize plugin tests', () => {
     it('removes the vertical inserter when moving the cursor out of the offset zone with culture language RTL', () => {
         const rect = initialize(true);
         runTest(
-            { x: rect.left, y: rect.top },
+            { x: rect.left + insideTheOffset, y: rect.top },
             { x: (rect.right - rect.left) / 2, y: rect.bottom },
             false
         );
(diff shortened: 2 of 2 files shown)


JiuqingSong:
@flyingbee2012 please review and test this change.

packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
```diff
@@ -0,0 +1,78 @@
+import * as TestHelper from '../../../roosterjs-editor-api/test/TestHelper';
```
flyingbee2012, line 1: thanks for adding the test!

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -286,6 +292,7 @@ export default class TableResize implements EditorPlugin {
             this.currentInsertTd = td;
             if (this.currentInsertTd) {
                 this.inserter = this.createInserter(tableRect);
+                this.inserter.addEventListener('mouseout', this.onMouseOutInserter);
```
flyingbee2012, line 295: let's move this inside createInserter where we add the click event handler as well.

packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
```diff
@@ -0,0 +1,78 @@
...
+        editorDiv.dispatchEvent(mouseEvent);
+        expect(body.innerHTML.includes(ADD_BUTTON)).toBe(expectedResult);
+    }
+
+    it('removes the inserter when moving the cursor out of the offset zone', () => {
+        const rect = initialize();
+        runTest({ x: rect.right, y: rect.top }, { x: rect.right / 2, y: rect.bottom }, false);
+    });
+
+    it('keeps the inserter when moving the cursor inside the sage zone', () => {
```
flyingbee2012, line 70: can you add comment about whether this is testing the vertical or horizonal inserter?

packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        mouseEvent = new MouseEvent('mousemove', {
+            clientX: mouseEnd.x,
+            clientY: mouseEnd.y,
+        });
+        editorDiv.dispatchEvent(mouseEvent);
+        expect(body.innerHTML.includes(ADD_BUTTON)).toBe(expectedResult);
+    }
+
+    it('removes the inserter when moving the cursor out of the offset zone', () => {
```
flyingbee2012, line 65: same here :-)

packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
```diff
@@ -0,0 +1,96 @@
...
+    it('removes the horizontal inserter when moving the cursor out of the offset zone', () => {
+        const rect = initialize();
+        runTest(
+            { x: rect.left, y: rect.bottom },
+            { x: (rect.right - rect.left) / 2, y: (rect.bottom - rect.top) / 2 },
+            false
+        );
+    });
+
+    it('keeps the horizontal inserter when moving the cursor out of the offset zone', () => {
```
flyingbee2012, line 88: '....inside the safe zone'?

packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
```diff
@@ -0,0 +1,96 @@
...
+        const rect = initialize();
+        runTest(
+            { x: rect.left, y: rect.bottom },
+            { x: rect.left, y: rect.bottom - insideTheOffset },
```
flyingbee2012, line 92: If my understanding is correct, this applies to the scenario where the culture language is left to right? but if it is right to left, the test for { x: rect.left, y: rect.bottom - insideTheOffset } will fail, and { x: rect.right, y: rect.bottom - insideTheOffset } will succeed.

I think we can add one more check here in this test case, something like:

runTest(
            { x: rect.left, y: rect.bottom },
            { x: isRTL ? rect.right : rect.left, y: rect.bottom - insideTheOffset },

you can use below to get isRTL

isRTL = getComputedStyle(this.editor.getDocument().body, 'direction') == 'rtl';

Lufedi: It's correct this is only testing left to right scenarios, I will add more test to cover right to left culture language

packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
```diff
@@ -0,0 +1,135 @@
...
+        runTest(
+            { x: rect.right, y: rect.bottom },
+            { x: (rect.right - rect.left) / 2, y: (rect.bottom - rect.top) / 2 },
+            false
+        );
+    });
+
+    it('keeps the horitzontal inserter when moving the cursor inside the safe zone with culture language RTL', () => {
+        const rect = initialize(true);
+        runTest(
```
flyingbee2012, line 129: Since we don't show horizontal inserter for the very first line, rect.top will fail this test, using rect.bottom should fix it. 
 ```
runTest(
            { x: rect.right, y: rect.bottom },
            { x: rect.right + insideTheOffset, y: rect.bottom },
            true
        );
```

Comment (Lufedi) [reply packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts:129]: Good catch!, I was using the wrong coordinates here. I will fix this using the right+bottom corner to test the horizontal inserter in RTL mode.

File: packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts [hunk 8]
```diff
@@ -0,0 +1,135 @@
...
+            { x: rect.left, y: rect.top },
+            { x: (rect.right - rect.left) / 2, y: rect.bottom },
+            false
+        );
+    });
+
+    it('keeps the vertical inserter when moving the cursor inside the safe zone with culture language RTL', () => {
+        const rect = initialize(true);
+        runTest(
+            { x: rect.left, y: rect.top },
```
flyingbee2012, line 112: It seems that it won't be able to display the vertical inserter if we land the cursor exactly on the left border, the pixel might be too small to be detected in that scenario. 

notshown

let's not land the cursor exactly on the left border, something like below can fix it.

{ x: rect.left + insideTheOffset, y: rect.top }

packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
```diff
@@ -0,0 +1,135 @@
...
+            { x: rect.left, y: rect.bottom },
+            { x: rect.left, y: rect.bottom - insideTheOffset },
+            true
+        );
+    });
+
+    it('removes the vertical inserter when moving the cursor out of the offset zone with culture language RTL', () => {
+        const rect = initialize(true);
+        runTest(
+            { x: rect.left, y: rect.top },
```
flyingbee2012, line 103: It seems that it won't be able to display the vertical inserter if we land the cursor exactly on the left border, the pixel might be too small to be detected in that scenario. 

notshown

let's not land the cursor exactly on the left border, something like below can fix it.

{ x: rect.left + insideTheOffset, y: rect.top }

packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
```diff
@@ -0,0 +1,135 @@
...
+            { x: (rect.right - rect.left) / 2, y: rect.bottom },
+            false
+        );
+    });
+
+    it('keeps the vertical inserter when moving the cursor inside the safe zone with culture language RTL', () => {
+        const rect = initialize(true);
+        runTest(
+            { x: rect.left, y: rect.top },
+            { x: rect.left, y: rect.top + insideTheOffset },
```
flyingbee2012, line 113: same here

 runTest(
            { x: rect.left + insideTheOffset, y: rect.top },
            { x: rect.left + insideTheOffset, y: rect.top + insideTheOffset },
            true
        );

can fix it

flyingbee2012 (approved):

PR #579, opened 2021-07-07
8.5.0 Fix resize border

JiuqingSong:
This is a temp fix to show resize border when single direction image resize is not enabled.

(This fix will be overwritten by the new image edit plugin in #572 )


flyingbee2012 (approved):

JiuqingSong:
#1145

PR #584, opened 2021-07-14
Image Edit Step 2: Add Image Edit Info API

JiuqingSong:
(For a full code change, see here: https://github.com/microsoft/roosterjs/pull/572)

This is the 2nd step, it adds several API to operate on a new data structure "Image Edit Info" which stored the original image src and edited information of an image. We use this information to know how image is edited, so that we can generate new image using Canvas.


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/checkEditInfoState.ts
```diff
@@ -0,0 +1,78 @@
...
+        ROTATE_KEYS.every(key => areSameNumber(compareTo[key], 0)) &&
+        CROP_KEYS.every(key => areSameNumber(editInfo[key], compareTo[key]))
+    ) {
+        return ImageEditInfoState.SameWithLast;
+    } else {
+        return ImageEditInfoState.FullyChanged;
+    }
+}
+
+function isNumber(o: any): o is number {
```
flyingbee2012, line 72: interesting. shall we do `function isNumber(o: any): boolean {....}` ?

Lufedi (approved):

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/checkEditInfoState.ts
Lufedi: I thought the same but I found this [link](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
```diff
@@ -0,0 +1,81 @@
...
+        case ImageEditInfoState.ResizeOnly:
+            // For resize only case, no need to generate a new image, just reuse the original one
+            newSrc = editInfo.src;
+            break;
+        case ImageEditInfoState.SameWithLast:
+            // For SameWithLast case, image may be resized but the content is still the same with last one,
+            // so no need to create a new image, but just reuse last one
+            newSrc = previousSrc;
+            break;
+        case ImageEditInfoState.FullyChanged:
```
juliaroldi, line 40: What happens when the case is "Invalid"?

juliaroldi (approved):

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/generateDataURL.ts
```diff
@@ -0,0 +1,49 @@
...
+ * It is possible to throw exception since the original image may not be able to read its content from
+ * the code, so better check canRegenerateImage() of the image first.
+ */
+export default function generateDataURL(image: HTMLImageElement, editInfo: ImageEditInfo): string {
+    const {
+        angle,
+        width,
+        height,
+        bottom,
+        left,
```
flyingbee2012, line 20: left/right here actually indicates the distance as percentage from the left/right border of the image, correct? could we give them a different name as it would make people think that's the position of the left/right border but it is not. maybe something like distFromLeft/distFromRight. Same as top/bottom.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/getTargetSizeByPercentage.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @internal
+ * Get target size of an image with a percentage
+ * @param editInfo
+ * @param percentage
+ * @returns [width, height] array
+ */
+export default function getTargetSizeByPercentage(
+    editInfo: ImageEditInfo,
+    percentage: number
+): number[] {
```
flyingbee2012, line 13: can we return an object with key of width/height, as index 0/1 would not tell too much info about what it indicates.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/saveEditInfo.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * @internal
+ * Save edit info to image
+ * @param image The image to save edit info to
+ * @param editInfo The edit info to save
+ */
+export default function saveEditInfo(image: HTMLImageElement, editInfo: ImageEditInfo) {
+    if (image) {
+        image.dataset[IMAGE_EDIT_INFO_NAME] = JSON.stringify(editInfo);
```
flyingbee2012, line 11: could we possibly use object url here as editInfo could be extremely long data, maybe we can do that later?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
JiuqingSong: It means we got dirty data which may not be generated from our code. So we need to ignore it

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/checkEditInfoState.ts
JiuqingSong: exactly as Felipe said.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/generateDataURL.ts
JiuqingSong: Actually here I'm trying to save some bytes since those names won't be compressed.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/checkEditInfoState.ts
flyingbee2012: ah good to know that

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/saveEditInfo.ts
JiuqingSong: We have EditImage event which will give the flexibility for application to relace dataUrl to be ObjectUrl

PR #585, opened 2021-07-14
Image Edit Step 3: Add image edit public API

JiuqingSong:
(For a full code change, see here: https://github.com/microsoft/roosterjs/pull/572)

This is the 3rd step of image edit, it adds public API for image editing based on ImageEditInfo. In further steps we will use these API to perform image editing. Customer application can also call them to do image editing from their own code.


juliaroldi (approved):

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/generateDataURL.ts
```diff
@@ -0,0 +1,49 @@
...
+import ImageEditInfo from '../types/ImageEditInfo';
+
+/**
+ * @internal
+ * Generate new dataURL from an image and edit info
+ * @param image The image to generate data URL from. It is supposed to have original src loaded
+ * @param editInfo Edit info of the image
+ * @returns A BASE64 encoded string with image prefix that represents the content of the generated image.
+ * If there are rotate/crop/resize info in the edit info, the generated image will also reflect the result.
+ * It is possible to throw exception since the original image may not be able to read its content from
```
Lufedi, line 11: As this function may throw an exception, do you think it's good idea to include the [@throws](https://jsdoc.app/tags-throws.html) tag in the function documentation?

packages/roosterjs-editor-plugins/test/imageEdit/applyChangeTest.ts
```diff
@@ -0,0 +1,397 @@
...
+xdescribe('applyChange', () => {
+    let img: HTMLImageElement;
+    let editor: IEditor;
+    let triggerPluginEvent: jasmine.Spy;
+
+    beforeEach(async () => {
+        img = await loadImage(IMG_SRC);
+        document.body.appendChild(img);
+        triggerPluginEvent = jasmine.createSpy('triggerPluginEvent');
+        editor = <IEditor>(<any>{
```
Lufedi, line 21: Just for curiosity, is this another way of creating mocks?

Lufedi (approved):

flyingbee2012 (approved):

packages/roosterjs-editor-plugins/test/imageEdit/applyChangeTest.ts
JiuqingSong: Maybe we can have a utility later.

PR #590, opened 2021-07-15
bugfix(handle-delete-before-1st-item): create feature to handle delet…

juliaroldi:
Create a feature to handle delete key press before the first item of a list.


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -92,6 +92,22 @@ const OutdentWhenBackOn1stEmptyLine: BuildInEditFeature<PluginKeyboardEvent> = {
...
 };
 
+/**
+ * IndentWhenDeleteBefore1stItem edit feature, provides the ability to indent the list if user press
+ * DELETE before the first item of a list
+ */
+const IndentWhenDeleteBefore1stItem: BuildInEditFeature<PluginKeyboardEvent> = {
+    keys: [Keys.DELETE],
+    shouldHandleEvent: (event, editor) => {
+        let li = editor.getElementAtCursor('LI', null /*startFrom*/, event);
```
JiuqingSong, line 102: can we also check if this is at the end of current line, and next line is under LI? Otherwise each time we press DELETE this code will be executed, which may have impact to performance.

juliaroldi: Good point!! I'll fix it.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -92,6 +92,22 @@ const OutdentWhenBackOn1stEmptyLine: BuildInEditFeature<PluginKeyboardEvent> = {
     handleEvent: toggleListAndPreventDefault,
 };
 
+/**
+ * IndentWhenDeleteBefore1stItem edit feature, provides the ability to indent the list if user press
+ * DELETE before the first item of a list
+ */
+const IndentWhenDeleteBefore1stItem: BuildInEditFeature<PluginKeyboardEvent> = {
```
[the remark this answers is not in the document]
juliaroldi: Is MaintainListChainWhenDelete better?

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -199,6 +223,17 @@ function getListChains(editor: IEditor) {
     return VListChain.createListChains(editor.getSelectedRegions());
 }
 
+function getCacheNextSibiling(event: PluginKeyboardEvent, editor: IEditor) {
+    const element = cacheGetEventData(event, 'nextSibiling', () => {
+        const range = editor.getSelectionRange();
+        const pos = Position.getEnd(range).normalize();
+        const traverser = editor.getBodyTraverser(pos.node);
+        const blockElement = traverser.getNextBlockElement().collapseToSingleElement();
```
JiuqingSong, line 231: I suggest we don't call collapseToSingleElement here because that may change the dom structure.

Instead, you can just use getStartNode(), then find if it is LI node or contained by LI node. You can call findClosestElementAncestor(startNode, 'li') to check.

juliaroldi: I'll fix!

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -199,6 +225,16 @@ function getListChains(editor: IEditor) {
     return VListChain.createListChains(editor.getSelectedRegions());
 }
 
+function getCacheNextSibiling(event: PluginKeyboardEvent, editor: IEditor) {
+    const element = cacheGetEventData(event, 'nextSibiling', () => {
+        const range = editor.getSelectionRange();
+        const pos = Position.getEnd(range).normalize();
+        const traverser = editor.getBodyTraverser(pos.node);
```
JiuqingSong, line 232: Better do null check for traverser and traverser.getNextBlockElement(). In same cases they can return null.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -92,6 +94,30 @@ const OutdentWhenBackOn1stEmptyLine: BuildInEditFeature<PluginKeyboardEvent> = {
...
+ */
+const MaintainListChainWhenDelete: BuildInEditFeature<PluginKeyboardEvent> = {
+    keys: [Keys.DELETE],
+    shouldHandleEvent: (event, editor) => {
+        const li = editor.getElementAtCursor('LI', null /*startFrom*/, event);
+        if (li) {
+            return false;
+        }
+        const nextSibiling = getCacheNextSibiling(event, editor);
+        const nextSibilingIsLI = findClosestElementAncestor(nextSibiling, null, 'li');
```
JiuqingSong, line 109: Consider use editor.getElementAtCursor() instead, it wraps findClosestElementAncestor(), and limit the search scope within editor

JiuqingSong: If nextSibiling is null, no need to do further action since we know there isn't LI after current position

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -92,6 +94,30 @@ const OutdentWhenBackOn1stEmptyLine: BuildInEditFeature<PluginKeyboardEvent> = {
...
+        const isAtEndAndBeforeLI = nextSibilingIsLI
+            ? Position.getEnd(editor.getSelectionRange()).isAtEnd
+            : false;
```
JiuqingSong, line 112: I prefer do this check first since it is cheaper than the traverser thing.

So first, check is in LI, second, check if is at end, third, check if next block is LI.

We already do cheapest check first, and let it fail fast.

JiuqingSong (approved):

PR #595, opened 2021-07-19
Loosen strict callback type requirements in chainSanitizerCallback

Adjective-Object:
Most of the time in javascript, a caller is allowed to pass a callback that is looser than the destination type. For example, you are not always required to declare the index variable `someArray.map((x: T, i: number) => ...)` when mapping over an array.

However, when chainSanitizerCallback is called with a record of functions of form `(a:A,b:B,c:C) => D`, and a callback to chain of form `(a:A, b:B) => D`, it throws a type error. This is because the way the function signature is currently declared enforces the argument list be exactly the same as the members of the callback record it is passed. See [Current behaviour and the proposed solution in typescript playground](
https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbzgBQDYFcDmwB2BRANwFMcYAaFDbfY0gFQE8wiKBlGKXTAWQEMw4AXzgAzKBBBwA5OIgBnGESgArOQFoiAE2AxoamEyJypAKFCRYiOAGMAFr1yteOHcABeSgMK9UqAEa81gDWQqLikjIQ8ooq6lo6epoSUgDcJibWEDgKcADqABIAknR4APqsyACCnmWerKylAHKV3HhwALzSAO62OkRqcmCBRKnpAPRjcJ7oUFAk8JmaRHAAFEriUHJw6DhLUHAKnNbwIjvHwFlwBsxyAJQmpzjnl1loWLiE82sAXJTvNPNbogTHBQXBgCJVkQAHREWgwRjMDqdN7UT70QzQgBCRBE0CIyF4CiIQIQILBFLsDhwThcMHcXh8-kCQRW5IpHLgMLkzlcbi4AHkwPSstDrHI5OwGKgiN5fAFgnIKBM4KxbBB0KhNDgpPAYL0tn5lplSFBeAReJxnPAIMR9vrlnNMlBNHAQEQQEbNgB+dmcsEFYplCrVWr1JotPBkP3+1YWjAsOClIgy92kCj64CSgwyoHtAB8wNjsZV7GAvgORD1tmW8fQy10VwNUplBwgTd48B0NmcB3pFZgvCCy1xIiIx1WczAqGGByG1iMV3bADIcH5Bil7sX-Sqseh4HMYDMcKIfHIG+25iBbctu2IJHB8nRuAAZJdwc0QYCuzx0VgfzRtBwTAuQADw4XguXAAw4FQXAjBjbcmyzFsiAAbUDEpyiqGpSjqBpmlaABdDo4DrIg0iQjlD2PMifHrOAAEJOikMA5lGKjBEQ0FbkosEuK4kwVWQcRIHPV05AgDARRwdIlmsGc5lEM4ZJsexHF5ekPCgOVmWCAAmAAeOgwMUXYtmcBg0KIig6E8dScC0AAxE8iHAkhNC2FZoR8y1MDkX46DzQsACUKBC-M2QpEB+F+ELx2gTRDMOLgKG83yoH8wLgrgCLowpHBeHdX4UuA-KwUcrpdIVIJAvs6lnNk25fgIL9NDSITJmmWZ5hsCAlihWZoC2HY9j7I4ThUi4T2uIx7keZ4T1eKgPnhfSfj+NF4VJP0IShWF4URZZ2hRFaAQxZhsVxfFCWJHbYypDS6QZHSmRq9buI5blNPcQVhWmsUJVQ6qWSVOBS3VTVtV1ZDDWNLIIItK1SDgG97RrOAnUSt0PS9ORfSozDgxwsMCMjcrtxWciKGTVN5gzZscxJDpCzJKiKVLftUErata3oi9YdQtsOy7BZewUctucHYcuREMcJxWKcZwXOdhi2RtV3XMBN0+ktJj3A8q1okQzwFq8b3BE5wkfZ830bT9vymP8AKAkC3IgqDhQYWD4LkXX-UzbNpXQonsNDPDw0IvASM6ci+PZzGjagE9yKYli2JGePiy42NeL9ASTCAA) for reference. (it may take a bit for the roosterjs types to load into the editor).

This change allows for callbacks that are a subtype of the record's function type, which brings behaviour in line with the the javascript ecosystem.

This change also fixes a bug in dts.js where it would fail to match functions with `=>` in their type arguments


tools/dts.js
```diff
@@ -1,323 +1,323 @@
-'use strict';
```
Adjective-Object, line 1: looks like something made a bunch of whitespace changes here, will revert so the diff is readable.

tools/dts.js
```diff
@@ -10,7 +10,7 @@ var multiLineComment = /(^\/\*(\*(?!\/)|[^*])*\*\/\s*)/m;
 // 1. [export ][default |declare ](class|interface) <NAME>[ extends| implements <BASECLASS>] {...}
 var regClassInterface = /(\/\*(\*(?!\/)|[^*])*\*\/\s*)?(export\s+)?(default\s+|declare\s+)?(interface|class)\s+([a-zA-Z0-9_]+(\s*<[^>]+>)?)((\s+extends|\s+implements)(\s[0-9a-zA-Z_\.\s,]+(\s*<[^{]+>)?))?\s*{/g;
 // 2. [export ][default |declare ]function <NAME>(...)[: <TYPE>];
-var regFunction = /(\/\*(\*(?!\/)|[^*])*\*\/\s*)?(export\s+)?(default\s+|declare\s+)?function\s+([a-zA-Z0-9_]+(\s*<[^>]+>)?)\s*(\([^;]+;)/g;
+var regFunction = /(\/\*(\*(?!\/)|[^*])*\*\/\s*)?(export\s+)?(default\s+|declare\s+)?function\s+([a-zA-Z0-9_]+(\s*<(?:[^>]|=>)+>)?)\s*(\([^;]+;)/g;
```
Adjective-Object, line 13: I used a noncapturing group here `(?: ... )` because I didn't want to mess up the match group indexing below

JiuqingSong (approved):

PR #596, opened 2021-07-22
8.7.0

Lufedi:
### Improvements and bugfixes:
* Add Auto detect mode to clearFormat API
* Create a feature to handle delete key press before the first item of a list.
* Restore default style in a div after typed a semantic element
* Loosen strict callback type requirements in chainSanitizerCallback #595
* Replace fromHtml with createElement
* Support TrustedTypes with new trustedHTMLHandler property in IEditor
* Fix setHtmlWithselectionPath with DOMPurity
* Fixe the EntityId Regex

### Interface changes
* clearBlockFormat is decrepated, use clearFormat(ClearFormatMode.Block)  instead
* EditorOptions new field `trustedHTMLHandler`
* IEditor new method `getTrustedHTMLHandler`


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "roosterjs",
-    "version": "8.6.1",
+    "version": "8.7.1",
```
JiuqingSong, line 3: Please use 8.7.**0**

JiuqingSong (approved):

JiuqingSong:
#1145

PR #602, opened 2021-08-04
Support TrustedType step 4: Finally enable CSP in demo site

### LINKED ISSUE #581: Support Trusted Types
**Describe the bug**
RoosterJS is not usable in a webpage which has [Trusted Types](https://web.dev/trusted-types/) enforcement.
For example, there are many [_innerHTML_](https://github.com/microsoft/roosterjs/search?q=innerHTML+-filename%3Atest+-path%3Ademo) usage. Those `innerHTML` usage needs migration to safer practices (e.g. `textContent`, sanitization before assignment to `innerHTML`).

**To Reproduce**
1. Enforce Trusted Types on a page (i.e. `<meta http-equive="Content-Security-Policy" content="require-trusted-types-for 'script';">`).
2. Load RoosterJS on the page and observe that RoosterJS won't work.

**Expected behavior**
RoosterJS should be compatible with Trusted Types.

**Device Information**
 - OS: Windows 10
 - Browser: Chrome/Edge
 - Version: 91

JiuqingSong:
Fix #581 

Fix all places where innerHTML and DOMParser are used, and enable CSP in demo site.


Lufedi (approved):

demo/scripts/utils/trustedHTMLHandler.ts
```diff
@@ -0,0 +1,13 @@
+import * as DOMPurify from 'dompurify';
+
+const policy = !!window.trustedTypes
+    ? window.trustedTypes.createPolicy('sanitizeHtml', {
+          createHTML: (input: string) => DOMPurify.sanitize(input), //
+      })
```
shhnjk, line 6: This portion is unnecessary.
You can call `DOMPurify.sanitize(input, {RETURN_TRUSTED_TYPE: true})` and [DOMPurify would return TrustedHTML](https://github.com/cure53/DOMPurify#what-about-dompurify-and-trusted-types).
So, you don't have to create Trusted Type policy.

JiuqingSong: Good to know.

This is code just for demo site, not in production. It can also show how to create a trusted policy if user doesn't want to use DOMPurify.

shhnjk: Oh, okay :) Sounds good then :)

PR #605, opened 2021-08-06
Fix setHtmlWithSelectionPath with DOMPurify

JiuqingSong:
With DOMPurify, comment will be removed when sanitize html, so that selection path won't work since it is based on HTML commment. To fix it, let's add a fallback logic in setHtmlWithSelectionPath in case comment is removed.

Related test added.

### DIFF
--- a/demo/scripts/utils/trustedHTMLHandler.ts
+++ b/demo/scripts/utils/trustedHTMLHandler.ts
@@ -2,9 +2,7 @@ import * as DOMPurify from 'dompurify';
 
 const policy = !!window.trustedTypes
     ? window.trustedTypes.createPolicy('sanitizeHtml', {
-          createHTML: (input: string) => {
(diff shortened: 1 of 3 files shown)


demo/scripts/utils/trustedHTMLHandler.ts
```diff
@@ -2,7 +2,9 @@ import * as DOMPurify from 'dompurify';
 
 const policy = !!window.trustedTypes
     ? window.trustedTypes.createPolicy('sanitizeHtml', {
-          createHTML: (input: string) => DOMPurify.sanitize(input), //
+          createHTML: (input: string) => {
+              return DOMPurify.sanitize(input);
+          }, //
```
flyingbee2012, line 7: remove // ?

flyingbee2012 (approved):

PR #607, opened 2021-08-12
Bugfix/query command state

juliaroldi:
New API getPendingFormatState to replace the use of queryCommandState in execCommand() API and getFormatState API.


packages/roosterjs-editor-core/lib/coreApi/getPendingFormatState.ts
```diff
@@ -0,0 +1,71 @@
...
+                break;
+            }
+            currentElement = currentElement.parentElement;
+        }
+    }
+    return isFormatted;
+}
+
+function getPendableFormatState(): PendableFormatState {
+    let keys = Object.keys(PendableFormatCommandMap) as PendableFormatNames[];
```
JiuqingSong, line 66: I can understand you are doing a loop like:
```
  for each style {
    for each node { 
      ...
   }
}
```

But I recommand to reverse the loop order, let's loop dom node first, then styles:
```
for each node {
   for each style { 
      ...
  }
}
```

packages/roosterjs-editor-core/lib/coreApi/getPendingFormatState.ts
```diff
@@ -0,0 +1,71 @@
...
+    return keys.reduce((state, key) => {
+        state[key] = queryCommandState(PendableFormatCommandMap[key]);
```
JiuqingSong, line 68: And instead of passing in PendableFormatCommandMap[key], I suggest just pass in the key, so that the receiving parameter type can just use "PendableFormatNames" to reduce the chance of mistake.

packages/roosterjs-editor-core/lib/coreApi/getPendingFormatState.ts
```diff
@@ -0,0 +1,71 @@
...
+    const isSamePosition = range && currentPosition.equalTo(cachedPosition);
+    if (range && cachedPendableFormatState && range.collapsed && isSamePosition) {
+        return cachedPendableFormatState;
+    } else {
+        return getPendableFormatState();
+    }
+}
+
+function isAPendableFormat(element: HTMLElement, format: string): boolean {
+    switch (format) {
```
JiuqingSong, line 27: For these checking function, how about make them as a function map:
```
const StyleCheckers: Record<PendableFormatNames, (tagName: string, style: CSSStyleDeclaration) => boolean> {
 isBold: (tag, style) => tag == 'B' || tag == 'STRONG' || style.fontWeight >= 700,
 ...
};
```

Then you just need a very simple loop to check all the styles:
```
  while (contains(core.contentDiv, node)) {  // We just need to check the node under editor, no need to loop to root node of the page
     const tag = getTagOfNode(node);
     const style = node.nodeType == NodeType.Element && (<Element>node).style;

     if (tag & style) { // Only need to check for Element
        Object.keys(StyleCheckers).forEach((style: PendableFormatNames) => {
           formatState[style] = formatState[style] || StyleCheckers[style](tag, style);
        });
    }
    node = node.parentNode; // Let's use parentNode instead of parentElement since parentElement may have some issue in IE
  }
```

packages/roosterjs-editor-api/lib/utils/execCommand.ts
```diff
@@ -25,11 +26,10 @@ export default function execCommand(editor: IEditor, command: DocumentCommand) {
     if (range && range.collapsed) {
         editor.addUndoSnapshot();
         formatter();
-
         if (isPendableFormatCommand(command)) {
             // Trigger PendingFormatStateChanged event since we changed pending format state
             editor.triggerPluginEvent(PluginEventType.PendingFormatStateChanged, {
-                formatState: getPendableFormatState(editor.getDocument()),
+                formatState: setPendableFormatState(editor.getPendingFormatState(), command),
```
JiuqingSong, line 32: Here you need to get the format state before running formatter(), then set it to opposite value. Otherwise we rely on an assumption that formatter() will not change DOM structure. Although it is true, but we should not rely on it.

```
const formatState = editor.getPendingFormatState();
formatter();

const formatName = Object.keys(PendableFormatCommandMap).filter(
        (x: PendableFormatNames) => PendableFormatCommandMap[x] == command
    )[0];

if (formatName ) {
   fomatState[formatName ] = !formatState[formatName ];
    editor.triggerPluginEvent(...)
}
```

packages/roosterjs-editor-types/lib/interface/EditorCore.ts
```diff
@@ -264,6 +277,13 @@ export interface CoreApiMap {
      */
     getStyleBasedFormatState: GetStyleBasedFormatState;
 
+    /**
+     * Get style based format state from current selection, including font name/size and colors
+     * @param core The EditorCore objects
+     * @param node The node to get style from
+     */
+    getPendingFormatState: GetPendingFormatState;
```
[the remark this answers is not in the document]
juliaroldi: Should we remove the existing getPendableFormatState? I just tested getPendingFormatState for the execCommand scenario. The getPendableFormatState function is used in createPasteFragment and select Range and I didn't test getPendingFormarState in this scenarios.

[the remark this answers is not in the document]
JiuqingSong: You bring up a very good question.

createPasteFragment calls getPendableFormatState() to determine what format to merge. So it should always read format from DOM tree, even if there is cached pending format. So that we need to add a parameter to the new API to allow force getting stats from DOM tree, then let createPasteFragment use it.

After that, the original getPengingFormatState function will not be used any more, so we can mark it as deprecated, but keep it for now. We will remove it in next major release.

packages/roosterjs-editor-api/lib/utils/execCommand.ts
```diff
@@ -45,3 +45,19 @@ function isPendableFormatCommand(command: DocumentCommand): boolean {
...
+        state[key] =
+            pendableFormatKeyToCommand(key) === command
+                ? !pendableFormats[key]
+                : pendableFormats[key];
+        return state;
+    }, <PendableFormatState>{});
+}
+
+function pendableFormatKeyToCommand(key: string) {
+    let newKey = key.replace('is', '');
```
JiuqingSong, line 61: Do not do this string operation. You can use PendableFormatCommandMap to convert between command and style name.

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -0,0 +1,62 @@
...
+    } else {
+        return queryCommandState(core, currentPosition);
+    }
+}
+
+const StyleCheckers: Record<
+    PendableFormatNames,
+    (tagName: string, style: CSSStyleDeclaration) => boolean
+> = {
+    isBold: (tag, style) => tag == 'B' || tag == 'STRONG' || style.fontWeight >= '700',
```
JiuqingSong, line 37: If I specify font-weight to "bold", what will it return here?

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -0,0 +1,62 @@
...
+
+function queryCommandState(core: EditorCore, currentPosition: NodePosition): PendableFormatState {
+    let node = currentPosition && currentPosition.node;
+    let formatState: PendableFormatState = {};
+    if (node) {
+        while (contains(core.contentDiv, node)) {
+            const tag = getTagOfNode(node);
+            const style = node.nodeType == NodeType.Element && (node as HTMLElement).style;
+            if (tag && style) {
+                Object.keys(StyleCheckers).forEach((key: PendableFormatNames) => {
```
JiuqingSong, line 54: I think we still need to consider another case:
<span style="font-weight: bold"><span style="font-weight: normal"> test...

With current code it will return true since there is font weight to bold outside, but actually the inner style will override the outer style.

juliaroldi: I tested it and it is really happening. I'll try to fix this.

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
juliaroldi: It is considering  font-weight normal as bold

packages/roosterjs-editor-types/lib/interface/IEditor.ts
```diff
@@ -499,6 +499,13 @@ export default interface IEditor {
      */
     getStyleBasedFormatState(node?: Node): StyleBasedFormatState;
 
+    /**
+     * Get the pendable format such as underline and bold
+     * @param forceGetStateFromDOM If set to true, will force get the format state from DOM tree.
```
JiuqingSong, line 504: Also explain what is the default behavior.

JiuqingSong: We don't need to mention the cache thing in the comment for this whole function, but we can explain how it works for this parameter, so explain cache here is ok

packages/roosterjs-editor-dom/lib/utils/getPendableFormatState.ts
```diff
@@ -47,7 +47,6 @@ export const PendableFormatCommandMap: { [key in PendableFormatNames]: DocumentC
  */
 export default function getPendableFormatState(document: Document): PendableFormatState {
     let keys = Object.keys(PendableFormatCommandMap) as PendableFormatNames[];
-
```
JiuqingSong, line 50: No need to remove this empty line

packages/roosterjs-editor-types/lib/interface/IEditor.ts
```diff
@@ -499,6 +499,13 @@ export default interface IEditor {
...
+     * @return The pending format state of editor.
+     */
+    getPendableFormatState(forceGetStateFromDOM: boolean): PendableFormatState;
```
JiuqingSong, line 507: forceGetStateFromDOM should still be optional for IEditor and Editor. We just need to make it required for core API.

The principle here is, for common public API such as Editor API, format API, we should make it easy to use. For core API which is more "internally used", we want to make it strict and explicit.

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * @param core The EditorCore object
+ * @param forceGetStateFromDOM If set to true, will force get the format state from DOM tree.
+ * @returns The cached format state if it exists. If the cached postion do not exist, search for pendable elements in the DOM tree and return the pendable format state.
+ */
+export function getPendableFormatState(
+    core: EditorCore,
+    forceGetStateFromDOM: boolean = false
+): PendableFormatState {
+    const range = core.api.getSelectionRange(core, true /** tryGetFromCache*/);
```
JiuqingSong, line 13: Use "/*", not "/**" here

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -0,0 +1,84 @@
...
+
+const PendableStyleCheckers: Record<
+    PendableFormatNames,
+    (tagName: string, style: CSSStyleDeclaration) => boolean
+> = {
+    isBold: (tag, style) =>
+        tag == 'B' ||
+        tag == 'STRONG' ||
+        parseInt(style.fontWeight) >= 700 ||
+        ['bold', 'bolder'].includes(style.fontWeight),
```
JiuqingSong, line 35: Array.includes() is not supported by IE. Use Array.indexOf(...)>=0 instead.

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -0,0 +1,84 @@
...
+    PendableFormatNames,
+    (style: CSSStyleDeclaration) => boolean
+> = {
+    isBold: style =>
+        (style.fontWeight !== '' && parseInt(style.fontWeight) < 700) ||
+        style.fontWeight === 'normal',
+    isUnderline: style =>
+        style.textDecoration !== '' && style.textDecoration.indexOf('underline') < 0,
+    isItalic: style => style.fontStyle !== '' && style.fontStyle !== 'italic',
+    isSubscript: style => style.verticalAlign !== '' && style.verticalAlign === 'sub',
```
JiuqingSong, line 54: !== 'sub'

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -0,0 +1,84 @@
...
+): PendableFormatState {
+    let node = isCollapsed ? currentPosition?.node : document.getSelection().focusNode;
+    let formatState: PendableFormatState = {};
+    let pendablekeys: PendableFormatNames[] = [];
+    while (contains(core.contentDiv, node)) {
+        const tag = getTagOfNode(node);
+        const style = node.nodeType == NodeType.Element && (node as HTMLElement).style;
+        if (tag && style) {
+            Object.keys(PendableStyleCheckers).forEach((key: PendableFormatNames) => {
+                if (!pendablekeys.includes(key)) {
```
JiuqingSong, line 73: same here

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -0,0 +1,84 @@
...
+        ['bold', 'bolder'].includes(style.fontWeight),
+    isUnderline: (tag, style) => tag == 'U' || style.textDecoration.indexOf('underline') >= 0,
+    isItalic: (tag, style) => tag == 'I' || tag == 'EM' || style.fontStyle === 'italic',
+    isSubscript: (tag, style) => tag == 'SUB' || style.verticalAlign === 'sub',
+    isSuperscript: (tag, style) => tag == 'SUP' || style.verticalAlign === 'super',
+    isStrikeThrough: (tag, style) =>
+        tag == 'S' || tag == 'STRIKE' || style.textDecoration.indexOf('line-through') >= 0,
+};
+
+const NonPendableStyleCheckers: Record<
```
JiuqingSong, line 44: We need a better name. How about "CssFalsyCheckers", and add some comment to explain why we need this

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -0,0 +1,84 @@
+import { contains, getTagOfNode, PendableFormatNames, Position } from 'roosterjs-editor-dom';
+import { EditorCore, NodePosition, NodeType, PendableFormatState } from 'roosterjs-editor-types';
+
+/**
+ * @param core The EditorCore object
+ * @param forceGetStateFromDOM If set to true, will force get the format state from DOM tree.
+ * @returns The cached format state if it exists. If the cached postion do not exist, search for pendable elements in the DOM tree and return the pendable format state.
+ */
+export function getPendableFormatState(
```
JiuqingSong, line 9: Please add test case for this function

JiuqingSong (approved):

PR #608, opened 2021-08-13
feat(align-table): new feature to align table

juliaroldi:
New feature to align the current table left, center, right.


packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -286,6 +286,16 @@ export default class VTable {
                     });
                 }
                 break;
+
+            case TableOperation.AlignCenter:
```
JiuqingSong, line 290: Please test the feature in OWA:
1. New compose
2. Insert a table, center the table
3. Send it out
4. When receive, see if the table is rendered correctly, read in both OWA and Outlook Desktop
If everything works well, we can go with this.

juliaroldi: Table Aligment

JiuqingSong: This looks nice. Please also try send the email out after align table to right, and check the receiving result, see if it still looks correctly.

juliaroldi: I tested in OWA and Outlook Desktop. I didn't find any bug in OWA, but for Outlook desktop, if I send a email with more than one table the second table loses the formatting 

image

juliaroldi: Even if it is not aligned 
image

juliaroldi: It works fine, if the cells are filled.
image

JiuqingSong (approved):

PR #610, opened 2021-08-16
Enable Check Dependency feature when build

JiuqingSong:
There are some irregular import statements in our code which is hard to find. We already have a function our build script to check dependency issues. Let's improve it and enable it when build, so that all those kind of dependency issues can be found.


packages/roosterjs-editor-dom/lib/utils/adjustInsertPosition.ts
```diff
@@ -1,32 +1,30 @@
...
+import VTable from '../table/VTable';
+import wrap from './wrap';
+import { NodePosition, NodeType, PositionType, QueryScope } from 'roosterjs-editor-types';
+import { splitBalancedNodeRange } from './splitParentNode';
 
 const adjustSteps: ((
     root: HTMLElement,
     nodeToInsert: Node,
-    position: Position,
+    position: NodePosition,
```
juliaroldi, line 25: Why use NodePosition instead of Position? I thought NodePosition was only for new instances

JiuqingSong: In most cases, if we can use the interface instead of the class, we should use interface. NodePosition is the interface of Position, although the name is not good, this is for history reason. In next major release we can consider rename them.

juliaroldi (approved):

PR #611, opened 2021-08-16
U/bvalverde/reset list number

BryanValverdeU:
Editor Improvement reset list number


JiuqingSong:
Reset number doesn't work if point to middle of a list. Expect the list to be splitted and the second one start from 1.
resetlist

demo/scripts/controls/contextMenu/ResetListPlugin.ts
```diff
@@ -22,17 +23,35 @@ export default class ResetListPlugin implements ContextMenuProvider<ContextMenuI
...
-                        }, ChangeSource.Format);
+            if (safeInstanceOf(list, 'HTMLOListElement')) {
+                items.push(
+                    {
+                        key: 'resetList',
+                        name: 'Reset list number',
+                        onClick: () => {
+                            this.editor?.addUndoSnapshot(() => {
+                                this.editor.focus();
+                                setOrderedListNumbering(this.editor, list, 1);
```
JiuqingSong, line 34: Since we are adding a new format API, this API should handle everything such as focus to editor, and add undo snapshot.

demo/scripts/controls/contextMenu/ResetListPlugin.ts
```diff
@@ -22,17 +23,35 @@ export default class ResetListPlugin implements ContextMenuProvider<ContextMenuI
...
+                            }, ChangeSource.Format);
+                        },
                     },
-                });
+                    {
+                        key: 'setNumberingValue',
+                        name: 'Set Numbering Value',
+                        onClick: () => {
+                            this.editor?.addUndoSnapshot(() => {
```
JiuqingSong, line 42: same here

demo/scripts/controls/contextMenu/ResetListPlugin.ts
```diff
@@ -22,17 +23,35 @@ export default class ResetListPlugin implements ContextMenuProvider<ContextMenuI
...
+                                this.editor.focus();
+
+                                try {
+                                    let value = parseInt(prompt('Set Value to...', '1'));
+                                    setOrderedListNumbering(this.editor, list, value);
+                                } catch (error) {
```
JiuqingSong, line 48: If we hit any error, we can just do nothing

packages/roosterjs-editor-api/lib/format/setOrderedListNumbering.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @param editor The editor instance
+ * @param list The List that is going to be split
+ * @param startNumber The number of that the splitted list should start
+ */
+export default function setOrderedListNumbering(
+    editor: IEditor,
+    list: HTMLOListElement,
+    startNumber?: number
+) {
+    splitList(editor, list, startNumber);
```
JiuqingSong, line 15: Unless this splitList() function is reused by other functions, we don't need to have a separate util file. Just put your code here is ok.

packages/roosterjs-editor-api/lib/utils/splitList.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * @internal
+ */
+export default function splitList(
+    editor: IEditor,
+    list: HTMLOListElement,
+    startNumber?: number
+): void;
+
+export default function splitList(editor: IEditor, list: HTMLOListElement, startNumber?: number) {
```
JiuqingSong, line 14: this parameter "list" is not used at all.

JiuqingSong: You can declare startNumber default to be 1:

function splitlist(..., startNumber: number = 1) {

packages/roosterjs-editor-api/lib/utils/splitList.ts
```diff
@@ -0,0 +1,22 @@
...
+    blockFormat(editor, (region, start) => {
```
JiuqingSong, line 15: blockFormat relies on current selection in editor, and it will handle complex selection case such as select across a table. Here we already know which list to operate on, so no need to call blockFormat, and we should use the list or li parameter directly.

JiuqingSong: And actually I think we need to pass in the LI node instead list, then we can restart the list number from the passed in li node, which means if we have a list "1 2 3 4", and I point to "3" and reset the number, it should become "1 2 1 2".

packages/roosterjs-editor-api/lib/utils/splitList.ts
```diff
@@ -0,0 +1,22 @@
...
+        const vList = createVListFromRegion(region, false /*includeSiblingLists*/);
+
+        if (vList) {
+            vList.splitWriteBack(start.element, startNumber ?? 1);
```
JiuqingSong, line 19: try reuse the existing writeBack function(), but not add a new writeBack function.

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -210,6 +212,47 @@ export default class VList {
...
     }
 
+    /**
+     * Splits the VList in two, using the element provided in the parameter as separator
+     * Write the result back into DOM tree
+     * After that, this VList becomes unavailable because we set this.rootList to null
+     * @param separator The HTML element that indicates when to split the VList
+     * @param startNumber The number that the new list is going to start
+     */
+    splitWriteBack(separator: HTMLElement, startNumber: number) {
```
JiuqingSong, line 222: I wish the API should be something like:
```ts
const vlist = new VList(...);
vlist.setStartNumber(li, num);
vlist.writeBack();
```
So that it can match the pattern of existing VList functions.

demo/scripts/controls/contextMenu/ResetListPlugin.ts
```diff
@@ -22,17 +23,35 @@ export default class ResetListPlugin implements ContextMenuProvider<ContextMenuI
...
+                            this.editor?.addUndoSnapshot(() => {
+                                this.editor.focus();
+                                setOrderedListNumbering(this.editor, list, 1);
+                            }, ChangeSource.Format);
+                        },
                     },
-                });
+                    {
+                        key: 'setNumberingValue',
+                        name: 'Set Numbering Value',
```
Lufedi, line 40: Just for curiosity, is this going to be localized in the future for another languages?

JiuqingSong: no. This is for demo site only. Demo site is not production code, and will not be localized.

demo/scripts/controls/contextMenu/ResetListPlugin.ts
BryanValverdeU: Added the editor.focus and the addUndoSnapshot to the function in this commit.
https://github.com/microsoft/roosterjs/commit/d48eaee2a387d9f40c59a8e3e5b378d5a9941952

demo/scripts/controls/contextMenu/ResetListPlugin.ts
BryanValverdeU: Added the editor.focus and the addUndoSnapshot to the function in this commit.
d48eaee

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -19,6 +19,8 @@ import {
     NodeType,
 } from 'roosterjs-editor-types';
 
+// const CHAIN_DATASET_NAME = 'listchain';
```
JiuqingSong, line 22: remove this comment

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -210,6 +212,29 @@ export default class VList {
...
+        if (!this.rootList) {
+            throw new Error('rootList must not be null');
+        }
+
+        //Traverse the items of the VList, when the separator is found, add a dummy element
+        for (let index = 0; index < this.items.length; index++) {
+            if (this.items[index].getNode() == separator) {
+                let dummyElement = this.items[index].getNode().cloneNode(false);
+
+                separator.parentElement.append(dummyElement);
```
JiuqingSong, line 229: Let's use parentNode instead of parentElement since it doesn't work well in IE

JiuqingSong: And use appendChild instead of append due to IE.

JiuqingSong: Actually, no need to appendChild, just do insertBefore() directly is ok

JiuqingSong: After debug with your code, I think you do not need to clone and append the list node. Instead, you can add some property in VListItem to specify the list number to reset, by default its value is undefined, which means no need to reset. But when its number is different than the start number we get from this list (line 180), we break current list and start a new one.

so that in this function, we just need to find the existing VListItem object from the element you passed in, and set its number (if it is different than the current one). Then let writeBack() handle everything else.

packages/roosterjs-editor-api/lib/format/setOrderedListNumbering.ts
```diff
@@ -0,0 +1,35 @@
...
+): void;
+export default function setOrderedListNumbering(
+    editor: IEditor,
+    separator: HTMLLIElement,
+    startNumber: number = 1
+) {
+    editor.addUndoSnapshot(() => {
+        editor.focus();
+        const regions = editor.getSelectedRegions();
+        regions.forEach(region => {
```
JiuqingSong, line 23: You should only find one region. So no need to do a loop

JiuqingSong: Just check if regions[0] is defined.

packages/roosterjs-editor-api/lib/format/setOrderedListNumbering.ts
```diff
@@ -0,0 +1,35 @@
...
+            const vList = createVListFromRegion(region, true, separator /*includeSiblingLists*/);
+            if (vList) {
+                if (separator != vList.items[0].getNode()) {
```
JiuqingSong, line 26: Even if separater == vlist.item[0], we can still do this operation. Because the first list item number can be greater than 1, but we can set it to 1, or any other number.

packages/roosterjs-editor-api/lib/format/setOrderedListNumbering.ts
```diff
@@ -0,0 +1,35 @@
...
+                    vList.split(separator);
+                    vList.writeBack();
+                }
+
+                separator.parentElement.setAttribute('start', startNumber.toString());
```
JiuqingSong, line 31: As I commented in VList.ts, we should do this when writeBack.

packages/roosterjs-editor-dom/lib/list/VList.ts
BryanValverdeU: I added the new property to the VListItem, I called it "newListStart", with its getter and setter.

What I am having a little issue with is changing the WriteBack function:
image

In the first change I remove the items when the item.getNewListStart() returns something (I am assiging this value on the "split method"), then just change the parent List's start property. I am not sure if there is any feedback on this change of the WriteBack Function

packages/roosterjs-editor-api/lib/format/setOrderedListNumbering.ts
```diff
@@ -0,0 +1,35 @@
+import { ChangeSource, IEditor } from 'roosterjs-editor-types';
+import { createVListFromRegion } from 'roosterjs-editor-dom';
+
+/**
+ * Resets Ordered List Numbering back to the value of the parameter startNumber
+ * @param editor The editor instance
+ * @param separator The HTML element that indicates when to split the VList
+ * @param startNumber The number of that the splitted list should start
+ */
+export default function setOrderedListNumbering(
```
JiuqingSong, line 10: no need to have this extra function declaration

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -210,6 +218,25 @@ export default class VList {
...
+    split(separator: HTMLElement, startNumber: number) {
+        if (!this.rootList) {
+            throw new Error('rootList must not be null');
+        }
+
+        //Traverse the items of the VList, when the separator is found, set the New List Start Property
+        this.items.forEach(element => {
+            if (element.getNode() == separator) {
+                element.setNewListStart(startNumber);
+                return;
```
JiuqingSong, line 235: this return statement won't work as expected because you are in a forEach loop.
To make it work, you need a regular for-loop

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -182,10 +182,18 @@ export default class VList {
         this.rootList.parentNode.replaceChild(placeholder, this.rootList);
 
         this.items.forEach(item => {
+            if (item.getNewListStart()) {
```
JiuqingSong, line 185: please also check if this number is different than exist "start" number. And we only need to split when they are different.

JiuqingSong: Then set the "start" number to be this value, so that existing code below can use it

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -182,10 +182,18 @@ export default class VList {
...
+                listStack.splice(1, listStack.length - 1);
+            }
             item.writeBack(listStack, this.rootList);
             const topList = listStack[1];
 
-            if (safeInstanceOf(topList, 'HTMLOListElement')) {
+            if (item.getNewListStart()) {
```
JiuqingSong, line 191: See my comment above. Once you set the start number above, no need to add extra code here, existing code should be able to handle it.

packages/roosterjs-editor-api/test/format/setOrderedListNumberingTest.ts
```diff
@@ -0,0 +1,81 @@
+import * as TestHelper from '../TestHelper';
+import { IEditor, PositionType } from 'roosterjs-editor-types';
+import { safeInstanceOf } from 'roosterjs-editor-dom/lib';
+import { setOrderedListNumbering } from 'roosterjs-editor-api';
+
+describe('setOrderedListNumberingTest()', () => {
```
JiuqingSong, line 6: Suggest move most of the test code to test the split and writeBack function for VList (in VListTest), and for this test, we just need to make sure we call VList with correct parameters.

packages/roosterjs-editor-api/lib/format/setOrderedListNumbering.ts
```diff
@@ -0,0 +1,35 @@
...
+    separator: HTMLLIElement,
+    startNumber: number = 1
+) {
+    editor.addUndoSnapshot(() => {
+        editor.focus();
+        const regions = editor.getSelectedRegions();
+        if (regions[0]) {
+            const vList = createVListFromRegion(
+                regions[0],
+                true /*includeSiblingLists*/,
```
JiuqingSong, line 26: I feel we can pass false here. 
includeSiblingLists is used when we want to merge the list with existing items before/after this list. But for this function we don't need to do that.

JiuqingSong:
Generally looks good. Found a bug here: https://github.com/microsoft/roosterjs/issues/614 But it can also repro without your change. So I assigned the bug to you so you can fix it separately.

JiuqingSong:
Found another bug:

Use the HTML code below as snapshot and restore into editor:

```html
<div><ol><li><span>line 1</span></li><li><span>line 2</span></li></ol><div><span>test</span></div></div><ol start="3"><li><span></span>line 3</li></ol>
```

Then right click "2." and choose "reset list number".

Expect: list numbers becomes "1, 1, 2"
Actual: list number becomes "1, 1, 3"

It seems there isn't a very easy fix yet. The VListChain class can't handle this case now. So we can fix it separately.

JiuqingSong (approved):

PR #613, opened 2021-08-18
Paste table from PowerPoint fix

BryanValverdeU:
Fix Table being pasted as image under some scenarios, I think this change also is going to fix when pasting gifs to the editor

TableImg


packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
```diff
@@ -52,7 +52,7 @@ export const createPasteFragment: CreatePasteFragment = (
     let doc: HTMLDocument;
 
     // Step 2: Fill the BeforePasteEvent object, especially the fragment for paste
-    if (!pasteAsText && !text && imageDataUri) {
+    if (!pasteAsText && !text && imageDataUri && !rawHtml) {
```
JiuqingSong, line 55: I'm concerned that this may bring some other issue since it is possible that we still want to paste image when there is HTML which just has the IMG tag to the image file.

Actually we have code to handle pasting from PowerPoint here: https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs-editor-plugins/lib/plugins/Paste/pptConverter/convertPastedContentFromPowerPoint.ts

However, in this case it can't work because we have a check to only do this for PowerPoint: https://github.com/microsoft/roosterjs/blob/fc2dcd6f900a1ca93915da73bf189771920b1d58/packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts#L78

And because some code bug in createPasteFragment, we didn't retrieve the correct attribute for PPT, so that the function to handle PPT paste is not executed. You need to make the code https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts#L67-L75  always executed when possible, so that the related attribute (as well as meta tag) can be parsed correctly.

BryanValverdeU: I was debugging and it seems like the rawHtml when performing the paste, do have meta Tags but are not equal to "POWERPOINT_ATTRIBUTE_VALUE", this is the rawHtml from the clipboardData,:

`
<html xmlns:m=\"http://schemas.microsoft.com/office/2004/12/omml\"\r\nxmlns=\"http://www.w3.org/TR/REC-html40\">\r\n\r\n
  <head>\r\n
    <meta http-equiv=Content-Type content=\"text/html; charset=utf-8\">\r\n
    <meta name=ProgId content=PowerPoint.Slide>\r\n
    <meta name=Generator content=\"Microsoft PowerPoint 15\">\r\n`

after parsing the Html from the string the meta tags are lost:

image

I tried this also copying and pasting content from word and I got the same error, the meta tags of the rawHtml are not included in the parsed Html.

I think we could do some manipulation on the rawHtml to retrieve the tags from when since the DomParser().parseFromString is not returning the meta tags.

Any feedback?

JiuqingSong: POWERPOINT_ATTRIBUTE_VALUE  is the const name of PowerPoint.Slide:
const POWERPOINT_ATTRIBUTE_VALUE = 'PowerPoint.Slide';

See https://github.com/microsoft/roosterjs/blob/1a1b87916f2e5253da363c1e51f015626aadbe7d/packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts#L19

The content is right, and the meta value is right, too. The problem is the code to retrieve the meta value is not run.

BryanValverdeU: After some debugging, I noticed that the HEAD Tag is being removed from the Html when sanitizing the rawHtml (line) for instance, this is the rawHtml that is provided in the ClipboardData...
image


but after Sanitizing this string in core.trustedHTMLHandler(rawHtml), only the Table tag is returned and the Head is removed.... this is the result after running trustedHTMLHandler:

image

JiuqingSong: Please try let DOMPurify allow Head tag similar with: https://stackoverflow.com/questions/60299226/how-to-allow-an-iframe-tag-in-dompurify-including-all-of-its-attributes 
by changing https://github.com/microsoft/roosterjs/blob/master/demo/scripts/utils/trustedHTMLHandler.ts

If that does not work, try remove DOMPurify and directly return the source HTML in turstedHTMLHandler.

BryanValverdeU: With this DOMPurity Configuration the head and the meta tags are now included to the sanitized html

image

Thanks

packages/roosterjs-editor-types/lib/enum/ClipboardSource.ts
```diff
@@ -0,0 +1,10 @@
+export const enum ClipboardSource {
```
JiuqingSong, line 1: add comment to this enum and all members. Public members need comment

packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
```diff
@@ -3,21 +3,22 @@ import convertPastedContentFromExcel from './excelConverter/convertPastedContent
...
 
-const WORD_ATTRIBUTE_NAME = 'xmlns:w';
-const WORD_ATTRIBUTE_VALUE = 'urn:schemas-microsoft-com:office:word';
-const EXCEL_ATTRIBUTE_NAME = 'xmlns:x';
-const EXCEL_ATTRIBUTE_VALUE = 'urn:schemas-microsoft-com:office:excel';
-const PROG_ID_NAME = 'ProgId';
-const EXCEL_ONLINE_ATTRIBUTE_VALUE = 'Excel.Sheet';
-const POWERPOINT_ATTRIBUTE_VALUE = 'PowerPoint.Slide';
-const GOOGLE_SHEET_NODE_NAME = 'google-sheets-html-origin';
+const WORD_ATTRIBUTE_NAME = ClipboardSource.WORD_ATTRIBUTE_NAME;
```
JiuqingSong, line 14: Since we already have these enums, no need to have these constants again

packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
```diff
@@ -51,19 +52,12 @@ export const createPasteFragment: CreatePasteFragment = (
...
-    if (!pasteAsText && !text && imageDataUri) {
-        // Paste image
-        const img = document.createElement('img');
-        img.style.maxWidth = '100%';
-        img.src = imageDataUri;
-        fragment.appendChild(img);
-    } else if (
+    // Paste HTML
+    if (
         !pasteAsText &&
```
JiuqingSong, line 57: I think we don't need to check pasteAsText here. We can retrieve the meta tags as long as there is rawHtml

packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
```diff
@@ -51,19 +52,12 @@ export const createPasteFragment: CreatePasteFragment = (
...
 
-    // Step 2: Fill the BeforePasteEvent object, especially the fragment for paste
-    if (!pasteAsText && !text && imageDataUri) {
-        // Paste image
-        const img = document.createElement('img');
-        img.style.maxWidth = '100%';
-        img.src = imageDataUri;
-        fragment.appendChild(img);
-    } else if (
+    // Paste HTML
```
JiuqingSong, line 55: this comment is not accurate now. We are not pasting HTML here, we are just retrieving some metadata from HTML

packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
```diff
@@ -173,3 +176,24 @@ function createBeforePasteEvent(core: EditorCore, clipboardData: ClipboardData):
 function processStyles(node: ParentNode, callback: (style: HTMLStyleElement) => void) {
     toArray(node.querySelectorAll('style')).forEach(callback);
 }
+
+function validateClipboardImageSource(event: BeforePasteEvent, pasteAsText: boolean): boolean {
```
JiuqingSong, line 180: We already have code to do check for powerpoint in Paste plugin. In this function it should be neutral and do not do check for any specific scenario.

JiuqingSong: See https://github.com/microsoft/roosterjs/blob/658dd377f31941b7422e1706d1ffe6fc65c2a72d/packages/roosterjs-editor-plugins/lib/plugins/Paste/pptConverter/convertPastedContentFromPowerPoint.ts#L18
It will rebuild the fragment to use HTML when paste from PPT.

JiuqingSong: createPasteFragment() only handles common cases, which means it should not know any special knowledge, just provide whatever it can provide such as html string, image source, ... and provide a document fragment with default paste result. Then let plugin decide what should be pasted finally. Plugin can overwrite the default result.
People can use our Paste plugin, and can also create their own plugin to decide what to paste.

packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
```diff
@@ -50,19 +50,11 @@ export const createPasteFragment: CreatePasteFragment = (
...
-    // Step 2: Fill the BeforePasteEvent object, especially the fragment for paste
-    if (!pasteAsText && !text && imageDataUri) {
-        // Paste image
-        const img = document.createElement('img');
-        img.style.maxWidth = '100%';
-        img.src = imageDataUri;
-        fragment.appendChild(img);
-    } else if (
-        !pasteAsText &&
+    // Retrieving Metadata from Html and the Html that was copied.
```
JiuqingSong, line 53: let comment this part as "step 2"

packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
```diff
@@ -95,7 +87,16 @@ export const createPasteFragment: CreatePasteFragment = (
             // and the nodes under HEAD are still used when convert global CSS to inline
             processStyles(doc.body, style => style.parentNode?.removeChild(style));
         }
+    }
 
+    // Step 2: Fill the BeforePasteEvent object, especially the fragment for paste
```
JiuqingSong, line 92: and this becomes step 3 now

packages/roosterjs-editor-types/lib/enum/ClipboardSource.ts
JiuqingSong: Since these values are only used in one single file now, Do we still need to make them public?

JiuqingSong (approved):

PR #617, opened 2021-08-23
new api to rotate resize handles

juliaroldi:
New api to rotate the resize handles when the image is rotated. 

rotateHandles

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/api/resizeHandlesRotator.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/api/resizeHandlesRotator.ts
@@ -1,0 +1,122 @@
+const Rotation30to60DegreesHandlesSequence = [
+    'w-resize',
+    'n-resize',
+    's-resize',
+    'e-resize',
+    'nw-resize',
+    'sw-resize',
+    'ne-resize',
+    'se-resize',
+];
+
+const Rotation60to120DegreesHandlesSequence = [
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/api/resizeHandlesRotator.ts
```diff
@@ -0,0 +1,122 @@
+const Rotation22to67DegreesHandlesSequence = [
```
JiuqingSong, line 1: I also want to simplify this part, something like

```ts
const DirectionOrder = ['nw', 'n', 'ne', 'e', 'se', 's', 'sw', w']; // This makes them like a circle
...
const originalDirection = element.dataset.y + element.dataset.x; // this will make it become one of the values above
const originalIndex = DirectionOrder.indexOf(originalDirection);
const rotatedIndex = originalIndex + radIndex; // radIndex is retrieved from the function in the comment below
return DirectionOrder[rotatedIndex  % Directions]; //Directions is 8
```

The idea is, suppose for the top handler, dataset.y+dataset.x will be "n", so originalIndex will be 1.
Then when rotated 90 deg (pi / 2), we get the radIndex from:
 radIndex = round(angle / (pi / 4)) % 8 = 2,
So rotatedIndex = originalIndex + radIndex = 1 + 2 = 3
So the returned direction will be DirectionOrder[3] = 'e', which means we should show "e-resize" for top handle now.

PR #618, opened 2021-08-23
U/bvalverde/editor improvement adjust list indentation

BryanValverdeU:
editor improvement adjust list indentation

### DIFF
--- a/demo/scripts/controls/contextMenu/ContextMenuProvider.tsx
+++ b/demo/scripts/controls/contextMenu/ContextMenuProvider.tsx
@@ -8,6 +8,7 @@ export interface ContextMenuItem {
     key: string;
     name: string;
     onClick: () => void;
+    dropDownItems?: ContextMenuItem[];
 }
 
 export const CONTEXT_MENU_DATA_PROVIDER: ContextMenuOptions<ContextMenuItem> = {
@@ -22,6 +23,7 @@ export const CONTEXT_MENU_DATA_PROVIDER: ContextMenuOptions<ContextMenuItem> = {
 function ContextMenu(props: { items: ContextMenuItem[]; onDismiss: () => void }) {
     const div = React.useRef<HTMLDivElement>();
     const { items, onDismiss } = props;
+    const [mouseOverKey, setMouseOverKey] = React.useState('');
 
     React.useEffect(() => {
         const doc = div.current.ownerDocument;
@@ -34,17 +36,31 @@ function ContextMenu(props: { items: ContextMenuItem[]; onDismiss: () => void })
     }, []);
     return (
         <div className={styles.menu} ref={div}>
-            {items.map(
-                item =>
+            {items?.map(item => {
+                return (
                     item && (
-                        <div
-                            className={styles.menuItem}
-                            key={item.key}
-                            onClick={() => item.onClick()}>
-                            {item.name}
-                        </div>
+                        <>
+                            <div
+                                className={styles.menuItem}
+                                key={item.key}
+                                onMouseEnter={() => setMouseOverKey(item.key)}
+                                onMouseLeave={() => setMouseOverKey(item.key)}
+                                onClick={() => item.onClick()}>
+                                {item.name}
+                            </div>
+                            {item.dropDownItems && item.key === mouseOverKey ? (
+                                <div className={item.dropDownItems ? styles.deepMenuItem : ''}>
+                                    <ContextMenu
+                                        items={item.dropDownItems}
+                                        onDismiss={onDismiss}></ContextMenu>
+                                </div>
+                            ) : (
+                                <></>
+                            )}
+                        </>
                     )
-            )}
+                );
+            })}
         </div>
     );
 }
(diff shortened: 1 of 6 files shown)


demo/scripts/controls/contextMenu/ContextMenuProvider.tsx
```diff
@@ -34,17 +36,31 @@ function ContextMenu(props: { items: ContextMenuItem[]; onDismiss: () => void })
...
-                            key={item.key}
-                            onClick={() => item.onClick()}>
-                            {item.name}
-                        </div>
+                        <>
+                            <div
+                                className={styles.menuItem}
+                                key={item.key}
+                                onMouseEnter={() => setMouseOverKey(item.key)}
+                                onMouseLeave={() => setMouseOverKey(item.key)}
```
JiuqingSong, line 47: Why set key when mouse leave?

demo/scripts/controls/contextMenu/ResetListPlugin.ts
```diff
@@ -1,7 +1,11 @@
...
 import { ContextMenuProvider, IEditor } from 'roosterjs-editor-types';
 import { safeInstanceOf } from 'roosterjs-editor-dom';
-import { setOrderedListNumbering } from 'roosterjs-editor-api';
+import {
+    setListIndent,
+    setListNumberPosition,
+    setOrderedListNumbering,
+} from 'roosterjs-editor-api';
 
 export default class ResetListPlugin implements ContextMenuProvider<ContextMenuItem> {
```
JiuqingSong, line 10: We can rename this plugin now since it doesn't only do reset list

demo/scripts/controls/contextMenu/ResetListPlugin.ts
```diff
@@ -42,6 +46,75 @@ export default class ResetListPlugin implements ContextMenuProvider<ContextMenuI
...
+                                    case 'px':
+                                        setListNumberPosition(this.editor, list, value, 'px');
+                                        break;
+                                    case '%':
+                                        setListNumberPosition(this.editor, list, value, '%');
+                                        break;
+                                    case 'in':
+                                        setListNumberPosition(this.editor, list, value, 'in');
+                                        break;
+                                    default:
```
JiuqingSong, line 80: You can merge them into:
case 'em':
case 'px':
case '%':
case 'in':
  setListNumberPosition(this.editor,list,value, unitValue);
  break;

demo/scripts/controls/contextMenu/ResetListPlugin.ts
```diff
@@ -42,6 +46,75 @@ export default class ResetListPlugin implements ContextMenuProvider<ContextMenuI
...
+                            onClick: () => {
+                                let value = parseInt(prompt('Set Value to...', '1'));
+
+                                let unitValue = prompt(
+                                    "Set Value to... ('em' | 'px' | '%' | 'in')",
+                                    'em'
+                                );
+
+                                switch (unitValue) {
+                                    case 'em':
```
JiuqingSong, line 97: same here

packages/roosterjs-editor-api/lib/format/setListNumberPosition.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export default function setListNumberPosition(
+    editor: IEditor,
+    list: HTMLElement,
+    numberPosition: number,
+    unit: 'em' | 'px' | '%' | 'in'
+) {
+    editor.addUndoSnapshot((start, end) => {
+        editor.focus();
+        list.style.marginLeft = numberPosition.toString() + unit;
```
JiuqingSong, line 18: What about RTL? Maybe we should set to both marginLeft and marginRight?

packages/roosterjs-editor-api/lib/format/setListNumberPosition.ts
```diff
@@ -0,0 +1,22 @@
+import { ChangeSource, IEditor } from 'roosterjs-editor-types';
+
+/**
+ * Set the List Number Position (Margin) of a List
+ * @param editor The editor instance
+ * @param list The HTML element
+ * @param numberPosition The number position value to be set to the list
+ * @param unit The Unit to be used.
+ */
+export default function setListNumberPosition(
```
JiuqingSong, line 10: Need a better name, this does not only work for numbering list, but it actually works for all block elements.

packages/roosterjs-editor-api/lib/format/setListIndent.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export default function setListIndent(
+    editor: IEditor,
+    list: HTMLElement,
+    textIdent: number,
+    unit: 'em' | 'px' | '%' | 'in'
+) {
+    editor.addUndoSnapshot((start, end) => {
+        editor.focus();
+        list.style.textIndent = textIdent.toString() + unit;
```
JiuqingSong, line 18: Actually these two function are very similar. Then works in different way when this is a list.

And we already have another format API setIndentation. We need to think about whether we can merge these 3 together, and make it smart. For example, if a block already have a BLOCKQUOTE-based indentation, then user want to adjust its indentation by px, how should we handle it?

BryanValverdeU: We could use a object parameter on the setIndentation like this one, or add it as a not required parameter so we have backward compatibility

image

And when any property of the object is not null set the styles on the element.

PR #620, opened 2021-08-24
bugfix pastegifs leave a static image

BryanValverdeU:
When pasting an animated image, the result image was static. This change will fix the issue.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
@@ -2,6 +2,7 @@ import convertPastedContentForLI from './commonConverter/convertPastedContentFor
 import convertPastedContentFromExcel from './excelConverter/convertPastedContentFromExcel';
 import convertPastedContentFromPowerPoint from './pptConverter/convertPastedContentFromPowerPoint';
 import convertPastedContentFromWord from './wordConverter/convertPastedContentFromWord';
+import convertPastedImage from './imageConverter/convertPastedImage';
 import handleLineMerge from './lineMerge/handleLineMerge';
 import { EditorPlugin, IEditor, PluginEvent, PluginEventType } from 'roosterjs-editor-types';
 import { toArray } from 'roosterjs-editor-dom';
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -1,3 +1,4 @@
+import safeInstanceOf from '../utils/safeInstanceOf';
```
JiuqingSong, line 1: this is not used

packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
```diff
@@ -94,6 +95,8 @@ export default class Paste implements EditorPlugin {
                 }
             } else if (fragment.querySelector(GOOGLE_SHEET_NODE_NAME)) {
                 sanitizingOption.additionalTagReplacements[GOOGLE_SHEET_NODE_NAME] = '*';
+            } else if (clipboardData.html && clipboardData.image) {
```
JiuqingSong, line 98: It is a very common case that both html and image exist. So this check causes Line 100 cannot execute in a lot of cases.

JiuqingSong: And since this will change existing paste behavior, suggest adding a new experimental feature so that we can turn it off anytime.

packages/roosterjs-editor-plugins/lib/plugins/Paste/imageConverter/convertPastedImage.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+export default function convertPastedImage(
+    event: BeforePasteEvent,
+    trustedHTMLHandler: TrustedHTMLHandler
+) {
+    const { fragment, clipboardData } = event;
+
+    if (clipboardData.html && clipboardData.image) {
+        //If there are Html in the clipboard, and the html body only have one img children, use the HTML
+        const doc = new DOMParser().parseFromString(
```
JiuqingSong, line 17: This means we need to parse the HTML content again in a very comment case, but in most case after parsing we will see that there will not only be an IMAGE, so we should do nothing. That means this check will be wasted in most time.

Consider add some hint info in the ClipboardData object so that when createPasteFragment we can set the hint info there since we have the parsed doc at the time. Then here we can read it and when it meet some condition, we do this work here. That can save more than 99% parsing cases.

PR #621, opened 2021-08-25
U/bvalverde/hyperlinks style getting removed

BryanValverdeU:
InnerHtml of Anchor elements styles get removed when removing format of a selected text.

Can use this snapshot to repro the issue.
`<div><div style="font-family: Calibri, Arial, Helvetica, sans-serif; font-size: 12pt; color: rgb(0, 0, 0); background-color: rgb(255, 255, 255);"><span style="font-size: 72pt;">This is a </span><a href="http://microsoft.com" title="microsoft.com"><span style="font-size: 72pt;">link</span></a><span style="font-size: 72pt;">.</span></div></div><!--{"start":[0,0,0],"end":[0,0,3]}-->`

Hyperlink Style after


packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -191,10 +190,17 @@ function clearInlineFormat(editor: IEditor) {
                 setFontSize(editor, defaultFormat.fontSize);
             }
             if (defaultFormat.textColor) {
+                const setColorIgnoredElements = editor.queryElements<HTMLElement>(
+                    'a *, a',
+                    QueryScope.OnSelection
+                );
+                const shouldApplyInlineStyle = (element: HTMLElement) =>
```
JiuqingSong, line 197: If setColorIgnoredElements is empty, no need to have this callback

JiuqingSong (approved):

Issue #614, opened 2021-08-20
Sibling lists with incontiguous numbers get merged when exit the list


JiuqingSong:
1. Start demo site locally
2. Go to http://localhost:3000/#Snapshot
3. Copy the code below, and paste into the right box of demo site, then click "Restore snapshot"
```html
<div><ol><li><span>line 1</span></li><li><span>line 2</span></li></ol><ol><li><span>line 3</span></li><li><span>line 4</span></li></ol><div><span><br></span></div></div><!--{"start":[0,1,1,0,0,6],"end":[0,1,1,0,0,6]}-->
```
4. Press Enter key twice

Expect: there should be still two lists, and numbers are "1,2,1,2"
Actual: it is merged into one single list, and numbers are "1,2,3,4"

JiuqingSong:
Fixed in #626

PR #626, opened 2021-08-31
U/bvalverde/sibling list merged

BryanValverdeU:
This pull request fixes the issue described on #614 
bug

### DIFF
--- a/packages/roosterjs-editor-api/lib/utils/toggleListType.ts
+++ b/packages/roosterjs-editor-api/lib/utils/toggleListType.ts
@@ -3,20 +3,22 @@ import { createVListFromRegion, getBlockElementAtNode } from 'roosterjs-editor-d
 import { IEditor, ListType } from 'roosterjs-editor-types';
 
 /**
- * @internal
+ * Toggle List Type at selection
+ * If ListType Provided is Ordered:
+ *      If selection contains numbering in deep level, toggle numbering will decrease the numbering level by one
+ *      If selection contains bullet list, toggle numbering will convert the bullet list into number list
+ *      If selection contains both bullet/numbering and normal text, the behavior is decided by corresponding
+ *       realization of browser execCommand API
+ * If ListType Provided is Unordered:
+ *      If selection contains bullet in deep level, toggle bullet will decrease the bullet level by one
+ *      If selection contains number list, toggle bullet will convert the number list into bullet list
+ *      If selection contains both bullet/numbering and normal text, the behavior is decided by corresponding
+ *      browser execCommand API
+ * @param editor The editor instance
+ * @param listType The list type to toggle
+ * @param startNumber (Optional) Start number of the list
+ * @param includeSiblingLists Sets wether the operation should include Sibling Lists, by default true
  */
-export default function toggleListType(
-    editor: IEditor,
-    listType: ListType,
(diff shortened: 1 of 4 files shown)


packages/roosterjs-editor-api/lib/utils/toggleListType.ts
```diff
@@ -11,8 +11,19 @@ export default function toggleListType(
     listType: ListType.Ordered,
     startNumber: number
 ): void;
+export default function toggleListType(
```
JiuqingSong, line 14: All these 3 overrides can compatible with each other, so no need to have 3 different overrides, just one override with optional parameter should work fine.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -248,15 +248,19 @@ function prepareAutoBullet(editor: IEditor, range: Range) {
...
+) {
     let listInfo = cacheGetListElement(event, editor);
     if (listInfo) {
         let listElement = listInfo[0];
         let tag = getTagOfNode(listElement);
         if (tag == 'UL') {
             toggleBullet(editor);
         } else if (tag == 'OL') {
-            toggleNumbering(editor);
+            toggleNumbering(editor, null, includeSiblingList);
```
JiuqingSong, line 263: I'm thinking about, if we should export that much parameters to a public format API (toggleNumbering). How about call toggleListType directly instead? So that no need to change the parameters for toggleNumbering.

And actually this bug happens for bullet list as well, it is just not that easy to repro. So maybe we should fix both of them, and calling toggleListType seems better here

packages/roosterjs-editor-api/lib/utils/toggleListType.ts
JiuqingSong: I mean, we can delete line 8 - line 18

JiuqingSong: And now that the function is exported, we need a regular comment, and remove "@intenal

JiuqingSong: And the comment need to reflect the default value of includeSiblingLists is true

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -248,16 +250,24 @@ function prepareAutoBullet(editor: IEditor, range: Range) {
...
+        let handler = (listType: ListType) =>
+            toggleListType(editor, listType, null, includeSiblingLists);
+
         if (tag == 'UL') {
-            toggleBullet(editor);
+            handler(ListType.Unordered);
         } else if (tag == 'OL') {
-            toggleNumbering(editor);
+            handler(ListType.Ordered);
         }
```
JiuqingSong, line 269: why not just 
```ts
if (tag == 'UL' || tag == 'OL') {
toggleListType(editor, tag == 'UL' ? ListType.Unordered : ListType.Ordered, null /* comment*/,  includeSiblingLists)
}
```
?

JiuqingSong (approved):

PR #627, opened 2021-08-31
U/juliaroldi/auto change

juliaroldi:
Added shouldReplace callback to ensure only characters that must be replaced to be replaced.
Also replaced space in charCode 160 with charCode 32.
auto change sup feat


packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -90,14 +96,23 @@ export default class CustomReplacePlugin implements EditorPlugin {
         if (range == null) {
             return;
         }
-        const searcher = this.editor.getContentSearcherOfCursor(event);
-        const stringToSearch = searcher.getSubStringBefore(this.longestReplacementLength);
 
+        const searcher = this.editor.getContentSearcherOfCursor(event);
+        const stringToSearch = searcher
+            .getSubStringBefore(this.longestReplacementLength)
+            .replace(/\s/g, ' ');
```
JiuqingSong, line 103: suggest move the replace call into getMAtchingReplacment(). It should be part of the matching logic.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -90,14 +96,23 @@ export default class CustomReplacePlugin implements EditorPlugin {
...
         const replacement = this.getMatchingReplacement(stringToSearch);
         if (replacement == null) {
             return;
         }
 
+        if (
+            replacement.shouldReplace &&
+            !replacement.shouldReplace(searcher.getWordBefore(), replacement.sourceString)
```
JiuqingSong, line 111: Can we just pass replacement as the first parameter here?

JiuqingSong: Open question: what about pass the whole replacement object as the second parameter?

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -131,9 +146,8 @@ export default class CustomReplacePlugin implements EditorPlugin {
             const [sourceMatch, replacementMatch] = replacement.matchSourceCaseSensitive
                 ? [stringToSearch, replacement.sourceString]
                 : [lowerCaseStringToSearch, replacement.sourceString.toLocaleLowerCase()];
-
             if (
-                sourceMatch.substring(sourceMatch.length - replacementMatch.length) ==
+                sourceMatch.substring(sourceMatch.length - replacementMatch.length) ===
```
JiuqingSong, line 150: Just curious, is there any case broken without this change?

juliaroldi: The tests cases didn't break, I think is because the replacementMatch and the sourceString are defined as strings.

juliaroldi: But its better to let the way it was, since I didn't test all scenarios

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -129,9 +143,8 @@ export default class CustomReplacePlugin implements EditorPlugin {
         const lowerCaseStringToSearch = stringToSearch.toLocaleLowerCase();
         for (const replacement of this.replacements) {
             const [sourceMatch, replacementMatch] = replacement.matchSourceCaseSensitive
-                ? [stringToSearch, replacement.sourceString]
+                ? [stringToSearch.replace(/\s/g, ' '), replacement.sourceString]
```
JiuqingSong, line 146: Need to move this out of the loop, otherwise the replace will be executed many times, it is a waste

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -129,9 +143,8 @@ export default class CustomReplacePlugin implements EditorPlugin {
         const lowerCaseStringToSearch = stringToSearch.toLocaleLowerCase();
```
JiuqingSong, line 143: lowerCaseStringToSearch should be calculated from the replaced result as well

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -98,6 +105,13 @@ export default class CustomReplacePlugin implements EditorPlugin {
             return;
         }
 
+        if (
+            replacement.shouldReplace &&
+            !replacement.shouldReplace(replacement, searcher.getWordBefore())
```
JiuqingSong, line 110: You already have stringToSearch, no need to call getWordBefore() again

juliaroldi: I'm only replacing st, nd, rd, th, then I'm using getWordBefore() to find if there's a number before st, nd, rd, th.

juliaroldi: Then I' don't have to replace  from 1st to 9th.

JiuqingSong: can you give me an example, when I type "1st", what are the values of stringToSearch and getWordBefore()?

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -131,9 +145,9 @@ export default class CustomReplacePlugin implements EditorPlugin {
             const [sourceMatch, replacementMatch] = replacement.matchSourceCaseSensitive
                 ? [stringToSearch, replacement.sourceString]
                 : [lowerCaseStringToSearch, replacement.sourceString.toLocaleLowerCase()];
-
+            const cleanSourceMatch = sourceMatch.replace(/\s/g, ' ');
```
JiuqingSong, line 148: reg exp replacing is an expensive operation, so we should move it out of the loop.
For example:
const originalStringToSearch = stringToSearch.replace(/s/g, ' ');
const lowerCaseStringToSearch = originalStringToSearch .toLocalLowerCase();
for (... ) {
   const [sourceMatch, replacementMatch] = replacement.matchSourceCaseSensitive ? [originalStringToSearch , replacement.sourceString] : [lowerCaseStringToSearch , replacement.sourceString.toLocaleLowerCase()];

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
juliaroldi: 1. Type "1st "
2.  The stringToSearch will be "st " and the getWordBefore() will be "1st "
3. The callback will use getWordBefore() to check if there's a 1 before "st "
4. If true, replace st for 1<sup>st</sup>&nbsp;
5. If not, do not replace.

JiuqingSong (approved):

PR #628, opened 2021-09-01
Clear Format On Exit Quote

BryanValverdeU:
Clear the format when user press ENTER on an empty line of a BLOCKQUOTE.


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/quoteFeatures.ts
```diff
@@ -74,16 +76,36 @@ function splitQuote(event: PluginKeyboardEvent, editor: IEditor) {
...
+        element.style.borderColor &&
+        element.style.paddingLeft &&
+        element.style.color
+    ) {
+        return true;
+    }
+    return false;
+};
+
 /**
```
BryanValverdeU, line 109: Added this check, so we verify that the QuoteElement is styled and it is not a blockquote used in indentation

JiuqingSong:
What is the original bug you are fixing? And what it broken without this fix?

BryanValverdeU:
> What is the original bug you are fixing? And what it broken without this fix?

This is the Bug: https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/125548
It seems that when exiting a Quote did not default format the text back to the default styles.

JiuqingSong:
> > What is the original bug you are fixing? And what it broken without this fix?
> 
> This is the Bug: https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/125548
> It seems that when exiting a Quote did not default format the text back to the default styles.

Can you repro the original bug?

BryanValverdeU:
Yes, I was able to repro, when the format of a quote text style is different: this is a snapshot.
`<blockquote style="border-left: 3px solid rgb(200, 200, 200); border-top-color: rgb(200, 200, 200); border-right-color: rgb(200, 200, 200); border-bottom-color: rgb(200, 200, 200); padding-left: 10px; color: rgb(102, 102, 102);"><div><span style="font-size: 36pt; color: rgb(243, 144, 29); font-family: Tahoma;">This is a quote</span></div></blockquote><div><span style="font-size: 36pt; color: rgb(243, 144, 29); font-family: Tahoma;">Text after exiting a quote with enter key</span></div><div>Text with Format Cleaned</div><!--{"start":[1,0,0,41],"end":[1,0,0,41]}-->`

JiuqingSong (approved):

PR #638, opened 2021-09-08
Improve the build tool to force comments Step 5: Finally improve the build tool

JiuqingSong:
This is the last step of comment improvement. For a full change, see here: https://github.com/microsoft/roosterjs/pull/630

Step 5: Improve build tool, fix export issue for ImageResize (deprecated) plugin

After all these changes, we will force add comment for:

Exported elements (interface/class/function/...) must have comments
Local exported elements (exported from its file but not exported from index.ts) must be marked as @internal
Public exported elements (exported from index.ts) must not be marked as @internal
Also fixed all existing violations.


tools/dts.js
```diff
@@ -205,44 +209,48 @@ function parseExportFrom(content, currentFileName, queue, baseDir, projDir) {
...
-
-    // 5. Parse functions
-    content = parseFunctions(content, item.elements);
-
-    // 6. Parse enum
-    content = parseEnum(content, item.elements);
-
-    // 7. Parse type
-    content = parseType(content, item.elements);
+    // 2. Remove imports
```
juliaroldi, line 230: is there no need to parse type, functions, classes and interfaces?

romanisa (approved):
Thank you making this improvement to the build tools!

tools/dts.js
```diff
@@ -205,44 +209,48 @@ function parseExportFrom(content, currentFileName, queue, baseDir, projDir) {
...
+    content = content.replace(singleLineComment, '');
 
     if (content.trim() != '') {
         err('File ' + currentFileName + ' contains unrecognized content:\r\n' + content);
     }
 }
 
+function publicElement(element) {
+    return element.map(x => {
+        x.published = true;
```
Lufedi, line 249: is this really necessary if you are returning x.text here?

Lufedi: oh nevermind I see that you are using the published prop in another parts of the code

tools/dts.js
```diff
@@ -298,6 +305,30 @@ function output(targetDir, library, isAmd, queue) {
...
 
+    // Check comments
+    for (var i = 0; i < queue.length; i++) {
+        var { filename, elements } = queue[i];
+
+        for (var name in elements) {
+            elements[name].forEach(({ published, text, comment }) => {
+                var code = text.replace(namePlaceholder, name);
+                if (!comment) {
+                    err(`Exported element must have comment. File: ${filename} Code: ${code}`);
```
Lufedi, line 316: Nice! this is a very good linting feature here.

Lufedi (approved):

tools/dts.js
Lufedi: I think he is parsing everything that has the `export` statement

JiuqingSong: Yes, that is on line 234

Issue #631, opened 2021-09-03
getFormateState API changes?


Yat3s:
**Describe the bug**
getFormateState()  return the format state of the whole dom instead of content.

**To Reproduce**
```
<div id="text1">
   <strong>Content</strong>
</div>

...
const element = document.querySelector('#text1');
const range = document.createRange();
range.selectNodeContents(element);
document.getSelection().removeAllRanges();
document.getSelection().addRange(range);

...
const format = getFormateState();
```

<=8.7.0:  
`format.isBold => true`

8.8.0:  
`format.isBold => false`

I didn't saw any change log for this API, is it regression or by design?

JiuqingSong:
Thanks for reporting this issue. It sounds like a regression. @juliaroldi  please take a look.

BTW, suggest use editor.select() instead of getSelection().addRange() so that editor can know that the selection is changed.

juliaroldi:
Bug fixed. See [PR 639](https://github.com/microsoft/roosterjs/pull/639)

PR #639, opened 2021-09-08
Fixed current position of getPendableFormatState

juliaroldi:
Fixed getPendableFormatState for always get the current node from the selected range.


packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -18,16 +18,15 @@ export const getPendableFormatState: GetPendableFormatState = (
...
     const cachedPosition = core.pendingFormatState.pendableFormatPosition?.normalize();
-    const currentPosition = isRangeCollapsed && Position.getStart(range).normalize();
+    const currentPosition = range && Position.getStart(range).normalize();
     const isSamePosition = currentPosition && currentPosition.equalTo(cachedPosition);
 
     if (range && cachedPendableFormatState && isSamePosition && !forceGetStateFromDOM) {
         return cachedPendableFormatState;
     } else {
-        return queryCommandStateFromDOM(core, isRangeCollapsed, currentPosition);
+        return queryCommandStateFromDOM(core, currentPosition);
```
JiuqingSong, line 29: when range is null, currentPosition is also null. Actually in that case, we can just return "{}" since no selection means no format.

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -18,16 +18,15 @@ export const getPendableFormatState: GetPendableFormatState = (
     forceGetStateFromDOM: boolean
 ): PendableFormatState => {
     const range = core.api.getSelectionRange(core, true /* tryGetFromCache*/);
-    const isRangeCollapsed = range && range.collapsed;
     const cachedPendableFormatState = core.pendingFormatState.pendableFormatState;
     const cachedPosition = core.pendingFormatState.pendableFormatPosition?.normalize();
-    const currentPosition = isRangeCollapsed && Position.getStart(range).normalize();
+    const currentPosition = range && Position.getStart(range).normalize();
     const isSamePosition = currentPosition && currentPosition.equalTo(cachedPosition);
```
JiuqingSong, line 24: Now currentPosition will have value even range is not collapsed. So you also need to check range is collapsed here

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -67,10 +66,9 @@ const CssFalsyCheckers: Record<PendableFormatNames, (style: CSSStyleDeclaration)
...
 function queryCommandStateFromDOM(
     core: EditorCore,
-    isCollapsed: boolean,
     currentPosition: NodePosition
 ): PendableFormatState {
-    let node = isCollapsed ? currentPosition?.node : document.getSelection().focusNode;
+    let node = currentPosition?.node;
     let formatState: PendableFormatState = {};
     let pendablekeys: PendableFormatNames[] = [];
     while (contains(core.contentDiv, node)) {
```
JiuqingSong, line 74: As I comment above, when currentPosition is null, node is also null, so this line will throw exception

packages/roosterjs-editor-core/lib/coreApi/getPendableFormatState.ts
```diff
@@ -15,19 +15,19 @@ import {
  */
 export const getPendableFormatState: GetPendableFormatState = (
     core: EditorCore,
-    forceGetStateFromDOM: boolean
+    forceGetStateFromDOM: boolean = false
```
JiuqingSong, line 18: Do not mark parameter of core API to optional. We expect all parameters of core API to be explicitly passed.

packages/roosterjs-editor-core/test/coreApi/getPendableFormatStateTest.ts
```diff
@@ -15,10 +15,31 @@ describe('getPendableFormatState', () => {
 
     it('underline with u tag', () => {
         const core = createEditorCore(div, {});
-        div.innerHTML =
-            '<div style="font-family: arial; font-size: 12pt; color: black; background-color: white"><u>test</u></div>';
-        core.api.focus(core);
-        const format = getPendableFormatState(core, false);
+
+        core.api.setContent(
```
JiuqingSong, line 19: Consider write a common "runTest" function to extract all common logic to make the test code easier to read.

JiuqingSong (approved):

juliaroldi:
Fixed current position of getPendableFormatState.
Bugfix for Issue [631](https://github.com/microsoft/roosterjs/issues/631)

juliaroldi:
Fixed getPendableFormatState for always get the current node from the selected range.
Bugfix for Issue [631](Fixed current position of getPendableFormatState)

PR #644, opened 2021-09-14
Set List Item ::marker styles to be equal as the content when starting new List

BryanValverdeU:
Set the List Item number or bullet, to have the same Font SIze, Color and Font Family as the Childrens, when starting a new List.

Resize Marker According


JiuqingSong:
general question, after apply the styles to list number, can I change the font/size to the numbers? I tried in Word and it can change number font when I select the whole list item

BryanValverdeU:
> general question, after apply the styles to list number, can I change the font/size to the numbers? I tried in Word and it can change number font when I select the whole list item

At the moment no
But we could call the setListItemStyle method from the format functions like setFontSize, setFontColor and setFontName, if the parent of the element is a LI Element, so we modify the marker styles too

Should I add this this functionality?

JiuqingSong:
> > general question, after apply the styles to list number, can I change the font/size to the numbers? I tried in Word and it can change number font when I select the whole list item
> 
> At the moment no
> But we could call the setListItemStyle method from the format functions like setFontSize, setFontColor and setFontName, if the parent of the element is a LI Element, so we modify the marker styles too
> 
> Should I add this this functionality?

let's try that.

BryanValverdeU:
Added the feature to change the list item marker when modifying all the inline childs for the FontName, Color and FontSize Format functions, but when the fontSize is really big some of the markers/bullets are Overflowing to the Left, not sure if that is an issue.

Change the Marker Style1

packages/roosterjs-editor-dom/lib/utils/setListItemStyle.ts
```diff
@@ -0,0 +1,56 @@
...
+    let currentInlineElement: InlineElement;
+
+    while (contentTraverser.currentInlineElement != currentInlineElement) {
+        currentInlineElement = contentTraverser.currentInlineElement;
+        let currentNode = currentInlineElement.getContainerNode();
+
+        if (currentNode.nodeName != 'BR') {
+            if (currentNode.nodeName == '#text') {
+                currentNode = currentNode.parentNode;
+            }
```
JiuqingSong, line 44: you can use findClosestElementAncestor

packages/roosterjs-editor-dom/lib/utils/setListItemStyle.ts
```diff
@@ -0,0 +1,56 @@
...
+        }
+    });
+    setStyles(element, stylesToApply);
+}
+function getInlineChildElementsStyle(element: HTMLElement) {
+    const result: Record<string, string>[] = [];
+    const contentTraverser = ContentTraverser.createBodyTraverser(element);
+    let currentInlineElement: InlineElement;
+
+    while (contentTraverser.currentInlineElement != currentInlineElement) {
```
JiuqingSong, line 37: Interesting, I thought when go beyond the boundary it currentInlineElement should return null.
Maybe it is a bug?

packages/roosterjs-editor-api/lib/format/setFontName.ts
```diff
@@ -14,5 +15,9 @@ export default function setFontName(editor: IEditor, fontName: string) {
     // for here, we use CSS font-family style
     applyInlineStyle(editor, (element, isInnerNode) => {
         element.style.fontFamily = isInnerNode ? '' : fontName;
+
+        if (safeInstanceOf(element.parentElement, 'HTMLLIElement')) {
+            setListItemStyle(element.parentElement, ['font-family']);
```
JiuqingSong, line 20: If there are multiple elements are involved for this action, this function will be run multiple times, but actually we only need it run once. So you can cache all parent nodes. Then get all distinct LI items and run them one by one.

packages/roosterjs-editor-api/lib/format/setFontName.ts
```diff
@@ -14,5 +15,9 @@ export default function setFontName(editor: IEditor, fontName: string) {
     // for here, we use CSS font-family style
     applyInlineStyle(editor, (element, isInnerNode) => {
         element.style.fontFamily = isInnerNode ? '' : fontName;
+
+        if (safeInstanceOf(element.parentElement, 'HTMLLIElement')) {
```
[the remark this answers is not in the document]
JiuqingSong: If there are nested nodes, LI may not be the direct parent node.

packages/roosterjs-editor-dom/lib/utils/setListItemStyle.ts
BryanValverdeU: Yes, maybe something is up there. 
If the currentInlineElement was the last element, the method getNextInlineElement was returning the last element instead of null sometimes.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -164,27 +164,25 @@ const AutoBullet: BuildInEditFeature<PluginKeyboardEvent> = {
...
-                    prepareAutoBullet(editor, rangeToDelete);
+                    prepareAutoBullet(editor, textRange);
                     toggleNumbering(editor);
                 } else if ((regions = editor.getSelectedRegions()) && regions.length == 1) {
                     const num = parseInt(textBeforeCursor);
-                    prepareAutoBullet(editor, rangeToDelete);
+                    prepareAutoBullet(editor, textRange);
                     toggleNumbering(editor, num);
                 }
+                searcher.getRangeFromText(textBeforeCursor, true /*exactMatch*/)?.deleteContents();
```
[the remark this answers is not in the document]
BryanValverdeU: I moved this from the prepareAutoBullet function, because before, we deleted the inline element tag with the styles applied to it, so we did not have styles to use in the List Item. 
So, I moved this after toggling the list type so we maintain the styles in the list.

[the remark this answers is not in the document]
JiuqingSong: Makes sense. Thanks.

packages/roosterjs-editor-dom/lib/utils/setListItemStyle.ts
JiuqingSong: Please open an issue to track. We need to revisit all places where this method is used.

packages/roosterjs-editor-api/lib/format/setFontName.ts
BryanValverdeU: Yes you are right.
I will add an array to store the unique parent nodes. So after we finish applying the styles on the child elements, calling this action only once for each individual Node.

packages/roosterjs-editor-api/lib/format/setFontName.ts
```diff
@@ -8,11 +9,18 @@ import { IEditor } from 'roosterjs-editor-types';
...
 export default function setFontName(editor: IEditor, fontName: string) {
+    const parentNodes: Node[] = [];
     fontName = fontName.trim();
     // The browser provided execCommand creates a HTML <font> tag with face attribute. <font> is not HTML5 standard
     // (http://www.w3schools.com/tags/tag_font.asp). Use applyInlineStyle which gives flexibility on applying inline style
     // for here, we use CSS font-family style
     applyInlineStyle(editor, (element, isInnerNode) => {
         element.style.fontFamily = isInnerNode ? '' : fontName;
+
+        if (parentNodes.indexOf(element.parentNode) === -1) {
```
JiuqingSong, line 20: This still can't sole the problem when LI is not the direct parent. e.g.
```html
<ol>
<li><div><span>aaa</span></div>
</li>
<ol>
```

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -7,6 +7,7 @@ import safeInstanceOf from '../utils/safeInstanceOf';
 import toArray from '../utils/toArray';
 import unwrap from '../utils/unwrap';
 import wrap from '../utils/wrap';
+import { default as setListItemStyle } from '../utils/setListItemStyle';
```
JiuqingSong, line 10: Just import setListItemStyle from ....

packages/roosterjs-editor-dom/lib/utils/setListItemStyle.ts
```diff
@@ -0,0 +1,58 @@
...
+        currentInlineElement = contentTraverser.currentInlineElement;
+        let currentNode = currentInlineElement.getContainerNode();
+
+        if (currentNode.nodeName != 'BR') {
+            if (currentNode.nodeName == '#text') {
```
JiuqingSong, line 44: no need to do this check. findClosestElementAncestor will do the check for you

packages/roosterjs-editor-api/lib/format/setFontName.ts
BryanValverdeU: Fixed this adding a new function to get the closest element ancestor of a specific tag.

packages/roosterjs-editor-dom/lib/utils/setListItemStyle.ts
BryanValverdeU: I think that if the inline element is a span, the findClosestElementAncestor would return the List Item, and we want to get the styles of the childs of the inline childrens. I think we only need to execute  findClosestElementAncestor  when the node it s text node

packages/roosterjs-editor-api/lib/format/setFontName.ts
BryanValverdeU: Also added a test case for this specific scenario

packages/roosterjs-editor-dom/lib/utils/setListItemStyle.ts
JiuqingSong: findClosestElementAncestor already does this for you. If you check here, it means we actually checked twice

packages/roosterjs-editor-api/lib/format/setFontSize.ts
```diff
@@ -1,4 +1,6 @@
 import applyInlineStyle from '../utils/applyInlineStyle';
+import { applyStyleToListItems } from '../utils/applyStyleToListItems';
+import { findClosestElementTypeAncestor } from 'roosterjs-editor-dom';
```
JiuqingSong, line 3: you can also use findClosestElementAncestor here. It has a "selector" parameter:

export default function findClosestElementAncestor(
    node: Node,
    root?: Node,
    selector?: string
): HTMLElement {

JiuqingSong: And these 3 files are almost the same. Consider add a common wrapper for them to do the duplicated part work

packages/roosterjs-editor-api/lib/utils/applyStyleToListItems.ts
```diff
@@ -0,0 +1,14 @@
+import { safeInstanceOf, setListItemStyle } from 'roosterjs-editor-dom';
+/**
+ * @internal
+ * Checks if the parent element is a List Item, if it is, apply the style elements to the list item
+ * @param parentNodes parentNodes to apply the style
+ * @param styles styles to apply to the List Item Element
+ */
+export function applyStyleToListItems(parentNodes: Node[], styles: string[]) {
```
JiuqingSong, line 8: Then this file can be part of the wrapper, no need to have a separate file

packages/roosterjs-editor-dom/lib/utils/findClosestElementTypeAncestor.ts
```diff
@@ -0,0 +1,25 @@
...
+import { NodeType } from 'roosterjs-editor-types';
+
+/**
+ * Find closest element ancestor of a specific tag element
+ * @param node Find ancestor start from this node
+ * @param root Root node where the search should stop at. The return value can never be this node
+ * @param tag Tag of the element to find.
+ * @returns An HTML element which is the closest ancestor with the provided Tag, returns null if not found and reached the root element
+ */
+export default function findClosestElementTypeAncestor(node: Node, root: Node, tag: string): Node {
```
JiuqingSong, line 11: this file is almost the same with findClosestElementAncestor, and findClosestElementAncestor can already cover this file. the selector parameter already work for tag name

packages/roosterjs-editor-api/lib/format/setFontName.ts
JiuqingSong: Now this check is not right. You need to check parentListItem instead of element.parentNode

packages/roosterjs-editor-dom/test/DomTestHelper.ts
```diff
@@ -113,3 +113,12 @@ export function itFirefoxOnly(
     const func = __karma__.config.browser == 'Chrome' ? xit : it;
     return func(expectation, assertion, timeout);
 }
+
+export function itChromeOnly(
```
Lufedi, line 117: I'd remove the `Only` word here just itChome

packages/roosterjs-editor-dom/lib/utils/setListItemStyle.ts
BryanValverdeU: Removing this check

packages/roosterjs-editor-api/lib/format/setFontSize.ts
BryanValverdeU: I see, I will remove the added function and use this instead.
Adding a wrapper too

packages/roosterjs-editor-dom/lib/utils/findClosestElementTypeAncestor.ts
BryanValverdeU: Removing this.

packages/roosterjs-editor-api/lib/format/setFontName.ts
```diff
@@ -8,11 +10,17 @@ import { IEditor } from 'roosterjs-editor-types';
...
+    const parentNodes: Node[] = [];
     fontName = fontName.trim();
+    const contentDiv = editor.getSelectedRegions()[0]?.rootNode;
     // The browser provided execCommand creates a HTML <font> tag with face attribute. <font> is not HTML5 standard
     // (http://www.w3schools.com/tags/tag_font.asp). Use applyInlineStyle which gives flexibility on applying inline style
     // for here, we use CSS font-family style
     applyInlineStyle(editor, (element, isInnerNode) => {
         element.style.fontFamily = isInnerNode ? '' : fontName;
+
+        setElementsToVerifyStyle(parentNodes, element, contentDiv, 'LI');
```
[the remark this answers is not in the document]
BryanValverdeU: Got it, will do that

packages/roosterjs-editor-api/lib/utils/applyStyleToListItems.ts
JiuqingSong: suggest move this function into applyListItemStyleWrap.ts and no need to export it.

JiuqingSong (approved):

PR #645, opened 2021-09-16
Keep image resposive, if not resized.

juliaroldi:
- Bugfix for when user select the image without change the size, image keep responsive.


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -504,6 +506,7 @@ export default class ImageEdit implements EditorPlugin {
                     rotateHandle
                 );
 
+                checkIftheImageWasResized(this.image, context);
```
JiuqingSong, line 509: Actually, when run into line 489 (or line 487 before your change), it means user has dragged resize handle and we can treat it as image resized.

You can add some new optional property to ImageEditInfo to save this resized state. Then only set image width/height when this value is true, or remove it when this is false. So no need to have additional "wasReset" or "wasResize" attribute

But if the image already has width/height initially, always set this value to true.

juliaroldi: I added this attribute in case of the image was deleted, so if the image is deleted, the attribute is deleted too, then I don't have to reset a property in case of the image is deleted.  Also, I'm not using the image size to check if it was resizes, because if the user changes the screen size, it will change the size of the image without resize plugin be called.

JiuqingSong: I'm not suggesting comparing image size. Instead, I mean we check if the resize handle is dragged by user, that will result in line 489 to be called.

Here's the flow:
- User drag  resize handle => 
- Resize.onDrag() is called and return true => 
- DragAndDropHelper call onSubmit() => 
- ImageEditPlugin.updateWrapper() is called with parameter "context" => 
- Line 489 is triggered

So that it is best place to remember that user has really "resized" the image. If user change screen size, since user didn't drag the resize handle, those code won't be triggered.

What is your concern of image is deleted?

juliaroldi: I changed the code to check if the image is resized after line 489. 
My concern is to use a property to store if the image was resized and even if delete the image, this property keep stores true as value.

JiuqingSong: If you store the value in ImageEditInfo, it will be serialized to the image. So when image is deleted, the serialized result is deleted as well.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -486,13 +488,18 @@ export default class ImageEdit implements EditorPlugin {
...
+                    const wasResized = this.image.getAttribute('wasResized');
                     doubleCheckResize(
                         this.editInfo,
                         this.options.preserveRatio,
                         clientWidth,
                         clientHeight
                     );
 
+                    if (!wasResized) {
+                        this.image.setAttribute('wasResized', 'true');
```
JiuqingSong, line 500: we already have ImageEditInfo object, so no need to add one more attribute to the image. You just need to add one more property to ImageEditInfo and save this result there.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/api/resetImage.ts
```diff
@@ -13,6 +13,7 @@ export default function resetImage(editor: IEditor, image: HTMLImageElement) {
         image.style.maxWidth = '100%';
         image.removeAttribute('width');
         image.removeAttribute('height');
+        image.setAttribute('wasReset', 'true');
```
JiuqingSong, line 16: If you use ImageEditInfo to save the result, you don't need to set anything here because next line (line 17) will delete the image edit info object so that image resized flag is also deleted.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
juliaroldi: Oh, I understood, I'll fix this.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -395,10 +398,8 @@ export default class ImageEdit implements EditorPlugin {
     private removeWrapper = (wrapper: HTMLElement) => {
         const parent = wrapper?.parentNode;
         const img = wrapper?.querySelector('img');
-
```
JiuqingSong, line 398: do not remove this empty line

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -486,13 +487,18 @@ export default class ImageEdit implements EditorPlugin {
                 if (context?.elementClass == ImageEditElementClass.ResizeHandle) {
                     const clientWidth = wrapper.clientWidth;
                     const clientHeight = wrapper.clientHeight;
+                    const wasResized = this.editInfo.wasResized;
```
JiuqingSong, line 490: No need to read original value, just always set to true

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -605,3 +611,18 @@ function updateHandleCursor(handles: HTMLElement[], angleRad: number) {
...
+/**
+ * Check if the image must be responsive
+ * @param img The current image.
+ * @param imgInfo the current edit state of the image
+ */
+function setImageResponsive(img: HTMLImageElement, imgInfo: ImageEditInfo) {
+    const wasResized = imgInfo.wasResized;
+    if (!wasResized) {
+        img.style.maxWidth = '100%';
+        img.style.height = 'auto';
```
JiuqingSong, line 624: what about width?

Also need to remove image.width and image.height.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -605,3 +611,18 @@ function updateHandleCursor(handles: HTMLElement[], angleRad: number) {
...
+    } else {
+        img.style.maxWidth = '';
```
JiuqingSong, line 626: initial

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
JiuqingSong: What about use "initial" instead of auto?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
```diff
@@ -55,8 +55,8 @@ export default function applyChange(
         newSrc = event.newSrc;
     }
 
-    if (newSrc == editInfo.src) {
-        // If newSrc is the same with original one, it means there is only size change, but no rotation, no cropping,
+    if (newSrc == editInfo.src && !editInfo.wasResized) {
```
JiuqingSong, line 58: Even if resized, you can still delete edit info. Because image will have width/height value. It can help us regenerate wasResized

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/editInfo.ts
```diff
@@ -46,6 +46,7 @@ function getInitialEditInfo(image: HTMLImageElement): ImageEditInfo {
         heightPx: image.clientHeight,
         naturalWidth: image.naturalWidth,
         naturalHeight: image.naturalHeight,
+        wasResized: false,
```
JiuqingSong, line 49: set to true if image has width or height. Including:
img.width
img.height
img.style.width
img.style.height
If one of them has meaningful value (not initial, not auto, ...), we should treat it as resized

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -280,6 +285,9 @@ export default class ImageEdit implements EditorPlugin {
                 this.removeWrapper(wrapper);
             }
 
+            // Keep Image responsive, if not resized
+            setImageResponsive(this.image, this.wasResized);
```
JiuqingSong, line 289: this logic should be part of applyChange(). We set image width/height there, so we should change it according to the value of wasResized.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -295,7 +303,16 @@ export default class ImageEdit implements EditorPlugin {
             this.image = image;
 
             // Get initial edit info
-            this.editInfo = getEditInfoFromImage(image);
+            if (this.editInfo) {
```
JiuqingSong, line 306: Isn't this always false?  We set editInfo to null in line 296.

You need to check if any of the following exists in the image when getEditInfoFromImage, and maybe still make wasResized as part of ImageEditInfo (sorry for the back and force idea change, but you can also argue if you also fount it makes more sense to put wasResized into ImageEditInfo):
- img.style.width
- img.style.height
- img.width
- img.height

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -605,3 +622,17 @@ function updateHandleCursor(handles: HTMLElement[], angleRad: number) {
...
+
+/**
+ * Check if the image must be responsive
+ * @param img The current image.
+ * @param imgInfo the current edit state of the image
+ */
+function setImageResponsive(img: HTMLImageElement, wasResized: boolean) {
+    if (!wasResized) {
+        img.style.maxWidth = '100%';
+        img.style.height = 'initial';
```
JiuqingSong, line 634: as I mentioned, you need to set all four values. For css values, you can set to initial, or even null. For HTML attributes, just delete them from the image.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -295,7 +303,16 @@ export default class ImageEdit implements EditorPlugin {
             this.image = image;
 
             // Get initial edit info
-            this.editInfo = getEditInfoFromImage(image);
+            if (this.editInfo) {
+                this.wasResized = true;
+            } else if (!this.editInfo && this.image.style.maxWidth === 'initial') {
```
JiuqingSong, line 308: do not check maxWidth, but check the four values above instead

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -296,6 +303,12 @@ export default class ImageEdit implements EditorPlugin {
 
             // Get initial edit info
             this.editInfo = getEditInfoFromImage(image);
+            if (this.image.height && this.image.width && this.image.style.maxWidth === '100%') {
```
JiuqingSong, line 306: if height/width exists, wasResized should be true
Also need to check style.width and style.height.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
```diff
@@ -73,9 +74,28 @@ export default function applyChange(
...
+ * @param wasResized the current resize state of the image
+ */
+function setImageResponsive(img: HTMLImageElement, wasResized: boolean) {
+    if (wasResized) {
+        img.style.maxWidth = 'initial';
+        img.removeAttribute('height');
+        img.removeAttribute('width');
+    } else {
+        img.style.maxWidth = '100%';
+        img.style.height = 'initial';
```
JiuqingSong, line 99: set width as well

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
```diff
@@ -73,9 +74,28 @@ export default function applyChange(
     image.width = targetWidth;
     image.height = targetHeight;
```
JiuqingSong, line 75: these 4 lines can be integrated into setImageResponsive()

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
```diff
@@ -73,9 +74,28 @@ export default function applyChange(
...
+
+/**
+ * Check if the image must be responsive
+ * @param img The current image.
+ * @param wasResized the current resize state of the image
+ */
+function setImageResponsive(img: HTMLImageElement, wasResized: boolean) {
+    if (wasResized) {
+        img.style.maxWidth = 'initial';
+        img.removeAttribute('height');
```
JiuqingSong, line 95: make sure this won't cause any exception, test in all browsers (Chrome, Safari, Firefox, Edge, IE)

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
juliaroldi: I cannot remove the width/height/style.width/style.height, because if they are remove will be not possible to use resize by the percentage. The image has to have  width/height/style.width/style.height, even if not resized.

JiuqingSong: that is not true.  you can try insert a new image, at beginning it doesn't have these attributes/CSS at all, but you can still resize by percentage.Resizing by percentage is using image's naturalWidth/Height which is not an HTML attribute.

juliaroldi: `export default function resizeByPercentage(
    editor: IEditor,
    image: HTMLImageElement,
    percentage: number,
    minWidth: number,
    minHeight: number
) {
    const editInfo = getEditInfoFromImage(image);

    if (!isResizedTo(image, percentage)) {
        loadImage(image, editInfo.src, () => {
            if (!editor.isDisposed() && editor.contains(image)) {
                const lastSrc = image.src;
                const { width, height } = getTargetSizeByPercentage(editInfo, percentage);
                editInfo.widthPx = Math.max(width, minWidth);
                editInfo.heightPx = Math.max(height, minHeight);

                    applyChange(editor, image, editInfo, lastSrc);
                }, ChangeSource.ImageResize);
            }
        });
    }
}
`

resizeByPercentage uses applyChanges, so applyChanges needs to set the value o width/height/style;width/style.height, even if it is not resized by the resizer handles

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
juliaroldi: I tested in all browsers.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
```diff
@@ -68,14 +69,40 @@ export default function applyChange(
...
         editInfo.heightPx != initEditInfo.heightPx
     );
 }
+
+/**
+ * Check if the image must be responsive
+ * @param img The current image.
+ * @param wasResized the current resize state of the image
+ */
+function setImageResponsive(
```
JiuqingSong, line 88: Now this function is not only for responsive, so better name is setImageSize

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -296,6 +303,18 @@ export default class ImageEdit implements EditorPlugin {
 
             // Get initial edit info
             this.editInfo = getEditInfoFromImage(image);
+            if (
+                this.image.style.maxWidth === 'initial' &&
+                this.image.style.height !== 'initial' &&
```
JiuqingSong, line 308: here you check height !== initial, but for style.width you only do null check. Any reason not doing the same check?

Actually checking for initial is not always accurate, its value can also be "auto" or "inherit".

I think what you can try is to parse the value to be a number, then check if the result is a valid number, but not NaN:

```ts
function isFixedNumberValue(v: string) {
  return parseInt(v) !== NaN;
}
```

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -605,3 +616,27 @@ function updateHandleCursor(handles: HTMLElement[], angleRad: number) {
...
+ * @returns if the user resized the image, returns true, otherwise, returns false
+ */
+function checkIfImageWasResized(image: HTMLImageElement): boolean {
+    const { width, height, style } = image;
+    if (
+        style.maxWidth === 'initial' &&
+        isFixedNumberValue(style.height) &&
+        isFixedNumberValue(style.width) &&
+        width &&
+        height
```
JiuqingSong, line 632: 1. we also need to call isFixedNumberValue for width and height
2. These 4 should be connected using "or". Any of them exists, we need to treat it as resized

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
```diff
@@ -68,14 +69,36 @@ export default function applyChange(
...
+        image.width = targetWidth;
+        image.height = targetHeight;
+        image.style.width = targetWidth + 'px';
+        image.style.height = targetHeight + 'px';
+    } else {
+        image.style.maxWidth = '100%';
+        image.style.height = 'initial';
+        image.removeAttribute('height');
+        image.removeAttribute('width');
+        image.style.width = '';
```
JiuqingSong, line 102: why set style.height to "initial" but set style.width to "" ? They should be the same.

JiuqingSong (approved):

juliaroldi:
Bugfix for [Bug 12701118](https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/127018/)

PR #646, opened 2021-09-17
Shadow DOM entity

JiuqingSong:
This is the change to introduce Shadow DOM entity. Shadow DOM entity is the new functionality of entity API. It allows hydrate an entity under a shadow DOM root, and benefit React Entity, undo snapshot, memory management.

The new API is fully compatible with existing API. Existing code based on Entity API can still work.

To use shadow DOM entity, just need to handle EntityOperation event (event.operation==EntityOperation.NewEntity), and put the hydrated content under event.contentForShadowEntity. Then Entity API will mount it and do memory management.

Two new EntityOperation enum values are added:
- AddShadowRoot triggers when a new shadowRoot is added into entity
- RemoveShadowRoot triggers when an existing shadowRoot is removed

### DIFF
--- a/packages/roosterjs-editor-dom/test/DomTestHelper.ts
+++ b/packages/roosterjs-editor-dom/test/DomTestHelper.ts
@@ -113,3 +113,12 @@ export function itFirefoxOnly(
     const func = __karma__.config.browser == 'Chrome' ? xit : it;
     return func(expectation, assertion, timeout);
 }
+
+export function itChromeOnly(
+    expectation: string,
+    assertion?: jasmine.ImplementationCallback,
+    timeout?: number
+) {
+    const func = __karma__.config.browser == 'Chrome' ? it : xit;
+    return func(expectation, assertion, timeout);
+}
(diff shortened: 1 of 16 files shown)


packages/roosterjs-editor-dom/test/DomTestHelper.ts
```diff
@@ -113,3 +113,12 @@ export function itFirefoxOnly(
     const func = __karma__.config.browser == 'Chrome' ? xit : it;
     return func(expectation, assertion, timeout);
 }
+
+export function itChromeOnly(
```
romanisa, line 117: nit: possible typo? isChromeOnly?

BryanValverdeU (approved):

packages/roosterjs-editor-dom/test/DomTestHelper.ts
JiuqingSong: no, "it" is the function name of running a test case.

PR #653, opened 2021-09-28
Fix #652

### LINKED ISSUE #652: Start new list before another numbering list makes the list number incorrect
listchain

JiuqingSong:
#652 We should check the start number, if it is a new list (start from 1), do not change numbers of existing lists.

### DIFF
--- a/packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
+++ b/packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
@@ -1141,7 +1141,7 @@ describe('Table Resizer/Inserter tests', () => {
         resizeTAbleWiderTest(2);
     });
 
-    it('resizes the table to be narrower correctly with default table', () => {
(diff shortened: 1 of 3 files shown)


flyingbee2012 (approved):

packages/roosterjs-editor-plugins/test/TableResize/tableResizeTest.ts
```diff
@@ -1141,7 +1141,7 @@ describe('Table Resizer/Inserter tests', () => {
         resizeTAbleWiderTest(2);
     });
 
-    it('resizes the table to be narrower correctly with default table', () => {
+    xit('resizes the table to be narrower correctly with default table', () => {
```
flyingbee2012, line 1144: that's weird...the same test was working yesterday but not today

flyingbee2012 (approved):

BryanValverdeU (approved):

PR #654, opened 2021-09-30
fix image resized when copied

juliaroldi:
- Set images that are not inserted as resized. 
- Only keep responsive images that were inserted as inline image.  

copiedImage2

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
@@ -619,22 +619,17 @@ function updateHandleCursor(handles: HTMLElement[], angleRad: number) {
 
 /**
  * Check if the current image was resized by the user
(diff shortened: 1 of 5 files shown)


JiuqingSong:
Please also mark those unstable test cases as "xit" so it won't impact further tests.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -625,7 +625,7 @@ function updateHandleCursor(handles: HTMLElement[], angleRad: number) {
 function checkIfImageWasResized(image: HTMLImageElement): boolean {
     const { width, height, style } = image;
     if (
-        style.maxWidth === 'initial' &&
+        style.maxWidth !== '100%' &&
```
JiuqingSong, line 628: What about other values, such as '90%' or '100px'?

JiuqingSong (approved):

PR #658, opened 2021-09-30
Fix Paste Gifs when HTML only have one IMG Tag.

BryanValverdeU:
This is a fix for the a past PR #620 in some scenarios.

When pasting a gif from a another page, the body of the document sometime have empty #text Nodes (\r\n), which do not have a tag, for example:

image

We did not verify if the the current node had a Tag, so multiple elements with empty string were added to the htmlFirstLevelChildTags, clipboard property: 

image

Since we have more that one element in the htmlFirstLevelChildTags property, we did not get to the function to process when the body only have one IMG  #Tag. 
packages/roosterjs-editor-plugins/lib/plugins/Paste/imageConverter/convertPasteContentForSingleImage.ts

-----
Code Change:

Verify that the nodes have a tag before adding them to clipboard.htmlFirstLevelChildTags  property

Added a Test case for this scenario


packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
```diff
@@ -67,9 +67,12 @@ export const createPasteFragment: CreatePasteFragment = (
         }, event.htmlAttributes);
 
         clipboardData.htmlFirstLevelChildTags = [];
-        doc?.body.childNodes.forEach(node =>
-            clipboardData.htmlFirstLevelChildTags.push(getTagOfNode(node))
-        );
+        doc?.body.childNodes.forEach(node => {
+            const nodeTag = getTagOfNode(node);
```
JiuqingSong, line 71: what if there is a valid text node as well?

suggest do doc?.body.normalize() then empty text node will be removed.

BryanValverdeU: It seems like normalize did not remove nodes with \r\n, I added a replace statement to remove this characters.
Also added a test case when the clipboard data have textNodes with Value and a IMG Tag

JiuqingSong (approved):

PR #683, opened 2021-10-29
8.11.2 Fix a bug for Safari

JiuqingSong:
In 8.11.1 there is a change to remove webkitUserSelect style when init. This causes sometimes use can't type in the editor in Safari. So let's add it back


BryanValverdeU (approved):

JiuqingSong:
#1145

PR #687, opened 2021-11-08
Table Cell Selection

BryanValverdeU:
This PR is just to demo a possible approach to table selection in Rooster, also added a feature to apply  background color 

Apply a background color to table cells when it is selected.

- Support selection with keyboard and also mousemove event
MouseMoveSelection


packages/roosterjs-editor-api/lib/table/applyCellShading.ts
```diff
@@ -0,0 +1,24 @@
+import { ChangeSource, IEditor, ModeIndependentColor } from 'roosterjs-editor-types';
+import { VTable } from 'roosterjs-editor-dom';
+
+/**
+ * Apply cell shading on a vTable selection
```
JiuqingSong, line 5: Is this function for cell or the whole table?

packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
```diff
@@ -144,6 +150,7 @@ export default class DOMEventPlugin implements PluginWithState<DOMEventPluginSta
 
     private onFocus = () => {
         this.editor.select(this.state.selectionRange);
+        setSelectedTableCells(this.editor);
```
JiuqingSong, line 153: Normally, code from roosterjs-editor-dom package should not know anything about editor.

This looks should be something done by a core plugin, but not a DOM level API

packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
```diff
@@ -161,6 +168,7 @@ export default class DOMEventPlugin implements PluginWithState<DOMEventPluginSta
     private cacheSelection = () => {
         if (!this.state.selectionRange) {
             this.state.selectionRange = this.editor.getSelectionRange(false /*tryGetFromCache*/);
+            clearSelectedTableCells(this.editor, false);
```
JiuqingSong, line 171: same here

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -244,6 +245,7 @@ export default class Editor implements IEditor {
      * @returns HTML string representing current editor content
      */
     public getContent(mode: GetContentMode = GetContentMode.CleanHTML): string {
+        clearSelectedTableCells(this);
```
JiuqingSong, line 248: Do not clear selection here. You need to do this in a plugin to handle ExtractContentWithDOM event

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -62,6 +66,7 @@ export default class VTable {
                                 spanAbove: rowSpan > 0,
                                 width: hasTd ? rect.width : undefined,
                                 height: hasTd ? rect.height : undefined,
+                                selected: !!td.classList.contains(TABLE_CELL_SELECTED_CLASS),
```
JiuqingSong, line 69: I suggest we don't put selected into VCell. Because when the following result is illegal logically but is legal to this data structure:
+---------+----------+
| selected | not         |
+---------+---------+
| not        | selected |
+--------+----------+

Instead of letting VCell know the selection, we can put the selection info into VTable only. say:
 selectedRanges: [[0,0],[1,1]].

We just need to remember the selected cell indexes, then we can easily get all selected cells, or check if a cell is selected.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -404,6 +409,96 @@ export default class VTable {
         return this.getTd(this.row, this.col);
     }
 
+    highlightSelection(start: HTMLTableCellElement, end: HTMLTableCellElement) {
```
JiuqingSong, line 412: same here, use table cell indexes (row, col) instead of the actual TD element.

Because table cell can be merged. Consider:
+--------+---------+---------+
|  cell 1  |  cell 2   |  cell 3 |
+-------+---------+--------+
| cell 4  | cell 5 merge 6 |
+-------+---------------+

Since cell 5 and 6 are merged, they share the same TD. when we specify that TD, it is hard to tell if cell 3 should be selected.

JiuqingSong: The top benefit of using VTable is to allow you forget about TD, but always use index. TD is only used when writeBack() is called.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -404,6 +409,96 @@ export default class VTable {
...
+        return selectedCells;
+    }
+
+    selectAll() {
+        for (let indexY = 0; indexY < this.cells.length; indexY++) {
+            for (let indexX = 0; indexX < this.cells[indexY].length; indexX++) {
+                let element = this.cells[indexY][indexX].td as HTMLElement;
+
+                if (element) {
+                    element.style.backgroundColor = 'rgb(9, 109, 202)';
```
JiuqingSong, line 495: Try not hard code color, especially for dark mode.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -517,6 +612,34 @@ export default class VTable {
...
+        }
+
+        const handler = (cell: VCell) => {
+            setColor(cell.td, backgroundColor, true, isInDarkMode);
+
+            const colorString = typeof backgroundColor === 'string' ? backgroundColor.trim() : '';
+            const modeIndependentColor =
+                typeof backgroundColor === 'string' ? null : backgroundColor;
+
+            cell.td.dataset[TEMP_BACKGROUND_COLOR] =
```
JiuqingSong, line 628: not every cell has td

packages/roosterjs-editor-dom/lib/utils/clearSelectedTableCells.ts
```diff
@@ -0,0 +1,109 @@
...
+
+const TABLE_CELL_SELECTED_CLASS = 'TableCellSelected';
+const TEMP_BACKGROUND_COLOR = 'tempBackgroundColor';
+const ON_FOCUS_CACHE = 'onFocusCache';
+const SELECTION_COLOR_OPACITY = 0.7;
+/**
+ * Remove the selected style of the cells
+ * @param editor Editor Instance
+ */
+export function clearSelectedTableCells(editor: IEditor, shouldRemoveClass: boolean = true) {
```
JiuqingSong, line 12: Use VTable to do this.

packages/roosterjs-editor-dom/lib/utils/clearSelectedTableCells.ts
```diff
@@ -0,0 +1,109 @@
...
+    const color = getColor(colorString);
+
+    if (color) {
+        return color;
+    }
+
+    return '';
+}
+
+function getColor(colorString: string, prefix: string = 'rgb') {
```
JiuqingSong, line 98: Why need to handle the color values? We should just use whatever it has, without any change

packages/roosterjs-editor-dom/lib/utils/clearSelectedTableCells.ts
JiuqingSong: And do not let code from dom package know editor

packages/roosterjs-editor-dom/lib/utils/highlightTableSelection.ts
```diff
@@ -0,0 +1,107 @@
...
+import getTagOfNode from './getTagOfNode';
+import safeInstanceOf from './safeInstanceOf';
+import VTable from '../table/VTable';
+import { IContentTraverser } from 'roosterjs-editor-types';
+
+/**
+ * Highlight the table cells when are in the selection range
+ * @param selection Wether a Selection from the document when browser is Firefox or a Content Traverser
+ */
+export const highlightTableSelection = (selection: Selection | IContentTraverser) => {
```
JiuqingSong, line 11: do this in VTable

packages/roosterjs-editor-dom/lib/utils/highlightTableSelection.ts
```diff
@@ -0,0 +1,107 @@
...
+    let firstTDSelected: HTMLTableCellElement;
+    let lastTDSelected: HTMLTableCellElement;
+    let table: HTMLTableElement;
+
+    if (selection instanceof Selection) {
```
JiuqingSong, line 16: instanceof won't work correctly in projection popout.

Do not use suggest not use Selection as parameter

packages/roosterjs-editor-dom/lib/utils/highlightTableSelection.ts
```diff
@@ -0,0 +1,107 @@
...
+        for (let index = 0; index < selection.rangeCount; index++) {
+            const range = selection.getRangeAt(index);
+            const container = range.startContainer;
+            if (getTagOfNode(container) == 'TR') {
+                let element = container.childNodes[range.startOffset] as HTMLElement;
```
JiuqingSong, line 21: VTable has utils to help you find correct TD.

And do not check tag name "TD", it can also be "TH"

packages/roosterjs-editor-dom/lib/utils/highlightTableSelection.ts
```diff
@@ -0,0 +1,107 @@
...
+                        break;
+                    }
+                } else {
+                    if (table) {
+                        setTableSelectedRange(table, firstTDSelected, lastTDSelected);
+                    }
+                }
+            }
+        }
+    } else {
```
JiuqingSong, line 51: I didn't see anything shared between "if" and "else" in this function. They should be two separated functions

packages/roosterjs-editor-plugins/lib/plugins/TableSelection/TableSelection.ts
```diff
@@ -0,0 +1,195 @@
...
+    Browser,
+    clearSelectedTableCells,
+    highlightTableSelection,
+    setTableSelectedRange,
+} from 'roosterjs-editor-dom';
+
+/**
+ * TableSelectionPlugin help highlight table cells
+ */
+export default class TableSelectionPlugin implements EditorPlugin {
```
JiuqingSong, line 12: Consider make this a core plugin, because the table selection is also something editor core should know

packages/roosterjs-editor-plugins/lib/plugins/TableSelection/TableSelection.ts
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * Initialize this plugin. This should only be called from Editor
+     * @param editor Editor instance
+     */
+    initialize(editor: IEditor) {
+        this.editor = editor;
+        console.log(this.editor.getScrollContainer());
+        setTimeout(() => {
+            const styleElement = document.createElement('style');
+            styleElement.id = 'Style4TableSelection';
```
JiuqingSong, line 34: Consider not adding style, as long as there is other solution

packages/roosterjs-editor-plugins/lib/plugins/TableSelection/TableSelection.ts
```diff
@@ -0,0 +1,195 @@
...
+
+            let sheet = styleElement.sheet;
+            // sheet.insertRule(
+            //     'td.TableCellSelected *::selection { background-color: red !important;}'
+            // );
+            // sheet.insertRule(
+            //     'td.TableCellSelected::selection { background-color: red !important;}'
+            // );
+
+            sheet.insertRule(
```
JiuqingSong, line 45: As I said, when select table sell, do not let browser handle selection. So we just need to call event.preventDefault() to disable it. Then we know which cells mouse is over and we use our own data structure to manage selection.

packages/roosterjs-editor-plugins/lib/plugins/TableSelection/TableSelection.ts
```diff
@@ -0,0 +1,195 @@
...
+        }, 500);
+    }
+
+    /**
+     * Dispose this plugin
+     */
+    dispose() {
+        this.removeMouseUpEventListener();
+        this.editor = null;
+        document
```
JiuqingSong, line 61: This will be an issue when:
1. projection popout
2. multiple instances of editor

JiuqingSong: Suggest editor only manage things inside editor, but do not touch global window/document if we can

packages/roosterjs-editor-plugins/lib/plugins/TableSelection/TableSelection.ts
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /**
+     * Handle events triggered from editor
+     * @param event PluginEvent object
+     */
+    onPluginEvent(event: PluginEvent) {
+        if (event.eventType == PluginEventType.BeforeSetContent) {
+            this.clearTableCellSelection(true /** forceClear */);
+            this.editor.select(this.editor.getElementAtCursor('td'));
```
JiuqingSong, line 73: why?

JiuqingSong: If there is a reason, add some comment to explain

packages/roosterjs-editor-types/lib/interface/VCell.ts
```diff
@@ -26,4 +26,9 @@ export default interface VCell {
      * The height in pixel of the actual td (including border and padding)
      */
     height?: number;
+
+    /**
+     * Selected
+     */
+    selected?: boolean;
```
JiuqingSong, line 33: As I said, do not let VCell know selection

packages/roosterjs-editor-api/lib/table/applyCellShading.ts
BryanValverdeU: Only cell and if there is a selection of cells

packages/roosterjs-editor-plugins/lib/plugins/TableSelection/TableSelection.ts
BryanValverdeU: On what element should we add the event listener?
I think the event selection change, on the document element it is not cancellable.
Also adding a select event listener on the elements is not cancellable, 
https://developer.mozilla.org/en-US/docs/Web/API/Element/select_event 

I tried adding prevent default on the mouse move event and it is not preventing the selection.

I also tried the selectstart event, but if the user is starting the selection outside of the table it is not working.

JiuqingSong: table selection

I did a little change on your branch:

image

PR #688, opened 2021-11-11
Predefined Table Styles

juliaroldi:
TO DO:
Test cases
Add support to banded rows and banded columns at the same time


packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -99,15 +99,16 @@ export default class VTable {
      * Apply the given table format to this virtual table
      * @param format Table format to apply
      */
-    applyFormat(format: Partial<TableFormat>) {
+    applyFormat(
+        format: Partial<TableFormat>,
```
JiuqingSong, line 103: Can we put everything into TableFormat, so no need to change the parameter here?

I suppose a tableFormat object should be self-contained and has everything we need.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -118,6 +119,102 @@ export default class VTable {
...
+     * @returns
+     */
+    setFirstColumnFormat(format: Partial<TableFormat>) {
+        if (!format) {
+            return;
+        }
+        this.cells.forEach(row => {
+            row.filter(cell => row.indexOf(cell) === 0).forEach(cell => {
+                if (row.indexOf(cell) === 0) {
+                    cell.td.style.borderTop = getBorderStyle(format.topBorderColor);
```
JiuqingSong, line 180: not every cell has TD

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -118,6 +119,102 @@ export default class VTable {
...
+            );
+        });
+    }
+
+    /**
+     * Set borders sides to table
+     * @param format
+     * @returns
+     */
+    setSideBorderFormat(format: Partial<TableFormat>) {
```
JiuqingSong, line 151: Use one single function to handle all formats.

BryanValverdeU:
Good job, remember to change the target branch to the https://github.com/microsoft/roosterjs/tree/tableStylingFeature

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -104,22 +105,193 @@ export default class VTable {
...
+            row.filter(cell => cell.td).forEach(
+                cell => (cell.td.style.backgroundColor = format.bgColorEven)
+            );
+        });
+        oddRows.forEach(row => {
+            row.filter(cell => cell.td).forEach(
+                cell => (cell.td.style.backgroundColor = format.bgColorOdd)
+            );
+        });
+    }
```
BryanValverdeU, line 155: nit: You could do something like this so only traverse the cells once instead of 3 times.

image

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -104,22 +105,193 @@ export default class VTable {
...
+     */
+    private setTableColumnsColor(format: Partial<TableFormat>) {
+        if (!format.bandedColumns) {
+            return;
+        }
+        this.cells.forEach(row => {
+            row.filter(cell => cell.td).forEach(cell => {
+                if (row.indexOf(cell) % 2 === 0) {
+                    cell.td.style.backgroundColor = format.bgColumnColorEven;
+                } else {
```
BryanValverdeU, line 169: No need to filter, just check if the cell have a td inside of the foreach. Using the filter you would traverse the table twice, one to get the filtered array and another to apply the styles.
Also you could add an additional parameter to the foreach like this (cell, index) and use the index inside instead of getting the index with indexOf()

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -104,22 +105,193 @@ export default class VTable {
...
+    }
+
+    /**
+     * Apply custom design to the Header Row
+     * @param format
+     * @returns
+     */
+    private setHeaderRowFormat(format: Partial<TableFormat>) {
+        if (!format.headerRow) {
+            format.headerRowColor = format.bgColor;
```
BryanValverdeU, line 286: If no header row, you should just return and don't apply any styles or you would override the first column style

Example:
image

packages/roosterjs-editor-dom/lib/utils/tableInfo.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Save the format info of a table
+ * @param table The table the info will be saved
+ * @param format The format of the table
+ */
+export function saveTableInfo(table: HTMLTableElement, format: Partial<TableFormat>) {
+    if (table) {
+        table.dataset[TABLE_STYLE_INFO] = JSON.stringify(format);
+    }
+}
+
```
BryanValverdeU, line 15: I think you could call this in the applyFormat in VTable, so there is no need to call separately it every time when changing the table style.

demo/scripts/controls/sidePane/apiPlayground/vtable/VTablePane.tsx
```diff
@@ -1,364 +1,776 @@
...
+                            </tr>
+                            <tr>
+                                <th colSpan={2}>Format Table</th>
+                            </tr>
+                            <tr>
+                                <td>State:</td>
+                                <td>{this.renderSetHeaderRowButton(editor)}</td>
+                                <td>{this.renderSetFirstColumnButton(editor)}</td>
+                                <td>{this.renderSetBandedColumnButton(editor)}</td>
+                                <td>{this.renderSetBandedRowButton(editor)}</td>
```
BryanValverdeU, line 398: I think this 4 buttons should be in the same td

demo/scripts/controls/sidePane/apiPlayground/vtable/VTablePane.tsx
```diff
@@ -1,364 +1,776 @@
...
+        firstColumn: firstColumn,
+        tableBorderFormat: borderFormat,
+    };
+}
+
+function setHeaderRow(table: HTMLTableElement): Partial<TableFormat> {
+    const format = getTableFormatInfo(table);
+    format.headerRow = !format.headerRow;
+    if (format.headerRow) {
+        format.headerRowColor = format.bottomBorderColor;
```
BryanValverdeU, line 696: I think you should add the header color always when creating the table format.
So you don't need to modify the TableFormat here. and just toggle the 'header row' state.

If you dont want to apply the header row when setting a predefined table styles, set the headerRow state by default as false.

packages/roosterjs-editor-dom/lib/utils/tableInfo.ts
BryanValverdeU: Also check if the format param have value.
Otherwise a undefined would be added to the dataset

demo/scripts/controls/sidePane/apiPlayground/vtable/VTablePane.tsx
```diff
@@ -1,364 +1,776 @@
...
+            true /** bandedRows */,
+            false /** bandedColumns */,
+            true /** headerRow */,
+            false /** firstColumn */,
+            TableBorderFormat.onlyMiddleBorders /** tableBorderFormat */,
+            lightColor /** bgColorEven */,
+            null /** bgColorOdd */,
+            null /** bgColumnColorOdd */,
+            null /** bgColumnColorEven */,
+            null /** headerRowColor */
```
BryanValverdeU, line 32: I would always add values here. this values wont be applied to the table format because the banded columns is false

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -121,22 +131,221 @@ export default class VTable {
             return;
```
BryanValverdeU, line 131: When using adding columns and writeback, on a table that I applied styles, the header row height is decreased and the styles removed.
I think we should call the applyformat inside of the writeback if the table have tableInfo
TABL1
o

PR #690, opened 2021-11-15
Update Format APIs to work with Table Cell Selection

BryanValverdeU:


packages/roosterjs-editor-api/lib/utils/applyInlineStyle.ts
```diff
@@ -50,18 +57,59 @@ export default function applyInlineStyle(
...
+            const selectedCells = Array.from(getSelectedTableCells(editor));
+
             let contentTraverser = editor.getSelectionTraverser();
             let inlineElement = contentTraverser && contentTraverser.currentInlineElement;
+
             while (inlineElement) {
                 let nextInlineElement = contentTraverser.getNextInlineElement();
                 inlineElement.applyStyle((element, isInnerNode) => {
-                    callback(element, isInnerNode);
+                    const closestTD =
```
JiuqingSong, line 69: too complex.

You just need to:

1. check if there is table selection. If so, get all selected cells (TD elements)
2. for each TD, editor.select(td);
3. Go with original code
4. if no table selection, go with original code.

packages/roosterjs-editor-api/lib/utils/blockFormat.ts
```diff
@@ -18,7 +23,17 @@ export default function blockFormat(
     editor.focus();
     editor.addUndoSnapshot((start, end) => {
         if (!beforeRunCallback || beforeRunCallback()) {
-            const regions = editor.getSelectedRegions();
+            const regions = editor.getSelectedRegions().filter(region => {
```
JiuqingSong, line 26: when there is table selection, it means we can already treat each select table cell as a region. So it should be come simpler here.

packages/roosterjs-editor-api/lib/utils/execCommand.ts
```diff
@@ -32,6 +44,61 @@ export default function execCommand(editor: IEditor, command: DocumentCommand) {
             });
         }
     } else {
-        editor.addUndoSnapshot(formatter, ChangeSource.Format);
+        const handler = (start: NodePosition, end: NodePosition) => {
```
JiuqingSong, line 47: same here. Just need to for each selected table cells, call editor.select(td), then go with original code

PR #701, opened 2021-12-09
Fix Narrator not announcing on delete at the last char

BryanValverdeU:
Issue:
The Narrator is not announcing the deleted charater when on the last element.

Before the change: (Notice how the narrator is not announcing the deleted charater)
Sppech on deleting B4

After
Sppech on deleting after

Code Changes:

Added a function to announce the deleted charater, only when is at the end of the line.
Also a new content edit feature to announce the message to the narrator.

This seems to be a browser issue. I tried to repro the behavior and I was able to repro an a bare HTML Div contenteditable.

### DIFF
--- a/packages/roosterjs-editor-dom/lib/utils/narratorMessage.ts
+++ b/packages/roosterjs-editor-dom/lib/utils/narratorMessage.ts
@@ -1,0 +1,54 @@
+import { ContentPosition, IEditor } from 'roosterjs-editor-types';
+
+const MESSAGE_CONTAINER_ID = 'roosterjs_NarratorMessage';
+let MESSAGE_JOB: number;
+/**
+ * Makes the narrator announce a message
+ * @param file The file to read
+ * @param message Message to announce
+ * @param priority priority (non mandatory): "polite" (by default) or "assertive"
+ */
+export default function narratorMessage(
+    editor: IEditor,
+    message: string,
+    priority: 'polite' | 'assertive' = 'polite'
+) {
+    const editorDocument = editor.getDocument() || document;
+    const currentWindow = editorDocument.defaultView || window;
+
+    currentWindow.clearTimeout(MESSAGE_JOB);
+
+    const el = document.createElement('div');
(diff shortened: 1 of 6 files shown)


packages/roosterjs-editor-dom/lib/utils/narratorMessage.ts
```diff
@@ -0,0 +1,54 @@
+import { ContentPosition, IEditor } from 'roosterjs-editor-types';
+
+const MESSAGE_CONTAINER_ID = 'roosterjs_NarratorMessage';
+let MESSAGE_JOB: number;
+/**
+ * Makes the narrator announce a message
+ * @param file The file to read
```
Lufedi, line 7: wrong parameter

packages/roosterjs-editor-dom/lib/utils/narratorMessage.ts
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    container.appendChild(el);
+
+    MESSAGE_JOB = currentWindow.setTimeout(function () {
+        editorDocument.getElementById(id).textContent = message;
+    }, 200);
+
+    currentWindow.setTimeout(function () {
+        const element = editorDocument.getElementById(id);
```
Lufedi, line 50: Is it possible to move this to the  MESSAGE_JOB setTimeOut? or why do we need to have two ´setTimeouts´ ?

packages/roosterjs-editor-types/lib/interface/ContentEditFeatureSettings.ts
```diff
@@ -201,6 +201,33 @@ export interface TableFeatureSettings {
     upDownInTable: boolean;
 }
 
+/**
+ * Settings for table features
+ */
+export interface TableFeatureSettings {
```
Lufedi, line 207: I this this is not part of this PR

BryanValverdeU:
Seems like this is an error in Browser. Closing this PR for now.

Issue #698, opened 2021-12-07
Add an option for restoring selection after cut/copy


Rain-Zheng:
**Describe the bug**
Currently roosterjs restores selection after handling cut or copy events. However, the default behavior of Android WebView is to clear the selection after cut or copy. Can we add an option for the selection restoration in `CopyPastePlugin`? so that we can disable it on Android app.

**To Reproduce**
1. Open roosterjs demo page on Android device.
2. Input some texts and long press to select them. 
3. Click the 'Copy' option in context menu.
4. See error.

**Expected behavior**
The selection should be cleared after copy on Android device.

**Screenshots**
Default behavior of Android WebView, the selection is cleared:

https://user-images.githubusercontent.com/67583056/145038474-c950a37e-ec0f-45da-948c-eab8a7470ce2.mp4

The behavior of roosterjs, the selection is restored:

https://user-images.githubusercontent.com/67583056/145038512-7df0415e-086c-41c5-bb6c-4c9e3034c390.mp4

BryanValverdeU:
Thank you for reporting I will check it

BryanValverdeU:
This has been solved, will release the version next Friday.

PR #702, opened 2021-12-09
Collapse selection after performing Copy when Android

BryanValverdeU:
#698 

Collapse the selection after copy in Android, so the selection does not persist as explained in the issue above

### DIFF
--- a/packages/roosterjs-editor-dom/lib/utils/Browser.ts
+++ b/packages/roosterjs-editor-dom/lib/utils/Browser.ts
@@ -61,3 +61,16 @@ export function getBrowserInfo(userAgent: string, appVersion: string): BrowserIn
 export const Browser = window
     ? getBrowserInfo(window.navigator.userAgent, window.navigator.appVersion)
     : {};
+
+/**
+ * Obtains wether the current device is Android
+ * @param window current window
+ * @returns true if is android, otherwise false
+ */
+export const isAndroid = (window: Window) => {
+    var userAgent = window.navigator.userAgent || navigator.vendor;
+    if (/android/i.test(userAgent)) {
+        return true;
+    }
+    return false;
+};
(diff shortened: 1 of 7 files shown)


packages/roosterjs-editor-dom/lib/utils/Browser.ts
```diff
@@ -61,3 +61,16 @@ export function getBrowserInfo(userAgent: string, appVersion: string): BrowserIn
...
     : {};
+
+/**
+ * Obtains wether the current device is Android
+ * @param window current window
+ * @returns true if is android, otherwise false
+ */
+export const isAndroid = (window: Window) => {
+    var userAgent = window.navigator.userAgent || navigator.vendor;
+    if (/android/i.test(userAgent)) {
```
Lufedi, line 72: Change this to  ´return /android/i.test(userAgent)´

Lufedi (approved):

packages/roosterjs-editor-dom/lib/utils/Browser.ts
```diff
@@ -61,3 +61,13 @@ export function getBrowserInfo(userAgent: string, appVersion: string): BrowserIn
 export const Browser = window
     ? getBrowserInfo(window.navigator.userAgent, window.navigator.appVersion)
     : {};
+
+/**
+ * Obtains wether the current device is Android
+ * @param window current window
+ * @returns true if is android, otherwise false
+ */
+export const isAndroid = (window: Window) => {
```
JiuqingSong, line 70: this should be part of Browser object, and do not export it separately

JiuqingSong: See how we define Browser.isMac , isWindows, ...

JiuqingSong (changes requested):

JiuqingSong (approved):

PR #715, opened 2022-01-07
8.12.0 Fix typescript version issue

JiuqingSong:
When a customer project uses a different version of typescript than roosterjs, it is possible that there are some unrecognized types in TargetWindowBase.g.ts. 

To fix it:
1. Remove all types that not quite often used
2. Allow passing in a type parameter in safeInstanceOf() so that those removed types still have a chance to be used
3. Do not auto generate TargetWindowBase class any more.


flyingbee2012 (approved):

JiuqingSong:
#1145

PR #719, opened 2022-01-14
Support scale in ImageEdit plugin

JiuqingSong:
- Adding a new parameter in ImageEdit plugin and DragAndDropHelper to help transform the size to make the size calculation working correctly when editor is put under a scaled container
- Fix some potential bugs in editor event system for shadow edit that I found when debug the image resize thing
- Fix context menu code for scaled container
- Add related UI in demo site

imageresize


demo/scripts/controls/editor/Editor.tsx
```diff
@@ -92,6 +93,7 @@ export default function Editor(props: EditorProps) {
             experimentalFeatures: experimentalFeatures,
             undoSnapshotService: props.snapshotService,
             trustedHTMLHandler: trustedHTMLHandler,
+            sizeTransformer: size => size / scale,
```
Lufedi, line 96: is there a validation on the scale prop to make sure it's greater than 0?
or maybe `size => (scale > 0 ? size / scale : size)`

Lufedi (approved):

demo/scripts/controls/editor/Editor.tsx
JiuqingSong: For demo site, I'd like to keep it simple and do not eat too much errors. If scale is 0, there must be code bug and we just let it throw

flyingbee2012 (approved):

PR #721, opened 2022-01-18
PoC of Table styling feature and table selection

BryanValverdeU:
Table Styling contains the features of: 
- Table cell selection 
cellSelection
- Customize predefined styles
- Manage style state of the table 
predefinedStyles


demo/scripts/controls/MainPane.scss
```diff
@@ -16,6 +16,12 @@
 }
 
 .editor {
+    :global(table._tableSelected td *::selection) {
+        background-color: transparent;
+    }
+    :global(table._tableSelected td::selection) {
+        background-color: transparent;
```
JiuqingSong, line 23: Does it mean any project that use roosterjs need to add this css to their page?

packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -78,6 +79,10 @@ function clearNodeFormat(node: Node): boolean {
             // 3. Otherwise, remove all attributes
             clearAttribute(node as HTMLElement);
         }
+
+        if ((node as HTMLElement).dataset[TableMetadata.TEMP_BACKGROUND_COLOR]) {
```
JiuqingSong, line 83: if node is not HTMLElement, it doesn't have dataset. Then there will be exception

packages/roosterjs-editor-dom/lib/utils/tableInfo.ts
```diff
@@ -0,0 +1,42 @@
+import { TableFormat } from 'roosterjs-editor-types';
+
+const TABLE_STYLE_INFO = 'roosterTableInfo';
+
+/**
+ * Save the format info of a table
+ * @param table The table the info will be saved
+ * @param format The format of the table
+ */
+export function saveTableInfo(table: HTMLTableElement, format: Partial<TableFormat>) {
```
JiuqingSong, line 10: If we can move the plugin to be regular, these code can be shared with ImageEditPlugin

packages/roosterjs-editor-dom/lib/utils/getTableOriginalColor.ts
```diff
@@ -0,0 +1,13 @@
...
+
+const SELECTION_COLOR_OPACITY = TableMetadata.SELECTION_COLOR_OPACITY;
+
+/**
+ * @internal
+ * Retrieve the color to be applied when a cell is selected
+ * @param colorString Color
+ * @returns color to be applied when a cell is selected
+ */
+export function getHighlightColor() {
```
JiuqingSong, line 11: move this function to VTable since it is not used in other places, and no need to export

demo/scripts/controls/MainPane.scss
BryanValverdeU: This is only when the selection starts outside of a table and moves in a table, for selection inside a single table this is not taking effect
Next gif is without this css rules.
So it is not required to add this to any project css.

If the :selection and the cell style we are applying to cells is to noisy, we can remove the code that handle highlighting cells when the selection is outside of a table
TABL1

packages/roosterjs-editor-api/lib/utils/tableSelectionBeforeCallBackWrap.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+export default function tableSelectionBeforeCallBackWrap(
+    editor: IEditor,
+    elementTag: string,
+    beforeRunCallback: () => boolean,
+    forEachCallback: (node: HTMLElement) => any
+) {
+    const tableSelection = editor.getTableSelection();
+    if (!tableSelection?.vSelection) {
+        return beforeRunCallback();
```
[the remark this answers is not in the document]
BryanValverdeU: This function is only called on the toggle code block and quoteblock, In the original api, if there are elements with blockquote/codeblock, the code toggle off the style applied, but if there is not elements, code proceed and add the style to the selection

packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
```diff
@@ -113,6 +113,19 @@ export default class Paste implements EditorPlugin {
...
+            ) {
+                const div = this.editor.getElementAtCursor('div');
+                if (getTagOfNode(div.lastChild) == 'BR' && div.childElementCount == 1) {
+                    div.removeChild(div.lastChild);
+                }
+                const divC = this.editor.getDocument().createElement('div');
+                divC.appendChild(this.editor.getDocument().createElement('br'));
+                div.parentNode.insertBefore(divC, div.nextSibling);
+            }
+
```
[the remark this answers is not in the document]
BryanValverdeU: Tested and this is an unneeded change, removing it.

packages/roosterjs-editor-api/lib/table/insertTable.ts
```diff
@@ -32,21 +57,33 @@ export default function insertTable(
...
-        editor.insertNode(fragment);
+        if (!isInsideTable) {
+            editor.insertNode(fragment, {
+                insertOnNewLine: true,
+                replaceSelection: true,
+                position: ContentPosition.SelectionStart,
+                updateCursor: true,
+            });
+        } else {
+            editor.insertNode(fragment);
```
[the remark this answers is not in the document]
BryanValverdeU: Checked and this is an unneeded change, removing it.

packages/roosterjs-editor-dom/lib/utils/tableInfo.ts
juliaroldi: do you mean create a plugin to save data?

PR #723, opened 2022-01-20
Add A PoC for SelectionRangeEx

BryanValverdeU:
Add a Editor APi for getSelectionRangeEx, so we can use this api to use different type of ranges.

At the moment, only have 2 types of selections Normal, using the selection of the window, and Table selection.

Also added an example with the inlineStyle to apply the formating to the changes independently of the type of selection range returned from  getSelectionRangeEx.

### DIFF
--- a/packages/roosterjs-editor-core/lib/coreApi/getSelectionRangeEx.ts
+++ b/packages/roosterjs-editor-core/lib/coreApi/getSelectionRangeEx.ts
@@ -1,32 +1,22 @@
-import {
-    EditorCore,
-    GetSelectionRangeEx,
-    SelectionRangeEx,
-    TableSelectionPluginState,
-} from 'roosterjs-editor-types';
+import { EditorCore, GetSelectionRangeEx, SelectionRangeEx } from 'roosterjs-editor-types';
 import {
     contains,
     createRange,
-    findClosestElementAncestor,
     NormalSelectionRange,
     safeInstanceOf,
+    TableMetadata,
     TableSelectionRange,
+    VTable,
 } from 'roosterjs-editor-dom';
 
 /**
  * @internal
  * Get current or cached selection range
  * @param core The EditorCore object
- * @param tryGetFromCache Set to true to retrieve the selection range from cache if editor doesn't own the focus now
  * @returns A Range object of the selection range
  */
-export const getSelectionRangeEx: GetSelectionRangeEx = (
-    core: EditorCore,
-    tryGetFromCache: boolean
-) => {
+export const getSelectionRangeEx: GetSelectionRangeEx = (core: EditorCore) => {
     let result: SelectionRangeEx = null;
-
-    const tSelection = core.tableSelection;
     if (core.lifecycle.shadowEditFragment) {
         const shadowRange =
             core.lifecycle.shadowEditSelectionPath &&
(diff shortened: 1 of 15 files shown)


packages/roosterjs-editor-core/lib/coreApi/getSelectionRangeEx.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * @internal
+ * Get current or cached selection range
+ * @param core The EditorCore object
+ * @param tryGetFromCache Set to true to retrieve the selection range from cache if editor doesn't own the focus now
+ * @returns A Range object of the selection range
+ */
+export const getSelectionRangeEx: GetSelectionRangeEx = (
+    core: EditorCore,
+    tryGetFromCache: boolean
```
JiuqingSong, line 25: Maybe we can totally remove this parameter. I checked the code and there are only 3 places passing false to this param.

For those cases we can check one by one on how to do without this parameter

packages/roosterjs-editor-dom/lib/selection/NormalSelectionRange.ts
```diff
@@ -0,0 +1,18 @@
...
+    constructor(ranges: Range[]) {
+        this.ranges = ranges;
+        this.type = SelectionRangeTypes.Normal;
+    }
+
+    type: SelectionRangeTypes.Normal;
+    ranges: Range[];
+
+    isCollapsed = () => {
+        return this.ranges[0].collapsed;
```
JiuqingSong, line 16: If we make isCollapse as a simple boolean value, no need to have this class, but just make it a simple object

packages/roosterjs-editor-dom/lib/selection/TableSelectionRange.ts
```diff
@@ -0,0 +1,29 @@
...
+    ) {
+        this.vTable = new VTable(tableElement);
+        this.vTable.startRange = startRange;
+        this.vTable.endRange = endRange;
+
+        this.ranges = this.vTable.getSelectedRanges();
+        this.type = SelectionRangeTypes.VSelection;
+    }
+
+    vTable: VTable;
```
JiuqingSong, line 21: how do you want to handle selection across multiple tables?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -1091,6 +1090,37 @@ export default class VTable {
...
+            });
+
+            if (firstSelected) {
+                rowRange.setStartBefore(firstSelected);
+                rowRange.setEndAfter(lastSelected);
+                ranges.push(rowRange);
+            }
+        }
+
+        return ranges.length > 0 ? ranges : null;
```
JiuqingSong, line 1122: I prefer return empty array when nothing selected. So that we don't need always check null when use the return value

BryanValverdeU: Ok, that makes sense!

packages/roosterjs-editor-types/lib/interface/SelectionRangeEx.ts
```diff
@@ -0,0 +1,72 @@
...
+     */
+    ranges: Range[];
+}
+
+/**
+ * Types of Selection Ranges that the SelectionRangeEx can return
+ */
+export enum SelectionRangeTypes {
+    Normal,
+    VSelection,
```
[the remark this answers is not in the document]
BryanValverdeU: We can change this to TableSelection. I did not know how to call the selection made inside of a table.

Across multiple table can be a normal selection. Checking in Word for example,
Multiple options for table design are only available when selecting inside of a single table (the tabs above).
TABLw

When there are multiple tables in a selection, we can only keep it as a Normal Selection.
TABLe

What do you think? Is there any other scenario?

packages/roosterjs-editor-dom/lib/selection/NormalSelectionRange.ts
BryanValverdeU: I agree, This is a better approach setting this property as a simple boolean.
Or even removing it as we could just check for each of the ranges inside of the array if the actual Range is collapsed.

packages/roosterjs-editor-types/lib/interface/SelectionRangeEx.ts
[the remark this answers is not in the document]
JiuqingSong: It sounds good to have normal selection for cross-table selection. Then the question is, do we still need to show the gray background color in that case? 

I think if we make it as normal selection, just save a normal range should be good enough an no need to have the info of gray background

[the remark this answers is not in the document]
BryanValverdeU: Sounds good, we can remove the gray background once the selection is normal. Will do that.

BryanValverdeU:
Moving this code changes in this PR to https://github.com/microsoft/roosterjs/pull/725/files so we dont have the table styling changes in the same branch

PR #724, opened 2022-01-20
Table Styling Feature

juliaroldi:
- Customize predefined styles
- Manage style state of the table
predefinedStyles


packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -90,6 +96,10 @@ export default class VTable {
                     }
                 });
             });
+            const format = getTableFormatInfo(this.table);
```
JiuqingSong, line 99: Suggest we read format info in the constructor of VTable. So that we can change its value directly if we want, then writeBack() will apply the change

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -104,10 +114,21 @@ export default class VTable {
             return;
         }
         this.table.style.borderCollapse = 'collapse';
-        this.trs[0].style.backgroundColor = format.bgColorOdd || 'transparent';
-        if (this.trs[1]) {
-            this.trs[1].style.backgroundColor = format.bgColorEven || 'transparent';
-        }
+        this.setBorderColors(format);
```
JiuqingSong, line 117: Suggest not change existing applyFormat(), but add a new API, and mark old one as @deprecated.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -116,10 +137,200 @@ export default class VTable {
                     cell.td.style.borderBottom = getBorderStyle(format.bottomBorderColor);
                     cell.td.style.borderLeft = getBorderStyle(format.verticalBorderColor);
                     cell.td.style.borderRight = getBorderStyle(format.verticalBorderColor);
+                    cell.td.style.borderBottom = getBorderStyle(format.bottomBorderColor);
```
JiuqingSong, line 140: You set borderBottom twice

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -104,10 +114,21 @@ export default class VTable {
...
-        if (this.trs[1]) {
-            this.trs[1].style.backgroundColor = format.bgColorEven || 'transparent';
-        }
+        this.setBorderColors(format);
+        this.setTableRowColor(format);
+        this.setTableColumnsColor(format);
+        this.setBorderType(format);
+        this.setFirstColumnFormat(format);
+        this.setHeaderRowFormat(format);
+        saveTableInfo(this.table, format);
```
JiuqingSong, line 123: Suggest move this line into writeBack()

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -116,10 +137,200 @@ export default class VTable {
...
+                        ? (cell.td.style.backgroundColor = format.bgColorEven)
+                        : ''
+                );
+            } else {
+                row.forEach(cell =>
+                    cell.td && format.bgColorOdd
+                        ? (cell.td.style.backgroundColor = format.bgColorOdd)
+                        : ''
+                );
+            }
```
JiuqingSong, line 167: This can be merged into:
const color = index % 2 === 0 ? format.bgColorEven : format.bgColorOdd;
fow.forEach(cell => {
  if (cell.td && color) {
     cell.td.style.backgroundColor = color;
  }
});

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -116,10 +137,200 @@ export default class VTable {
...
+            return;
+        }
+
+        this.cells.forEach(row => {
+            row.forEach((cell, index) => {
+                if (index % 2 === 0 && cell.td && format.bgColumnColorEven) {
+                    cell.td.style.backgroundColor = format.bgColumnColorEven;
+                } else if (index % 2 === 1 && cell.td && format.bgColumnColorOdd) {
+                    cell.td.style.backgroundColor = format.bgColumnColorOdd;
+                }
```
JiuqingSong, line 186: same here

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -116,10 +137,200 @@ export default class VTable {
...
+        });
+    }
+
+    /**
+     * Organize the borders of table
+     * @param format
+     * @returns
+     */
+    private setBorderType(format: Partial<TableFormat>) {
+        if (!format.tableBorderFormat) {
```
JiuqingSong, line 197: Maybe just merge this check into the switch block below:

switch (tableborderformat) {
  ... 
  case TableBorderFormat.None;
    break; // no no
}

Then you can add a "None" value of TableBorderFormat

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -104,10 +114,21 @@ export default class VTable {
...
+        this.setHeaderRowFormat(format);
+        saveTableInfo(this.table, format);
+    }
+
+    /**
+     * Set color to borders of an table
+     * @param format
+     * @returns
+     */
+    private setBorderColors(format: Partial<TableFormat>) {
```
JiuqingSong, line 131: Once we apply the default format when read, we can use "Required" instead of "Partial" here since we are sure we always pass full format object

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -116,10 +137,200 @@ export default class VTable {
...
+     * @returns
+     */
+    private setBorderType(format: Partial<TableFormat>) {
+        if (!format.tableBorderFormat) {
+            return;
+        }
+        switch (format.tableBorderFormat) {
+            case TableBorderFormat.onlyExternalBorders:
+                this.cells.forEach(row => {
+                    row.filter(cell => cell.td).forEach(cell => {
```
JiuqingSong, line 203: This pattern appeared many times, need to extract as a function

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -116,10 +137,200 @@ export default class VTable {
...
+                        if (this.cells.indexOf(row) !== 0) {
```
JiuqingSong, line 204: This result is used multiple times, cache its value as local var first

JiuqingSong: You can do 
this.cells.forEach((row, index) => { ....
  if (index !== 0) ...

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -116,10 +137,200 @@ export default class VTable {
...
+                            cell.td.style.borderTopColor = TRANSPARENT;
+                        }
+                        if (this.cells.indexOf(row) !== this.cells.length - 1) {
+                            cell.td.style.borderBottomColor = TRANSPARENT;
+                        }
+                        if (row.indexOf(cell) !== 0) {
```
JiuqingSong, line 210: same here

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -116,10 +137,200 @@ export default class VTable {
...
+                    }
+                });
+                this.forEachCellOfColumn(this.cells[0].length - 1, (cell, row, i) => {
+                    if (cell.td) {
+                        cell.td.style.borderRightColor = TRANSPARENT;
+                    }
+                });
+        }
+    }
+
```
JiuqingSong, line 270: Generally, consider refactor this function, such as:

```ts
const TableBorderFormatters: {[format in TableBorderFormat]: (td: HtmlTableCellElement, isFirstRow: boolean, isLastRow, isFirstColumn, isLastColumn)} = {
  [TableBorderFormat.onlyMiddleBorders]: (td, isFirstRow, isLastRow,isFirstColumn,isLastColumn) => {
     if (isFirstRow) { td.style.borderLeftColor = .... }
}, { ... }
};


private setBorderType(format: Partial<TableFormat>) {
   const formatter = TableBorderFormatters[format];
   this.cells.forEach((row, rowIndex) => row.forEach(cell, cellIndex) => {
     if (cell.td)
     formatter(cell.td, rowIndex ==0, rowIndex == cells.length -1, cellIndex == 0, cellIndex == row.length - 1);
 }
}
```

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -116,10 +137,200 @@ export default class VTable {
...
+     */
+    private setFirstColumnFormat(format: Partial<TableFormat>) {
+        if (!format.firstColumn) {
+            this.forEachCellOfColumn(0, (cell, row, i) => {
+                if (cell.td) {
+                    cell.td = changeElementTag(
+                        cell.td,
+                        TABLE_CELL_TAG_NAME
+                    ) as HTMLTableCellElement;
+                    cell.td.scope = '';
```
JiuqingSong, line 284: What is this line for?

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
```diff
@@ -1,17 +1,13 @@
+import { TableBorderFormat } from '../enum/TableBorderFormat';
+
 /**
  * Table format
  */
 export default interface TableFormat {
-    /**
-     * Background color for even rows
-     */
-    bgColorEven: string;
```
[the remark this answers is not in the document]
juliaroldi: I didn't remove, I just changed the position in the list. We still need this property.

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
```diff
@@ -25,5 +21,52 @@ export default interface TableFormat {
     /**
      * Vertical border color for each row
      */
-    verticalBorderColor: string;
+    verticalBorderColor: string | null;
```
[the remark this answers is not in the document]
juliaroldi: Ok, I'll create a new interface and use the default format as default values.

packages/roosterjs-editor-dom/lib/table/VTable.ts
juliaroldi: Since we are applying the style after we create a VTable, if we try to get the current format in the constructor, the current format will be always the default

packages/roosterjs-editor-dom/lib/table/VTable.ts
juliaroldi: To remove the scope, when remove first row or first column.

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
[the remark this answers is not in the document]
JiuqingSong: Still need to make all properties optional, e.g.:

bgColor?: string

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -90,12 +113,17 @@ export default class VTable {
...
+            if (formatInfo) {
+                this.applyFormatAndStyle(formatInfo);
+            }
         } else if (this.table) {
             this.table.parentNode.removeChild(this.table);
         }
     }
 
     /**
+     * @deprecated
```
JiuqingSong, line 126: If there is no conflict betwen applyFormat and applyFormatAndStyle, no need to create the new API, but just reuse the existing one. Ping me if you have any question about this part.

JiuqingSong: And we should still accept "Partial<TableFormat>"

packages/roosterjs-editor-dom/lib/table/VTable.ts
JiuqingSong: This is still active

packages/roosterjs-editor-api/lib/table/formatTable.ts
```diff
@@ -9,14 +9,14 @@ import { VTable } from 'roosterjs-editor-dom';
  */
 export default function formatTable(
     editor: IEditor,
-    format: Partial<TableFormat>,
+    format: Required<TableFormat>,
```
JiuqingSong, line 12: Why change this?

We should keep public function to use Partial so that people can call it with fewer parameters, but change private method to use Required so that we are sure all properties have value

packages/roosterjs-editor-dom/lib/index.ts
```diff
@@ -53,6 +53,7 @@ export { default as adjustInsertPosition } from './utils/adjustInsertPosition';
 export { default as createElement, KnownCreateElementData } from './utils/createElement';
 export { default as moveChildNodes } from './utils/moveChildNodes';
 export { default as setListItemStyle } from './utils/setListItemStyle';
+export { default as getTableFormatInfo } from './utils/getFormatTableInfo';
```
JiuqingSong, line 56: Move this function into VTable.ts and do not export it unless we really want people to operate on this property

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -120,6 +148,247 @@ export default class VTable {
...
+        });
+    }
+
+    private formatBorders(
+        borderType: TableBorderFormat,
+        td: HTMLTableCellElement,
+        isNotFirstRow: boolean,
+        isNotLastRow: boolean,
+        isNotFirstColumn: boolean,
+        isNotLastColumn: boolean,
```
JiuqingSong, line 234: We usually use "isSomething" but not "isNotSomething".

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -120,6 +148,247 @@ export default class VTable {
...
+        isFirstCell: boolean,
```
JiuqingSong, line 235: IsFirstCell means isFirstRow && isFirstColumn, right? So no need have these two parameters

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -120,6 +148,247 @@ export default class VTable {
...
+        }
+        this.forEachCellOfColumn(0, (cell, row, i) => {
+            if (cell.td) {
+                if (i !== 0) {
+                    cell.td.style.borderTopColor = TRANSPARENT;
+                    cell.td.style.backgroundColor = TRANSPARENT;
+                }
+                if (i !== this.cells.length - 1 && i !== 0) {
+                    cell.td.style.borderBottomColor = TRANSPARENT;
+                }
```
JiuqingSong, line 344: Have we already do this in formatBorders()?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -120,6 +148,247 @@ export default class VTable {
...
+        this.cells.forEach(row => {
+            row.forEach((cell, index) => {
+                const backgroundColor = color(index);
+                if (cell.td && backgroundColor) {
+                    cell.td.style.backgroundColor = backgroundColor;
+                }
+            });
+        });
+    }
+
```
JiuqingSong, line 227: Can these two functions be merged? Both of them are setting background color, if we have two of them, is it possible to have some conflict?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -90,12 +113,17 @@ export default class VTable {
                     }
                 });
             });
+            const formatInfo = getTableFormatInfo(this.table);
```
JiuqingSong, line 116: It is possible the retrieved format info is invalid. At least we should make sure it has all values for all properties.

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
```diff
@@ -1,17 +1,13 @@
...
-     * Background color for even rows
-     */
-    bgColorEven: string;
-
     /**
      * Background color for odd rows
      */
-    bgColorOdd: string;
-
+    bgColor: string | null;
```
JiuqingSong, line 10: Comment doesn't match.

Question: Since we already have bgColorOdd and bgColorEven, why need this bgColor? They can be conflict

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
```diff
@@ -25,5 +21,52 @@ export default interface TableFormat {
     /**
      * Vertical border color for each row
      */
-    verticalBorderColor: string;
+    verticalBorderColor: string | null;
+
+    /**
+     * Set header row
+     */
+    headerRow: boolean;
```
JiuqingSong, line 29: Better name it "hasHeaderRow"

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
```diff
@@ -25,5 +21,52 @@ export default interface TableFormat {
...
+
+    /**
+     * Header row background color for even cells
+     */
+    headerRowColor: string | null;
+    /**
+     * Set first column
+     */
+    firstColumn: boolean;
```
JiuqingSong, line 38: hasFirstColumn

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
```diff
@@ -25,5 +21,52 @@ export default interface TableFormat {
...
+    /**
+     * Set banded columns
+     */
+    bandedColumns: boolean;
```
JiuqingSong, line 42: hasBandedColumns

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
```diff
@@ -25,5 +21,52 @@ export default interface TableFormat {
...
+
+    /**
+     * Background color for even columns
+     */
+    bgColumnColorEven: string | null;
```
JiuqingSong, line 47: Same here, can this be conflict with bgColorOdd/bgColorEven ?

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
```diff
@@ -25,5 +21,52 @@ export default interface TableFormat {
...
+
+    /**
+     * Background color for odd columns
+     */
+    bgColumnColorOdd: string | null;
```
JiuqingSong, line 52: same

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
```diff
@@ -25,5 +21,52 @@ export default interface TableFormat {
...
+    /**
+     * Set banded rows
+     */
+    bandedRows: boolean;
```
JiuqingSong, line 56: hasBandedRows

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
```diff
@@ -1,17 +1,13 @@
+import { TableBorderFormat } from '../enum/TableBorderFormat';
+
 /**
  * Table format
  */
 export default interface TableFormat {
```
JiuqingSong, line 6: Generall, design this interface is very similar with design a database.
We need to follow database normal form such as BCNF or 3NF. We need to check each property if they are necessary, if they have conflict, if its value can be calculated from other properties. Then make sure all combinations of the values are meaningful.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -120,6 +145,237 @@ export default class VTable {
...
+            }
+        });
+    }
+
+    /**
+     * Save the format info of a table
+     * @param table The table the info will be saved
+     * @param format The format of the table
+     */
+    saveTableInfo(table: HTMLTableElement, format: TableFormat) {
```
JiuqingSong, line 373: Move this function out. It can be put together with getTableFormatInfo. So you only need to define the TABLE_STYLE_INFO const once

packages/roosterjs-editor-dom/lib/utils/getTableFormatInfo.ts
```diff
@@ -0,0 +1,81 @@
...
+const TABLE_STYLE_INFO = 'roosterTableInfo';
+
+/**
+ * Get the format info of a table
+ * If the table does not have a info saved, it will be retrieved from the css styles
+ * @param table The table that has the info
+ */
+export default function getTableFormatInfo(table: HTMLTableElement) {
+    const obj = safeParseJSON(table?.dataset[TABLE_STYLE_INFO]) as TableFormat;
+    return checkIfTableFormatIsValid ? obj : undefined;
```
JiuqingSong, line 11: checkIfTableFormatIsValid(obj)

packages/roosterjs-editor-dom/lib/utils/getTableFormatInfo.ts
```diff
@@ -0,0 +1,81 @@
...
+        bgColorOdd,
+        bgColorEven,
+        hasBandedColumns,
+        hasBandedRows,
+        hasFirstColumn,
+        hasHeaderRow,
+        tableBorderFormat,
+    } = format;
+    const colorsValues = [
+        ...Object.values({
```
JiuqingSong, line 28: Object.values is not supported by Chrome 49 and IE.
image

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -120,6 +145,237 @@ export default class VTable {
...
 
+    /**
+     * Apply the given table format to this virtual table
+     * @param format Table format to apply
+     */
+    applyFormatAndStyle(format: TableFormat) {
+        if (!this.table) {
+            return;
+        }
+        format = { ...DEFAULT_FORMAT, ...format };
```
JiuqingSong, line 156: Here we should use current format of the table instead of default format. Otherwise if I pass in a partial format, all other properties will be replaced with default value

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -90,12 +110,17 @@ export default class VTable {
                     }
                 });
             });
+            const formatInfo = getTableFormatInfo(this.table);
```
JiuqingSong, line 113: getTableFormatInfo() should be called when create VTable, and then save format in writeBack(). Then during the whole lifecycle of VTable we know its current format

JiuqingSong: And if the format is still the default value when save, no need to save it. Any existing format can be removed. This can happen for a table never be formatted

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -120,6 +145,237 @@ export default class VTable {
...
+            });
+        });
+        if (hasBandedColumns) {
+            this.cells.forEach(row => {
+                row.forEach((cell, index) => {
+                    const backgroundColor = color(index);
+                    if (cell.td && backgroundColor) {
+                        cell.td.style.backgroundColor = backgroundColor;
+                    }
+                });
```
JiuqingSong, line 194: My understanding is hasBandedRows and hasBandedColumns are actually conflict, right? When they are both true, bandedColumns will overwrite bandedRows.

In that case, we should not have both of them, but use a enum instead. Because two booleans has 4 combinations, but "true, true" is not valid, we only allow 3 combinations

JiuqingSong: Then these two loops can be merged into one

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -120,6 +145,237 @@ export default class VTable {
...
+        this.cells.forEach((row, rowIndex) => {
+            row.forEach((cell, cellIndex) => {
+                if (cell.td) {
+                    this.formatBorders(
+                        format,
+                        cell.td,
+                        rowIndex === 0,
+                        rowIndex === this.cells.length - 1,
+                        cellIndex === 0,
+                        cellIndex === this.cells[0].length - 1,
```
JiuqingSong, line 297: Although in most cases all rows have same count of cells, but let's do not assume that. You can check cellIndex == row.length - 1 here, same for next line

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -120,6 +145,237 @@ export default class VTable {
...
+                        rowIndex === this.cells[0].length - 1
```
JiuqingSong, line 298: what does this mean? Why compare row index with column count?

JiuqingSong:
Please resolve conflict and fix test case

packages/roosterjs-editor-dom/lib/index.ts
```diff
@@ -50,6 +50,7 @@ export { default as setColor } from './utils/setColor';
 export { default as matchesSelector } from './utils/matchesSelector';
 export { default as createElement, KnownCreateElementData } from './utils/createElement';
 export { default as moveChildNodes } from './utils/moveChildNodes';
+export { saveTableInfo, getTableFormatInfo } from './utils/tableFormatInfo';
```
JiuqingSong, line 53: Suggest do not export these 2 functions from index.ts. We don't expect customer app call these 2 functions directly.

For demo site, you can do deep import to that file. It is ok to do some kind of hack in demo site since it is not production code

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -36,6 +43,8 @@ export default class VTable {
 
     private trs: HTMLTableRowElement[] = [];
 
+    formatInfo: TableFormat;
```
JiuqingSong, line 46: comment

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -36,6 +43,8 @@ export default class VTable {
 
     private trs: HTMLTableRowElement[] = [];
```
JiuqingSong, line 44: better move this private member down after other public members

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -107,6 +116,10 @@ export default class VTable {
                     }
                 });
             });
+            if (this.formatInfo) {
+                this.applyFormat(this.formatInfo);
+                saveTableInfo(this.table, this.formatInfo);
```
JiuqingSong, line 121: If formatInfo has the same value with default format, do not save and delete any existing format info from the table

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -117,25 +130,275 @@ export default class VTable {
      * @param format Table format to apply
      */
     applyFormat(format: Partial<TableFormat>) {
-        if (!format || !this.table) {
+        if (!this.table) {
             return;
         }
+        if (format) {
```
JiuqingSong, line 136: Do we want to allow null value for format here? If so, what does it mean?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -117,25 +130,275 @@ export default class VTable {
      * @param format Table format to apply
      */
     applyFormat(format: Partial<TableFormat>) {
-        if (!format || !this.table) {
+        if (!this.table) {
             return;
         }
+        if (format) {
+            this.formatInfo = format;
```
JiuqingSong, line 137: And since format is Partial type here, we need to make sure all properties have meaningful value.

you can do 
```ts
this.formatInfo = {
   ...defaultFormat, // Make sure everything has value
   ...(this.formatInfo || {}), // Apply existing format if any
   ...(format || {})  // Then overwrite with the passed-in value
};
```

packages/roosterjs-editor-dom/lib/utils/tableFormatInfo.ts
```diff
@@ -0,0 +1,111 @@
...
+        a === TableBorderFormat.NO_SIDE_BORDERS ||
+        a === TableBorderFormat.NO_HEADER_BORDERS ||
+        a === TableBorderFormat.ESPECIAL_TYPE_1 ||
+        a === TableBorderFormat.ESPECIAL_TYPE_2 ||
+        a === TableBorderFormat.ESPECIAL_TYPE_3
+    ) {
+        return true;
+    }
+    return false;
+}
```
BryanValverdeU, line 92: Is it possible to change the parameter type from `any `to `TableBorderFormat`?
Also if you could change the variable name to inputFormat, 'a' does not provide a meaningful description of what is being passed into the function. Maybe something like inputBorder

packages/roosterjs-editor-dom/lib/utils/tableFormatInfo.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    return true;
+}
+
+function isAValidType(a: any) {
+    if (a === null || a === undefined || typeof a === 'string') {
+        return true;
+    }
+    return false;
+}
```
BryanValverdeU, line 69: If possible add a name with more defintion other than `a`, maybe something like colorInput.
Name of the function could be isAValidColor.

colorValues is a string[], so I think we could change the parameter type to string or is there a specific scenario where is going to be a different type?

packages/roosterjs-editor-dom/lib/utils/tableFormatInfo.ts
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * @internal
+ * Save the format info of a table
+ * @param table The table the info will be saved
+ * @param format The format of the table
+ */
+export function saveTableInfo(table: HTMLTableElement, format: TableFormat) {
+    if (JSON.stringify(format) === JSON.stringify(DEFAULT_FORMAT)) {
```
JiuqingSong, line 101: If properties sequences are different, this will return false.

you can compare each property one by one.

And when it returns true, need to delete existing format info from table

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -117,25 +133,276 @@ export default class VTable {
      * @param format Table format to apply
      */
     applyFormat(format: Partial<TableFormat>) {
-        if (!format || !this.table) {
+        if (!this.table) {
             return;
         }
+        this.formatInfo = {
+            ...(this.formatInfo || {}),
```
JiuqingSong, line 140: What if this.formatInfo is null and format only contains part of the properties?

You need to apply default format first.

JiuqingSong (approved):

BryanValverdeU (approved):

PR #725, opened 2022-01-21
Add getSelectionRangeEx Editor Api

BryanValverdeU:
Add a Editor APi for getSelectionRangeEx, so we can use this api to use different type of ranges.

At the moment, only have 2 types of selections Normal, using the selection of the window, and Table selection.


packages/roosterjs-editor-types/lib/interface/SelectionRangeEx.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Represents normal selection
+ */
+export interface INormalSelectionRange extends ISelectionRangeEx<SelectionRangeTypes.Normal> {}
+
+/**
+ * Represents normal selection
+ */
+export interface ISelectionRangeEx<T> {
```
JiuqingSong, line 19: Let's remove the prefix "I", as long as it is not conflict with existing types. Same for other interfaces

packages/roosterjs-editor-types/lib/interface/SelectionRangeEx.ts
```diff
@@ -0,0 +1,53 @@
+import IVTable from './IVTable';
```
JiuqingSong, line 1: Can we just name it "Table"?

packages/roosterjs-editor-types/lib/interface/SelectionRangeEx.ts
JiuqingSong, line 19:
```ts
<T extends SelectionRangeTypes>
```

packages/roosterjs-editor-types/lib/interface/IVTable.ts
```diff
@@ -0,0 +1,120 @@
...
+     * Get the TableCell in a provided coordinate
+     * @param row
+     * @param col
+     */
+    getTd(row: number, col: number): HTMLTableCellElement;
+
+    /**
+     * Transforms the selected cells to Ranges.
+     * For Each Row a Range with selected cells, a Range is going to be returned.
+     * @returns
```
JiuqingSong, line 117: returns what?

packages/roosterjs-editor-types/lib/interface/IEditor.ts
```diff
@@ -445,7 +453,7 @@ export default interface IEditor {
     /**
      * Get a content traverser for current selection
      */
-    getSelectionTraverser(): IContentTraverser;
+    getSelectionTraverser(range?: Range): IContentTraverser;
```
JiuqingSong, line 456: We can make this change in next PR.

packages/roosterjs-editor-types/lib/interface/SelectionRangeEx.ts
JiuqingSong, line 19: Move this interface to the top

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -28,13 +29,27 @@ export default class VTable {
      */
     col: number;
 
+    /**
+     * Start of the  Selection Range
+     */
+    startRange: number[];
```
JiuqingSong, line 35: number[] means people can put numbers with any length.

Consider something like:
```ts
interface TableSelection {
   firstCol: number;
   firstRow: number;
   lastCol: number;
   lastRow: number;
}

interface Table {
   ...
   selection: TableSelection;
}
```

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -28,13 +29,27 @@ export default class VTable {
...
 
     /**
      * Create a new instance of VTable object using HTML TABLE or TD node
      * @param node The HTML Table or TD node
      */
-    constructor(node: HTMLTableElement | HTMLTableCellElement, normalizeSize?: boolean) {
+    constructor(
+        node: HTMLTableElement | HTMLTableCellElement,
+        normalizeSize?: boolean,
+        setSelectedRange?: boolean
```
JiuqingSong, line 51: Need comment

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -63,6 +78,17 @@ export default class VTable {
                                 width: hasTd ? rect.width : undefined,
                                 height: hasTd ? rect.height : undefined,
                             };
+
+                            if (
+                                setSelectedRange &&
```
JiuqingSong, line 83: If the cost of setting the selection range is low, we can always set it, no need to add parameter

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -63,6 +78,17 @@ export default class VTable {
                                 width: hasTd ? rect.width : undefined,
                                 height: hasTd ? rect.height : undefined,
                             };
+
+                            if (
+                                setSelectedRange &&
+                                td.classList.contains(TableMetadata.TABLE_CELL_SELECTED)
```
JiuqingSong, line 84: Does this work in IE11?

For now roosterjs still supports IE11. We can stop the supporting in next major release

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -520,6 +572,25 @@ export default class VTable {
         this.normalizeTableCellSize();
         setHTMLElementSizeInPx(this.table); // Make sure table width/height is fixed to avoid shifting effect
     }
+
+    /**
+     * Check if the cell is inside of the selection range
+     * @returns true if it is inside, otherwise false
+     */
+    private isInsideOfSelection(x: number, y: number) {
```
JiuqingSong, line 580: suggest inline the logic into getSelectedRanges. and some of them only need to do once.

packages/roosterjs-editor-dom/lib/selection/TableSelectionRange.ts
```diff
@@ -0,0 +1,24 @@
+import VTable from '../table/VTable';
+import { ITableSelectionRange, SelectionRangeTypes } from 'roosterjs-editor-types';
+/**
+ * Table selection Range used in the getSelectedRangeEx editor Api
+ * Can create a object with an array of ranges depending on a vTable range provided.
+ */
+export default class TableSelectionRange implements ITableSelectionRange {
```
JiuqingSong, line 7: This class doesn't bring any value. No need to have it. TableSelectionRange is just a pure interface without any member function

JiuqingSong: If need you can create a helper function to create instance of TableSelectionRange and calculate areAllCollapsed. The put it to the place where it is used, and no need to export it

packages/roosterjs-editor-dom/lib/selection/NormalSelectionRange.ts
```diff
@@ -0,0 +1,20 @@
+import { INormalSelectionRange, SelectionRangeTypes } from 'roosterjs-editor-types';
+
+/**
+ * Normal selection Range used in the getSelectedRangeEx editor Api
+ */
+export default class NormalSelectionRange implements INormalSelectionRange {
```
JiuqingSong, line 6: same here. No need to have this class

packages/roosterjs-editor-core/lib/coreApi/getSelectionRangeEx.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (core.lifecycle.shadowEditFragment) {
+        const shadowRange =
+            core.lifecycle.shadowEditSelectionPath &&
+            createRange(
+                core.contentDiv,
+                core.lifecycle.shadowEditSelectionPath.start,
+                core.lifecycle.shadowEditSelectionPath.end
+            );
+
+        if (shadowRange.collapsed) {
```
JiuqingSong, line 29: why?

For shadow edit, we should not touch anything from the real DOM

packages/roosterjs-editor-dom/lib/table/tableMetadata.ts
```diff
@@ -0,0 +1,25 @@
...
+     */
+    TABLE_CELL_SELECTED = '_tableCellSelected',
+    /**
+     * Data set applied on a selected table cell to store the original background color
+     */
+    TEMP_BACKGROUND_COLOR = 'originalBackgroundColor',
+    /**
+     * Table cell selector
+     */
+    TABLE_CELL_SELECTOR = 'td,th',
```
JiuqingSong, line 20: I think we don't need to have a special const for this one. Just use it directly

packages/roosterjs-editor-dom/lib/table/tableMetadata.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Selectors for Table Selection
+ */
+export const enum TableMetadata {
```
JiuqingSong, line 4: For now let's do not add this enum. I think in this PR only TABLE_CELL_SELECTED and TABLE_SELECTED is used, right?

We can defined them in where they are used and do not export. Later we can move them. 

Adding new type/interface/enum is easy, removing/changing is hard.

packages/roosterjs-editor-core/lib/coreApi/getSelectionRangeEx.ts
BryanValverdeU: When the new API and Table Selection, the selection is always going to be collapsed, as it just stays in the first cell as we just change the styles of the cells. So I think we will need to check if the shadow range is collapsed, if there is a table selection. 

Otherwise if there is a better approach we would need to do it in the lifecycle plugin

JiuqingSong: Better approach should be change the shadowEditSelectionPath, ... to make it knows table selection. But we can do this later.

packages/roosterjs-editor-dom/lib/table/VTable.ts
BryanValverdeU: Yes, IE11 only have partial support when using classlist on SVG elements
https://developer.mozilla.org/en-US/docs/Web/API/Element/classList
image

packages/roosterjs-editor-types/lib/interface/SelectionRangeEx.ts
BryanValverdeU: This one have conflict with the union type setting it as SelectionRangeExBase

packages/roosterjs-editor-types/lib/interface/SelectionRangeEx.ts
```diff
@@ -0,0 +1,53 @@
...
+     */
+    areAllCollapsed: boolean;
+}
+
+/**
+ * Represents the selection made inside of a table.
+ */
+export interface TableSelectionRange
+    extends SelectionRangeExBase<SelectionRangeTypes.TableSelection> {
+    vTable: Table;
```
JiuqingSong, line 28: comment

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -28,14 +30,26 @@ export default class VTable {
...
     private trs: HTMLTableRowElement[] = [];
 
     /**
      * Create a new instance of VTable object using HTML TABLE or TD node
      * @param node The HTML Table or TD node
+     * @param normalizeSize Whether table size needs to be normalized
      */
     constructor(node: HTMLTableElement | HTMLTableCellElement, normalizeSize?: boolean) {
         this.table = safeInstanceOf(node, 'HTMLTableElement') ? node : getTableFromTd(node);
+        this.selection = {
```
JiuqingSong, line 47: How about just set the whole selection object to null when there is no table selection?
What do you think?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -63,6 +77,12 @@ export default class VTable {
                                 width: hasTd ? rect.width : undefined,
                                 height: hasTd ? rect.height : undefined,
                             };
+                            if (td.classList.contains(TABLE_CELL_SELECTED)) {
+                                this.selection.firstCol = this.selection.firstCol ?? targetCol;
```
JiuqingSong, line 81: this.selection.firstCol will be treated as false if its value is 0.
Same for this.selection.firstRow

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -74,6 +94,37 @@ export default class VTable {
         }
     }
 
+    getSelectedRanges(): Range[] {
```
JiuqingSong, line 97: Comments

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -407,7 +458,7 @@ export default class VTable {
         return this.getTd(this.row, this.col);
     }
 
-    private getTd(row: number, col: number) {
+    getTd(row: number, col: number) {
```
JiuqingSong, line 461: comment

packages/roosterjs-editor-core/lib/coreApi/getSelectionRangeEx.ts
```diff
@@ -0,0 +1,84 @@
...
+            const tableSelected = getTableSelected(core.contentDiv);
+
+            if (tableSelected) {
+                return createTableSelectionEx(tableSelected);
+            }
```
JiuqingSong, line 33: Let's just remove this for now. We can go directly with better approach when we want to fix it.

Take some note that we have this pending work to fix

packages/roosterjs-editor-core/lib/coreApi/getSelectionRangeEx.ts
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        return createNormalSelectionEx([shadowRange]);
+    } else {
+        if (core.api.hasFocus(core)) {
```
JiuqingSong, line 38: If not has focus, need to get from cache.

In this API we can always assume "tryGetFromCache" to be true

BryanValverdeU: I think we have this behavior already using the selected range from the Dom plugin if no focus.

packages/roosterjs-editor-dom/lib/table/VTable.ts
BryanValverdeU: Only when firstCol or firstRow are null is going to use the right Expr

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -72,6 +92,51 @@ export default class VTable {
...
+            };
+        }
+    }
+
+    /**
+     * Transforms the selected cells to Ranges.
+     * For Each Row a Range with selected cells, a Range is going to be returned.
+     * @returns Array of ranges from the selected table cells.
+     */
+    getSelectedRanges(): Range[] {
```
JiuqingSong, line 111: Prefer adding a test case for this function

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -28,14 +30,26 @@ export default class VTable {
...
     private trs: HTMLTableRowElement[] = [];
 
     /**
      * Create a new instance of VTable object using HTML TABLE or TD node
      * @param node The HTML Table or TD node
+     * @param normalizeSize Whether table size needs to be normalized
      */
     constructor(node: HTMLTableElement | HTMLTableCellElement, normalizeSize?: boolean) {
         this.table = safeInstanceOf(node, 'HTMLTableElement') ? node : getTableFromTd(node);
+        this.selection = null;
```
JiuqingSong, line 47: Add a test case

JiuqingSong (approved):

PR #726, opened 2022-01-21
Upgrade the RoosterJS version to 8.13.0

romanisa:
Upgrading the RoosterJS version to 8.12.1 to include the latest improvements and bug fixes:

Add GetContentMode.PlainTextFast - A faster way to get plain text content
se image.getAttribute('src') instead of image.src to avoid browser translate relative path to absolute path
Fix a bunch of test case issues
Adjust the execution order of core plugins and customized plugins
Support scale in ImageEdit plugin


JiuqingSong:
Let's use 8.13.0 instead.

JiuqingSong (approved):

PR #728, opened 2022-01-25
Table Cell Selection

BryanValverdeU:
Add a new Plugin for Table Selection.

+Ability to vertical and horizontal selection inside of a table.
+Ability to copy and paste selection made with the plugin selection
+Two content edit features to handle DELETE and BACKSPACE Keys when table selection.

Pending changes ->

-  change API package to handle selection inside of table.
- Table Selector using DragAndDropHelper approach.


JiuqingSong:
Not a must, but consider adding options and sample code for the new plugin in demo site:

image

packages/roosterjs-editor-plugins/lib/plugins/TableSelection/TableCellSelection.ts
```diff
@@ -0,0 +1,850 @@
...
+        }
+        clearSelectedTableCells(event.clonedRoot);
+    }
+
+    //#region Key events
+    /**
+     * Handles the on key event.
+     * @param event the plugin event
+     */
+    private handleKeyDownEvent(event: PluginKeyDownEvent | PluginKeyUpEvent) {
```
JiuqingSong: It seems some of the code from this function can be replaced with Region API.

const regions = editor.getSelectedRegions();
Then check if there are multiple regions, and if they are in the same table.

packages/roosterjs-editor-plugins/lib/plugins/TableSelection/TableCellSelection.ts
```diff
@@ -0,0 +1,850 @@
...
+
+    private handleMouseUpEvent(event: PluginMouseUpEvent) {
+        this.range = this.editor.getSelectionRange();
+        if (
+            event.isClicking &&
+            this.range.collapsed &&
+            !this.tableSelection &&
+            event.rawEvent.which != RIGHT_CLICK
+        ) {
+            this.clearTableCellSelection();
```
JiuqingSong: why?

packages/roosterjs-editor-plugins/lib/plugins/TableSelection/TableCellSelection.ts
```diff
@@ -0,0 +1,850 @@
...
+        // If the event target is not contained in the editor.
+        if (
+            (this.lastTarget.lastChild == this.lastTarget.firstChild &&
+                getTagOfNode(this.lastTarget.lastChild) == 'TABLE' &&
+                getTagOfNode(this.lastTarget) == 'DIV') ||
+            !this.contentDiv.contains(this.lastTarget)
+        ) {
+            event.preventDefault();
+            return;
+        }
```
JiuqingSong: First, we should only handle mousemove when left button is down, right?

packages/roosterjs-editor-plugins/lib/plugins/TableSelection/TableCellSelection.ts
```diff
@@ -0,0 +1,850 @@
...
+        const isNewTDContainingFirstTable = safeInstanceOf(this.lastTarget, 'HTMLTableCellElement')
+            ? contains(this.lastTarget, this.firstTable)
+            : false;
+
+        if (this.firstTable && (targetIsValid || isNewTDContainingFirstTable)) {
+            //When starting selection inside of a table and ends inside of the same table.
+            this.selectionInsideTableMouseMove(event);
+        } else if (this.tableSelection) {
+            clearSelectedTableCells(this.contentDiv);
+            this.tableSelection = false;
```
JiuqingSong: sees some of the code can be replaced with editor.getSelectedRegions()

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -471,6 +487,277 @@ export default class VTable implements Table {
...
+                callback(this.cells[indexY][indexX], indexX, indexY);
+            }
+        }
+    }
+
+    /**
+     * Remove the cells outside of the selection.
+     * @param outsideOfSelection whether to remove the cells outside or inside of the selection
+     */
+    removeCellsBySelection(outsideOfSelection: boolean = true) {
```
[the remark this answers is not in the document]
JiuqingSong: Actually it seems all these functions can be moved to plugin folder. They are not likely to be called from other places, and actually we don't even want people call them from other places

packages/roosterjs-editor-core/lib/corePlugins/CopyPastePlugin.ts
```diff
@@ -75,7 +76,16 @@ export default class CopyPastePlugin implements PluginWithState<CopyPastePluginS
...
+        const selection = this.editor.getSelectionRangeEx();
+        let originalRange: Range = new Range();
+
+        if (selection.type == SelectionRangeTypes.TableSelection && selection.vTable) {
+            originalRange.selectNode(selection.vTable.table);
+            this.editor.select(originalRange);
+        } else if (selection.type == SelectionRangeTypes.Normal) {
+            originalRange = selection.ranges[0];
+        }
+
```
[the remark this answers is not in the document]
BryanValverdeU: Just removed the code from 82 to 88 and it is working fine as you mentioned, thanks.

packages/roosterjs-editor-plugins/lib/plugins/TableSelection/TableCellSelection.ts
BryanValverdeU: Added getSelectedRegions to the code as recommended.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/utils/highlightCellHandler.ts
```diff
@@ -0,0 +1,27 @@
...
+        (!element.dataset[tableCellSelectionCommon.TEMP_BACKGROUND_COLOR] ||
+            element.dataset[tableCellSelectionCommon.TEMP_BACKGROUND_COLOR] == '')
+    ) {
+        element.dataset[tableCellSelectionCommon.TEMP_BACKGROUND_COLOR] =
+            element.style.backgroundColor ?? element.style.background ?? '';
+    }
+    element.style.backgroundColor = tableCellSelectionCommon.HIGHLIGHT_COLOR;
+    element.classList.add(tableCellSelectionCommon.TABLE_CELL_SELECTED);
+
+    element.querySelectorAll('table').forEach(table => {
```
JiuqingSong, line 23: Why need to call to all tables?

new VTable has some cost. So we should avoid use it too much

JiuqingSong (approved):

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/utils/highlightCellHandler.ts
BryanValverdeU: I agree VTable should not be required in this case.
Refactoring so we dont use VTable.

PR #729, opened 2022-01-26
Remove VTable from SelectionRangeEx

BryanValverdeU:
Remove VTable from SelectionRangeEx and include logic to get ranges from table without VTable 

Added test cases for the new functionality in the api.


packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -2,14 +2,12 @@ import moveChildNodes from '../utils/moveChildNodes';
...
+import { TableFormat, TableOperation, TableSelection, VCell } from 'roosterjs-editor-types';
 
 /**
  * A virtual table class, represent an HTML table, by expand all merged cells to each separated cells
  */
-export default class VTable implements Table {
+export default class VTable {
     /**
      * The HTML table object
      */
```
JiuqingSong, line 13: It seems VTable.selection is not used for now. We can remove it. Later if we still need it, we can add it back

BryanValverdeU: Done on latest commit

JiuqingSong: I can see it is still on line 34

JiuqingSong (approved):

PR #732, opened 2022-01-28
Refactor TableResize plugin for zoom

JiuqingSong:
Create 3 table resize handlers:
- TableResizer
- CellResizer
- TableInserter
And a TableEditor class to manager these handlers. Then TableResize plugin only handle event and create/dispose TableEditor.

Test case need to be fixed, will be done in next change.


packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
```diff
@@ -0,0 +1,207 @@
...
+        return this.isChanged;
+    }
+
+    onMouseMove(x: number, y: number) {
+        for (let i = 0; i < this.table.rows.length; i++) {
+            const tr = this.table.rows[i];
+            let j = 0;
+            for (; j < tr.cells.length; j++) {
+                const td = tr.cells[j];
+                const tdRect = normalizeRect(td.getBoundingClientRect());
```
BryanValverdeU, line 84: I think we need a null conditional here, if table is too long there is an exception

JiuqingSong: sounds good

flyingbee2012:
Thanks @JiuqingSong for the refactoring. Changes look good to me. It looks like currently we have a regression in our sample site where the table resizer is mispositioned, and in your changes that issue is fixed.

flyingbee2012 (approved):

BryanValverdeU (approved):

PR #735, opened 2022-01-28
Add format APIs support Table Selection

BryanValverdeU:
+Modified getSelectionRange to getSelectionRangeEx in multiple places to use selection made when using Table Cell Selection Plugin. 
+Added multiple test cases.


packages/roosterjs-editor-api/lib/utils/execCommand.ts
```diff
@@ -32,6 +38,17 @@ export default function execCommand(editor: IEditor, command: DocumentCommand) {
             });
         }
     } else {
-        editor.addUndoSnapshot(formatter, ChangeSource.Format);
+        editor.addUndoSnapshot(() => {
+            if (selection.type == SelectionRangeTypes.TableSelection) {
```
JiuqingSong, line 42: Can we always use selection.ranges, no matter what selection type is?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -209,11 +209,22 @@ export default class Editor implements IEditor {
...
     ) {
+        const result: HTMLElement[] = [];
         let scope = scopeOrCallback instanceof Function ? QueryScope.Body : scopeOrCallback;
         callback = scopeOrCallback instanceof Function ? scopeOrCallback : callback;
 
-        let range = scope == QueryScope.Body ? null : this.getSelectionRange();
-        return queryElements(this.core.contentDiv, selector, callback, scope, range);
+        let selectionEx = scope == QueryScope.Body ? null : this.getSelectionRangeEx();
+        if (selectionEx) {
+            selectionEx.ranges.forEach(range => {
```
JiuqingSong, line 218: same here

packages/roosterjs-editor-api/lib/utils/execCommand.ts
BryanValverdeU: Sounds good.

packages/roosterjs-editor-core/lib/editor/Editor.ts
BryanValverdeU: Here we are check whether the selection is null, if it is null the QueryScope will be Body type. 
I think we don't need range  if the Scope is body.

packages/roosterjs-editor-api/lib/utils/execCommand.ts
```diff
@@ -32,6 +38,22 @@ export default function execCommand(editor: IEditor, command: DocumentCommand) {
...
+
+                formatter();
+
+                tempRange = range;
+            });
+
+            if (tempRange && selection.type == SelectionRangeTypes.TableSelection) {
+                tempRange.collapse();
+                editor.select(tempRange);
+            }
```
JiuqingSong, line 56: what is this part for?

BryanValverdeU: Since table selection is not using actual browser range if we keep this, the last range of the table is going to be selected.
I think it could be a little confusing to the user if they see it, but we can remove it.

image

JiuqingSong (approved):

PR #742, opened 2022-02-02
Change how we add styles for Table Cell Selection

BryanValverdeU:
Changing how we are adding the styles when selecting table.

Using a style tag as recommended.

Most of the previously added Table cell selection tests are broken after this change, will fix them on future PR

+Added a property in the domEvent Plugin so we can save the table selection data.
+Modified the SelectionRangePath when in ShadowEdit from Path to Path[], that way if is table selection we just add the ranges from the added domEvent property.
+New Editor API so we can communicate with domEvent Plugin and save Table Selection (setTableSelection)
+Fixed tests from before Table Selection.

+Removed the styles used from Table Selection and now using new API.

asd

Splitted this PR in 3 for easier review

#748 
#749 
#750


packages/roosterjs-editor-types/lib/interface/IEditor.ts
```diff
@@ -298,6 +299,11 @@ export default interface IEditor {
      */
     select(path: SelectionPath): boolean;
 
+    /**
+     * Select content using the Table Selection
+     */
+    setTableSelection(table: HTMLTableElement, coordinates: TableSelection): boolean;
```
JiuqingSong, line 305: Prefer reuse existing select API, to let it accept TableSelection as parameter

packages/roosterjs-editor-types/lib/event/ShadowEditEvent.ts
```diff
@@ -14,7 +14,7 @@ export interface EnterShadowEditEvent extends BasePluginEvent<PluginEventType.En
     /**
      * The selection path of original editor content
      */
-    selectionPath: SelectionPath;
+    selectionPath: SelectionPath[];
```
JiuqingSong, line 17: Do not change existing var definition, you can add additional property TableSelection

packages/roosterjs-editor-types/lib/corePluginState/LifecyclePluginState.ts
```diff
@@ -45,5 +45,5 @@ export default interface LifecyclePluginState {
     /**
      * Cached selection path for original content
      */
-    shadowEditSelectionPath: SelectionPath;
+    shadowEditSelectionPath: SelectionPath[];
```
JiuqingSong, line 48: Same here, do not change existing var def

packages/roosterjs-editor-dom/lib/selection/selectTable.ts
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * Select a table
+ * @param table table to select
+ * @param coordinates first and last cell of the selection, if this parameter is null, instead of
+ * selecting, will unselect the table.
+ * @param document owner document of the table, if added will append a style element with styles to add to the table
+ * @returns TableSelectionRange if selected a new table, if unselecting the table, null
+ */
+export default function selectTable(
```
JiuqingSong, line 14: Suggest to put this into core package, and make it a core api, can either integrate together with select, or a separate one

packages/roosterjs-editor-dom/lib/selection/selectTable.ts
```diff
@@ -0,0 +1,158 @@
...
+    table: HTMLTableElement,
+    coordinates?: TableSelection,
+    document?: Document
```
JiuqingSong, line 17: no need to pass in document

packages/roosterjs-editor-dom/lib/selection/selectTable.ts
```diff
@@ -0,0 +1,158 @@
...
+    return { firstCell: newFirst, lastCell: newLast };
+}
+
+function min(input1: number, input2: number) {
+    return input1 > input2 ? input2 : input1;
+}
+
+function max(input1: number, input2: number) {
+    return input1 < input2 ? input2 : input1;
+}
```
JiuqingSong, line 138: use Math.min, Math.max

JiuqingSong:
Is there any existing code we can remove since the selection functionality is moved into editor core?

BryanValverdeU:
> Is there any existing code we can remove since the selection functionality is moved into editor core?

Probably most of the utils inside of the table selection will be removed if we can do this change.

PR #746, opened 2022-02-04
fix conflict between cellshade and applyFormat

juliaroldi:
- Fix conflict between cell shade feature and apply format. 
cellShade
cellShade4


packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -120,8 +122,9 @@ export default class VTable {
                 });
             });
             if (this.formatInfo) {
```
BryanValverdeU, line 124: I think if you modify this if to
`if(this.formatInfo && !isResized)`
we wont apply format when resizing. And no need to modify applyFormat nor add shouldApplyFormatToCell

juliaroldi: The cell shade is also been replace when we edit the table and after editing the table we need apply the styles and 
 do not replace the cell shade.
cellShade2

packages/roosterjs-editor-dom/lib/utils/tableFormatInfo.ts
```diff
@@ -23,7 +10,18 @@ const DEFAULT_FORMAT: TableFormat = {
  */
 export function getTableFormatInfo(table: HTMLTableElement) {
```
JiuqingSong, line 11: this file should be moved to table folder

packages/roosterjs-editor-dom/lib/utils/applyTableFormat.ts
```diff
@@ -0,0 +1,299 @@
...
+const TABLE_HEADER_TAG_NAME = 'TH';
+const CELL_SHADE = 'cellShade';
+
+/**
+ * @internal
+ * Apply the given table format to this virtual table
+ * @param format Table format to apply
+ */
+export default function applyTableFormat(cells: VCell[][], format: Partial<TableFormat>) {
+    if (format) {
```
JiuqingSong, line 14: format or !format?

Is this code tested?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -105,6 +101,7 @@ export default class VTable {
 
     /**
      * Write the virtual table back to DOM tree to represent the change of VTable
+     * @param isResized if the table was resized, the shaded cell should not be colored
```
JiuqingSong, line 104: remove this comment

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -121,7 +118,6 @@ export default class VTable {
             });
             if (this.formatInfo) {
                 this.applyFormat(this.formatInfo);
```
JiuqingSong, line 120: here we should call applyTableFormat() instead, but don't call the public one

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -131,276 +127,14 @@ export default class VTable {
     /**
      * Apply the given table format to this virtual table
      * @param format Table format to apply
+     * @param isResized if the table was resized, the shaded cell should not be colored
```
JiuqingSong, line 130: remove comment

packages/roosterjs-editor-dom/lib/index.ts
```diff
@@ -50,6 +50,7 @@ export { default as setColor } from './utils/setColor';
 export { default as matchesSelector } from './utils/matchesSelector';
 export { default as createElement, KnownCreateElementData } from './utils/createElement';
 export { default as moveChildNodes } from './utils/moveChildNodes';
+export { saveTableInfo } from './utils/tableFormatInfo';
```
JiuqingSong, line 53: still not prefer export it.

This should be handled all by VTable. You can always let VTable call saveTableInfo() from VTable.applyFormat()

packages/roosterjs-editor-api/lib/table/insertTable.ts
```diff
@@ -35,8 +55,9 @@ export default function insertTable(
 
     editor.focus();
     editor.addUndoSnapshot(() => {
+        saveTableInfo(table, format || DEFAULT_FORMAT);
```
JiuqingSong, line 58: This should be moved into VTable.applyFormat()

packages/roosterjs-editor-api/lib/table/formatTable.ts
```diff
@@ -15,11 +18,23 @@ export default function formatTable(
     table = table || (editor.getElementAtCursor('TABLE') as HTMLTableElement);
     if (table) {
         editor.addUndoSnapshot((start, end) => {
+            saveTableInfo(table, format);
             let vtable = new VTable(table);
+            deleteCellShadeDataset(vtable);
```
JiuqingSong, line 23: line 21 and 23 should be moved into VTable.applyFormat()

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -41,11 +39,6 @@ export default class VTable {
      */
     col: number;
 
-    /**
-     * Current format of the table
-     */
-    formatInfo: TableFormat;
```
JiuqingSong, line 47: why remove? We should keep this one

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -119,9 +111,9 @@ export default class VTable {
                     }
                 });
             });
-            if (this.formatInfo) {
-                this.applyFormat(this.formatInfo);
-                saveTableInfo(this.table, this.formatInfo);
+            const formatInfo = this.getTableFormatInfo();
```
JiuqingSong, line 114: Please see the design, we should still get formatTable in constructor, then never read it from table again.

When writeBack(), call applyTableFormat() if tableFormat exists.

JiuqingSong: image

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -133,274 +125,33 @@ export default class VTable {
      * @param format Table format to apply
      */
     applyFormat(format: Partial<TableFormat>) {
-        if (!this.table) {
+        if (!this.table || !format) {
```
JiuqingSong, line 128: what if format is incomplete? We can keep the default format here and use it when applyFormat is called.

applyFormat() can pass in a partial format object, then we need to combine it with default format and existing format to make sure it has a complete value:

```ts
format = { ...defaultFormat, ...(this.formatInfo || {}), ...(format || {}) };
```

In this way it is guaranteed existing format (if any) is respected and new format is complete

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -133,274 +125,33 @@ export default class VTable {
...
         }
-
-        this.formatInfo = format ? format : this.formatInfo;
-        this.table.style.borderCollapse = 'collapse';
-        this.setBordersType(this.formatInfo);
-        this.setColor(this.formatInfo);
-        this.setFirstColumnFormat(this.formatInfo);
-        this.setHeaderRowFormat(this.formatInfo);
+        this.deleteCellShadeDataset(this.cells);
+        applyTableFormat(this.table, this.cells, format);
```
JiuqingSong, line 132: you can call writeBack() to trigger the applyTableFormat() call. See the design chart.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -133,274 +125,33 @@ export default class VTable {
...
+        saveTableInfo(this.table, format);
```
JiuqingSong, line 133: same, this can be done in writeBack()

packages/roosterjs-editor-dom/lib/table/VTable.ts
juliaroldi: We should save the format before we create the table, otherwise formatInfo will be undefined, then we cannot not change the state of the table.

JiuqingSong: That is right. And that is why you need to read it in constructor, and save into this.tableFormat.

JiuqingSong (approved):

PR #749, opened 2022-02-07
Change how we add Style in Table Selection (Step 2)

BryanValverdeU:
Add a new Core and Editor API to handle the selection.


packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -0,0 +1,206 @@
...
+            ranges.push(rowRange);
+        }
+    }
+
+    css += '{background-color: rgba(198,198,198,0.7) !important;}';
+
+    return { css, ranges };
+}
+
+function select(table: HTMLTableElement, coordinates: TableSelection, core: EditorCore): Range[] {
```
JiuqingSong, line 119: suggest make core as the first parameter, to follow the same pattern with other functions

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -406,6 +406,12 @@ export default class Editor implements IEditor {
     }
 
     public select(arg1: any, arg2?: any, arg3?: any, arg4?: any): boolean {
+        if ((<HTMLTableElement>arg1)?.rows != undefined) {
+            const selection = this.core.api.selectTable(this.core, arg1, arg2);
+            this.core.domEvent.tableSelectionRange = selection;
+            return;
+        }
```
JiuqingSong, line 413: question: how can I unselect a table?

We should always unselect existing table if any when we call select. Say currently  a table is selected, then I want to select something else, then we can call selectTable() with null parameter to unselect first, then call selectRange(), same for the other way

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -0,0 +1,206 @@
...
+ * @returns true if successful
+ */
+export const selectTable: SelectTable = (
+    core: EditorCore,
+    table: HTMLTableElement,
+    coordinates?: TableSelection
+) => {
+    const doc = core.contentDiv.ownerDocument;
+
+    if (coordinates) {
```
JiuqingSong, line 29: When table or cooridinates is null, we can just unselect then exist. So that we can always unselect then reselect from Editor class, and in this core API we just need to do either select or unselect

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -0,0 +1,206 @@
...
+function select(table: HTMLTableElement, coordinates: TableSelection, core: EditorCore): Range[] {
+    const doc = core.contentDiv.ownerDocument;
+    const contentDivId = core.contentDiv.id;
+
+    const contentDivSelector = '#' + contentDivId;
+    let { css, ranges } = buildCss(table, coordinates, contentDivSelector);
+    if (doc) {
+        unselect();
+        if (styleElement && styleElement.ownerDocument != doc) {
+            styleElement.parentElement.removeChild(styleElement);
```
JiuqingSong, line 128: When there are multiple roosterjs editor in the same page, this will impact other editor.

We can just got those CSS rules for this editor and delete them, but keep this STYLE tag

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -0,0 +1,206 @@
...
+    coordinates: TableSelection,
+    contentDivSelector: string
+): { css: string; ranges: Range[] } {
+    coordinates = normalizeTableSelection(coordinates);
+    const tr1 = coordinates.firstCell.y + 1;
+    const td1 = coordinates.firstCell.x + 1;
+    const tr2 = coordinates.lastCell.y + 1;
+    const td2 = coordinates.lastCell.x + 1;
+    const ranges: Range[] = [];
+    const tbodySelector = table.querySelector('tbody') ? '>tbody' : '';
```
JiuqingSong, line 62: From my test, no matter if there is tbody, the ">tbody" in the selector always works.

PS, also need to consider THead and TFoot scenario

JiuqingSong: and better not use querySelector here since if there is nested table inside, it will also be selected

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -0,0 +1,206 @@
...
+                table.id +
+                tbodySelector +
+                ' >tr:nth-child(' +
+                i +
+                ')>td:nth-child(' +
+                j +
+                ')';
+            if (table.querySelector(selector)) {
+                css += selector;
+            } else {
```
JiuqingSong, line 92: not sure if something like  (td,th):nth-child() works. We can try.

But if it not work, the index of TD need to be adjusted if there is TH before.

And what if there are multiple TH? Although it is not per standard. Maybe we can change all further TH to TD? What do you think

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -0,0 +1,206 @@
...
+                selector = selector.replace('td', 'th');
+                if (table.querySelector(selector)) {
```
JiuqingSong, line 94: Maybe we don't need this check, but just add whatever we have. Because if it fails the check, we still can do nothing

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -0,0 +1,206 @@
...
+        const lastSelector = `${rowSelector}td:nth-child(${td2}),${rowSelector}th:nth-child(${td2})`;
+
+        firstSelected = table.querySelector(firstSelector);
+        lastSelected = table.querySelector(lastSelector)!;
+        if (firstSelected && lastSelected) {
+            const rowRange = new Range();
+            rowRange.setStartBefore(firstSelected);
+            rowRange.setEndAfter(lastSelected);
+            ranges.push(rowRange);
+        }
```
JiuqingSong, line 111: what is this part for?

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+}
+
+/**
+ * Make the first Cell of a table selection always be on top of the last cell.
+ * @param input Table selection
+ * @returns Table Selection where the first cell is always going to be first selected in the table
+ * and the last cell always going to be last selected in the table.
+ */
+function normalizeTableSelection(input: TableSelection): TableSelection {
```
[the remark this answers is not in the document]
BryanValverdeU: I think we do that in tableCellSelection plugin, but might as well add it here since this is going to be a api that is going to be usable from all the Editor.

I will add this.

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
BryanValverdeU: Ah I see, good catch.
After doing some research seems like adding tr: nth-child() > *:nth-child() works fine.
Since we are checking we are under a TR this should not fail, unless there is a really weird scenario.
What do you think?

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
BryanValverdeU: This is to return the ranges and save the info to the property in domEvent, where we save the tableSelection.

packages/roosterjs-editor-core/lib/editor/Editor.ts
BryanValverdeU: I see, adding this if the first param in select api is not table, unselect the table selection.

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
BryanValverdeU: Added some checks to see whether the table contains a tbody, thead and tfoot from line 70 to 94

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -406,6 +406,14 @@ export default class Editor implements IEditor {
     }
 
     public select(arg1: any, arg2?: any, arg3?: any, arg4?: any): boolean {
+        if ((<HTMLTableElement>arg1)?.rows != undefined) {
```
JiuqingSong, line 409: although this check also work, but suggest to make a more robust check like 
```ts
if (!!(<HTMLTableElement>arg1)?.rows) { ...
```

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -0,0 +1,202 @@
...
+    const td2 = coordinates.lastCell.x + 1;
+    const ranges: Range[] = [];
+
+    let firstSelected: HTMLTableCellElement | null = null;
+    let lastSelected: HTMLTableCellElement | null = null;
+    let css = '';
+    let isFirst = true;
+
+    // Get whether table has thead, tbody or tfoot.
+    const tableChilds = toArray(table.childNodes).filter(
```
JiuqingSong, line 66: tableChildren

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -0,0 +1,202 @@
...
+    SelectionRangeTypes,
+    TableSelection,
+    SelectTable,
+    Coordinates,
+} from 'roosterjs-editor-types';
+
+const TABLE_ID = 'tableSelected';
+const CONTENT_DIV_ID = 'contentDiv_';
+
+let styleElement: HTMLStyleElement;
```
JiuqingSong, line 13: having a global reference of STYLE node can be a problem when there is projection popout, because each window need its own STYLE node, but a global one stops creating more STYLE node for child windows. And currently the code will remove all style rules from the STYLE tag when unselect, that will also remove rules from other editor in the same window

packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -0,0 +1,202 @@
...
+    let isFirst = true;
+
+    // Get whether table has thead, tbody or tfoot.
+    const tableChilds = toArray(table.childNodes).filter(
+        node => ['THEAD', 'TBODY', 'TFOOT'].indexOf(getTagOfNode(node)) > -1
+    );
+
+    // Set the start and end of each of the table childs, so we can build the selector according the element between the table and the row.
+    let cont = 0;
+    const indexes = tableChilds.map(node => {
```
JiuqingSong, line 72: The calculation of indexes can be an issue when table has merged cells. We need VTable to help get correct TD/TH here. And I think the easiest way to handle multiple THEAD/TBODY/TFOOT/TH is to just make a full loop when we generate the CSS

JiuqingSong:
Suggest add test case for the CSS generation part to test different kinds of TABLE

JiuqingSong (approved):

PR #750, opened 2022-02-07
Change how we add Style in Table Selection (Step 3)

BryanValverdeU:
Add changes to Table Cell Selection and some Editor APIs
Remove utils that are going to be no longer needed.

Modify multiple functions to support the new Table Selection

Splitting #742 for easier review


packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/utils/forEachSelectedCell.ts
```diff
@@ -10,18 +10,13 @@ import { VTable } from 'roosterjs-editor-dom';
  */
 export function forEachSelectedCell(vTable: VTable, callback: (cell: VCell) => void): number {
     let selectedCells = 0;
+    vTable.selection = normalizeTableSelection(vTable.selection, vTable.table);
```
JiuqingSong, line 13: Suggest not change VTable for this function since it looks like just a reader, but not a writer.

You can do either
1. When set selection, already normalize it. or
2. Save the normalize result to a local var

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/utils/forEachSelectedCell.ts
```diff
@@ -10,18 +10,13 @@ import { VTable } from 'roosterjs-editor-dom';
...
 
     const { lastCell, firstCell } = vTable.selection;
 
     for (let y = 0; y < vTable.cells.length; y++) {
```
JiuqingSong, line 17: then why not just
for (i from selecton.row1 to row2)
  for (j from selection.col1 to col2) ...

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/utils/forEachSelectedCell.ts
```diff
@@ -10,18 +10,13 @@ import { VTable } from 'roosterjs-editor-dom';
...
-            let element = vTable.cells[y][x].td as HTMLElement;
-            if (
-                element?.classList.contains(tableCellSelectionCommon.TABLE_CELL_SELECTED) ||
-                (((y >= firstCell.y && y <= lastCell.y) || (y <= firstCell.y && y >= lastCell.y)) &&
-                    ((x >= firstCell.x && x <= lastCell.x) ||
-                        (x <= firstCell.x && x >= lastCell.x)))
-            ) {
+            if (y >= firstCell.y && y <= lastCell.y && x >= firstCell.x && x <= lastCell.x) {
                 selectedCells += 1;
                 callback(vTable.cells[y][x]);
```
JiuqingSong, line 21: It is possible this cell doesn't have TD. is that ok?

packages/roosterjs-editor-dom/lib/utils/normalizeTableSelection.ts
```diff
@@ -0,0 +1,31 @@
...
+    };
+    let newLast = {
+        x: Math.max(firstCell.x, lastCell.x),
+        y: Math.max(firstCell.y, lastCell.y),
+    };
+
+    const checkIfExists = (coord: Coordinates) => table?.rows?.item(coord.y)?.cells.item(coord.x);
+
+    if (!checkIfExists(newFirst) || !checkIfExists(newLast)) {
+        throw new Error('Table selection provided is not valid');
```
JiuqingSong, line 27: We can just ignore it here, no need to throw

JiuqingSong: Just make sure the return value fall into the table is good enough.

Need to consider if the range is totally outside the table, what we should do?

packages/roosterjs-editor-dom/lib/index.ts
```diff
@@ -50,6 +50,7 @@ export { default as setColor } from './utils/setColor';
 export { default as matchesSelector } from './utils/matchesSelector';
 export { default as createElement, KnownCreateElementData } from './utils/createElement';
 export { default as moveChildNodes } from './utils/moveChildNodes';
+export { default as normalizeTableSelection } from './utils/normalizeTableSelection';
```
JiuqingSong, line 53: If we can guarantee the selection is always normalized before assign to anywhere, do we still need to export this function?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -409,6 +409,12 @@ export default class Editor implements IEditor {
         if (!!(<HTMLTableElement>arg1)?.rows) {
             const selection = this.core.api.selectTable(this.core, arg1, arg2);
             this.core.domEvent.tableSelectionRange = selection;
+
+            const range = new Range();
+            range.selectNodeContents(<HTMLTableElement>arg1);
+            range.collapse(true);
+
+            this.core.api.selectRange(this.core, range);
```
JiuqingSong, line 417: this can be simplified to 
```ts
this.core.api.selectRange(createRange(new Position(arg1, PositionType.Begin)));
```

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -84,6 +79,9 @@ export default class TableCellSelection implements EditorPlugin {
      * Dispose this plugin
      */
     dispose() {
+        if (this.firstTable) {
+            this.editor.select(this.firstTable, null);
```
JiuqingSong, line 83: Do we always need to pass in the table object even we want to unselect?

BryanValverdeU: AH yes, we can just pass in a null parameter to remove the styles added if any.
Doing that change

packages/roosterjs-editor-dom/lib/utils/normalizeTableSelection.ts
BryanValverdeU: Sounds good, will do that

JiuqingSong (approved):

PR #751, opened 2022-02-10
8.15.0 Fix table resize bug

JiuqingSong:
Due to a conflict with table format feature, table cell is not working now. Fix it by preserving the same node when changeElementTag to the same tag name


packages/roosterjs-editor-dom/lib/utils/changeElementTag.ts
```diff
@@ -26,6 +26,13 @@ export default function changeElementTag(element: HTMLElement, newTag: string):
         return null;
     }
 
+    const origianlTag = getTagOfNode(element);
```
BryanValverdeU, line 29: Change to original

BryanValverdeU (approved):

JiuqingSong:
#1145

PR #754, opened 2022-02-11
Fix predefined styles in darkMode

juliaroldi:
Fix predefined styles behavior when switch from dark to light mode.


BryanValverdeU (approved):

packages/roosterjs-editor-core/lib/coreApi/transformColor.ts
```diff
@@ -1,128 +1,130 @@
...
-        const allChildren = rootNode.getElementsByTagName('*');
-        arrayPush(result, toArray(allChildren));
-    } else if (safeInstanceOf(rootNode, 'DocumentFragment')) {
-        const allChildren = rootNode.querySelectorAll('*');
-        arrayPush(result, toArray(allChildren));
-    }
-
-    return result;
-}
+import { arrayPush, getComputedStyles, safeInstanceOf, toArray } from 'roosterjs-editor-dom';
```
JiuqingSong, line 1: What is the changed part in this file? it is hard to review

packages/roosterjs-editor-core/lib/coreApi/transformColor.ts
```diff
@@ -1,128 +1,130 @@
...
+    elementsToTransform.forEach(
+        element => element?.dataset && element.style && transformFunction(element)
+    );
+};
+
+function transformToLightMode(element: HTMLElement) {
+    ColorAttributeName.forEach(names => {
+        // Reset color styles based on the content of the ogsc/ogsb data element.
+        // If those data properties are empty or do not exist, ignore them.
+        if (element.dataset[names[ColorAttributeEnum.CssDataSet]]) {
```
JiuqingSong, line 65: do you still need this change?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -29,6 +29,11 @@ const DEFAULT_FORMAT: TableFormat = {
     tableBorderFormat: TableBorderFormat.DEFAULT,
 };
 
+/**
+ * Check if the VTable is being created in darkMode, if yes, apply the styles for darkMode
+ */
+let isVTableInDarkMode: boolean;
```
JiuqingSong, line 35: why global variable?

packages/roosterjs-editor-dom/lib/table/tableFormatInfo.ts
```diff
@@ -61,8 +61,20 @@ function checkIfTableFormatIsValid(format: TableFormat) {
     return true;
 }
 
+function checkIfColorIsModeIndependentColor(color: any) {
+    if (color.lightModeColor && color.darkModeColor) {
```
JiuqingSong, line 65: if color is a number, or some other type, this may throw?

Need to be safer

packages/roosterjs-editor-dom/lib/utils/applyTableFormat.ts
```diff
@@ -13,16 +19,17 @@ const CELL_SHADE = 'cellShade';
 export default function applyTableFormat(
     table: HTMLTableElement,
     cells: VCell[][],
-    format: Partial<TableFormat>
+    format: Partial<TableFormat>,
```
JiuqingSong, line 22: For this function, it should accept Required`<TableFormat>`

For the functions face to other developer, we can loose the type and accept `Partial<>`, but for internal API that only called by our code, we need to make it strong-typed and only accept well-formed data.

packages/roosterjs-editor-dom/lib/utils/applyTableFormat.ts
```diff
@@ -286,10 +310,19 @@ function setHeaderRowFormat(cells: VCell[][], format: TableFormat) {
     }
     cells[0]?.forEach(cell => {
         if (cell.td && format.headerRowColor) {
-            cell.td.style.backgroundColor = format.headerRowColor;
-            cell.td.style.borderRightColor = format.headerRowColor;
-            cell.td.style.borderLeftColor = format.headerRowColor;
-            cell.td.style.borderTopColor = format.headerRowColor;
+            setColor(cell.td, format.headerRowColor, true, isDarkMode);
+            cell.td.style.borderRightColor = setHeaderRowBorderColors(
```
JiuqingSong, line 314: actually our color transformation code won't transform border color. So let's always use light mode color here. Otherwise it is another bug

juliaroldi: If we do not change the border color for the header row, we can see a difference between the background aborder color. 
Screen Shot 2022-02-21 at 16 54 19

juliaroldi: Do we use always light mode for header row background color?

packages/roosterjs-editor-core/lib/coreApi/transformColor.ts
juliaroldi: Yes, because we need to set the color for light mode, if we remove this code, when w switch back to light mode, the dark to light transform will not happen.

packages/roosterjs-editor-dom/lib/utils/applyTableFormat.ts
JiuqingSong: Background color need to be transformed to dark color. And in theory, we need to do it for border color too. But it is not down in transformColor

packages/roosterjs-editor-dom/lib/utils/applyTableFormat.ts
JiuqingSong: insertTable and formatTable will call VTable.applyFormat(), right? Then VTable.applyFormat() can take the responsibility to full fill this object

packages/roosterjs-editor-dom/lib/utils/tsconfig.child.json
```diff
@@ -3,6 +3,6 @@
         "strict": true
     },
     "extends": "../../../tsconfig.json",
-    "include": ["./**/*.ts"],
+    "include": ["./**/*.ts", "../table/applyTableFormat.ts"],
```
JiuqingSong, line 6: why? Which file causes this dependency?

packages/roosterjs-editor-dom/test/utils/applyTableFormatTest.ts
```diff
@@ -1,4 +1,4 @@
-import applyTableFormat from '../../lib/utils/applyTableFormat';
+import applyTableFormat from '../../lib/table/applyTableFormat';
```
JiuqingSong, line 1: need to move to test/table folder too.

packages/roosterjs-editor-types/lib/interface/TableFormat.ts
```diff
@@ -7,45 +8,45 @@ export default interface TableFormat {
     /**
      * Top border color for each row
      */
-    topBorderColor?: string | null;
+    topBorderColor: string | null;
```
JiuqingSong, line 11: Why need to change this file?
The template class `Required<T>` can make everything required, no need to change this file

packages/roosterjs-editor-core/lib/coreApi/transformColor.ts
JiuqingSong: But the change is to check if there is data-ogsc, and not run if it doesn't have. That means after this change there will be fewer chance to run this code (only when data-ogsc exist).

JiuqingSong:
tableformat

See the GIF. I inserted a table in dark mode, and applied format, it shows correct background color. Then I switch to light mode, the background color is gone.  This is on your branch.

PR #755, opened 2022-02-15
Add zoom scale number support

JiuqingSong:
Add a new way to support zoom scale: instead of passing in a size transformer function, we now support passing in the scale number, as well as setter/getter the scaler number and an event. So that we can change the scale value without recreating editor object. Also deprecate original sizeTransformer function.


packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -879,11 +882,43 @@ export default class Editor implements IEditor {
...
+    getZoomScale(): number {
+        return this.core.zoomScale;
+    }
+
+    /**
+     * Set current zoom scale, default value is 1
+     * When editor is put under a zoomed container, need to pass the zoom scale number using EditorOptions.zoomScale
+     * to let editor behave correctly especially for those mouse drag/drop behaviors
+     */
+    setZoomScale(scale: number): void {
```
BryanValverdeU, line 906: nit: param is missing in the comment

BryanValverdeU (approved):

PR #758, opened 2022-02-15
Add  Table Selector

BryanValverdeU:
image

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
@@ -1,8 +1,9 @@
 import createCellResizer from './CellResizer';
 import createTableInserter from './TableInserter';
 import createTableResizer from './TableResizer';
+import createTableSelector from './TableSelector';
 import TableEditFeature, { disposeTableEditFeature } from './TableEditorFeature';
-import { ChangeSource, IEditor, NodePosition } from 'roosterjs-editor-types';
+import { ChangeSource, IEditor, NodePosition, TableSelection } from 'roosterjs-editor-types';
 import { getComputedStyle, normalizeRect } from 'roosterjs-editor-dom';
 
 const INSERTER_HOVER_OFFSET = 5;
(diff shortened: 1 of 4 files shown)


packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
```diff
@@ -46,6 +47,9 @@ export default class TableEditor {
     // 5 - Resize whole table
     private tableResizer: TableEditFeature;
 
+    // 6 - Select whole table
```
JiuqingSong, line 50: Please also add this one as comment in line 15 to explain what it is.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
```diff
@@ -219,4 +232,24 @@ export default class TableEditor {
         this.disposeTableResizer();
         this.onFinishEditing();
     };
+
+    private onSelect = (table: HTMLTableElement) => {
+        this.editor.focus();
+        if (table) {
+            const rows = table.rows.length - 1;
+            const lastCellIndex = table.rows.item(rows).cells.length - 1;
```
JiuqingSong, line 240: what if last row has merged cell?

suggest use VTable to get column count

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableSelector.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const TABLE_SELECTOR_LENGTH = 12;
+const TABLE_SELECTOR_ID = '_Table_Selector';
+
+/**
+ * @internal
+ */
+export default function createTableSelector(
+    table: HTMLTableElement,
+    sizeTransformer: SizeTransformer,
```
JiuqingSong, line 14: So this will be a little conflict with my PR https://github.com/microsoft/roosterjs/pull/755.

Whoever merge the last needs to fix this part to avoid use sizeTransformer.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
BryanValverdeU: Sounds good adding VTable here.

JiuqingSong (approved):

PR #759, opened 2022-02-15
Fix Issue when Creating a link.

BryanValverdeU:
Fix issue when creating a link.
Exec command is transforming the selecting into multiple anchors instead of one.

To repro use this snapshot

`<div><span style="font-size:medium;box-sizing:border-box;color:rgba(0, 0, 0, 0.9);background-color:rgb(255, 255, 255);font-family:Calibri, Helvetica, sans-serif;display:inline !important">Hello<span style="box-sizing:border-box">&nbsp;</span></span><b style="font-size:medium;box-sizing:border-box;color:rgba(0, 0, 0, 0.9);background-color:rgb(255, 255, 255);font-family:Calibri, Helvetica, sans-serif">world<span style="box-sizing:border-box">&nbsp;</span></b><span style="font-size:medium;box-sizing:border-box;color:rgba(0, 0, 0, 0.9);background-color:rgb(255, 255, 255);margin:0px;font-family:Calibri, Helvetica, sans-serif">🙂</span><span style="font-size:medium;box-sizing:border-box;color:rgba(0, 0, 0, 0.9);background-color:rgb(255, 255, 255);font-family:Calibri, Helvetica, sans-serif;display:inline !important">&nbsp;this is a test</span><br></div><!--{"start":[0,3,0,15],"end":[0,3,0,15]}-->`

1. Select first two words.
2. Create link.
3. Instead of only one anchor element created there are two anchor elements.

Fix done:
Traverse the selection and remove additional elements after the first element.


packages/roosterjs-editor-api/lib/format/createLink.ts
```diff
@@ -83,6 +83,13 @@ export default function createLink(
             } else {
                 // the selection is not collapsed, use browser execCommand
                 editor.getDocument().execCommand(DocumentCommand.CreateLink, false, normalizedUrl);
+                const traverser = editor.getSelectionTraverser();
+                let currentInline = traverser.getNextInlineElement();
+                while (currentInline) {
+                    const a = currentInline.getContainerNode();
+                    a.parentElement.removeChild(a);
```
JiuqingSong, line 90: So this looks like a bug from execCommand, it creates extra node, is that right?

Then this code can only work if this bug exists. If somehow browser fixed the bug, (or try other browsers like Firefox, Safari, Opera, see if they have the same issue), then this code may become another bug?

BryanValverdeU: I see, debugging this.

BryanValverdeU: I tried in multiple browsers and the bug that is causing execCommand to create multiple anchor elements instead of 1 is in Edge, Chrome, Opera and Firefox.

With the fix in this branch it is fixing the issue in all the above browsers.

If there is a fix for this bug in execCommand, I think this code should not create another issue since we only remove the elements that are in selection and that are after the first anchor element. 

What do you think?

JiuqingSong (approved):

PR #771, opened 2022-02-21
Fix onFocus Event for TableSelection

BryanValverdeU:
When onFocus we need to re-select the table selection if any, if we dont, there is going to be a conflict with the previous code, since it is going to remove the table selection.


packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -31,6 +31,14 @@ export const selectTable: SelectTable = (
         ensureUniqueId(core.contentDiv, CONTENT_DIV_ID);
 
         const ranges = select(core, table, coordinates);
+
+        const tempRange = new Range();
+        tempRange.setStart(
+            table.rows.item(coordinates.firstCell.y).cells.item(coordinates.firstCell.x),
+            0
+        );
```
JiuqingSong, line 39: use createRange()

packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
```diff
@@ -144,8 +144,13 @@ export default class DOMEventPlugin implements PluginWithState<DOMEventPluginSta
     };
 
     private onFocus = () => {
+        const { table, coordinates } = this.state.tableSelectionRange || {};
         this.editor.select(this.state.selectionRange);
```
JiuqingSong, line 148: so if table selection exists, we should not call this line

JiuqingSong (approved):

PR #772, opened 2022-02-21
U/juliaroldi/table align

juliaroldi:
Align table using setAlignment API. 

tableAlignStyle


packages/roosterjs-editor-api/lib/format/setAlignment.ts
```diff
@@ -3,34 +3,127 @@ import {
...
+ * @param addUndoSnapshot check if this function is being called by addUndoSnapshot
+ */
+function alignElement(
+    editor: IEditor,
+    element: HTMLElement,
+    elementType: string,
+    alignment: Alignment,
+    addUndoSnapshot?: boolean
+) {
+    if (elementType === TABLE && editor.isFeatureEnabled(ExperimentalFeatures.TableAlignment)) {
```
BryanValverdeU: We can check whether the whole table is selected using the editor api getSelectionRangeEx().

```
const selection = editor.getSelectionRangeEx();
if (selection.type == SelectionRangeTypes.TableSelection) {
 const vTable = new VTable(selection.table)
 const { firstCell, lastCell } = selection.coordinates;
    const rowsLength = vTable.cells.length - 1;
    const colIndex = vTable.cells[rowsLength].length - 1;
```

    const firstX = firstCell.x;
    const firstY = firstCell.y;
    const lastX = lastCell.x;
    const lastY = lastCell.y;

    const selectedAllTable = firstX == 0 && firstY == 0 && lastX == colIndex && lastY == rowsLength;
```
 if(selectedAllTable) { alignAllTable() }
}
```

juliaroldi: I see, if I select only some cells, the editor.getElementAtCursor returns Table. I will fix that.

packages/roosterjs-editor-api/lib/format/setAlignment.ts
```diff
@@ -1,36 +1,146 @@
...
+    if (selection && selectionType === SelectionRangeTypes.Normal) {
+        return TEXT;
+    } else if (
+        editor.isFeatureEnabled(ExperimentalFeatures.TableAlignment) &&
+        selection &&
+        selectionType === SelectionRangeTypes.TableSelection &&
+        isWholeTableSelected(selection)
+    ) {
+        return TABLE;
+    }
```
JiuqingSong, line 55: else?

You can simplify your function to be:

return (type == table && isFeatureEnabled && isWholeTableSelected);

Making it returning a boolean should be easier to use

packages/roosterjs-editor-api/lib/format/setAlignment.ts
```diff
@@ -1,36 +1,146 @@
...
+    elementType: string,
+    alignment: Alignment,
+    addUndoSnapshot?: boolean
+) {
+    if (elementType === TABLE) {
+        alignTable(editor, element, alignment);
+    } else {
+        alignText(editor, element, alignment, addUndoSnapshot);
+    }
+}
```
JiuqingSong, line 78: I think all 3 functions above can be merged together:

```ts
function setAlignment(...) {
  if (type == table && isFeatureEnabled) {
    alignTable(..);
  } else {
    alignText(...);
}
```

packages/roosterjs-editor-api/lib/format/setAlignment.ts
```diff
@@ -1,36 +1,146 @@
...
 export default function setAlignment(editor: IEditor, alignment: Alignment) {
-    let command = DocumentCommand.JustifyLeft;
-    let align = 'left';
+    const element = editor.getElementAtCursor();
+    const elementType = isATableOrText(element, editor);
+
+    editor.addUndoSnapshot(() => {
+        alignElement(editor, element, elementType, alignment);
+        editor.queryElements('[align]', QueryScope.OnSelection, node =>
+            alignElement(editor, node, elementType, alignment, true /** addUndoSnapshot */)
```
JiuqingSong, line 30: why we need to call alignElement twice in this funciton?

packages/roosterjs-editor-api/lib/format/setAlignment.ts
```diff
@@ -1,36 +1,146 @@
...
         command = DocumentCommand.JustifyCenter;
         align = 'center';
     } else if (alignment == Alignment.Right) {
         command = DocumentCommand.JustifyRight;
         align = 'right';
     }
-
-    editor.addUndoSnapshot(() => {
+    if (addUndoSnapshot) {
+        element.style.textAlign = align;
```
JiuqingSong, line 125: why "addUndosnapshot" can determin if we want to set textAlign?

packages/roosterjs-editor-api/lib/format/setAlignment.ts
```diff
@@ -1,36 +1,146 @@
...
-    }, ChangeSource.Format);
+    }
+}
+
+/**
+ * Check if the whole table is selected
+ * @param selection
+ * @returns
+ */
+function isWholeTableSelected(selection: TableSelectionRange) {
```
JiuqingSong, line 136: what if I select more than whole table?

juliaroldi: Only the text is aligned. 
tableAndText

JiuqingSong: I think we should center-align both text outside the table and the table itself

JiuqingSong: Ok, I checked the behavior or Word, it is the same with your current implementation. So I think it is fine here.

packages/roosterjs-editor-api/lib/format/setAlignment.ts
```diff
@@ -14,23 +18,77 @@ import {
...
 export default function setAlignment(editor: IEditor, alignment: Alignment) {
-    let command = DocumentCommand.JustifyLeft;
-    let align = 'left';
+    const element = editor.getElementAtCursor();
+    const selection = editor.getSelectionRangeEx();
+    const selectionType = selection.type;
+    editor.addUndoSnapshot(() => {
+        if (
+            editor.isFeatureEnabled(ExperimentalFeatures.TableAlignment) &&
+            selection &&
```
JiuqingSong, line 27: this need to be moved before line 23. you already used selection before check it

packages/roosterjs-editor-api/lib/format/setAlignment.ts
```diff
@@ -14,23 +18,77 @@ import {
...
 
+/**
+ * Align text using the margins
+ * @param editor
+ * @param element
+ * @param alignment
+ * @param addUndoSnapshot
+ * @returns
+ */
+function alignTable(editor: IEditor, element: HTMLElement, alignment: Alignment) {
```
JiuqingSong, line 46: editor is not used

BryanValverdeU (approved):

JiuqingSong (approved):

PR #774, opened 2022-02-22
fix position for images

juliaroldi:
Bug: Image was being inserted inside a text element. [Bug 139624](https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/139624/)
Repro:  
1. Type something
2. Highlight the text
3. Insert a inline image
4. Image is highlighted

Expected:
Image not highlighted. 

After fix: 
insertImage

### DIFF
--- a/packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
+++ b/packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
@@ -243,7 +243,7 @@ function adjustInsertPositionForMoveCursorOutOfALink(
 }
 
 /**
- * Adjust the position cursor position when a image is inserted to prevent it be inserted into another element.
+ * Adjust the cursor position when a image is inserted to prevent it be inserted into another element.
  */
 function adjustInsertPositionForImages(
     root: HTMLElement,
--- a/packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
+++ b/packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
@@ -16,8 +16,14 @@ import toArray from '../utils/toArray';
 import unwrap from '../utils/unwrap';
 import VTable from '../table/VTable';
 import wrap from '../utils/wrap';
-import { NodePosition, NodeType, PositionType, QueryScope } from 'roosterjs-editor-types';
 import { splitBalancedNodeRange } from '../utils/splitParentNode';
+import {
+    InlineElement,
+    NodePosition,
+    NodeType,
+    PositionType,
+    QueryScope,
+} from 'roosterjs-editor-types';
 
 const adjustSteps: ((
     root: HTMLElement,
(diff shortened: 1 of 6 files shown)


packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -235,9 +242,38 @@ function adjustInsertPositionForMoveCursorOutOfALink(
...
+    nodeToInsert: Node,
+    position: NodePosition,
+    range: Range
+): NodePosition {
+    const isImage = nodeToInsert.nodeName === 'IMG' ? true : false;
+    if (range && range.collapsed && isImage) {
+        const searcher = new PositionContentSearcher(root, Position.getStart(range));
+        const inlineElementBefore = searcher.getInlineElementBefore();
+        const inlineElementAfter = searcher.getInlineElementAfter();
+        const containerNodeType = (inlineElement: InlineElement) =>
```
JiuqingSong, line 259: Any reason we  need to have this inline function? Can we just do calculation directly instead?

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -235,9 +242,36 @@ function adjustInsertPositionForMoveCursorOutOfALink(
...
+    nodeToInsert: Node,
+    position: NodePosition,
+    range: Range
+): NodePosition {
+    const isImage = nodeToInsert.nodeName === 'IMG' ? true : false;
+    if (range && range.collapsed && isImage) {
+        const searcher = new PositionContentSearcher(root, Position.getStart(range));
+        const inlineElementBefore = searcher.getInlineElementBefore();
+        const inlineElementAfter = searcher.getInlineElementAfter();
+        const adjustedPosition = (inlineElement: InlineElement, positionType: PositionType) =>
```
JiuqingSong, line 259: I'm not a fan of this kind of inline function. Make it a regular function if we can.

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -235,9 +242,36 @@ function adjustInsertPositionForMoveCursorOutOfALink(
...
+/**
+ * Adjust the cursor position when a image is inserted to prevent it be inserted into another element.
+ */
+function adjustInsertPositionForImages(
+    root: HTMLElement,
+    nodeToInsert: Node,
+    position: NodePosition,
+    range: Range
+): NodePosition {
+    const isImage = nodeToInsert.nodeName === 'IMG' ? true : false;
```
JiuqingSong, line 254: When you want to get a boolean value, just directly retrieve from the condition, no need to have "true : false" at the end:

```ts
const isImage = nodeToInsert.nodeName === 'IMG';
```

And You can call getTagOfNode() to make it safer. It can handle all kind of input node including null.

JiuqingSong:
With this change, I can never insert image in the same line of text.

And if I try to insert image in middle of some text, it will also insert to next line.

PR #778, opened 2022-02-24
WIP: new color change event plugin

juliaroldi:
1-Do we use contentDiv, when the plugin is trigger by editor and not by a specific api? 
2- How do we use editor to trigger the predefined styles color change?

### DIFF
--- a/packages/roosterjs-editor-core/lib/coreApi/getContent.ts
+++ b/packages/roosterjs-editor-core/lib/coreApi/getContent.ts
@@ -1,3 +1,4 @@
+import ColorChangedEvent from 'roosterjs-editor-types/lib/event/ColorChangedEvent';
 import {
     ColorTransformDirection,
     EditorCore,
@@ -48,13 +49,14 @@ export const getContent: GetContent = (core: EditorCore, mode: GetContentMode): 
         const range = path && createRange(clonedRoot, path.start, path.end);
 
         if (core.lifecycle.isDarkMode) {
-            core.api.transformColor(
-                core,
-                clonedRoot,
-                false /*includeSelf*/,
-                null /*callback*/,
-                ColorTransformDirection.DarkToLight
-            );
+            const event: ColorChangedEvent = {
+                eventType: PluginEventType.ColorChanged,
+                rootNode: clonedRoot,
+                includeSelf: false,
+                callback: null,
+                direction: ColorTransformDirection.DarkToLight,
+            };
+            core.api.triggerEvent(core, event, false);
         }
 
         if (triggerExtractContentEvent) {
--- a/packages/roosterjs-editor-core/lib/coreApi/transformColor.ts
+++ b/packages/roosterjs-editor-core/lib/coreApi/transformColor.ts
@@ -61,12 +61,14 @@ export const transformColor: TransformColor = (
 function transformToLightMode(element: HTMLElement) {
     ColorAttributeName.forEach(names => {
         // Reset color styles based on the content of the ogsc/ogsb data element.
-        // If those data properties are empty or do not exist, set them anyway to clear the content.
-        element.style.setProperty(
-            names[ColorAttributeEnum.CssColor],
-            getValueOrDefault(element.dataset[names[ColorAttributeEnum.CssDataSet]], '')
-        );
-        delete element.dataset[names[ColorAttributeEnum.CssDataSet]];
+
+        if (element.dataset[names[ColorAttributeEnum.CssDataSet]]) {
+            element.style.setProperty(
+                names[ColorAttributeEnum.CssColor],
+                getValueOrDefault(element.dataset[names[ColorAttributeEnum.CssDataSet]], '')
+            );
+            delete element.dataset[names[ColorAttributeEnum.CssDataSet]];
+        }
 
         // Some elements might have set attribute colors. We need to reset these as well.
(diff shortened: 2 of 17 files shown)


packages/roosterjs-editor-types/lib/corePluginState/ColorChangedEventPluginState.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * The state object for ColorChangedEventPlugin
+ */
+export default interface ColorChangedEventPluginState {
+    /**
+     * Content div
+     */
+    contentDiv: HTMLElement;
```
JiuqingSong, line 8: Is it possible to not having this one here?

packages/roosterjs-editor-dom/test/table/applyTableFormatTest.ts
```diff
@@ -1,4 +1,4 @@
-import applyTableFormat from '../../lib/utils/applyTableFormat';
+import applyTableFormat from '../../lib/table/applyTableFormat';
```
JiuqingSong, line 1: suggest we don't change these files in this PR.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
```diff
@@ -68,6 +68,7 @@ export default class TableEditor {
             this.isRTL,
             this.onFinishEditing
         );
+
```
JiuqingSong, line 71: suggest we don't change these files in this PR.

packages/roosterjs-editor-dom/lib/table/applyTableFormat.ts
```diff
@@ -1,4 +1,5 @@
-import changeElementTag from './changeElementTag';
+import changeElementTag from '../utils/changeElementTag';
```
JiuqingSong, line 1: suggest we don't change these files in this PR.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -1,4 +1,4 @@
-import applyTableFormat from '../utils/applyTableFormat';
+import applyTableFormat from './applyTableFormat';
```
JiuqingSong, line 1: suggest we don't change these files in this PR.

packages/roosterjs-editor-core/lib/coreApi/transformColor.ts
```diff
@@ -61,12 +61,14 @@ export const transformColor: TransformColor = (
 function transformToLightMode(element: HTMLElement) {
```
JiuqingSong, line 61: can we delete this file now?

packages/roosterjs-editor-core/lib/coreApi/getContent.ts
```diff
@@ -48,13 +49,14 @@ export const getContent: GetContent = (core: EditorCore, mode: GetContentMode):
...
 
         if (core.lifecycle.isDarkMode) {
-            core.api.transformColor(
-                core,
-                clonedRoot,
-                false /*includeSelf*/,
-                null /*callback*/,
-                ColorTransformDirection.DarkToLight
-            );
+            const event: ColorChangedEvent = {
```
JiuqingSong, line 52: Now I think I understand the problem.

So there are two different scenarios:
1. A format API did some color change, and need the plugin do color transform if editor is in dark mode. This happens in:
   - Apply table format
   - Paste
   - insertNode
   - replaceNode
   - setTextColor and setBackgroundColor
   - setContent
2. Need to do a color transformation to a given direction so that we get the html content:
   - getContent

So we need the plugin for 1, because it can be triggered from multiple sources and we don't want all of them care about current color mode and the detail of dark mode.

But for 2, we just need a util function since it has very limited source and we clearly know the color mode.

So the general idea will be:
1. Keep the new plugin
2. Simplify the event to only take necessary parameters. It doesn't need to know the direction since this plugin will only do light to dark transformation.
3. Move transformColor function out to be a single function (possibly in dom package), then both the plugin and getContent() can call it to do color transformation.
4. Trigger ColorChanged event for those APIs to let plugin handle color transformation in dark mode.

packages/roosterjs-editor-core/lib/corePlugins/ColorChangedEventPlugin.ts
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /**
+     * Handle events triggered from editor
+     * @param event PluginEvent object
+     */
+    onPluginEvent(event: PluginEvent) {
+        if (event.eventType === PluginEventType.ColorChanged) {
+            const table = this.editor.getElementAtCursor('TABLE');
+            console.log(table);
```
JiuqingSong, line 95: remove

packages/roosterjs-editor-core/lib/corePlugins/ColorChangedEventPlugin.ts
```diff
@@ -0,0 +1,221 @@
...
+        return this.state;
+    }
+
+    /**
+     * Handle events triggered from editor
+     * @param event PluginEvent object
+     */
+    onPluginEvent(event: PluginEvent) {
+        if (event.eventType === PluginEventType.ColorChanged) {
+            const table = this.editor.getElementAtCursor('TABLE');
```
JiuqingSong, line 94: why the plugin need to care table here?

PR #785, opened 2022-03-02
Refactor Apply Table format

juliaroldi:
- Change applyTableFormat file to table folder 
- Add background color using setColor api


packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -133,7 +133,7 @@ export default class VTable {
             });
             if (this.formatInfo) {
                 saveTableInfo(this.table, this.formatInfo);
-                applyTableFormat(this.table, this.cells, this.formatInfo);
+                applyTableFormat(this.table, this.cells, this.formatInfo as Required<TableFormat>);
```
JiuqingSong, line 136: you can change the type declaration in line 64 to be Required ..., so that no need to cast here. And also force us always assign a valid value to it.

packages/roosterjs-editor-dom/lib/table/applyTableFormat.ts
```diff
@@ -13,14 +14,14 @@ const CELL_SHADE = 'cellShade';
 export default function applyTableFormat(
     table: HTMLTableElement,
     cells: VCell[][],
-    format: Partial<TableFormat>
+    format: Required<TableFormat>
```
JiuqingSong, line 17: this is good

packages/roosterjs-editor-dom/lib/table/VTable.ts
juliaroldi: the line 111 
`this.formatInfo = getTableFormatInfo(this.table);`
is getting the formatInfo as TableFormat

packages/roosterjs-editor-api/lib/table/editTable.ts
```diff
@@ -13,6 +13,9 @@ export default function editTable(editor: IEditor, operation: TableOperation) {
             let vtable = new VTable(td);
             vtable.edit(operation);
             vtable.writeBack();
+
+            editor.replaceNode(vtable.table, vtable.table, true);
```
JiuqingSong, line 17: please add comment why we need this line here. Same for other files

JiuqingSong: and add inline comment for "true"

packages/roosterjs-editor-dom/lib/table/VTable.ts
JiuqingSong: I know. And getTableFormatInfo() has validation code so it can guarantee the return value is a valid `Required<TableFormat>`. So you can change line 64 to enforce this and do not do type casting here.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/CellResizer.ts
```diff
@@ -102,7 +102,7 @@ function onDraggingHorizontal(
         }
     });
 
-    vTable.writeBack();
+    vTable.writeBack(true);
```
JiuqingSong, line 105: Add comment to explain why we pass true here. Same for the other files

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -148,12 +150,16 @@ export default class VTable {
         if (!this.table) {
             return;
         }
-        this.formatInfo = { ...DEFAULT_FORMAT, ...(this.formatInfo || {}), ...(format || {}) };
+        this.formatInfo = {
+            ...DEFAULT_FORMAT,
+            ...(this.formatInfo || {}),
+            ...(format || {}),
+        } as Required<TableFormat>;
```
JiuqingSong, line 157: you can change line 18 to be 
```ts
const DEFAULT_FORMAT: Required<TableFormat> = {
    topBorderColor: '#ABABAB',
    bottomBorderColor: '#ABABAB',
    verticalBorderColor: '#ABABAB',
    hasHeaderRow: false,
    hasFirstColumn: false,
    hasBandedRows: false,
    hasBandedColumns: false,
    bgColorEven: null,
    bgColorOdd: '#ABABAB20',
    headerRowColor: '#ABABAB',
};
```

then no need to do type cast here

packages/roosterjs-editor-dom/lib/table/tableFormatInfo.ts
```diff
@@ -9,7 +9,7 @@ const TABLE_STYLE_INFO = 'roosterTableInfo';
  * @param table The table that has the info
  */
 export function getTableFormatInfo(table: HTMLTableElement) {
-    const obj = safeParseJSON(table?.dataset[TABLE_STYLE_INFO]) as TableFormat;
+    const obj = safeParseJSON(table?.dataset[TABLE_STYLE_INFO]) as Required<TableFormat>;
```
JiuqingSong, line 12: you can change line 28 to be 
```ts
function checkIfTableFormatIsValid(format: any): format is Required<TableFormat> {
```

and do the type cast inside that function, then no need to cast here

JiuqingSong: Can you remove this type cast here? And you can change the type of format to be "any" in line 28

JiuqingSong (approved):

PR #799, opened 2022-03-08
Ribbon step 3: Get ready for more buttons

JiuqingSong:
1. Fix dts tool to support export types from 3rd part library (fluent ui)
2. Add types and utils for localized strings
3. Add zoom scale and dark mode in FormatState


JiuqingSong:
#798

PR #800, opened 2022-03-08
Ribbon step 4: Better localized string support

JiuqingSong:
- Let each button have a localized string key type parameter
- Ribbon now can accept strong-type string map
- Fix up each button for localized string key


JiuqingSong:
#798

BryanValverdeU (approved):

PR #806, opened 2022-03-08
Adapt font color to background

juliaroldi:
New api to adapt de font colors to darker or brighter  backgrounds.

In light mode: 

- Create a table
- Set black shade to a cell
- Type 
- The text is visible in font color white

POCAdaptiveColorsLight

In dark mode: 

- Create a table
- Set white shade to a cell
- Type 
- The text is visible in font color black

POCAdaptiveColorsDark
k


packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Change the font color to white or some other color, so the text can be visible with a darker background
+ * @param element The element that contains text.
+ * @param newColor The font color to be applied. If not defined, the font color will be white.
+ */
+export default function adaptFontColorToBackgroundColor(
+    element: HTMLElement,
+    newFontColorInDark?: string,
+    newFontColorInBright?: string
```
JiuqingSong, line 14: I didn't see when these parameters are used

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,76 @@
...
+) {
+    if (element.hasChildNodes() && element.firstElementChild?.hasAttribute('style')) {
```
JiuqingSong, line 16: you already have null check for element.firstElementChild, so no need to check element.hasChildNodes()

JiuqingSong: and why return if there is style?

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,76 @@
...
+}
+
+const applyFontColor: Record<
+    string,
+    (element: HTMLElement, newFontColorInDark?: string, newFontColorInBright?: string) => void
+> = {
+    DARK: (element, newFontColorInDark) => (element.style.color = newFontColorInDark || '#ffffff'),
+    BRIGHT: (element, newFontColorInBright) =>
+        (element.style.color = newFontColorInBright || '#000000'),
+    DEFAULT: element => (element.style.color = ''),
```
JiuqingSong, line 34: here you created 3 functions but only 1 of them will be used, the other 2 are wasted. I prefer just a regular if...else or switch...case check, no need to create those functions.

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,76 @@
+const TRANSPARENT = 'transparent';
+const DARK = 'DARK';
+const BRIGHT = 'BRIGHT';
+const DEFAULT = 'DEFAULT';
```
JiuqingSong, line 4: use const enum instead

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,76 @@
...
+    string,
+    (element: HTMLElement, newFontColorInDark?: string, newFontColorInBright?: string) => void
+> = {
+    DARK: (element, newFontColorInDark) => (element.style.color = newFontColorInDark || '#ffffff'),
+    BRIGHT: (element, newFontColorInBright) =>
+        (element.style.color = newFontColorInBright || '#000000'),
+    DEFAULT: element => (element.style.color = ''),
+};
+
+function isBrightOrDark(color: string) {
```
JiuqingSong, line 37: the function name makes me feel it should only return 2 possible values, but it actually can return 3

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,76 @@
...
+    const lightness = calculateLightness(color);
+    if (lightness < 19) {
+        return DARK;
+    } else if (lightness > 79) {
```
JiuqingSong, line 41: make  19, 79 constant outside, and add comment to explain why use these numbers

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,76 @@
...
+function calculateLightness(color: string) {
+    if (color === TRANSPARENT) {
+        return DEFAULT;
+    }
+    const isRGB = color.includes('rgb(');
+    const isRGBA = color.includes('rgba(');
+    let r;
+    let g;
+    let b;
+    if (isRGB || isRGBA) {
```
JiuqingSong, line 57: better use a regular expression to parse the values.

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,76 @@
...
+        b = parseInt(colors.split(',')[2]);
+    } else {
+        let colors = color.replace('#', '');
+        if (colors.length === 3) {
+            colors = colors.replace(/(.)/g, '$1$1');
+        }
+        r = parseInt(colors.substr(0, 2), 16);
+        g = parseInt(colors.substr(2, 2), 16);
+        b = parseInt(colors.substr(4, 2), 16);
+    }
```
JiuqingSong, line 71: what if color is string name, such as "white", or variable, such as "--ms-color-white"?

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,76 @@
...
+    r = r / 255;
+    g = g / 255;
+    b = b / 255;
+    return (Math.max(r, g, b) + Math.min(r, g, b)) * (1 / 2) * 100;
```
JiuqingSong, line 75: add comment about this equation.

packages/roosterjs-editor-dom/lib/table/applyTableFormat.ts
```diff
@@ -62,6 +63,7 @@ function setCellColor(cells: VCell[][], format: TableFormat) {
                 } else {
                     setColor(cell.td, TRANSPARENT, true /** isBackgroundColor*/);
                 }
+                adaptFontColorToBackgroundColor(cell.td);
```
JiuqingSong, line 66: I saw this function is always used after setColor is called, prefer integrate it into setColor() and add an optional parameter to specify if we need this operation, then no need to export this function

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
juliaroldi: I put this parameters in case the user whats to change the font color to other color rather than black/white

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,87 @@
...
+//If the value of the lightness is more than 80, the color is bright
+const BRIGHT_COLORS_LIGHTNESS = 80;
+
+/**
+ * @internal
+ * Change the font color to white or some other color, so the text can be visible with a darker background
+ * @param element The element that contains text.
+ * @param backgroundColor The backgroundColor of a element
+ */
+export default function adaptFontColorToBackgroundColor(
```
JiuqingSong, line 21: Since this function is only used by setColor(), you can move it into setColor.ts

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,87 @@
+const enum ColorTones {
+    BRIGHT,
+    DARK,
+}
+const LiteralDarkColor = ['black', '--ms-color-black'];
+const LiteralBrightColor = ['white', '--ms-color-white'];
```
JiuqingSong, line 6: RoosterJs should not know this kind of color variable. They are not standard.

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,87 @@
+const enum ColorTones {
+    BRIGHT,
+    DARK,
+}
```
JiuqingSong, line 4: If there are only two possible values, why not just use boolean?
We should only use enum when there are at least 3 values.

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,87 @@
...
+ * @param backgroundColor The backgroundColor of a element
+ */
+export default function adaptFontColorToBackgroundColor(
+    element: HTMLElement,
+    backgroundColor: string
+) {
+    if (element.firstElementChild?.hasAttribute('style')) {
+        return;
+    }
+    if (isADarkOrBrightColor(backgroundColor) === ColorTones.DARK) {
```
JiuqingSong, line 28: No need to calculate twice.

const brightness = isADarkOrBrightColor(backgroundColor);
if (...)

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,87 @@
...
+function isADarkOrBrightColor(color: string) {
+    let lightness = 50; // set 50, because is not bright or dark
+    if (isRGB(color) || isRGBA(color) || isHEX(color)) {
+        lightness = calculateLightness(color);
+    }
+    if (lightness < DARK_COLORS_LIGHTNESS || LiteralDarkColor.indexOf(color) > -1) {
+        return ColorTones.DARK;
+    } else if (lightness > BRIGHT_COLORS_LIGHTNESS || LiteralBrightColor.indexOf(color) > -1) {
+        return ColorTones.BRIGHT;
+    }
```
JiuqingSong, line 46: I think a better way is to use getComputedStyle since that will always give you rgb/rgba color (at least from what I see)

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
```diff
@@ -0,0 +1,87 @@
...
+ * @returns
+ */
+function calculateLightness(color: string) {
+    let r; // red
+    let g; // green
+    let b; // blue
+
+    if (isRGB(color) || isRGBA(color)) {
+        //if the color representation is RGB, extract the values of red, green and blue
+        const colors = color.match(/[\d\.]+/g) as RegExpMatchArray;
```
JiuqingSong, line 65: You already do isRGB or isRGBA test in line 39, no need to check again. 

Actually you can just use regex do check and parse together, no need to do isRGB/isRGBA check since regex can do that for you.

packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -11,19 +12,23 @@ export default function setColor(
     element: HTMLElement,
     color: string | ModeIndependentColor,
     isBackgroundColor: boolean,
-    isDarkMode?: boolean
+    isDarkMode?: boolean,
+    shouldAdaptTheFontColor?: boolean
```
JiuqingSong, line 16: Add parameter for shouldAdaptTheFontColor

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
juliaroldi: The color can be bright or dark or neither of them, so we have three options

JiuqingSong: But the enum ColorTones only has two values: BRIGHT and DARK.

packages/roosterjs-editor-dom/lib/utils/adaptFontColorToBackgroundColor.ts
juliaroldi: getComputedStyle will return the light gray color for when the element is selected

juliaroldi: Using getPropertyValue will always return RGB or RGBA

packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -36,3 +45,49 @@ export default function setColor(
...
+}
+
+function isADarkOrBrightColor(color: string): boolean | null {
+    let lightness = calculateLightness(color);
+    if (lightness < DARK_COLORS_LIGHTNESS) {
+        return true;
+    } else if (lightness > BRIGHT_COLORS_LIGHTNESS) {
+        return false;
+    }
+    return null;
```
JiuqingSong, line 76: I like the enum solution. You just need to define a 3 value enum:

const enum Darkness = {
 None,
 Light,
 Dark,
}

packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -36,3 +45,49 @@ export default function setColor(
...
+}
+
+/**
+ * Calculate the lightness of HSL (hue, saturation and lightness) representation
+ * @param color a RBG COLOR
+ * @returns
+ */
+function calculateLightness(color: string) {
+    let [r, g, b] = color.match(/[\d\.]+/g) as RegExpMatchArray;
```
JiuqingSong, line 85: the Regex should also include "rgb" and "rgba", for example:

/^rgba?\(([\d\.]+)\s*,\s*([\d\.]+)\s*,\s*([\d\.]+)(\s*,\s*[\d\.]+)?\)$/

packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -36,3 +45,49 @@ export default function setColor(
...
+    // Use the values of r,g,b to calculate the lightness in the HSl representation
+    //First calculate the fraction of the light in each color, since in css the value of r,g,b is in the interval of [0,255], we have
+    const red = parseInt(r) / 255;
+    const green = parseInt(g) / 255;
+    const blue = parseInt(b) / 255;
+    //Then the lightness in the HSL representation is the average between maximum fraction of r,g,b and the minimum fraction
+    return (Math.max(red, green, blue) + Math.min(red, green, blue)) * (1 / 2) * 100;
```
JiuqingSong, line 92: "* (1 / 2) * 100" can be " * 50", right?

packages-ui/roosterjs-react/lib/ribbon/component/buttons/cellShade.ts
```diff
@@ -0,0 +1,22 @@
...
+    BackgroundColorDropDownItems,
+    BackgroundColors,
+    getColorPickerDropDown,
+} from './colorPicker';
+
+/**
+ * @internal
+ * "Cell Shade" button on the format ribbon
+ */
+export const cellShade: RibbonButton<CellShadeButtonStringKey> = {
```
JiuqingSong, line 14: this button only works for table, right? So we should add isDisabled() function to check if focus is in table. You can add a "isInTable" property in ElementBasedFormatState, and set it in getElementBasedFormatState().

If that is too much change, you can do this in a separate PR, and do not add this button in getButtons() for now.

packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -23,7 +38,6 @@ export default function setColor(
                 ? modeIndependentColor?.darkModeColor
                 : modeIndependentColor?.lightModeColor) || colorString
         );
-
         if (element.dataset) {
```
JiuqingSong, line 41: nit, do not remove the empty line before "if", normally we should add empty line before those structural statement such as if, for, while, ...

packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -34,5 +48,66 @@ export default function setColor(
...
+    switch (isADarkOrBrightOrNone) {
+        case ColorTones.DARK:
+            element.style.color = WHITE;
+            break;
+        case ColorTones.BRIGHT:
+            element.dataset[DarkModeDatasetNames.OriginalStyleColor] = WHITE;
+            element.style.color = BLACK;
+            break;
+        default:
+            element.style.color = '';
```
JiuqingSong, line 80: why remove color?

PR #808, opened 2022-03-08
Fix: Do not remove cellshade when change table state

juliaroldi:
The cell shade should only be removed when a predefined styles is selected. 
If the user remove or add header row, first column, banded rows or columns, the cells must be still shaded. 

cellshadeWithHeader


BryanValverdeU:
Hello Julia,

I am getting this behavior from your branch:
CellShade1

We may need to save somewhere the cell Shade color when applying HeaderRow or First Color, so that when remove HeaderRow/First Color restore back the cell Shade.

PR #810, opened 2022-03-09
Add Tab functionalities 1 (Tab)

BryanValverdeU:
Add some Tab Functionalities, all this new features are under a experimental feature, to enable iit in demo check this:

image

- On Tab, Add spaces:
Tab

- On selecting all paragraph, indent
Tab behavior 1

-Add spaces on selection is not all the paragraph selected
Tab behavior 2

-----------------------------


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
```diff
@@ -0,0 +1,152 @@
...
+
+function isWholeParagraphSelected(editor: IEditor): boolean {
+    const regions = editor.getSelectedRegions();
+    let endPosition: NodePosition = null;
+    let startPosition: NodePosition = null;
+
+    regions.forEach(r => {
+        endPosition = r.fullSelectionEnd;
+        startPosition = startPosition || r.fullSelectionStart;
+    });
```
JiuqingSong, line 74: To get first and last position, just simply get them from the selection range. e.g.

```ts
const range = editor.getSelectionRange() // or use getSelectionRangeEx()
const start = Position.getStart(range);
const end=  Position.getEnd(range);
```

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    const isAtStart: boolean = checkIfIsAtStart(editor, startPosition);
+    const isAtEnd: boolean = checkIfIsAtEnd(editor, endPosition);
+
+    return isAtEnd && isAtStart;
+}
+
+function checkIfIsAtEnd(editor: IEditor, endPosition: NodePosition): boolean {
+    let isAtEnd: boolean;
+    let blockElement = editor.getBlockElementAtNode(endPosition.node).collapseToSingleElement();
```
JiuqingSong, line 84: you can use BlockElement.getEndNode(), try not use collapseToSingleElement if we haven't decided to do some change. This function has some side effect.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    if (tempChild == endPosition.node && endPosition.isAtEnd) {
+        isAtEnd = true;
+    }
+    return isAtEnd;
+}
+
+function checkIfIsAtStart(editor: IEditor, startPosition: NodePosition): boolean {
+    let isAtStart: boolean;
+    const blockElement = editor.getBlockElementAtNode(startPosition.node).collapseToSingleElement();
```
JiuqingSong, line 99: same here

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    tempChild = ignoreEmptySpans(tempChild);
+
+    if (tempChild == startPosition.node && startPosition.offset == 0) {
+        isAtStart = true;
+    }
+    return isAtStart;
+}
+
+function ignoreEmptySpans(tempChild: ChildNode) {
```
JiuqingSong, line 113: this makes me feel the code knows too much detail of the DOM tree, and this check is easy to fail if the tag is not span but something like B, I, U, ...

Can we try this:

Create a range from the real beginning of this block, to the selected position. Then if this range is empty (no text content, no image, no table, no list), then we just treat it as empty, treat the selection is at beginning of block? You can do the same for the end of the block.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
```diff
@@ -0,0 +1,146 @@
...
+    const range = createRange(tempChild, PositionType.Begin, tempChild, PositionType.End);
+    return range.collapsed;
+}
+
+function insertTab(editor: IEditor, event: PluginKeyboardEvent) {
+    const span = editor.getDocument().createElement('span');
+    let searcher = editor.getContentSearcherOfCursor(event);
+    const charsBefore = searcher.getSubStringBefore(Number.MAX_SAFE_INTEGER);
+
+    const numberOfChars = 6 - (charsBefore.length % 6);
```
JiuqingSong, line 133: Make "6" a constant

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
```diff
@@ -0,0 +1,146 @@
...
+            ? node.nextSibling || block.nextSibling
+            : node.previousSibling || block.previousSibling;
+    const getElementToCheck = (node: Node) => (start ? node.firstChild : node.lastChild);
+
+    let tempChild = block;
+    while (getElementToCheck(tempChild)) {
+        tempChild = getElementToCheck(tempChild);
+    }
+
+    while (isEmptyElement(tempChild)) {
```
JiuqingSong, line 113: Can this work correctly for image?

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
```diff
@@ -0,0 +1,146 @@
...
+ * @param pos Position to check if is at the limits of the block
+ * @param start If true, checks from the beginning of the block, if false check from the end of block
+ * @returns boolean. true if is at limit, false otherwise
+ */
+function checkIfIsAtBlockLimit(editor: IEditor, pos: NodePosition, start: boolean): boolean {
+    let isAtLimit: boolean;
+    let blockElement = editor.getBlockElementAtNode(pos.node).getEndNode();
+    const block = (start ? blockElement.firstChild : blockElement.lastChild) || blockElement;
+
+    const tempChild = ignoreEmptyElement(block, start);
```
JiuqingSong, line 87: Actually what I'm thinking is something like:

```ts
const blockStart = new Position(block.getStartNode(), PositionType.Begin);
const blockEnd = new Position(block.getEndNode(), PositionType.End);
const rangeBefore = createRange(blockStart, Position.getStart(range));
const rangeAfter = createRange(Position.getEnd(range), blockEnd);
return isRangeEmpty(rangeBefore, regionRoot) && isRangeEmpty(rangeAfter, regionRoot);

function isRangeEmpty(range, regionRoot) {
  return range.toString == '' && queryElements(regionRoot, 'img,table,ul,ol', null, QueryScope.InSelection, range).length == 0;
}
```

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
```diff
@@ -0,0 +1,146 @@
...
+ * @internal
+ */
+export const TextFeatures: Record<
+    keyof TextFeatureSettings,
+    BuildInEditFeature<PluginKeyboardEvent>
+> = {
+    indentWhenTabText: IndentWhenTabText,
+};
+
+function isWholeParagraphSelected(editor: IEditor, range: Range): boolean {
```
JiuqingSong, line 65: What if I select two whole blocks?

JiuqingSong: I tried in Word, if I select 1 and a half blocks, it will also indent both blocks. Can we do the same?

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
BryanValverdeU: I see, yes this did not work with a IMG at the end or at the start of the block element, but I think with the snippet you provided this is fixed.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
BryanValverdeU: I see, if we select 2 whole blocks with current implementation we can indent, but 1 and a half block we just remove the selection, I will check how to implement this.

BryanValverdeU: I just messaged Jorge to see his point of view, this is kind of weird because in Office Desktop Apps, if we select 1 and a half paragraphs we just remove all the selection (what we have in this branch), but in Office Web we do what you mentioned, we just indent.

I think that even in Word Web we just need to select 2 different blocks and on Tab it is going to indent both blocks

BryanValverdeU: Added functionality suggested, if we select more than one block and tab we indent all the blocks in selection
Tab behavior in Listx
,

JiuqingSong (approved):

PR #812, opened 2022-03-09
Roosterjs React step 8: Support generate demo code for roosterjs-react

JiuqingSong:
Support send code to codepen to generate sample code for roostejs-react

image
image


JiuqingSong:
#798

packages/roosterjs-editor-core/lib/coreApi/addUndoSnapshot.ts
```diff
@@ -81,3 +85,73 @@ export const addUndoSnapshot: AddUndoSnapshot = (
         }
     }
 };
+
+// interface SnapshotSelectionBase<T extends SelectionRangeTypes> {
```
BryanValverdeU, line 89: There is some commented code here, should we keep it?

JiuqingSong: oops, forget to remove

BryanValverdeU (approved):

PR #816, opened 2022-03-10
Add Tab Functionalities 3 (Tab / Shift + Tab in the First List Item)

BryanValverdeU:
Add functionality to Indent / Outdent a List, only when the impacted list item is the first item in the List
Tab behavior in List


packages/roosterjs-editor-api/lib/format/setIndentation.ts
```diff
@@ -40,6 +40,15 @@ export default function setIndentation(editor: IEditor, indentation: Indentation
 
             if (vList) {
                 blockGroups.push([]);
+
+                if (vList.getListItemIndex(startNode) == 0) {
+                    const list = editor.getElementAtCursor('OL,UL', startNode);
```
JiuqingSong, line 45: can we get list element from vList?

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -376,6 +376,23 @@ export default class VList {
             }
         });
     }
+
+    /**
+     * Get the index of the List Item in the current List
+     * @param input List item to find in the root list
+     */
+    getListItemIndex(input: Node) {
```
JiuqingSong, line 384: For a splitted list e.g. `<ol start=3>`, this will return 0 for actual "3".

Need to leverage getStart(), and you just need to check if the given list node is the first one of this.items. Try not access childNodes directly.

BryanValverdeU: This wont be needed anymore as we are only going to check if the first item of the vList is impacted by the indent operation. Removing it.

packages/roosterjs-editor-api/lib/format/setIndentation.ts
```diff
@@ -39,12 +39,18 @@ export default function setIndentation(editor: IEditor, indentation: Indentation
...
 
             if (vList) {
-                blockGroups.push([]);
                 while (blocks[i + 1] && vList.contains(blocks[i + 1].getStartNode())) {
                     i++;
                 }
-                vList.setIndentation(start, end, indentation);
-                vList.writeBack();
+
+                if (vList.items[0]?.getNode() == startNode) {
```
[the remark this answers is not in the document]
BryanValverdeU: Good point, added that.

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -324,6 +324,22 @@ export default class VList {
...
     }
 
+    /**
+     * Get the index of the List Item in the current List
+     * @param input List item to find in the root list
+     */
+    getListItemIndex(input: Node) {
+        if (this.items) {
+            for (let index = 0; index < this.items.length; index++) {
+                const child = this.items[index];
```
JiuqingSong, line 334: Need to check if this item is a Ordered list number and it is the first leve item. Something like:
```ts
let listIndex = this.getStart(this.listRoot) - 1;
if (child.getListType() == ListType.Ordered && child.getLevel() == 1 && !child.isDummy()) {
listIndex++;
}
if (child.getNode() == input) {
 return listIndex;
}
```

See VList.getLastItemNumber for more reference

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -324,6 +324,29 @@ export default class VList {
...
+        if (this.items) {
+            const start = getStart(this.rootList) - 1;
+
+            for (let index = 0; index < this.items.length; index++) {
+                const child = this.items[index];
+                if (
+                    child.getLevel() == 1 &&
+                    !child.isDummy() &&
+                    child.getListType() == ListType.Ordered &&
+                    child.getNode() == input
```
JiuqingSong, line 341: do not check getNode() == input here. It will return wrong result from the second child.

JiuqingSong:
I mean, as a new public API, getListItemIndex() should always return right value, not only for this feature. In your loop you checked getNode() == input, this will only be true for the inputted node. So suppose the list starts from 1, and no matter which node you pass in, as long as it is not the first node, it will always return 2.

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -324,6 +324,36 @@ export default class VList {
...
+
+                if (child.getNode() == input) {
+                    return listIndex;
+                }
+            }
+        }
+        return -1;
+    }
+
+    getStart(): number | undefined {
```
JiuqingSong, line 353: please add comment for this public API

JiuqingSong (approved):

PR #818, opened 2022-03-14
Add Tab Functionalities 4 (Alt + F10 to move to next element) 

BryanValverdeU:
When press Alt + F10, move focus to next element, 

We get the next element by checking the getCustomData and see if there is a element with Key FocusNextElement, if there is we call the focus function on this element.

Tab behaviorAltF10


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
```diff
@@ -87,6 +94,36 @@ const OutdentWhenTabText: BuildInEditFeature<PluginKeyboardEvent> = {
...
+            cacheGetEventData(event, FOCUS_NEXT_ELEMENT_KEY, () =>
+                editor.getCustomData(FOCUS_NEXT_ELEMENT_KEY)
+            )
+        );
+    },
+    handleEvent: (event, editor) => {
+        const element = cacheGetEventData(
+            event,
+            FOCUS_NEXT_ELEMENT_KEY,
+            editor.getCustomData(FOCUS_NEXT_ELEMENT_KEY)
```
JiuqingSong, line 118: This is too tricky, looks doesn't like a formal API.

And the next focusable control can change during editing, so that means caller need to monitor that change and update to editor all the time.

I strongly recommend we don't do this in rooster, but do it in a plugin in OWA instead.

If we really want to do in rooster, it should be an event and still let caller handle the event to do focus.

BryanValverdeU: Sounds good, and if a user wants to use the TabKeyFeatures, we can just mention to add this functionality to their current implementation.

PR #826, opened 2022-03-16
add adaptive font color

juliaroldi:
New api to adapt de font colors to darker or brighter  backgrounds.

In light mode: 

- Create a table
- Set black shade to a cell
- Type 
- The text is visible in font color white

POCAdaptiveColorsLight

In dark mode: 

- Create a table
- Set white shade to a cell
- Type 
- The text is visible in font color black

POCAdaptiveColorsDark

### DIFF
--- a/packages/roosterjs-editor-dom/lib/utils/setColor.ts
+++ b/packages/roosterjs-editor-dom/lib/utils/setColor.ts
@@ -1,7 +1,7 @@
 import { DarkModeDatasetNames, ModeIndependentColor } from 'roosterjs-editor-types';
 
 const WHITE = '#ffffff';
-const BLACK = '#000000';
+const GRAY = '#333333';
 const TRANSPARENT = 'transparent';
 const enum ColorTones {
     BRIGHT,
@@ -67,24 +67,29 @@ function adaptFontColorToBackgroundColor(element: HTMLElement, isDarkMode?: bool
         return;
     }
     const backgroundColor = element.style.getPropertyValue('background-color');
-    const lightModeBackgroundColor = isDarkMode
-        ? element.dataset[DarkModeDatasetNames.OriginalStyleBackgroundColor] ||
-          element.dataset[DarkModeDatasetNames.OriginalAttributeBackgroundColor]
-        : backgroundColor;
+    const lightModeBackgroundColor =
+        (isDarkMode &&
+            (element.dataset[DarkModeDatasetNames.OriginalStyleBackgroundColor] ||
+                element.dataset[DarkModeDatasetNames.OriginalAttributeBackgroundColor])) ||
+        backgroundColor;
     if (!lightModeBackgroundColor || lightModeBackgroundColor === TRANSPARENT) {
         return;
     }
     const isADarkOrBrightOrNone = isADarkOrBrightColor(lightModeBackgroundColor!);
     switch (isADarkOrBrightOrNone) {
         case ColorTones.DARK:
-            const colorDark = isDarkMode ? lightModeBackgroundColor : WHITE;
-            element.dataset[DarkModeDatasetNames.OriginalStyleColor] = WHITE;
-            setColor(element, colorDark, false);
+            const fontForDark: ModeIndependentColor = {
+                lightModeColor: WHITE,
+                darkModeColor: GRAY,
+            };
+            setColor(element, fontForDark, false /*isBackground*/, isDarkMode);
             break;
         case ColorTones.BRIGHT:
-            const colorBright = isDarkMode ? lightModeBackgroundColor : BLACK;
-            element.dataset[DarkModeDatasetNames.OriginalStyleColor] = BLACK;
-            setColor(element, colorBright, false);
+            const fontForLight: ModeIndependentColor = {
+                lightModeColor: GRAY,
+                darkModeColor: WHITE,
+            };
+            setColor(element, fontForLight, false /*isBackground*/, isDarkMode);
             break;
     }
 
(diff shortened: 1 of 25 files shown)


packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -34,5 +51,72 @@ export default function setColor(
...
+    return ColorTones.NONE;
+}
+
+/**
+ * Calculate the lightness of HSL (hue, saturation and lightness) representation
+ * @param color a RBG or RGBA COLOR
+ * @returns
+ */
+function calculateLightness(color: string) {
+    let [r, g, b] = color.match(/[\d\.]+/g) as RegExpMatchArray;
```
JiuqingSong, line 114: why change the regex back? we should use the long regex to check if it is rgb/rgba as well

packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -34,5 +51,72 @@ export default function setColor(
...
+
+    const isADarkOrBrightOrNone = isADarkOrBrightColor(backgroundColor);
+    const fontColorInDarkMode = element.dataset[DarkModeDatasetNames.OriginalStyleBackgroundColor];
+    switch (isADarkOrBrightOrNone) {
+        case ColorTones.DARK:
+            element.style.color = WHITE;
+            break;
+        case ColorTones.BRIGHT:
+            //Save this value so transformColor can adjust the color when switch from dark to light mode
+            element.dataset[DarkModeDatasetNames.OriginalStyleColor] = WHITE;
```
JiuqingSong, line 83: I believe you should either set this property for both case ColorTones.DARK and here, or neither of them.

And you can read the background color like:
const lightModeBackgroundColor = isDarkMode ? element.dataset[DarkModeDatasetNames.OriginalStyleBackgroundColor] || element.dataset[DarkModeDatasetNames.OriginalAttributeBackgroundColor] : backgroundColor;

Now you have the back color always for light mode, then calculate color tone from it, and set the text color to the cell, and delete element.dataset[DarkModeDatasetNames.OriginalStyleColor] and element.dataset[DarkModeDatasetNames.OriginalAttributeColor].

After that, the setShadeColor code can call dark mode transform code to transform to dark mode if necessary.

packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -34,5 +51,100 @@ export default function setColor(
...
+        : backgroundColor;
+    if (!lightModeBackgroundColor || lightModeBackgroundColor === TRANSPARENT) {
+        return;
+    }
+    const isADarkOrBrightOrNone = isADarkOrBrightColor(lightModeBackgroundColor!);
+    switch (isADarkOrBrightOrNone) {
+        case ColorTones.DARK:
+            const colorDark = isDarkMode ? lightModeBackgroundColor : WHITE;
+            element.dataset[DarkModeDatasetNames.OriginalStyleColor] = WHITE;
+            setColor(element, colorDark, false);
```
JiuqingSong, line 82: This is not right.

now you already know whether you should apply white color for light mode. What you need is just set white color. You can define a mode independent white color as:
```ts
const WHITE: ModeIndependentColor = {
  lightColor: '#FFFFFF',
  darkColor: '#333333', // This is the dark color value for white in dark mode
};
```

Then use setColor here to set this color:
```ts
setColor(element, WHITE, false /*isBackground*/ isDarkMode);
```

Similar for the Light case.

JiuqingSong: No need to manually set element.dataset[DarkModeDatasetNames.OriginalStyleColor] here because setColor will do that for you.

packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -34,5 +51,100 @@ export default function setColor(
...
+ * Calculate the lightness of HSL (hue, saturation and lightness) representation
+ * @param color a RBG or RGBA COLOR
+ * @returns
+ */
+function calculateLightness(color: string) {
+    let r: number;
+    let g: number;
+    let b: number;
+
+    if (color.includes('#')) {
```
JiuqingSong, line 117: Better check color.substring(0,1) == '#' instead.

packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -34,5 +51,105 @@ export default function setColor(
...
+        case ColorTones.DARK:
+            const fontForDark: ModeIndependentColor = {
+                lightModeColor: WHITE,
+                darkModeColor: GRAY,
+            };
+            setColor(element, fontForDark, false /*isBackground*/, isDarkMode);
+            break;
+        case ColorTones.BRIGHT:
+            const fontForLight: ModeIndependentColor = {
+                lightModeColor: GRAY,
```
JiuqingSong, line 89: When background color is bright, should text color be BLACK or GRAY in light mode? I think it should be BLACK (the default value).

Other colors look fine.

packages/roosterjs-editor-dom/lib/utils/setColor.ts
```diff
@@ -34,5 +51,105 @@ export default function setColor(
...
+                darkModeColor: WHITE,
+            };
+            setColor(element, fontForLight, false /*isBackground*/, isDarkMode);
+            break;
+    }
+
+    if (!isDarkMode) {
+        delete element.dataset[DarkModeDatasetNames.OriginalStyleColor];
```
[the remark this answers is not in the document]
juliaroldi: No, we do not need, setColor will remove in line 49. I just forgot to remove it.

Cearl03 (approved):

JiuqingSong (approved):

PR #847, opened 2022-03-18
Step 1 [Table Improvements]: Improve delete row and column operation to delete selected cells.

juliaroldi:
- Added support to delete selected rows

deleteRowsFixed

- Added support to delete selected columns

deleteColumnsFixed

- Added delete table operations to rooster-react, but this buttons was not added to ribbon. 

deleteRowsAndColumnsFixed


packages-ui/roosterjs-react/lib/ribbon/component/buttons/tableEditOperations.ts
```diff
@@ -0,0 +1,38 @@
...
+    deleteTable: TableOperation.DeleteTable,
+    deleteRow: TableOperation.DeleteRow,
+    deleteColumn: TableOperation.DeleteColumn,
+};
+
+/**
+ * @internal
+ * "TableEditOperations" button on the format ribbon
+ */
+export const tableEdit: RibbonButton<TableEditOperationsStringKey> = {
```
JiuqingSong, line 24: Do you want this become an office feature of roosterjs-react ribbon or just an entry of this feature for demo site?

If the latter, you can put this button into demo site path (similar with popout button) so that you can still use it in demo site but no need to publish it to other developers.

JiuqingSong, line 24: Do you want this become an office feature of roosterjs-react ribbon or just an entry of this feature for demo site?

If the latter, you can put this button into demo site path (similar with popout button) so that you can still use it in demo site but no need to publish it to other developers.

packages-ui/roosterjs-react/lib/ribbon/type/RibbonButtonStringKeys.ts
```diff
@@ -257,4 +262,5 @@ export type AllButtonStringKeys =
     | TextColorButtonStringKey
     | UnderlineButtonStringKey
     | UndoButtonStringKey
+    | TableEditOperationsStringKey
```
JiuqingSong, line 265: same for these types

JiuqingSong, line 265: same for these types

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -239,23 +240,51 @@ export default class VTable {
...
+                    let nextCell = this.getCell(rowIndex + 1, i);
                     if (cell.td && cell.td.rowSpan > 1 && nextCell.spanAbove) {
                         nextCell.td = cell.td;
                     }
-                });
-                this.cells.splice(this.row, 1);
+                };
+                if (this.selection) {
+                    const { firstCell, lastCell } = this.selection;
+                    for (let rowIndex = firstCell.y; rowIndex <= lastCell.y; rowIndex++) {
```
JiuqingSong, line 251: curious, do we need to check lastCell.y >= firstCell.y? Otherwise it is possible to go into infinite loop

JiuqingSong, line 251: curious, do we need to check lastCell.y >= firstCell.y? Otherwise it is possible to go into infinite loop

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -239,23 +240,51 @@ export default class VTable {
...
+                            deleteRowHandler(cell, i, rowIndex);
+                        });
+                    }
+                    this.cells.splice(firstCell.y, lastCell.y - firstCell.y + 1);
+                } else {
+                    this.forEachCellOfCurrentRow((cell, i) => {
+                        deleteRowHandler(cell, i, this.row);
+                    });
+                    this.cells.splice(this.row, 1);
+                }
```
JiuqingSong, line 262: So I'd like to see these if.. and else.. block merged together.

The difference of them are
1. Delete single row vs delete multiple row
2. Read current row from this.row vs from this.selection

You can read the row numbers first, then always assume delete multiple rows. When delete single, it can be treated the first row and last row are the same. So something like:
```ts
const firstRow = this.selection ? this.selection.firstCell.y : this.row;
const lastRow = this.selection ? this.selection.lastCell.y : this. row;
for (from firstRow to last row) ...
```

JiuqingSong, line 262: So I'd like to see these if.. and else.. block merged together.

The difference of them are
1. Delete single row vs delete multiple row
2. Read current row from this.row vs from this.selection

You can read the row numbers first, then always assume delete multiple rows. When delete single, it can be treated the first row and last row are the same. So something like:
```ts
const firstRow = this.selection ? this.selection.firstCell.y : this.row;
const lastRow = this.selection ? this.selection.lastCell.y : this. row;
for (from firstRow to last row) ...
```

JiuqingSong: Then no need to have the function "deleteRowHandler". It can be just inline into the loop

JiuqingSong: Then no need to have the function "deleteRowHandler". It can be just inline into the loop

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -239,23 +240,51 @@ export default class VTable {
...
-                    let nextCell = this.getCell(i, this.col + 1);
+                const deleteColumnsHandler = (cell: VCell, i: number, colIndex: number) => {
+                    let nextCell = this.getCell(i, colIndex + 1);
                     if (cell.td && cell.td.colSpan > 1 && nextCell.spanLeft) {
                         nextCell.td = cell.td;
                     }
-                    row.splice(this.col, 1);
-                });
+                };
+                if (this.selection) {
```
JiuqingSong, line 271: same for here

JiuqingSong, line 271: same for here

packages/roosterjs-editor-dom/lib/table/VTable.ts
juliaroldi: I don't think it is possible. The first selected cell will be ways the 0,0, so the coordinate of the other cells will be always bigger. 

`export function getCellCoordinates(vTable: VTable, cellInput: Node): Coordinates {
    let result: Coordinates;
    if (vTable?.cells) {
        for (let indexY = 0; indexY < vTable.cells.length; indexY++) {
            for (let indexX = 0; indexX < vTable.cells[indexY].length; indexX++) {
                if (cellInput == vTable.cells[indexY][indexX].td) {
                    result = {
                        x: indexX,
                        y: indexY,
                    };
                }
            }
        }
    }

    return result;
}`

JiuqingSong (approved):

PR #852, opened 2022-03-22
FHL: Add files in clipboard

zorauf:
# Intent
Users can copy a file and paste it in editor which the hosts can decide how to handle it. Currently the `ClipboardData` and corresponding extraction of items does not handle this scenario.

# Change
The change is to add a new `files` field in `ClipboardData` and when extracting data we get the list of files in the clipboard and pass that information so hosts can decide how to handle it. As user can pick multiple files hence the `files` field is an `array`.

# TODO
- [x] ~Update minor package version~ Will be done in a separate package
- [x] ~See if this needs to be handled in IE~ Not going to support IE for files

## Future Enhancements
Currently the `image` in `ClipboardData` only takes one image and ignores the next which means if user copies multiple images from computer and paste we don't add them as inline image.
With this change at least the next images will be added as files so host has an opportunity to handle them. Future improvement can involve making `image` field be an `images: File[]` so that it can handle multiple images, which should allow to insert multiple images inline in editor. 

As that would be a breaking change and requires multiple changes to existing behavior hence that is out of scope for this PR.


packages/roosterjs-editor-dom/lib/clipboard/extractClipboardItems.ts
```diff
@@ -1,131 +1,141 @@
-import readFile from '../utils/readFile';
```
zorauf, line 1: Please let me know how can I fix this formatting. I see that `pretty-quick` is part of pre-commit hook but even running manually it didn't fix. Is it because this file hasn't changed since prettier config changed and its doing the right thing?

packages/roosterjs-editor-dom/lib/clipboard/extractClipboardItems.ts
```diff
@@ -1,131 +1,141 @@
...
+                        readFile(data.image, dataUrl => {
+                            data.imageDataUri = dataUrl;
+                            resolve();
+                        });
+                    } else {
+                        resolve();
+                    }
+                });
+            } else if (item.kind == 'file') {
+                return new Promise<void>(resolve => {
```
zorauf, line 81: This is the change to add files to `ClipboardData`. I've added the fetching of files and putting it in clipboard in the promise, is there  any reason it should be before the promise?

zorauf:
+ @JiuqingSong , @BryanValverdeU , @juliaroldi would like your input on the proposed changes along with some questions I've left in the comments.

packages/roosterjs-editor-types/lib/interface/ClipboardData.ts
```diff
@@ -31,6 +31,11 @@ export default interface ClipboardData {
      */
     image: File | null;
 
+    /**
+     * General file from clipboard event
+     */
+    files: File[];
```
JiuqingSong, line 37: please make this optional otherwise it will be a breaking change.

zorauf: Good catch. Made it optional.

JiuqingSong (approved):

PR #862, opened 2022-03-28
Fix Issue when selecting all table and trying to Copy

BryanValverdeU:
Cut Copy was not working when we select the table with the table selector control and the performed a copy. This PR fixes this issue

### DIFF
--- a/packages/roosterjs-editor-core/lib/corePlugins/CopyPastePlugin.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/CopyPastePlugin.ts
@@ -17,6 +17,8 @@ import {
     ExperimentalFeatures,
     PluginWithState,
     KnownCreateElementDataIndex,
+    SelectionRangeEx,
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-core/lib/corePlugins/CopyPastePlugin.ts
```diff
@@ -162,16 +163,38 @@ export default class CopyPastePlugin implements PluginWithState<CopyPastePluginS
         return div;
     }
 
-    private cleanUpAndRestoreSelection(tempDiv: HTMLDivElement, range: Range, isCopy: boolean) {
-        if (isCopy && Browser.isAndroid) {
-            range.collapse();
+    private cleanUpAndRestoreSelection(
```
BryanValverdeU, line 166: Had to modify this because after copy/paste we were modifying from Table Selection to Normal Selection

juliaroldi (approved):

PR #868, opened 2022-03-29
Remove unneeded check in getListItemIndex in VList

BryanValverdeU:
We were checking whether the List Item was Ordered, but this is not required. 
Checking if the LI was a Ordered Item caused the API to not work with UL Lists, removing this check will fix this issue.

### DIFF
--- a/packages/roosterjs-editor-dom/lib/list/VList.ts
+++ b/packages/roosterjs-editor-dom/lib/list/VList.ts
@@ -330,7 +330,7 @@ export default class VList {
      */
     getListItemIndex(input: Node) {
         if (this.items) {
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -330,15 +330,11 @@ export default class VList {
      */
     getListItemIndex(input: Node) {
```
JiuqingSong, line 331: I can understand this change, but now this function makes people confusing. Need some comment to explain how it works for mixed list.

JiuqingSong (approved):

PR #871, opened 2022-03-30
Change conserveConstEnums for compatibility

sverrejoh:
This makes sure your interface is not using const enums for your external interface, which makes it compatible with all external projects. Notably Create React App is using `isolatedModules` which is not compatible with const enums. The TypeScript team outlines the issue in the const enums pitfalls documentation. https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls

### DIFF
--- a/packages/tsconfig.json
+++ b/packages/tsconfig.json
@@ -8,7 +8,7 @@
         "declaration": true,
         "removeComments": false,
         "noImplicitAny": true,
-        "preserveConstEnums": true,
+        "preserveConstEnums": false,
         "noUnusedLocals": true,
         "baseUrl": ".",
         "paths": {


JiuqingSong:
Instead of changing like this, I'd like to explore the solution of changing all const enums to use string value to help prevent bundle size increasing too much.

packages/tsconfig.json
```diff
@@ -8,7 +8,7 @@
         "declaration": true,
         "removeComments": false,
         "noImplicitAny": true,
-        "preserveConstEnums": true,
+        "preserveConstEnums": false,
```
JiuqingSong, line 11: Why change preserveConstEnums to false can fix this problem?

JiuqingSong:
https://github.com/microsoft/roosterjs/pull/911

JiuqingSong:
Close this one since it will be covered by #911

PR #874, opened 2022-03-30
Prevent List Item removal when Shift+Tab in a Level 1 LI

BryanValverdeU:
Add the ability to prevent the list item to be removed when pressing Shift + Tab or outdenting,
Feature is under TabKeyFeatures Experimental features so we dont remove the previous functionality.
Prevent list item to be removed

### DIFF
--- a/packages/roosterjs-editor-dom/lib/list/VList.ts
+++ b/packages/roosterjs-editor-dom/lib/list/VList.ts
@@ -17,12 +17,7 @@ import {
     NodePosition,
     PositionType,
     NodeType,
-    VListBehaviorConfig,
 } from 'roosterjs-editor-types';
-
-const DEFAULT_VLIST_CONFIG: VListBehaviorConfig = {
-    preventItemRemovalOnOutdent: false,
-};
 
 /**
  * Represent a bullet or a numbering list
@@ -67,7 +62,6 @@ const DEFAULT_VLIST_CONFIG: VListBehaviorConfig = {
  */
 export default class VList {
     public readonly items: VListItem[] = [];
-    private config: VListBehaviorConfig;
 
     /**
      * Create a new instance of VList class
@@ -137,15 +131,6 @@ export default class VList {
         queryElements(this.rootList, 'li', moveLiToList);
 
         this.populateItems(this.rootList);
-        this.config = DEFAULT_VLIST_CONFIG;
-    }
-
-    /**
-     * Overrides the configuration to be used for the current VList
-     * @param config Config to apply to the VList
-     */
-    setConfiguration(config: VListBehaviorConfig) {
-        this.config = config || DEFAULT_VLIST_CONFIG;
     }
 
     /**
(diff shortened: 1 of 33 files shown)


packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -12,6 +12,8 @@ import { KnownCreateElementDataIndex, ListType } from 'roosterjs-editor-types';
 
 const orderListStyles = [null, 'lower-alpha', 'lower-roman'];
 
+const MARGIN_BASE = '0in 0in 0in 0.5in';
+const NEGATIVE_MARGIN = '-.25in';
```
JiuqingSong, line 16: who will generate this value? What if other negative value?

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -131,6 +137,15 @@ export default class VList {
...
 
         this.populateItems(this.rootList);
+        this.config = DEFAULT_VLIST_CONFIG;
+    }
+
+    /**
+     * Overrides the configuration to be used for the current VList
+     * @param config Config to apply to the VList
+     */
+    setConfiguration(config: VListBehaviorConfig) {
```
JiuqingSong, line 147: Can we set this config in constructor instead?

packages/roosterjs-editor-types/lib/interface/VListBehaviorConfig.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Config to modify the behavior of VList
+ */
+export default interface VListBehaviorConfig {
```
JiuqingSong, line 4: I prefer "VListOptions"

packages/roosterjs-editor-dom/lib/list/VList.ts
JiuqingSong: Or, only pass preventItemRemovalOnOutdent as a parameter in setIndentation

packages/roosterjs-editor-dom/lib/list/VListItem.ts
BryanValverdeU: These values are set when trying to outdent the first list item and it is the first item in the list, I checked at what Word is doing to perform this action and this is the styles that they add to the List.

packages/roosterjs-editor-dom/lib/list/VList.ts
BryanValverdeU: Sounds good, I think only adding the parameter to setIndentation is better

packages/roosterjs-editor-dom/lib/list/VListItem.ts
BryanValverdeU: To give more context, this is what we are trying to achieve, instead of removing the bullet, we outdent using this margin.
prevent removal

PR #880, opened 2022-04-01
[Table Improvements] Step 4 - Merge multiple cells 

juliaroldi:
Add support to merge multiple cells with cell selection. 
mergeCellsRooster

### DIFF
--- a/demo/scripts/controls/ribbonButtons/tableEditOperations.ts
+++ b/demo/scripts/controls/ribbonButtons/tableEditOperations.ts
@@ -9,7 +9,10 @@ type TableEditOperationsKey =
     | 'insertAbove'
     | 'insertBelow'
     | 'insertLeft'
-    | 'insertRight';
+    | 'insertRight'
+    | 'merge'
+    | 'mergeAbove'
+    | 'mergeBelow';
 
 /**
  * Key of localized strings of Table Edit Operations button
@@ -21,9 +24,12 @@ const tableEditOperationsLabel: Record<TableEditOperationsKey, string> = {
     insertBelow: 'Insert Below',
     insertLeft: 'Insert Left',
     insertRight: 'Insert Right',
+    merge: 'Merge Cells',
     deleteTable: 'Delete Table',
     deleteRow: 'Delete Row',
     deleteColumn: 'Delete Column',
+    mergeAbove: 'Merge Above',
+    mergeBelow: 'Merge Below',
 };
 
 const tableEditOperations: Record<TableEditOperationsKey, TableOperation> = {
@@ -31,9 +37,12 @@ const tableEditOperations: Record<TableEditOperationsKey, TableOperation> = {
     insertBelow: TableOperation.InsertBelow,
     insertLeft: TableOperation.InsertLeft,
     insertRight: TableOperation.InsertRight,
+    merge: TableOperation.MergeCells,
     deleteTable: TableOperation.DeleteTable,
     deleteRow: TableOperation.DeleteRow,
     deleteColumn: TableOperation.DeleteColumn,
+    mergeAbove: TableOperation.MergeAbove,
+    mergeBelow: TableOperation.MergeBelow,
 };
 
 /**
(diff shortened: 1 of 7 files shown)


packages/roosterjs-editor-dom/lib/selection/Position.ts
```diff
@@ -40,7 +40,7 @@ export default class Position implements NodePosition {
         offsetOrPosType: number = 0,
         private readonly isFromEndOfRange?: boolean
     ) {
-        if ((<NodePosition>nodeOrPosition).node) {
+        if (nodeOrPosition && (<NodePosition>nodeOrPosition).node) {
```
JiuqingSong, line 43: This is interesting. In what case you see this is null? I'm afraid there maybe more serious bug hiding here

demo/scripts/controls/ribbonButtons/tableEditOperations.ts
```diff
@@ -21,19 +24,25 @@ const tableEditOperationsLabel: Record<TableEditOperationsKey, string> = {
     insertBelow: 'Insert Below',
     insertLeft: 'Insert Left',
     insertRight: 'Insert Right',
+    merge: 'Merge Cells',
     deleteTable: 'Delete Table',
     deleteRow: 'Delete Row',
     deleteColumn: 'Delete Column',
+    mergeAbove: 'Merge Above',
+    mergeBelow: 'Merge Below',
```
JiuqingSong, line 32: what about merge left and right?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -348,6 +348,33 @@ export default class VTable {
                 }
                 break;
 
+            case TableOperation.MergeCells:
```
JiuqingSong, line 351: Question: is it possible to just reuse MergeLeft and MergeAbove to do MergeCells? e.g.
case MergeCells:
  for (...) for (...) this.edit(MergeLeft) ;

demo/scripts/controls/ribbonButtons/tableEditOperations.ts
juliaroldi: I just added this two to test the behavior, I will add the other two.

packages/roosterjs-editor-dom/lib/selection/Position.ts
juliaroldi: Sometimes, after just hovering the mouse while using cell selection after merging some cells, some node returned null.

JiuqingSong: In that case I'd like to see why that can happen, and fix the root cause, but not just add a simple check here. We already enabled strict mode for this class and the in theory it should never be null.

juliaroldi: I found the problem in selectTable, in this code 
`table.rows.item(coordinates.firstCell.y).cells.item(coordinates.firstCell.x),`
the the coordinates of the merged cells are being passes, so the it returns null to new Position.
After merge, we need to recalculate the coordinates.

juliaroldi: I didn't create nothing to recalculate, just added a check to select a cell, if it exist.

packages/roosterjs-editor-dom/lib/table/VTable.ts
juliaroldi: I put the common code between the merges function inside a function

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -474,6 +465,21 @@ export default class VTable {
         }
     }
 
+    mergeCells(cell: VCell, nextCell: VCell, horizontally?: boolean) {
```
JiuqingSong, line 468: make it private

JiuqingSong (approved):

Issue #895, opened 2022-04-06
 Uncaught TypeError: Cannot read properties of undefined (reading 'x')


BryanValverdeU:
**Describe the bug**
Fix possible undefined first cell when trying to normalize the table selection

Uncaught TypeError: Cannot read properties of undefined (reading 'x')
TypeError: Cannot read properties of undefined (reading 'x')
at normalizeTableSelection (roosterjs-editor-plugins/lib/plugins/TableCellSelection/utils/normalizeTableSelection.js:17:31)

PR #897, opened 2022-04-06
Fix Uncaught TypeError: Cannot read properties of undefined (reading 'x') 

BryanValverdeU:
#895


juliaroldi (approved):

JiuqingSong:
Is this another bug that need to integrate back to OWA? If so, better do it now

BryanValverdeU:
This was reported by Julia; I did not see any reports of this issue. But I will increment the version again, so we fix this issue too.

PR #898, opened 2022-04-06
Add zIndex option for TableResizer

adrienWeiss:
Hello,

TableResizer overlay elements on hover are appended to the matching document `body` and may be hidden below the editor itself when there is a stacking context defined.
(since d2e134d0515eba8807af2f97d82f75af88359f6b)

I've been thinking of an easy solution, by adding a `class` to all overlay elements, so we can safely add z-index to them, but I assume that is not "the way" this lib solves such cases. If you think it's a better solution feel free to reject this PR and I will do it in another one.

Maybe we can also insert those elements right after the Editor but I think there is a reason it is now appended on the `body`.

Otherwise, here is my proposal : 

TableResize plugin can now accept an `options` object with a `zIndex`
property, which will add `z-index` style to all overlay elements.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableInserter.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableInserter.ts
@@ -14,6 +14,7 @@ const INSERTER_BORDER_SIZE = 1;
 export default function createTableInserter(
     editor: IEditor,
     td: HTMLTableCellElement,
+    zIndex: number,
     isRTL: boolean,
     isHorizontal: boolean,
     onInsert: (table: HTMLTableElement) => void
@@ -34,6 +35,10 @@ export default function createTableInserter(
             ),
             document
         ) as HTMLDivElement;
+
+        if (zIndex !== undefined) {
+            div.style.zIndex = zIndex.toString();
+        }
 
         if (isHorizontal) {
             div.style.left = `${
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableResizer.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableResizer.ts
@@ -12,6 +12,7 @@ const MIN_CELL_HEIGHT = 20;
  */
 export default function createTableResizer(
     table: HTMLTableElement,
+    zIndex: number,
     zoomScale: number,
     isRTL: boolean,
     onStart: () => void,
@@ -27,6 +28,11 @@ export default function createTableResizer(
 
     div.style.width = `${TABLE_RESIZER_LENGTH}px`;
     div.style.height = `${TABLE_RESIZER_LENGTH}px`;
+
+    if (zIndex !== undefined) {
+        div.style.zIndex = zIndex.toString();
+    }
+
     document.body.appendChild(div);
 
     const context: DragAndDropContext = {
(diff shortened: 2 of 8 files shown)


packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableInserter.ts
```diff
@@ -35,6 +36,10 @@ export default function createTableInserter(
             document
         ) as HTMLDivElement;
 
+        if (!zIndex !== undefined) {
```
BryanValverdeU, line 39: I think we can remove the `!` before zIndex.

adrienWeiss: oops sorry 😓 
commit amended & rebased

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableResizer.ts
```diff
@@ -27,6 +28,11 @@ export default function createTableResizer(
 
     div.style.width = `${TABLE_RESIZER_LENGTH}px`;
     div.style.height = `${TABLE_RESIZER_LENGTH}px`;
+
+    if (zIndex !== undefined) {
+        div.style.zIndex = zIndex.toString();
+    }
```
JiuqingSong, line 34: Instead of just pass in a zIndex, I suggest we make it more common so that if in the future people have more requirement to customize the elements, they don't need to add more parameters.

My suggestion is to allow passing in a callback to customized the createElement process. For example:

```ts
class TableResize(
...,
onCreateHelperElements: (createElementData: CreateElementData, helperType: 'CellResizer' | 'TableInserter' | 'TableResizer' | 'TableSelector') => void) {...}
```
Then in createTableResizer, you can do:

```ts
const data = isRTL ?  KnownCreateElementDataIndex.TableResizerRTL          : KnownCreateElementDataIndex.TableResizerLTR;
onCreateHelperElements?.(data, 'TableResizer');
const div = createElement(data, doc);
```

Then in your code:
```ts
const tabeResize = new TableResize( ..., data => {
   data.style = (data.style || '') + ';zIndex: 10';
});
```

With this way you can do any customization to the helper elements you need.

adrienWeiss: I think it's a good idea, but I'm having trouble coding it due to my lack of knowledge of Rooster codebase & Typescript. In your code sample above `data` is an Index. Maybe you meant to expose `div` in the callback, but I don't know how to properly handle all possible types for this element (looks like it's at least `CreateElementData` and `HTMLDivElement` 🤔 ) and how to correctly declare types for this callback.

You can close this PR if you want, and I'll open an issue so we can track this problem as other users may encounter z-index issues.

JiuqingSong: Ok I can make a PR later.

JiuqingSong: Please see #909

adrienWeiss: Thank you @JiuqingSong  🙏

adrienWeiss:
closing in favor of #909

PR #900, opened 2022-04-07
Fix #899

### LINKED ISSUE #899: Check if the TabKeyFeatures before modifying the ListPosition to be inside
This change only should be done when TabKeyFeatures is enabled.

BryanValverdeU:
Only apply style when TabKeyFeatures is enabled
#899

### DIFF
--- a/packages/roosterjs-editor-api/lib/format/setIndentation.ts
+++ b/packages/roosterjs-editor-api/lib/format/setIndentation.ts
@@ -58,10 +58,6 @@ export default function setIndentation(editor: IEditor, indentation: Indentation
                         ExperimentalFeatures.TabKeyTextFeatures
                     );
 
-                    if (isTabKeyTextFeaturesEnabled) {
-                        vList.rootList.style.listStylePosition = 'inside';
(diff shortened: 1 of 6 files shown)


packages/roosterjs-editor-api/lib/format/setIndentation.ts
```diff
@@ -58,7 +58,9 @@ export default function setIndentation(editor: IEditor, indentation: Indentation
                         ExperimentalFeatures.TabKeyTextFeatures
                     );
 
-                    vList.rootList.style.listStylePosition = 'inside';
+                    if (isTabKeyTextFeaturesEnabled) {
```
JiuqingSong, line 61: Need to bump version?

JiuqingSong (approved):

JiuqingSong:
try change the version of @types/react-dom to be 17.0.11 to fix the build

        "@types/react-dom": "17.0.11",

JiuqingSong:
https://github.com/microsoft/roosterjs/pull/901 this should fix it

PR #906, opened 2022-04-14
Update const enum to enum

nik-john:
## What?

Updating all `const enum`s to `enum`.

## Why?

`const` enums don't always transpile well on TypeScript projects like Yammer becayse of [this issue](https://ncjamieson.com/dont-export-const-enums/). Also, this should be a harmless change as there is no downside to using a regular `enum` instead.

## References

* Similar [issue](https://github.com/OfficeDev/microsoft-teams-library-js/issues/267) and [fix](https://github.com/OfficeDev/microsoft-teams-library-js/pull/387) in the microsoft-teams-library-js codebase

* [Discussion on TypeScript repo](https://github.com/Microsoft/TypeScript/issues/5243)

## TODO

* Also make corresponding changes at other spots where we use `const enum`

### DIFF
--- a/demo/scripts/controls/colorPicker/ColorPicker.tsx
+++ b/demo/scripts/controls/colorPicker/ColorPicker.tsx
@@ -10,7 +10,7 @@ export interface ColorPickerProps {
     className?: string;
 }
 
-const enum Keys {
+enum Keys {
     PageUp = 33,
     PageDown = 34,
     End = 35,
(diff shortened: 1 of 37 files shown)


JiuqingSong:
I have tried this solution and it causes 16KB+ bundles size increase which is not acceptable.

My plan is to change build script to publish two different versions, one with enum and one with const enum. Then everyone can pick up their preferred version.

To unblock your current development, you can use version 8.19.3-dev.1 (https://www.npmjs.com/package/roosterjs/v/8.19.3-dev.1) for now as a workaround. It has exactly same code with 8.19.3 and only change const enum to enum.

JiuqingSong:
https://github.com/microsoft/roosterjs/pull/911

PR #910, opened 2022-04-18
Fix table info 

juliaroldi:
If the table does no have table information, we should use the default format as format Info.


packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -109,7 +109,10 @@ export default class VTable {
                     }
                 }
             });
-            this.formatInfo = getTableFormatInfo(this.table);
+            this.formatInfo = {
+                ...DEFAULT_FORMAT,
```
JiuqingSong, line 113: No, if table doesn't have table info, it means it is not a table created by roosterjs. Then we should respect its original style, and do not format it using our code unless user explicitly want to do so

PR #911, opened 2022-04-18
Fix const enum issue (Fix #641)

### LINKED ISSUE #641: Use `export enum` instead of `export const enum`
Hi, I have run into a problem about `roosterjs-editor-types` package. It seems not to be able to import and use some enums which exported with `export const` if the project typescript complier has an `isolatedModules` option.

Here is the reference about the conflict of `isolatedModules` option and `const enum`. https://ncjamieson.com/dont-export-const-enums/

I wonder if it makes sense to refactor `export const enum` to `export enum`. Thanks a lot!

JiuqingSong:
We have got a lot of complain that exporting const enum will cause compatibility issue, e.g. #641 . To fix, we are doing this:
1. Replace current public const enum to enum to satisfy the requirement
2. Add build script to replace those enum back to const enum when build in a separate agent with a tagged version "*.*.*-constenum.0"
3. After 2, build and publish the code with tagged version
So that people can use code from @latest to get best compatibility, and we can also use code from @constenum to get best bundle size.

### DIFF
--- a/.github/workflows/publish-size-optimized-package.yml
+++ b/.github/workflows/publish-size-optimized-package.yml
@@ -1,4 +1,4 @@
-name: Publish smaller package
+name: Publish size optimized package
 on:
     push:
         branches:
--- a/.github/workflows/publish-smaller-pack.yml
+++ b/.github/workflows/publish-smaller-pack.yml
@@ -1,0 +1,30 @@
+name: Publish smaller package
+on:
+    push:
+        branches:
+            - master
+jobs:
+    publish-smaller-pack:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v2.3.1
+              with:
+                  persist-credentials: false
+
+            - name: Set Node Version
+              uses: actions/setup-node@v2
+              with:
+                  node-version: 'v14.18.2'
+
+            - name: Install dependencies
+              run: npm install
+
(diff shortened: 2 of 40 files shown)


nik-john:
Thanks for tackling this gnarly issue 👍 . One suggestion regarding naming - from an end user perspective, naming the releases after the difference in _effect_ between the two (i.e. **@latest** and **@size-optimized**) might possibly make more sense as we might add more than just the `const enum` fix to optimize bundle size. At that point it might not be accurate to name the release after just one of the changes. Just some food for thought.

JiuqingSong:
> Thanks for tackling this gnarly issue 👍 . One suggestion regarding naming - from an end user perspective, naming the releases after the difference in _effect_ between the two (i.e. **@latest** and **@size-optimized**) might possibly make more sense as we might add more than just the `const enum` fix to optimize bundle size. At that point it might not be accurate to name the release after just one of the changes. Just some food for thought.

Makes sense!

nik-john:
@JiuqingSong  Do we have a timeline for when this might be merged so that we can plan to bump up our RoosterJS version?

JiuqingSong:
> @JiuqingSong Do we have a timeline for when this might be merged so that we can plan to bump up our RoosterJS version?

Just need to wait for an approval

BryanValverdeU (approved):

PR #913, opened 2022-04-19
Add dataset to roosterJStables

juliaroldi:
Add dataset to identify tables that were created by roosterJS, so when the formatInfo is null, we can set set the defaultFormat.


JiuqingSong:
First question: Why we need to do this? What is the matter if format is null?

PR #915, opened 2022-04-20
Adapt resize handler to make it more accessible 

juliaroldi:
Changes: 
Wrapper border to 2px
Handler size to 10x10px

Added: 
Size adaptation: remove side handlers for small  images. 
Circular handlers: 

resizeHandler

Should we make the rotate handler smaller when the image is small?

### DIFF
--- a/demo/scripts/controls/sidePane/editorOptions/Plugins.tsx
+++ b/demo/scripts/controls/sidePane/editorOptions/Plugins.tsx
@@ -73,6 +73,7 @@ export default class Plugins extends React.Component<PluginsProps, {}> {
         moreOptions?: JSX.Element
     ): JSX.Element {
         const checked = this.props.state.pluginList[id];
+
         return (
             <tr>
                 <td className={styles.checkboxColumn}>
--- a/demo/scripts/controls/sidePane/editorOptions/Plugins.tsx
+++ b/demo/scripts/controls/sidePane/editorOptions/Plugins.tsx
@@ -14,6 +14,8 @@ export default class Plugins extends React.Component<PluginsProps, {}> {
     private linkTitle = React.createRef<HTMLInputElement>();
     private watermarkText = React.createRef<HTMLInputElement>();
     private forcePreserveRatio = React.createRef<HTMLInputElement>();
+    private sizeAdaptiveImageHandles = React.createRef<HTMLInputElement>();
+    private circularImageHandles = React.createRef<HTMLInputElement>();
 
     render() {
         return (
@@ -51,6 +53,26 @@ export default class Plugins extends React.Component<PluginsProps, {}> {
                             this.forcePreserveRatio,
                             this.props.state.forcePreserveRatio,
                             (state, value) => (state.forcePreserveRatio = value)
+                        )
+                    )}
+                    {this.renderPluginItem(
+                        'imageEdit',
+                        'Image Edit Plugin',
+                        this.renderCheckBox(
+                            'Size Adaptive Image Handles',
+                            this.sizeAdaptiveImageHandles,
+                            this.props.state.sizeAdaptiveImageHandles,
+                            (state, value) => (state.sizeAdaptiveImageHandles = value)
+                        )
+                    )}
+                    {this.renderPluginItem(
+                        'imageEdit',
+                        'Image Edit Plugin',
+                        this.renderCheckBox(
+                            'Circular Image Handles',
+                            this.circularImageHandles,
+                            this.props.state.circularImageHandles,
+                            (state, value) => (state.circularImageHandles = value)
                         )
                     )}
                     {this.renderPluginItem('cutPasteListChain', 'CutPasteListChainPlugin')}
--- a/demo/scripts/controls/sidePane/editorOptions/Plugins.tsx
+++ b/demo/scripts/controls/sidePane/editorOptions/Plugins.tsx
@@ -14,8 +14,6 @@ export default class Plugins extends React.Component<PluginsProps, {}> {
     private linkTitle = React.createRef<HTMLInputElement>();
     private watermarkText = React.createRef<HTMLInputElement>();
     private forcePreserveRatio = React.createRef<HTMLInputElement>();
-    private sizeAdaptiveImageHandles = React.createRef<HTMLInputElement>();
-    private circularImageHandles = React.createRef<HTMLInputElement>();
 
     render() {
         return (
@@ -55,26 +53,6 @@ export default class Plugins extends React.Component<PluginsProps, {}> {
                             (state, value) => (state.forcePreserveRatio = value)
                         )
                     )}
-                    {this.renderPluginItem(
-                        'imageEdit',
-                        'Image Edit Plugin',
-                        this.renderCheckBox(
-                            'Size Adaptive Image Handles',
-                            this.sizeAdaptiveImageHandles,
-                            this.props.state.sizeAdaptiveImageHandles,
-                            (state, value) => (state.sizeAdaptiveImageHandles = value)
-                        )
-                    )}
-                    {this.renderPluginItem(
-                        'imageEdit',
-                        'Image Edit Plugin',
-                        this.renderCheckBox(
-                            'Circular Image Handles',
-                            this.circularImageHandles,
-                            this.props.state.circularImageHandles,
-                            (state, value) => (state.circularImageHandles = value)
-                        )
-                    )}
                     {this.renderPluginItem('cutPasteListChain', 'CutPasteListChainPlugin')}
                     {this.renderPluginItem('tableResize', 'Table Resize Plugin')}
                     {this.renderPluginItem('pickerPlugin', 'Sample Picker Plugin')}
@@ -95,7 +73,6 @@ export default class Plugins extends React.Component<PluginsProps, {}> {
         moreOptions?: JSX.Element
     ): JSX.Element {
         const checked = this.props.state.pluginList[id];
-
         return (
             <tr>
                 <td className={styles.checkboxColumn}>
--- a/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
@@ -74,6 +74,8 @@ const DefaultOptions: Required<ImageEditOptions> = {
     minRotateDeg: 5,
     imageSelector: 'img',
     rotateIconHTML: null,
+    sizeAdaptiveHandlers: false,
+    circularHandlers: false,
 };
 
 /**
@@ -363,11 +365,14 @@ export default class ImageEdit implements EditorPlugin {
 
         // Get HTML for all edit elements (resize handle, rotate handle, crop handle and overlay, ...) and create HTML element
         const options: ImageHtmlOptions = {
+            editInfo: this.editInfo,
             borderColor: this.options.borderColor,
             rotateIconHTML: this.options.rotateIconHTML,
             rotateHandleBackColor: this.editor.isDarkMode()
                 ? DARK_MODE_BGCOLOR
                 : LIGHT_MODE_BGCOLOR,
+            sizeAdaptiveHandlers: this.options.sizeAdaptiveHandlers,
+            circularHandlers: this.options.circularHandlers,
         };
         const htmlData: CreateElementData[] = [];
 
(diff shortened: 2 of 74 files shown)


JiuqingSong:
Nice fix. Is there a bug related to this fix?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/imageEditors/Resizer.ts
```diff
@@ -108,59 +110,117 @@ export function doubleCheckResize(
...
+                      ),
                       className: ImageEditElementClass.ResizeHandle,
                       dataset: { x, y },
                   },
               ],
           };
 }
+
+function setHandlerStyle(
+    circularHandlers: boolean,
```
JiuqingSong, line 209: Please rename "Handle**r**s" to "Handles", same for others.

Handler usually means a callback function, here "Handle" means the UI element.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/imageEditors/Resizer.ts
```diff
@@ -108,59 +110,117 @@ export function doubleCheckResize(
...
+    isSideHandlers: boolean,
+    y: string,
+    borderColor: string,
+    direction: string,
+    topOrBottom: string,
+    leftOrRight: string
+) {
+    if (!circularHandlers) {
+        return `position:relative;width:${RESIZE_HANDLE_SIZE}px;height:${RESIZE_HANDLE_SIZE}px;background-color: ${borderColor};cursor:${direction}-resize;${topOrBottom}:-${RESIZE_HANDLE_MARGIN}px;${leftOrRight}:-${RESIZE_HANDLE_MARGIN}px;`;
+    } else if (!isSideHandlers) {
```
JiuqingSong, line 219: do not use the combination of boolean values to decide which value to return. It is very hard to understand. Use const enum instead.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/imageEditors/Rotator.ts
```diff
@@ -85,7 +85,7 @@ export function getRotateHTML({
                 {
                     tag: 'div',
                     className: ImageEditElementClass.RotateHandle,
-                    style: `position:absolute;background-color:${rotateHandleBackColor};border:solid 1px ${borderColor};border-radius:50%;width:${ROTATE_SIZE}px;height:${ROTATE_SIZE}px;left:-${handleLeft}px;cursor:move`,
+                    style: `position:absolute;background-color:${rotateHandleBackColor};border:solid 1px ${borderColor};border-radius:50%;width:${ROTATE_SIZE}px;height:${ROTATE_SIZE}px;left:-${handleLeft}px;cursor:move;z-index:2`,
```
JiuqingSong, line 88: Why need z-index here?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/types/ImageHtmlOptions.ts
```diff
@@ -19,4 +21,22 @@ export default interface ImageHtmlOptions {
...
      */
     rotateHandleBackColor: string;
+
+    /**
+     * Image editing data
+     */
+    editInfo: ImageEditInfo;
+
+    /**
+     * The number of handlers should be adapted to the image size
```
JiuqingSong, line 31: The comment says it is a number, but the actual type is a boolean. Which is right?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/types/ImageHtmlOptions.ts
```diff
@@ -19,4 +21,22 @@ export default interface ImageHtmlOptions {
...
+     * 8 handlers: 100x100px or bigger
+     * 4 handlers: 50x50px or bigger
+     * 1 handler: smaller than 50x50px
+     */
+    sizeAdaptiveHandlers?: boolean;
+
+    /**
+     * The handlers should have circular borders
+     */
+    circularHandlers?: boolean;
```
JiuqingSong, line 41: I think we don't need this one. Let's just change it to circular if we have decided

packages/roosterjs-editor-types/lib/interface/ImageEditOptions.ts
```diff
@@ -46,4 +46,16 @@ export default interface ImageEditOptions {
...
+     * The number of handlers should be adapted to the image size
+     * 8 handlers: 100x100px or bigger
+     * 4 handlers: smaller than 100x100px
+     */
+    sizeAdaptiveHandlers?: boolean;
+
+    /**
+     * The handlers should have circular borders
+     */
+    circularHandlers?: boolean;
```
JiuqingSong, line 60: same here

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/imageEditors/Resizer.ts
```diff
@@ -108,59 +110,117 @@ export function doubleCheckResize(
...
  */
 export function getSideResizeHTML({
+    editInfo: editInfo,
     borderColor: resizeBorderColor,
+    sizeAdaptiveHandlers: sizeAdaptiveHandlers,
+    circularHandlers: circularHandlers,
 }: ImageHtmlOptions): CreateElementData[] {
+    const { widthPx, heightPx } = editInfo;
+    if (sizeAdaptiveHandlers && widthPx && heightPx && imageArea(widthPx, heightPx) < 10000) {
+        return;
```
JiuqingSong, line 148: return what?

JiuqingSong: Why return when sizeAdaptiveHandlers is true?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/imageEditors/Rotator.ts
juliaroldi: So the handle do not be overlaid by the handle.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/types/ImageHtmlOptions.ts
juliaroldi: I think we can use this property to make a flight for this feature.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/imageEditors/Resizer.ts
juliaroldi: when sizeAdaptiveHandlers is true, we do not need to create the side handles, only the corners.

demo/scripts/controls/sidePane/editorOptions/Plugins.tsx
```diff
@@ -53,6 +55,26 @@ export default class Plugins extends React.Component<PluginsProps, {}> {
...
+                        'imageEdit',
+                        'Image Edit Plugin',
+                        this.renderCheckBox(
+                            'Size Adaptive Image Handles',
+                            this.sizeAdaptiveImageHandles,
+                            this.props.state.sizeAdaptiveImageHandles,
+                            (state, value) => (state.sizeAdaptiveImageHandles = value)
+                        )
+                    )}
+                    {this.renderPluginItem(
```
JiuqingSong, line 68: this causes 3 ImageResizePlugins:
image

packages/roosterjs-editor-types/lib/interface/ImageEditOptions.ts
```diff
@@ -46,4 +46,16 @@ export default interface ImageEditOptions {
...
+     * Adapt handles number to the image size
+     * 8 handlers: 100x100px or bigger
+     * 4 handlers: smaller than 100x100px
+     */
+    sizeAdaptiveHandles?: boolean;
+
+    /**
+     * The handlers should have circular borders
+     */
+    circularHandles?: boolean;
```
JiuqingSong, line 60: Instead of having a parameter here, I prefer using editor ExperimentalFeature. So once we feel it is good enough we can turn it on by default

packages/roosterjs-editor-types/lib/interface/ImageEditOptions.ts
```diff
@@ -46,4 +46,16 @@ export default interface ImageEditOptions {
      * @default A predefined SVG icon
      */
     rotateIconHTML?: string;
+
+    /**
+     * Adapt handles number to the image size
+     * 8 handlers: 100x100px or bigger
+     * 4 handlers: smaller than 100x100px
+     */
+    sizeAdaptiveHandles?: boolean;
```
JiuqingSong, line 55: We can set this always be true, and no need to have this parameter

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/types/ImageHtmlOptions.ts
```diff
@@ -19,4 +21,22 @@ export default interface ImageHtmlOptions {
...
+     * 8 handlers: 100x100px or bigger
+     * 4 handlers: 50x50px or bigger
+     * 1 handler: smaller than 50x50px
+     */
+    sizeAdaptiveHandles?: boolean;
+
+    /**
+     * The handlers should have circular borders
+     */
+    circularHandles?: boolean;
```
JiuqingSong, line 41: same here, no need to have these parameters

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/imageEditors/Rotator.ts
JiuqingSong: Can we solve this issue by change the rendering order? If possible, avoid using zIndex

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/imageEditors/Resizer.ts
```diff
@@ -109,58 +119,120 @@ export function doubleCheckResize(
...
 }: ImageHtmlOptions): CreateElementData[] {
+    const { widthPx, heightPx } = editInfo;
+    if (
+        sizeAdaptiveHandles &&
+        widthPx &&
+        heightPx &&
+        imageArea(widthPx, heightPx) < MAX_SMALL_SIZE_IMAGE
+    ) {
+        return null;
+    }
```
JiuqingSong, line 167: You can do this check outside this function, so no need to pass editInfo in.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -74,6 +74,8 @@ const DefaultOptions: Required<ImageEditOptions> = {
     minRotateDeg: 5,
     imageSelector: 'img',
     rotateIconHTML: null,
+    sizeAdaptiveHandles: false,
+    circularHandles: false,
```
JiuqingSong, line 78: consider make these options true by default

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -362,14 +368,22 @@ export default class ImageEdit implements EditorPlugin {
         this.image.style.maxWidth = null;
 
         // Get HTML for all edit elements (resize handle, rotate handle, crop handle and overlay, ...) and create HTML element
+
```
JiuqingSong, line 371: remove this empty line

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -362,14 +368,22 @@ export default class ImageEdit implements EditorPlugin {
...
         const options: ImageHtmlOptions = {
             borderColor: this.options.borderColor,
             rotateIconHTML: this.options.rotateIconHTML,
             rotateHandleBackColor: this.editor.isDarkMode()
                 ? DARK_MODE_BGCOLOR
                 : LIGHT_MODE_BGCOLOR,
+            isSmallImage: isASmallImage(
+                this.editInfo,
+                this.editor.isFeatureEnabled(ExperimentalFeatures.AdaptiveHandlesResizer)
```
JiuqingSong, line 380: This check also happens in line 382, save the result to a local var

JiuqingSong (approved):

Issue #872, opened 2022-03-30
Problem with setting content with colgroup


adrienWeiss:
**Describe the bug**
There is a bug I'm reproducing both on outlook web & on my project with setting initial content including a `table` with a `colgroup` definition.
This is notably affecting content pasted from Google Spreadsheets, after it was saved & restored as initial content.

**To Reproduce**
Init Rooster with content such as : 
`<table><thead><colgroup><col width="100" ><col width="100" ></thead><tbody><tr><td>col 1</td><td>col 2</td></tr><tr><td>col 1</td><td>col 2</td></tr></tbody></table>`

Both devtools inspector & `getContent()` will return 2 distinct tables
`<table><thead></thead><colgroup><col width="100"><col width="100"></colgroup></table><table><tbody><tr><td>col 1</td><td>col 2</td></tr><tr><td>col 1</td><td>col 2</td></tr></tbody></table>`

**Expected behavior**
Rooster should render & return the same injected HTML. Something is breaking somewhere. I have tried with and without self-closing `<col>` tags & with and without `<thead> / <tbody>` tags.

**Device Information**
 - OS: [e.g. iOS] MacOS 12.3
 - Browser [e.g. chrome, safari] Latest Chrome
 - Version [e.g. 22] Rooster 8.10.0 and 8.18.0

PR #916, opened 2022-04-20
Fix Problem with setting content with colgroup #872

BryanValverdeU:
#872

When setting initial content to start with a table with colgroup the ensureTypeInContainer selected the colgroup element as the initial position.

Then select the block element of the initial position is going to be the thead element that is containing the Colgroup Node. Then the collapseToSingleElement causes the thead to be moved to another table.

Changes:
If the initialContent is a table, try select the first td/th and use it as initial position.

### DIFF
--- a/packages/roosterjs-editor-core/lib/coreApi/ensureTypeInContainer.ts
+++ b/packages/roosterjs-editor-core/lib/coreApi/ensureTypeInContainer.ts
@@ -28,8 +28,9 @@ export const ensureTypeInContainer: EnsureTypeInContainer = (
     keyboardEvent?: KeyboardEvent
 ) => {
     const table = findClosestElementAncestor(position.node, core.contentDiv, 'table');
-    if (table) {
-        position = new Position(table.querySelector('td,th'), PositionType.Begin);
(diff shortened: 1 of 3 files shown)


packages/roosterjs-editor-core/lib/coreApi/ensureTypeInContainer.ts
```diff
@@ -26,7 +27,12 @@ export const ensureTypeInContainer: EnsureTypeInContainer = (
     position: NodePosition,
     keyboardEvent?: KeyboardEvent
 ) => {
+    const table = findClosestElementAncestor(position.node, core.contentDiv, 'table');
+    if (table) {
+        position = new Position(table.querySelector('td,th'), PositionType.Begin);
```
JiuqingSong, line 32: In some corner case, querySelector my return null? then position will be null

BryanValverdeU: Good point, fixing this.

JiuqingSong (approved):

PR #945, opened 2022-05-02
Apply Default Format when using Clear Block Format.

BryanValverdeU:
When using Clear block format, the default editor format is not applied.
Adding applyDefaultFormat after we clear all the Block Formats


packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -89,6 +89,8 @@ function clearAttribute(element: HTMLElement) {
     for (let attr of toArray(element.attributes)) {
         if (isTableCell && attr.name == 'style') {
             removeNonBorderStyles(element);
+        } else if (safeInstanceOf(element, 'HTMLTableElement') && attr.name == 'style') {
```
JiuqingSong, line 92: Like line 87, you can move safeInstanceOf check out of the for loop so you only need to check it once

JiuqingSong (approved):

PR #960, opened 2022-05-04
Mantain the list style type when it is already applied in the list.

BryanValverdeU:
If a list already contains list style, when pressing Enter in the list we modify the list item style depending on the list indentation.
This PR tries to maintain the applied list style, so the user does not see the list style changing when editing the list.

Before:
maintainListStyle2

After:
maintainListStyle


packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -228,17 +230,26 @@ export default class VListItem {
...
+            listStack[stackLength].appendChild(newList);
             listStack.push(newList);
-        }
 
+            //If the current node parent is in the same deep child index,
+            //apply the styles of the current parent list to the new list
+            if (this.getDeepChildIndex(originalRoot) == stackLength) {
+                const styles = getStyles(this.node.parentElement!);
+                if (styles['list-style-type']) {
+                    setStyles(newList, styles);
```
JiuqingSong, line 249: We should only apply list-style-type, and keep other styles no change

JiuqingSong (approved):

PR #964, opened 2022-05-05
Remove Duplicated Table Ids to avoid unexpected behavior in TableCellSelection

BryanValverdeU:
**Describe the bug**
Having multiple tables with the same ID can cause issues when selecting cells.

selectionBug

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
@@ -93,7 +93,7 @@ export default class TableCellSelection implements EditorPlugin {
         if (this.editor) {
             switch (event.eventType) {
                 case PluginEventType.EditorReady:
-                    this.editor.queryElements('table', table => this.removeDuplicatedId(table));
+                    this.editor.queryElements('table', table => this.preventDuplicatedId(table));
                     break;
                 case PluginEventType.EnteredShadowEdit:
                     const selection = this.editor.getSelectionRangeEx();
@@ -186,7 +186,7 @@ export default class TableCellSelection implements EditorPlugin {
                 clonedVTable.selection = this.tableRange;
                 removeCellsOutsideSelection(clonedVTable);
                 clonedVTable.writeBack();
-                this.removeDuplicatedId(clonedVTable.table);
+                this.preventDuplicatedId(clonedVTable.table);
 
                 event.range.selectNode(clonedTable);
 
@@ -199,7 +199,7 @@ export default class TableCellSelection implements EditorPlugin {
                 }
             }
         }
-        queryElements(event.clonedRoot, 'table', node => this.removeDuplicatedId(node));
+        queryElements(event.clonedRoot, 'table', node => this.preventDuplicatedId(node));
     }
 
     //#region Key events
@@ -737,7 +737,7 @@ export default class TableCellSelection implements EditorPlugin {
         }
     }
 
-    private removeDuplicatedId(node: HTMLElement) {
+    private preventDuplicatedId(node: HTMLElement) {
         if (this.editor?.queryElements('#' + node.id).length > 1) {
             node.removeAttribute('id');
         }
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
@@ -26,7 +26,6 @@ import {
     VTable,
     Position,
     contains,
-    queryElements,
 } from 'roosterjs-editor-dom';
 
 const TABLE_CELL_SELECTOR = 'td,th';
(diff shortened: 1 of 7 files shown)


packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -91,6 +92,9 @@ export default class TableCellSelection implements EditorPlugin {
     onPluginEvent(event: PluginEvent) {
         if (this.editor) {
             switch (event.eventType) {
+                case PluginEventType.EditorReady:
+                    this.editor.queryElements('table', table => table.removeAttribute('id'));
```
JiuqingSong, line 96: Suggest we don't remove existing id. Only if we found duplicated id, we can change them.

JiuqingSong: Duplicated id can be created any time.

Suggest we do the duplicate check when start to select. e.g.

```ts
function createUniqueId() {
  if (!node.id) { node.id = .... }
  else if (editor.queryElements('#' + node.id).length > 1) {
    // Change to another id since it is duplicated
  }
}
 ```

File: packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts [hunk 2]
```diff
@@ -182,7 +186,7 @@ export default class TableCellSelection implements EditorPlugin {
                 clonedVTable.selection = this.tableRange;
                 removeCellsOutsideSelection(clonedVTable);
                 clonedVTable.writeBack();
-                clonedVTable.table.id = '';
+                clonedVTable.table.removeAttribute('id');
```
JiuqingSong, line 189: same here

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -195,6 +199,7 @@ export default class TableCellSelection implements EditorPlugin {
                 }
             }
         }
+        queryElements(event.clonedRoot, 'table', element => element.removeAttribute('id'));
```
JiuqingSong, line 202: same here

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
BryanValverdeU: Sounds good, fixed in the latest push, I think we can just remove the ID if duplicated in the TableCell Plugin.
If we need to assign a new Id we can do it in the selectTable Api.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -91,6 +92,9 @@ export default class TableCellSelection implements EditorPlugin {
     onPluginEvent(event: PluginEvent) {
         if (this.editor) {
             switch (event.eventType) {
+                case PluginEventType.EditorReady:
+                    this.editor.queryElements('table', table => this.preventDuplicatedId(table));
+                    break;
```
JiuqingSong, line 97: do we still need this change?

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -182,7 +186,7 @@ export default class TableCellSelection implements EditorPlugin {
                 clonedVTable.selection = this.tableRange;
                 removeCellsOutsideSelection(clonedVTable);
                 clonedVTable.writeBack();
-                clonedVTable.table.id = '';
+                this.preventDuplicatedId(clonedVTable.table);
```
JiuqingSong, line 189: same here

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -195,6 +199,7 @@ export default class TableCellSelection implements EditorPlugin {
                 }
             }
         }
+        queryElements(event.clonedRoot, 'table', node => this.preventDuplicatedId(node));
```
JiuqingSong, line 202: same here

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -731,6 +736,12 @@ export default class TableCellSelection implements EditorPlugin {
             this.editor?.select(this.vTable.table, normalizeTableSelection(this.vTable));
         }
     }
+
+    private preventDuplicatedId(node: HTMLElement) {
```
JiuqingSong, line 740: same here

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
BryanValverdeU: I think you are right; we can just remove the duplicated id in selectTable, removing these changes here in the plugin

JiuqingSong (approved):

PR #967, opened 2022-05-10
Add metadata API for roosterjs

JiuqingSong:
In order to store/retrieve metadata for a given element, add types and functions below:
- API to get/set metadata `getMetadata`, `setMetadata`, `removeMetadata`
- API to validate metadata object `validate`
- Helper functions to create type definition `create*Definition`
- Types of definitions

When get a metadata, we can do:
```ts
const metadata = getMetadata(element, def);
```

Here `def` is an optional type definition object used for validating if the metadata object is valid. For example, we have a metadata type:

```ts
interface MyMetadata {
  x: number;
  y: string;
}
```
and when retrieve it, we can use a type definition object to help validate if the given element contains a valid object with this type:

```ts
const MyMetadataDefinition = createObjectDefinition<MyMetadata>({
  x: createNumberDefinition(),
  y: createStringDefinition(),
});

const metadata = getMetadata(element, MyMetadataDefinition );
```

If the given element doesn't contain a valid metadata string, it will return null.

We can also create definition object for other types like array, boolean. And there is also a CustomizeDefinition type to support customization of a validation.

Related test cases added.

### DIFF
--- a/packages/roosterjs-editor-dom/lib/metadata/metadata.ts
+++ b/packages/roosterjs-editor-dom/lib/metadata/metadata.ts
@@ -6,11 +6,17 @@ const MetadataDataSetName = 'editingInfo';
 /**
  * Get metadata object from an HTML element
  * @param element The HTML element to get metadata object from
- * @param def The type definition of this metadata used for validate this metadata object.
+ * @param definition The type definition of this metadata used for validate this metadata object.
  * If not specified, no validation will be performed and always return whatever we get from the element
+ * @param defaultValue The default value to return if the retrieved object cannot pass the validation,
+ * or there is no metadata object at all
  * @returns The strong-type metadata object if it can be validated, or null
  */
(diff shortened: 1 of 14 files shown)


packages/roosterjs-editor-dom/lib/metadata/metadata.ts
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * Set metadata object into an HTML element
+ * @param element The HTML element to set metadata object to
+ * @param metadata The metadata object to set
+ * @param def An optional type definition object used for validate this metadata object.
+ * If not specified, metadata will be set without validation
+ * @returns True if metadata is set, otherwise false
+ */
+export function setMetadata<T>(element: HTMLElement, metadata: T, def?: Definition<T>): boolean {
```
juliaroldi, line 50: Should I use this function to store the table style info?

JiuqingSong: yes

PR #969, opened 2022-05-10
U/juliaroldi/word tables

juliaroldi:
**Bug:** 
1. Copy a table from word online dark mode and default font color
2. Paste into RoosterJS editor in dark mode 
Actual: font color turn to black
Expect: font color to be default 

**Fix:** 
1. Identify a table copied from word online in Paste plugin
2. use the wordTableConverter to remove attribute color: windomtext.


packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
```diff
@@ -126,3 +131,10 @@ function isWordDocument(htmlAttributes: Record<string, string>) {
         htmlAttributes[PROG_ID_NAME] == WORD_PROG_ID
     );
 }
+
+function getWordOnlineTable(fragment: DocumentFragment) {
```
BryanValverdeU, line 135: I think we should check for all the tables, not just one.

packages/roosterjs-editor-plugins/lib/plugins/Paste/wordConverter/wordOnlineTableConverter.ts
```diff
@@ -0,0 +1,24 @@
...
+import { VTable } from 'roosterjs-editor-dom';
+
+/**
+ * @internal
+ * Fix styles of table copied from word online
+ * @param editor
+ * @param table copied from word online
+ */
+export default function wordOnlineTableConverter(editor: IEditor, table: HTMLTableElement) {
+    const vTable = new VTable(table as HTMLTableElement);
```
BryanValverdeU, line 11: Quick question, is it possible to just use querySelectorAll('td,th') or querySelectorAll('p') in this function and edit the style if needed. That way we don't need to build the VTable object.

packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
```diff
@@ -74,8 +77,10 @@ export default class Paste implements EditorPlugin {
             const { htmlAttributes, fragment, sanitizingOption, clipboardData } = event;
             const trustedHTMLHandler = this.editor.getTrustedHTMLHandler();
             let wacListElements: Node[];
-
-            if (isWordDocument(htmlAttributes)) {
+            const table = getWordOnlineTable(fragment);
+            if (table) {
+                wordOnlineTableConverter(this.editor, table);
```
JiuqingSong, line 82: When paste from Word Online, can it fall into line 98 or any other existing branch?

If so, this should be moved into that branch, otherwise it will cause that branch not executed.

packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
JiuqingSong: agree.

packages/roosterjs-editor-plugins/lib/plugins/Paste/wordConverter/wordOnlineTableConverter.ts
```diff
@@ -0,0 +1,24 @@
...
+    vTable.cells.forEach(row =>
+        row.forEach(cell => {
+            if (cell.td) {
+                const styledElement = cell.td.querySelector('p');
+                if (editor.isDarkMode()) {
```
JiuqingSong, line 16: This should be moved out to the top, if editor is not in dark mode, no need to do all the actions, right? Or do we always want to remove this style?

JiuqingSong: I feel we should always remove it. Otherwise later when user switch to dark mode, there can still be problem.

packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
```diff
@@ -109,6 +111,8 @@ export default class Paste implements EditorPlugin {
                 clipboardData.htmlFirstLevelChildTags[0] == 'IMG'
             ) {
                 convertPasteContentForSingleImage(event, trustedHTMLHandler);
+            } else if (isWordOnlineTables(fragment)) {
```
JiuqingSong, line 114: What if the copied content contains both table and list? Will it fall into line 91?

We need to run both list and table process when copy from word only, but not a "else if" condition.

juliaroldi:
Obsolete

PR #973, opened 2022-05-11
Adding custom attributes to image element.

marcinkiewicz:
Add optional `attributes` parameter to `insertImage` function to enable setting optional attributes on the `` element which allows specifying `aria-` and `alt` tags for accessibility support.


marcinkiewicz:
Fixed Lint errors introduced by reverted PR - use `Object.keys` instead of iterating through object properties using `for .. in ..` loop.

JiuqingSong (approved):

PR #975, opened 2022-05-12
Fix #974: change regex so that text node starts with \r\n should not be skipped

### LINKED ISSUE #974: apply format to a text node start with \r\n not work
Have a text node such as "\r\ntest", and try to select and format it. The function shouldSkipNode() will return true for this kind of node due to incorrect regular expression. So the node is skipped and not formatted.

JiuqingSong:
shouldSkipNode() function uses a regex to check if a text node contains only new line char (\r, \n). But the regex actually matches any string that starts with \r\n in a single line.

Change to do full string match instead.

This impact the format API to apply format to a selection with multiple paragraphs.


JiuqingSong:
Related VSO bug: 154098

PR #978, opened 2022-05-13
Add sanitize to remove deprecated colors

juliaroldi:
Add sanitize method to remove deprecated colors from pasted content. 

**Before**
wordOnlineBugBefore

**After**
wordOnlineBug


packages/roosterjs-editor-plugins/lib/plugins/Paste/sanitizeHtmlTextFromPastedContent/sanitizeHtmlTextFromPastedContent.ts
```diff
@@ -0,0 +1,27 @@
...
+import { DeprecatedColorList } from './deprecatedColorList';
+import { HtmlSanitizerOptions } from 'roosterjs-editor-types';
+
+/**
+ * @internal
+ * Remove the deprecated colors from pasted content
+ * @sanitizingOption
+ * */
+export default function sanitizeHtmlTextFromPastedContent(
+    fragment: DocumentFragment,
```
BryanValverdeU, line 11: nit: Param is missing in the comment

BryanValverdeU, line 11: nit: Param is missing in the comment

packages/roosterjs-editor-plugins/lib/plugins/Paste/sanitizeHtmlTextFromPastedContent/sanitizeHtmlTextFromPastedContent.ts
```diff
@@ -0,0 +1,27 @@
...
+    sanitizingOption: Required<HtmlSanitizerOptions>
+) {
+    const htmlElements = fragment.querySelectorAll('*') as NodeListOf<HTMLElement>;
```
BryanValverdeU, line 14: Not sure if I am correct, but if there are many repeated Elements with tags, we would add many repeated chainSanitizerCallbacks.

I think we can filter the unique tags doing:

```
const allTags = fragment.querySelectorAll('*').map(el => getTagOfNode(el));
const uniqueTags = allTags.filter((tag, index)=>  allTags.indexOf(tag) == index) )
```

And then use unique tags to add the chainSanitizeCallbacks as you are doing below.

BryanValverdeU, line 14: Not sure if I am correct, but if there are many repeated Elements with tags, we would add many repeated chainSanitizerCallbacks.

I think we can filter the unique tags doing:

```
const allTags = fragment.querySelectorAll('*').map(el => getTagOfNode(el));
const uniqueTags = allTags.filter((tag, index)=>  allTags.indexOf(tag) == index) )
```

And then use unique tags to add the chainSanitizeCallbacks as you are doing below.

packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
```diff
@@ -75,6 +76,8 @@ export default class Paste implements EditorPlugin {
             const trustedHTMLHandler = this.editor.getTrustedHTMLHandler();
             let wacListElements: Node[];
 
+            sanitizeHtmlColorsFromPastedContent(fragment, sanitizingOption);
```
JiuqingSong, line 79: minor: suggest to move this below all the if...else blocks, to make it not looks like that "high priority"

packages/roosterjs-editor-plugins/lib/plugins/Paste/sanitizeHtmlColorsFromPastedContent/sanitizeHtmlColorsFromPastedContent.ts
```diff
@@ -0,0 +1,31 @@
...
+export default function sanitizeHtmlColorsFromPastedContent(
+    fragment: DocumentFragment,
+    sanitizingOption: Required<HtmlSanitizerOptions>
+) {
+    const htmlElements = fragment.querySelectorAll('*') as NodeListOf<HTMLElement>;
+    const allTags = Array.from(htmlElements).map(el => getTagOfNode(el));
+    const uniqueTags = allTags.filter((tag, index) => allTags.indexOf(tag) == index);
+
+    uniqueTags.forEach(tag =>
+        chainSanitizerCallback(sanitizingOption.elementCallbacks, tag, element => {
```
JiuqingSong, line 20: Instead of using elementCallbacks, you can use cssStyleCallbacksm to specify just color and background color, so that no need to query all unique tag names.

packages/roosterjs-editor-plugins/lib/plugins/Paste/sanitizeHtmlColorsFromPastedContent/sanitizeHtmlColorsFromPastedContent.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @param sanitizingOption the sanitizingOption of BeforePasteEvent
+ * */
+export default function sanitizeHtmlColorsFromPastedContent(
+    sanitizingOption: Required<HtmlSanitizerOptions>
+) {
+    ['color', 'background-color'].forEach(property => {
+        chainSanitizerCallback(sanitizingOption.cssStyleCallbacks, property, (value: string) => {
+            if (DeprecatedColorList.indexOf(value) < 0) {
+                return true;
+            }
```
JiuqingSong, line 17: else?

you can do:
```ts
(value: string) => DeprecatedColorList.indexOf(value) < 0
```

JiuqingSong (approved):

PR #992, opened 2022-05-18
Add key binding for clearing format

ianeli1:
Currently we lack a keyboard shortcut for clearing the format of the selection. 
This change binds this action to the Ctrl + Spacebar press.


JiuqingSong (approved):

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/shortcutFeatures.ts
```diff
@@ -65,7 +67,17 @@ const commands: ShortcutCommand[] = [
  */
```
BryanValverdeU, line 67: nit: Add comment for the new Shortcut

BryanValverdeU (approved):

PR #1000, opened 2022-05-25
Auto format for numbered and bullet list

juliaroldi:
Auto format for list provide automatic formatting for list triggered by some combination of characters. 

AutoFormatList

AutoFormatListNumbering

**Part 1:** Trigger list 
AutoBullet will detect the characters type and check if a list should be triggered

**Part 2**: Identify the list style trigger by some combination of characters. 
`getListType` : provide the type of the list (ordered or unordered)
`getListyStyle`: provide the style of the list. If ordered, the style will be NumberingListType, if unordered, BulletListType. 

**Part 3**: Toggle List
After AutoBullet called the `getListType` and `getListyStyle`, AutoBullet will call `toggleBullet` for unordered lists or `toggleNumbering` for ordered lists. Either of this two functions will receive a style parameter `listStyle` and will call the 
`toggleListType` function, responsible for creating the list and creating the  a `vList`, the logical representation of the list element. 

**Part 4**: VList
The VList class has the `setListStyleType` method that sets the style info into the editing info metadata, then the style of the li element will always be known. The `applyListStyle` method of VListItem will verify if the li element has the editing info metadata and if it is at first level, it will add the right list-style-type css property using the `setBulletListMarkers` or `setNumberingListMarker`.


packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -340,6 +348,49 @@ export default class VList {
...
+        start: NodePosition,
+        end: NodePosition,
+        targetStyle:
+            | NumberingListType
+            | BulletListType
+            | CompatibleBulletListType
+            | CompatibleNumberingListType
+    ) {
+        this.findListItems(start, end, item => {
+            item.getNode().title = targetStyle.toString();
```
BryanValverdeU, line 368: Title prop adds a tooltip to the List Item, I wonder if we can find a way to instead of using this property to every List Item we can just apply use the metadata api and just apply it to the rootList.
That way we only add a single property to the List instead of adding properties to all the items.

BryanValverdeU, line 368: Title prop adds a tooltip to the List Item, I wonder if we can find a way to instead of using this property to every List Item we can just apply use the metadata api and just apply it to the rootList.
That way we only add a single property to the List instead of adding properties to all the items.

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -340,6 +348,49 @@ export default class VList {
...
+                if (item.getListType() === ListType.Unordered) {
+                    setBulletListMarkers(li, style as BulletListType);
+                } else if (item.getListType() === ListType.Ordered) {
+                    setNumberingListMarkers(li, style as NumberingListType, index);
+                }
+            } else {
+                li.style.removeProperty('list-style-type');
+            }
+        }
+    }
```
BryanValverdeU, line 392: We can add this to VListItem Class

packages/roosterjs-editor-dom/lib/list/VList.ts
juliaroldi: True, that was a bad approach. I changed to use setMetadata and getMetadata, so the style info will be store in the editingInfo dataset.

BryanValverdeU (approved):

PR #1006, opened 2022-05-30
Fix multileval for unordered lists

juliaroldi:
When indent a list item it should change to the proper list style. 
**Before:** 
multiLevelBug

**After:**
multiLevelFixed


packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -216,7 +218,11 @@ export default class VListItem {
      * @param index the list item index
      */
     applyListStyle(rootList: HTMLOListElement | HTMLUListElement, index: number) {
-        const style = getMetadata(rootList) ? getMetadata(rootList) : undefined;
+        const style = getMetadata(rootList)
+            ? (getMetadata(rootList, createNumberDefinition()) as
+                  | NumberingListType
+                  | BulletListType)
+            : undefined;
```
JiuqingSong, line 225: Why do like this?

You can just do
const style = getMetadata<NumberingListType|BulletListType>(...);

And if it doesn't have a valid matadata, it will just return null

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -360,13 +366,32 @@ function createListElement(
         result = doc.createElement(listType == ListType.Ordered ? 'ol' : 'ul');
     }
 
+    // Always maintain the metadata saved in the list
+    if (originalRoot && nextLevel == 1 && listType != getListTypeFromNode(originalRoot)) {
+        const style = getMetadata(originalRoot)
```
JiuqingSong, line 371: same here

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -360,13 +366,32 @@ function createListElement(
...
+            ? (getMetadata(originalRoot, createNumberDefinition()) as
```
JiuqingSong, line 372: No need to create definition every time, just use a const.

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -360,13 +366,32 @@ function createListElement(
...
+                  | NumberingListType
+                  | BulletListType)
+            : undefined;
+        if (style) {
+            setMetadata(result, style, createNumberDefinition());
```
JiuqingSong, line 377: same here

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -31,6 +31,13 @@ import type {
     CompatibleNumberingListType,
 } from 'roosterjs-editor-types/lib/compatibleTypes';
 
+const numberDefinition = createNumberDefinition(
+    false /** isOptional */,
+    undefined /** value */,
+    0,
+    19
```
JiuqingSong, line 38: Here is a trick:

Instead of hard code a number here, you can declare "MIN", "MAX" elements in the enum to have the same value with the last item, for example:
```ts
const enum MyEnum = {
 MIN = 1,
  A = 1,
  B = 2,
   ...
  Z = 26,
  Max = 26,
}
```
Then you can use the Max value to define the definition:
```ts
createNumberDefinition(false, undefined, 0, MyEnum.MIN, MyEnum.MAX);
```

So, if we want to add some new element to the enum someday, we know we also need to update the value of MAX, and no need to worry that we forget update the value here.

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -20,9 +21,17 @@ import {
 import type { CompatibleListType } from 'roosterjs-editor-types/lib/compatibleTypes';
 
 const orderListStyles = [null, 'lower-alpha', 'lower-roman'];
+const unorderedListStyles = ['disc', 'circle', 'square'];
+const numberDefinition = createNumberDefinition(
+    false /** isOptional */,
+    undefined /** value */,
+    0,
+    19
```
JiuqingSong, line 29: same

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -360,13 +369,31 @@ function createListElement(
         result = doc.createElement(listType == ListType.Ordered ? 'ol' : 'ul');
     }
 
+    // Always maintain the metadata saved in the list
+    if (originalRoot && nextLevel == 1 && listType != getListTypeFromNode(originalRoot)) {
+        const style = getMetadata<NumberingListType | BulletListType>(
+            originalRoot,
+            numberDefinition
+        );
+        if (style) {
```
JiuqingSong, line 378: same

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -216,7 +225,7 @@ export default class VListItem {
      * @param index the list item index
      */
     applyListStyle(rootList: HTMLOListElement | HTMLUListElement, index: number) {
-        const style = getMetadata(rootList) ? getMetadata(rootList) : undefined;
+        const style = getMetadata<NumberingListType | BulletListType>(rootList, numberDefinition);
         if (style) {
```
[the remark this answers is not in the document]
juliaroldi: I used the approach you suggested and added Min and Max to the enum, so the Min value will be 0 and the types of bullets or numbering will start at 1.

packages/roosterjs-editor-types/lib/enum/BulletListType.ts
```diff
@@ -32,4 +37,9 @@ export const enum BulletListType {
      * Bullet triggered by =>
      */
     UnfilledArrow,
+
+    /**
+     * Maximum value of the enum
+     */
+    Max,
```
JiuqingSong, line 44: This will cause Max has a different number value than UnfilledArrow, but it is still a valid value. What if people call an API and pass in Max or Min? There will not be build time error but runtime will show incorrect result.

juliaroldi: That's true, I fixed. I used max -1 to validate the number. Now I'm using the last max equal to the last number value of the enum

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -31,6 +31,16 @@ import type {
     CompatibleNumberingListType,
 } from 'roosterjs-editor-types/lib/compatibleTypes';
 
+const maxEnum =
+    BulletListType.Max > NumberingListType.Max ? BulletListType.Max : NumberingListType.Max;
```
JiuqingSong, line 35: nit: You can use Math.max()

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -20,9 +21,20 @@ import {
 import type { CompatibleListType } from 'roosterjs-editor-types/lib/compatibleTypes';
 
 const orderListStyles = [null, 'lower-alpha', 'lower-roman'];
+const unorderedListStyles = ['disc', 'circle', 'square'];
+const maxEnum =
+    BulletListType.Max > NumberingListType.Max ? BulletListType.Max : NumberingListType.Max;
+
+const numberDefinition = createNumberDefinition(
```
JiuqingSong, line 28: you just need to keep this one and export it, and use it from VList. No need to have two separate definitions. Not only for code size, but also they need to always be the same. If some day we change one of them for some reason and forget to change the other one, there will be bug

JiuqingSong: or, get/set the metadata always from a same file, don't split them.

packages/roosterjs-editor-types/lib/enum/BulletListType.ts
```diff
@@ -32,4 +32,14 @@ export const enum BulletListType {
      * Bullet triggered by =>
      */
     UnfilledArrow,
+
+    /**
+     * Minimum value of the enum
+     */
+    Min = 0,
```
JiuqingSong, line 39: Move Min to top

packages/roosterjs-editor-types/lib/enum/BulletListType.ts
```diff
@@ -32,4 +32,14 @@ export const enum BulletListType {
      * Bullet triggered by =>
      */
     UnfilledArrow,
```
JiuqingSong, line 34: Add number value for each values, so that we clearly know how Min and Max are decided

packages/roosterjs-editor-types/lib/enum/NumberingListType.ts
```diff
@@ -101,4 +101,14 @@ export const enum NumberingListType {
      * Numbering triggered by I-
      */
     UpperRomanDash,
+
+    /**
+     * Minimum value of the enum
+     */
+    Min = 0,
```
JiuqingSong, line 108: same here

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -214,10 +217,15 @@ export default class VListItem {
...
      * @param index the list item index
+     * @param numberDefinition the number definitions of BulletListType or NumberingListType
      */
-    applyListStyle(rootList: HTMLOListElement | HTMLUListElement, index: number) {
-        const style = getMetadata(rootList) ? getMetadata(rootList) : undefined;
-        if (style) {
+    applyListStyle(
+        rootList: HTMLOListElement | HTMLUListElement,
+        index: number,
+        numberDefinition: NumberDefinition
```
JiuqingSong, line 225: I don't think it is a good idea to expose the numberDefinition as a parameter of a public API. That means every one can call this API with a different value of definition, which is not correct.

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -234,8 +242,13 @@ export default class VListItem {
      * Write the change result back into DOM
      * @param listStack current stack of list elements
      * @param originalRoot Original list root element. It will be reused when write back if possible
+     * @param numberDefinition the number definition of BulletListType or NumberingListType saved in the style metadata
      */
-    writeBack(listStack: Node[], originalRoot?: HTMLOListElement | HTMLUListElement) {
+    writeBack(
+        listStack: Node[],
+        originalRoot?: HTMLOListElement | HTMLUListElement,
+        numberDefinition?: NumberDefinition
```
JiuqingSong, line 250: same here

JiuqingSong: You just need to declare numberDefinition in this file as global var, and use it directly from each function

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -360,13 +375,36 @@ function createListElement(
         result = doc.createElement(listType == ListType.Ordered ? 'ol' : 'ul');
     }
 
+    // Always maintain the metadata saved in the list
+    if (
+        numberDefinition &&
```
JiuqingSong, line 380: then no need to check null here because you know it can never be null

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -216,14 +243,25 @@ export default class VListItem {
...
-                    setBulletListMarkers(this.node, style as BulletListType);
-                } else if (this.listTypes[1] === ListType.Ordered) {
-                    setNumberingListMarkers(this.node, style as NumberingListType, index);
-                }
+            if (
+                isFirstLevel &&
+                this.listTypes[1] === ListType.Unordered &&
+                style.unOrderedStyleType
+            ) {
+                setBulletListMarkers(this.node, style.unOrderedStyleType as BulletListType);
```
JiuqingSong, line 254: no need to do type cast

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -216,14 +243,25 @@ export default class VListItem {
...
+            } else if (
+                isFirstLevel &&
+                this.listTypes[1] === ListType.Ordered &&
+                style.orderedStyleType
+            ) {
+                setNumberingListMarkers(
+                    this.node,
+                    style.orderedStyleType as NumberingListType,
```
JiuqingSong, line 262: same

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -216,14 +243,25 @@ export default class VListItem {
...
-            if (this.listTypes.length < 3) {
-                if (this.listTypes[1] === ListType.Unordered) {
-                    setBulletListMarkers(this.node, style as BulletListType);
-                } else if (this.listTypes[1] === ListType.Ordered) {
-                    setNumberingListMarkers(this.node, style as NumberingListType, index);
-                }
+            if (
+                isFirstLevel &&
+                this.listTypes[1] === ListType.Unordered &&
+                style.unOrderedStyleType
```
JiuqingSong, line 252: Need to handle value 0

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getListStyle.ts
```diff
@@ -123,7 +123,7 @@ export default function getListStyle(
 ): NumberingListType | BulletListType {
     if (listType === ListType.Ordered) {
         return identifyNumberingListType(textBeforeCursor);
-    } else {
+    } else if (listType === ListType.Unordered) {
         return identifyBulletListType(textBeforeCursor);
     }
```
JiuqingSong, line 128: else? need return value

packages/roosterjs-editor-types/lib/interface/ListStyleMetadata.ts
```diff
@@ -0,0 +1,10 @@
+import { BulletListType } from '../enum/BulletListType';
+import { NumberingListType } from '../enum/NumberingListType';
+
+/**
+ * Represents the metadata of the style of a list element
+ */
+export default interface ListStyleMetadata {
+    orderedStyleType?: NumberingListType;
+    unOrderedStyleType?: BulletListType;
```
JiuqingSong, line 9: "unordered" is one word, don't upper case "O"

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getListType.ts
```diff
@@ -17,9 +33,10 @@ function isANumberingList(textBeforeCursor: string) {
...
  */
 export default function getListType(textBeforeCursor: string): ListType {
-    if (isABulletList(textBeforeCursor)) {
+    const trigger = textBeforeCursor.replace(/\s/g, '');
+    if (isABulletList(trigger)) {
         return ListType.Unordered;
-    } else if (isANumberingList(textBeforeCursor)) {
+    } else if (isANumberingList(trigger)) {
         return ListType.Ordered;
     }
```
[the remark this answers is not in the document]
juliaroldi: It is to check if the character is triggering a order or unordered list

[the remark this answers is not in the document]
juliaroldi: We have a set of character that triggers numbering lists and another set of characters that triggers bullet list, this file is for identify if this characters were typed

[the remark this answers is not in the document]
juliaroldi: We have a set of character that triggers numbering lists and another set of characters that triggers bullet list, this file is for identify if this characters were typed

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -354,13 +354,18 @@ export default class VList {
...
      */
     setListStyleType(
-        targetStyle:
-            | NumberingListType
-            | BulletListType
-            | CompatibleBulletListType
-            | CompatibleNumberingListType
+        targetType: ListType | CompatibleListType,
+        orderedStyle: NumberingListType | CompatibleNumberingListType,
+        unorderedStyle: BulletListType | CompatibleBulletListType
```
JiuqingSong, line 359: I think we should make these two parameters optional

JiuqingSong, line 359: I think we should make these two parameters optional

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -354,13 +354,18 @@ export default class VList {
      * @param targetStyle Target list style
```
JiuqingSong, line 354: add comment for new parameter

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -548,3 +553,27 @@ function moveLiToList(li: HTMLElement) {
...
+    listType: ListType | CompatibleListType,
+    orderedStyle: NumberingListType,
+    unorderedStyle: BulletListType
+): ListStyleMetadata {
+    if (style) {
+        if (listType === ListType.Ordered && orderedStyle !== undefined) {
+            style.orderedStyleType = orderedStyle;
+        } else if (unorderedStyle !== undefined) {
+            style.unorderedStyleType = unorderedStyle;
+        }
```
JiuqingSong, line 568: Can we just update the style without checking list type? So that means even for a ordered list, it is still allowed to updated unordered style, as long as the value is not undefined

JiuqingSong, line 568: Can we just update the style without checking list type? So that means even for a ordered list, it is still allowed to updated unordered style, as long as the value is not undefined

packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -360,13 +395,28 @@ function createListElement(
         result = doc.createElement(listType == ListType.Ordered ? 'ol' : 'ul');
     }
 
+    // Always maintain the metadata saved in the list
+    if (originalRoot && nextLevel == 1 && listType != getListTypeFromNode(originalRoot)) {
+        const style = getMetadata<ListStyleMetadata>(originalRoot, ListStyleDefinitionMetadata);
+        if (style) {
+            setMetadata(result, style, ListStyleDefinitionMetadata);
```
JiuqingSong, line 402: I'm confused here.
So you call getMetadata(), then without any change to the style object you call setMetadata() to save it back. Why we need to do this? It seems nothing is changed at all.

JiuqingSong, line 402: I'm confused here.
So you call getMetadata(), then without any change to the style object you call setMetadata() to save it back. Why we need to do this? It seems nothing is changed at all.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getListStyle.ts
JiuqingSong: This comment is still active

JiuqingSong: This comment is still active

packages/roosterjs-editor-types/lib/interface/ListStyleMetadata.ts
```diff
@@ -0,0 +1,10 @@
+import { BulletListType } from '../enum/BulletListType';
+import { NumberingListType } from '../enum/NumberingListType';
+
+/**
+ * Represents the metadata of the style of a list element
+ */
+export default interface ListStyleMetadata {
```
JiuqingSong, line 7: This type is only used by VLIst and VListItem, and we don't expect other developer can access it. So you can declare it in VLIstItem, and make an internal export, no need to export from here.

JiuqingSong, line 7: This type is only used by VLIst and VListItem, and we don't expect other developer can access it. So you can declare it in VLIstItem, and make an internal export, no need to export from here.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getListType.ts
```diff
@@ -1,14 +1,30 @@
 import { ListType } from 'roosterjs-editor-types';
 
-function isABulletList(textBeforeCursor: string) {
-    const hasTriggers = ['*', '-', '>'].indexOf(textBeforeCursor[0]) > -1;
-    const REGEX: RegExp = /^(.*?)=>|^(.*?)->|^(.*?)-->|^(.*?)=>|^(.*?)--/;
-    return hasTriggers || REGEX.test(textBeforeCursor);
+function isABulletList(trigger: string) {
+    const triggerLength = trigger.length;
+    if (triggerLength === 1) {
+        const REGEX: RegExp = /^((.*?)-|^(.*?)*|^(.*?)>)$/;
```
JiuqingSong, line 6: If you already know the length is 1, just check if it is -, *, > should be good enough, right?

JiuqingSong, line 6: If you already know the length is 1, just check if it is -, *, > should be good enough, right?

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getListType.ts
```diff
@@ -1,14 +1,30 @@
...
+        return REGEX.test(trigger);
+    } else if (triggerLength === 2) {
+        const REGEX: RegExp = /^((.*?)=>|^(.*?)->|^(.*?)=>|^(.*?)--)$/;
```
JiuqingSong, line 9: Same here, if we only want to support =>, ->, =>, --, just directly check it, no need to use regex I think

JiuqingSong, line 9: Same here, if we only want to support =>, ->, =>, --, just directly check it, no need to use regex I think

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getListType.ts
```diff
@@ -1,14 +1,30 @@
...
+        return REGEX.test(trigger);
+    } else if (triggerLength === 3) {
+        const REGEX: RegExp = /^(.*?)-->$/;
```
JiuqingSong, line 12: same

JiuqingSong, line 12: same

JiuqingSong: Generally, you can just simply do
const BulletTriggers = ['*', '-', '>', '=>', ... ];
...

if (BulletTriggers.indexOf(trigger) >= 0) { ...

We should try to avoid using Regex when user is typing since regex matching is much slower than just string comparing.

JiuqingSong: Generally, you can just simply do
const BulletTriggers = ['*', '-', '>', '=>', ... ];
...

if (BulletTriggers.indexOf(trigger) >= 0) { ...

We should try to avoid using Regex when user is typing since regex matching is much slower than just string comparing.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getListType.ts
```diff
@@ -1,14 +1,30 @@
...
+    return false;
 }
 
-function isANumberingList(textBeforeCursor: string) {
-    const REGEX: RegExp = /^([1-9,a-z, i,A-Z,I]{1,2}\.|[1-9,a-z, i,A-Z,I]{1,2}\)|[1-9,a-z, i,A-Z,I]{1,2}\-|\([1-9,a-z, i,A-Z,I]{1,2}\))$/;
-    return REGEX.test(textBeforeCursor.replace(/\s/g, ''));
+function isANumberingList(trigger: string) {
+    const triggerLength = trigger.length;
+    if (triggerLength === 2) {
+        const REGEX: RegExp = /^([1-9,a-z,A-Z]{1,2}\.|[1-9,a-z,A-Z]{1,2}\)|[1-9,a-z,A-Z]{1,2}\-)$/;
```
JiuqingSong, line 21: why 1-9?  We don't support "10" ?

JiuqingSong, line 21: why 1-9?  We don't support "10" ?

JiuqingSong: And if you want to use regex, make it a global const, but no create it everytime

JiuqingSong: And if you want to use regex, make it a global const, but no create it everytime

JiuqingSong: And the first part seems all same, so you can do
`/^[0-9]{1,2}[\.\)\-]$/`

BTW, why we allow a-z together with 0-9? Is "1a" a valid match? Maybe what you want is
`([0-9]{1,2}|[a-z]{1,2}|[a-Z]{1,2})`

JiuqingSong: And as I said, regex should be avoid as much as possible, you can first check if the last char is ".", ")", "-", then use regex in order to reduce the chance to exec regex.

JiuqingSong: And as I said, regex should be avoid as much as possible, you can first check if the last char is ".", ")", "-", then use regex in order to reduce the chance to exec regex.

packages/roosterjs-editor-dom/lib/list/VListItem.ts
juliaroldi: The list can be reused a new one is created, so the metadata is lost, to prevent the lost of the metadata, get the metadata from the original root and pass it to the new one.

JiuqingSong: I see, got it.

PR #1010, opened 2022-05-31
Bump version to 8.22.2

ianeli1:


JiuqingSong:
Let's use 8.23.0 since a lot of new types/parameters were added during this period. Please refer to https://semver.org/ for the rule of semver.

JiuqingSong:
#1145

PR #1014, opened 2022-06-02
Refactor auto hyphen experimental flight

juliaroldi:
Refactor the AutoHyphen into a new plugin for Auto Format

autoHyphen

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/AutoFormat/AutoFormat.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/AutoFormat/AutoFormat.ts
@@ -1,0 +1,92 @@
+import {
+    EditorPlugin,
+    ExperimentalFeatures,
+    IEditor,
+    PluginEvent,
+    PluginEventType,
+    PositionType,
+} from 'roosterjs-editor-types';
+
+const specialCharacters = /[`!@#$%^&*()_+\-=\[\]{};':"\\|,.<>\/?~]/;
+
+/**
+ * Automatically transform -- into hyphen, if typed between two words.
+ */
+export default class AutoFormat implements EditorPlugin {
+    private editor: IEditor;
+    private isHyphenTyped = false;
+    private shouldFormat = false;
+
+    /**
+     * Get a friendly name of this plugin
+     */
+    getName() {
+        return 'AutoFormat';
+    }
+
+    /**
+     * Initialize this plugin
+     * @param editor The editor instance
+     */
+    initialize(editor: IEditor) {
+        this.editor = editor;
+    }
+
+    /**
+     * Dispose this plugin
+     */
+    dispose() {
+        this.editor = null;
+    }
+
+    /**
+     * Handle events triggered from editor
+     * @param event PluginEvent object
+     */
+    onPluginEvent(event: PluginEvent) {
+        if (
+            event.eventType === PluginEventType.KeyDown &&
+            this.editor.isFeatureEnabled(ExperimentalFeatures.AutoHyphen)
+        ) {
+            const keyTyped = event.rawEvent.key;
+
+            if (keyTyped === '-') {
+                this.isHyphenTyped = true;
+            }
+
+            if (this.isHyphenTyped && keyTyped === '-') {
+                this.shouldFormat = true;
+            } else {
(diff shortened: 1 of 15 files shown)


packages/roosterjs-editor-types/lib/interface/ContentEditFeatureSettings.ts
```diff
@@ -226,12 +226,6 @@ export interface TextFeatureSettings {
      *      If Whole Paragraph selected, outdent paragraph
      */
     outdentWhenTabText: boolean;
-
-    /**
-     * Requires @see ExperimentalFeatures.AutoHyphen to be enabled
-     * Automatically transform -- into hyphen, if typed between two words.
-     */
-    autoHyphen: boolean;
```
JiuqingSong, line 234: do not remove, mark as @deprecated instead, otherwise it will become a breaking change.

packages/roosterjs-editor-plugins/lib/plugins/AutoFormat/AutoFormat.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * Handle events triggered from editor
+     * @param event PluginEvent object
+     */
+    onPluginEvent(event: PluginEvent) {
+        if (
+            event.eventType === PluginEventType.KeyDown &&
+            this.editor.isFeatureEnabled(ExperimentalFeatures.AutoHyphen)
```
JiuqingSong, line 48: No need to check feature. If we don't want this feature, just don't add this plugin at all.

packages/roosterjs-editor-plugins/lib/plugins/AutoFormat/AutoFormat.ts
```diff
@@ -0,0 +1,86 @@
...
+                this.editor.addUndoSnapshot(
+                    () => {
+                        textRange.deleteContents();
+                        textRange.insertNode(nodeHyphen);
+                        this.editor.select(nodeHyphen, PositionType.End);
+                    },
+                    null /*changeSource*/,
+                    true /*canUndoByBackspace*/
+                );
+                this.lastKeyTyped = '';
```
JiuqingSong, line 80: why?

packages/roosterjs-editor-plugins/lib/plugins/AutoFormat/AutoFormat.ts
```diff
@@ -0,0 +1,86 @@
...
+            } else {
+                this.lastKeyTyped = keyTyped;
+            }
```
JiuqingSong, line 83: also need to clear lastKeyTyped when:
1. up/Down/left/right/pgup/pgdn/home/end key is pressed
2. mouse click
3. content changed event.

packages/roosterjs-editor-plugins/lib/plugins/AutoFormat/AutoFormat.ts
juliaroldi: After the change, the lastKey type must be cleaned.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/textFeatures.ts
```diff
@@ -111,7 +111,7 @@ const OutdentWhenTabText: BuildInEditFeature<PluginKeyboardEvent> = {
 };
 
 /**
- * Requires @see ExperimentalFeatures.AutoHyphen to be enabled
+ * @deprecated
```
JiuqingSong, line 114: it is ok to delete this object since we have never exported it.

juliaroldi: But remove it would be a breaking change?

JiuqingSong: To keep it not breaking, you just need to keep all published things not changed. For this one, it is not public. So we can remove it, and just create a dummy feature and assign to TextFeatures and make it does nothing. For example:

const dummyFeature = {
  keys: [],
  shouldHandle: () => false,
 ...
}

JiuqingSong (approved):

PR #1016, opened 2022-06-06
Check if the rotator is hidden, if yes, flip to the other side

juliaroldi:
If the rotator handle is at the top or in left side at the editor or is overlaid by the border or the rotator handle overlays the resize handle. 

**Before:**
rotatorBugsBefore

**After:**
rotatorBugs

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ There are also some extension packages to provide additional functionalities.
 2. [roosterjs-react](https://microsoft.github.io/roosterjs/docs/modules/roosterjs_react.html):
    Provide a React wrapper of roosterjs so it can be easily used with React.
 
-3. [roosterjs-editor-types-compatible](modules/roosterjs_editor_types_compatible.html):
+3. [roosterjs-editor-types-compatible](https://microsoft.github.io/roosterjs/docs/modules/roosterjs_editor_types_compatible.html):
    Provide types that are compatible with isolatedModules mode. When using isolatedModules mode,
    "const enum" will not work correctly, this package provides enums with prefix "Compatible" in
    their names and they have the same value with const enums in roosterjs-editor-types package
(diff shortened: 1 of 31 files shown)


JiuqingSong:
This is interesting. With this fix, I think you can remove the code to auto-lower-down the rotate handle to make it consistent.

juliaroldi:
> This is interesting. With this fix, I think you can remove the code to auto-lower-down the rotate handle to make it consistent.

I think it is better to let this feature, because if this user is dragging next to the top, the experience will be better. If the user drag the rotator very near the top, the rotator will be overlayed by the border.

JiuqingSong:
> > This is interesting. With this fix, I think you can remove the code to auto-lower-down the rotate handle to make it consistent.
> 
> I think it is better to let this feature, because if this user is dragging next to the top, the experience will be better. If the user drag the rotator very near the top, the rotator will be overlayed by the border without this feature. I think would be nice if we also add to detect the left side.

I mean when image is at the top of editor and rotate it, today we have code to auto lower the rotation handle:

image

compare to:

image

With this change, that feature is no needed any more since we will show resize handle at the bottom instead. You can remove the old code.

juliaroldi:
> > > This is interesting. With this fix, I think you can remove the code to auto-lower-down the rotate handle to make it consistent.
> > 
> > 
> > I think it is better to let this feature, because if this user is dragging next to the top, the experience will be better. If the user drag the rotator very near the top, the rotator will be overlayed by the border without this feature. I think would be nice if we also add to detect the left side.
> 
> I mean when image is at the top of editor and rotate it, today we have code to auto lower the rotation handle:
> 
> image
> 
> compare to:
> 
> image
> 
> With this change, that feature is no needed any more since we will show resize handle at the bottom instead. You can remove the old code.

I think if the user is dragging near the top, without clicking out of the rotator, the experience will be better, but I don't mind to remove it.

PR #1018, opened 2022-06-06
Fix ENOENT build error on WSL Ubuntu

ianeli1:
Using `yarn build:ci` is currently failing on WSL2 due to errors being thrown in the `normalize` and `checkDependency` steps. 

Setting default values to the return value of `readPackageJson` prevents the `Object.keys` statements from raising an exception.

Additionally, having `@types/dom-inputevent` as a dependency was causing the `normalize` step to look for a package called `dom-inputevent` inside `/packages/`

### DIFF
--- a/packages/roosterjs-editor-types/package.json
+++ b/packages/roosterjs-editor-types/package.json
@@ -1,7 +1,7 @@
 {
     "name": "roosterjs-editor-types",
     "description": "Type definition for roosterjs",
-    "dependencies": {
+    "devDependencies": {
         "@types/dom-inputevent": "^1.0.3"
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-types/package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "roosterjs-editor-types",
     "description": "Type definition for roosterjs",
-    "dependencies": {
+    "devDependencies": {
```
JiuqingSong, line 4: This is dependency, not devDependency. So we should not change this.

PR #1028, opened 2022-06-10
[Part3] Refactor Auto Format - AutoBulletList and AutoNumberingList

juliaroldi:
Added the AutoBulletList and AutoNumberingList to detect the numbering and bullet list. Deprecate AutoBullet content edit feature.


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -181,39 +181,102 @@ const AutoBullet: BuildInEditFeature<PluginKeyboardEvent> = {
...
+            null /*changeSource*/,
+            true /*canUndoByBackspace*/
+        );
+    },
+};
+
+/**
+ * AutoBulletList edit feature, provides the ability to automatically convert current line into a bullet list.
+ */
+const AutoBulletList: BuildInEditFeature<PluginKeyboardEvent> = {
```
JiuqingSong, line 214: Besides we can disable a feature separately, is there any other reason we want to split into two features?

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -181,39 +181,102 @@ const AutoBullet: BuildInEditFeature<PluginKeyboardEvent> = {
             () => {
                 let regions: RegionBase[];
                 let searcher = editor.getContentSearcherOfCursor();
-                let textBeforeCursor = searcher.getSubStringBefore(5);
+                let textBeforeCursor = searcher.getSubStringBefore(4);
```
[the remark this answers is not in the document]
juliaroldi: Since the new code is in the is in the two new objects, I revert this object to it was before my changes.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -156,21 +155,22 @@ function isAListPattern(textBeforeCursor: string) {
...
+        ) {
+            let searcher = editor.getContentSearcherOfCursor(event);
+            let textBeforeCursor = searcher.getSubStringBefore(4);
+
             // Auto list is triggered if:
             // 1. Text before cursor exactly matches '*', '-' or '1.'
             // 2. There's no non-text inline entities before cursor
-            return !searcher.getNearestNonTextInlineElement() && listTrigger(textBeforeCursor);
+            return isAListPattern(textBeforeCursor) && !searcher.getNearestNonTextInlineElement();
         }
```
JiuqingSong, line 174: can we not change the code of this one since we are about to deprecate it, as long as it still work. So that people who is still using this code won't be impacted.

PR #1033, opened 2022-06-13
fix wrap content when element is empty

FrancisMengx:


JiuqingSong (approved):

biubiubiu-boom:
What's the reason to add a `<br/>` into an empty node? @JiuqingSong  @FrancisMengx

JiuqingSong:
> What's the reason to add a `<br/>` into an empty node? @JiuqingSong @FrancisMengx image

Browser can't put cursor into empty `<DIV></DIV>`, which makes some other code not working correctly.

biubiubiu-boom:
Thanks for the explanation @JiuqingSong . I have another question: if I added some empty elements (not only the div) as a flag into editor, would those elements contain a `<br/>` as a child element and show a blank line in editor too?

JiuqingSong:
If it is intentional invisible, then no need to add BR

PR #1038, opened 2022-06-17
Inserting a table after a non content editable entity, makes the table not editable

BryanValverdeU:
To repro, restore the following snapshot:

`<div data-ogsc="rgb(0,0,0)" data-ogsb="rgb(255,255,255)" style="font-family: &quot;Times New Roman&quot;; font-size: 12pt; color: rgb(255, 255, 255); background-color: rgb(51, 51, 51);"><span class="_Entity _EType_ _EReadonly_1" contenteditable="false" style="display: inline-block;"><span data-hydrated-html=""><span><span><a data-ogsc="" data-ogsb="rgb(244,244,244)" style="pointer-events: auto; padding: 0px 1px; border-radius: 2px; user-select: all; background-color: rgb(243, 242, 241);" class="OWAAutoLink OxGIL wOeYL" id="OLK_Beautified_OWA63443add-d99e-6fa0-2620-c7a4898bd7f2" href="https://microsoft-my.sharepoint.com/personal/bvalverde_microsoft_com/_layouts/15/doc.aspx?sourcedoc={adf896dd-dad6-4422-a561-187bdf5e6c39}&amp;action=edit">Document44.docx<span class="owaAttachmentRemoveOnSend"></span></a></span></span></span></span><br></div><!--{"type":0,"isDarkMode":true,"start":[0,1],"end":[0,1]}-->`

Place cursor next to the not editable entity and then insert a table, notice how the table is not editable since it is inside of the read only entity.

This PR fixes this issue inserting the table after the not content editable element

### DIFF
--- a/packages/roosterjs-editor-api/lib/table/insertTable.ts
+++ b/packages/roosterjs-editor-api/lib/table/insertTable.ts
@@ -1,5 +1,13 @@
-import { ChangeSource, IEditor, PositionType, TableFormat } from 'roosterjs-editor-types';
-import { Position, VTable } from 'roosterjs-editor-dom';
+import { createRange, Position, VTable, wrap } from 'roosterjs-editor-dom';
+import {
+    ChangeSource,
+    ContentPosition,
+    IEditor,
+    PositionType,
+    TableFormat,
+} from 'roosterjs-editor-types';
+
+const NOT_EDITABLE_SELECTOR = '[contenteditable=false]';
 
 /**
  * Insert table into editor at current selection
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-api/lib/table/insertTable.ts
```diff
@@ -39,12 +47,46 @@ export default function insertTable(
         vtable.applyFormat(format);
         vtable.writeBack();
         editor.insertNode(fragment);
```
JiuqingSong, line 49: I think the problem is why we can insert node into a readonly node.

Try fix insertNode() to make sure not insert under the readonly node.

BryanValverdeU: If we are inside of a readonly node what should we do? I think inserting after the readonly should be okay right?

PR #1040, opened 2022-06-17
Fix Paste will not replace selected content on Android #925

BryanValverdeU:
Seems like Android is not removing the selected range after pasting.
This PR fixes the issue

Issue #925, opened 2022-04-26
Paste will not replace selected content on Android


yifax:
**Describe the bug**
Select some content inside editor and paste, selected content will not be replaced. Instead, content got pasted will be inserted at the end of selection range.

**To Reproduce**
1. Type something in RoosterJs Demo on Android device
2. Copy some text to clipboard
3. Select some content, long press, and click on `paste`
4. Check behavior

**Expected behavior**
Selected content is replaced by pasting content from clipboard

**Screenshots**
https://user-images.githubusercontent.com/14918900/165219749-379e8022-cadd-4657-b772-a808b718a2e0.mp4

**Device Information**
 - OS: Android
 - Chrome / Webview
 - Android 12 / Webview 100.0.4896.127 

**Additional context**
Not repro on PC and iOS. Not sure if related to this [issue](https://github.com/microsoft/roosterjs/issues/797)

PR #1043, opened 2022-06-20
Fix Paste will not replace selected content on Android #925 #1040

BryanValverdeU:
Seems like Android is not removing the selected range after pasting.
This PR fixes the issue

### DIFF
--- a/packages/roosterjs-editor-core/lib/corePlugins/CopyPastePlugin.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/CopyPastePlugin.ts
@@ -130,7 +130,8 @@ export default class CopyPastePlugin implements PluginWithState<CopyPastePluginS
                 removeTempDiv: div => {
                     this.cleanUpAndRestoreSelection(div, range, false /* isCopy */);
                 },
-            }
+            },
+            this.editor?.getSelectionRange()?.cloneRange()
         );
     };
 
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-core/lib/corePlugins/CopyPastePlugin.ts
```diff
@@ -130,7 +130,8 @@ export default class CopyPastePlugin implements PluginWithState<CopyPastePluginS
                 removeTempDiv: div => {
                     this.cleanUpAndRestoreSelection(div, range, false /* isCopy */);
                 },
-            }
+            },
+            this.editor?.getSelectionRange()?.cloneRange()
```
BryanValverdeU, line 134: Tested again and cloneRange was not needed.

packages/roosterjs-editor-dom/lib/clipboard/extractClipboardEvent.ts
```diff
@@ -24,16 +26,28 @@ interface WindowForIE extends Window {
...
+        extractClipboardItems(toArray(dataTransfer.items), options)
+            .then(callback)
+            .then(() => removeContents(rangeBeforePaste));
     } else {
         extractClipboardItemsForIE(dataTransfer, callback, options);
     }
 }
+
+function removeContents(range?: Range) {
+    return new Promise<void>(resolve => {
```
JiuqingSong, line 47: why need a promise here?

BryanValverdeU: Tested again and it is not needed. and we can actually it in the same promise as the one we call the callback.

JiuqingSong (approved):

Issue #1004, opened 2022-05-30
Table Selector is displaying incorrectly above the editors Div


BryanValverdeU:
**Describe the bug**
Table Selector is displaying incorrectly above the editors Div
 
![image](https://user-images.githubusercontent.com/8291124/171054357-c7d2311a-c60c-42c4-8a9a-1316c55768cf.png)

Repro steps:
Create a table
Add content in order to scroll the content in the editor div
hover the table.
See table selector above outside of the editor

PR #1050, opened 2022-06-23
Re enable Table Selector is displaying incorrectly above the editors Div

BryanValverdeU:
Renable https://github.com/microsoft/roosterjs/pull/1005 and fix null checks 

#1004

Before
image

After
image


packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
```diff
@@ -69,7 +76,8 @@ export default class TableEditor {
         private onShowHelperElement?: (
             elementData: CreateElementData,
             helperType: 'CellResizer' | 'TableInserter' | 'TableResizer' | 'TableSelector'
-        ) => void
+        ) => void,
+        private event?: MouseEvent
```
JiuqingSong, line 80: does this event object need to be a member?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
```diff
@@ -271,26 +280,51 @@ export default class TableEditor {
...
+            }
         }
     };
+
+    private getShouldShowTableSelectorHandler(
+        scrollContainer: HTMLElement,
+        ev?: MouseEvent
+    ): (rect: Rect) => boolean {
+        if (
+            ev?.currentTarget &&
```
JiuqingSong, line 312: if only currentTarget is needed, just pass it only, but not the whole event ojbect

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
```diff
@@ -69,7 +76,8 @@ export default class TableEditor {
         private onShowHelperElement?: (
             elementData: CreateElementData,
             helperType: 'CellResizer' | 'TableInserter' | 'TableResizer' | 'TableSelector'
-        ) => void
+        ) => void,
+        private eventTarget?: EventTarget
```
JiuqingSong, line 80: I only see eventTarget is used at line 98, so no need to add "private" to make it as a member

Issue #917, opened 2022-04-21
Creating link does not replace the whole selected text


mathildebm:
**Describe the bug**
Creating a link does not replace the whole selection range, when the selection includes several \<div\> or \<span\>

**To Reproduce**
It is reproducible on the [demo site](https://microsoft.github.io/roosterjs/index.html) 
1. Write some text on several lines, with some line breaks
2. Select lines of text with at least 2 line breaks
3. Click on button to create link
4. Write in the 'Web address (URL)' input 
5. Click on OK
6. See error: the whole selected text or part of the selected text is remaining. 
In the HTML we can see that 2 links with the same url have been created: one for the remaining text, one for the correct new text

**Expected behavior**
The link should replace the whole selected text, and create only one link

**Screenshots**
![Capture d’écran de 2022-04-15 12-10-04](https://user-images.githubusercontent.com/104149443/164473806-28060631-a0d5-4c0b-ae1d-70e6bc827d87.png)
![Capture d’écran de 2022-04-15 12-11-08](https://user-images.githubusercontent.com/104149443/164473816-a924f814-cbea-4086-85eb-ee1dc4a99876.png)
![Capture d’écran de 2022-04-15 12-11-27](https://user-images.githubusercontent.com/104149443/164473829-79f2cd07-e29e-420f-9c52-2eddc1043a37.png)

**Device Information**
 - Browser: Chrome, Firefox

**Additional context**
This is reproductible on the Demo site

JiuqingSong:
@BryanValverdeU  please try to repro, thanks.

PR #1057, opened 2022-06-27
Fixed issue with creating link not replacing all selected text - Creating link does not replace the whole selected text #917

Andres-CT98:
Fix to https://github.com/microsoft/roosterjs/issues/917

**Describe the bug**
Creating a link does not replace the whole selection range, when the selection includes several \<div\> or \<span\>

**To Reproduce**
It is reproducible on the [demo site](https://microsoft.github.io/roosterjs/index.html) 
1. Write some text on several lines, with some line breaks
2. Select lines of text with at least 2 line breaks
3. Click on button to create link
4. Write in the 'Web address (URL)' input 
5. Click on OK
6. See error: the whole selected text or part of the selected text is remaining. 
In the HTML we can see that 2 links with the same url have been created: one for the remaining text, one for the correct new text

**Expected behavior**
The link should replace the whole selected text, and create only one link

**Screenshots**
Capture d’écran de 2022-04-15 12-10-04
Capture d’écran de 2022-04-15 12-11-08
Capture d’écran de 2022-04-15 12-11-27

**Device Information**
 - Browser: Chrome, Firefox

**Additional context**
This is reproductible on the Demo site

**Fix**

Added all unwanted nodes to a list for deletion from the editor. Previous verison only deleted one node.


demo/scripts/controls/sidePane/editorOptions/EditorOptionsPlugin.ts
```diff
@@ -1,64 +1,64 @@
...
-    }
-
-    getComponentProps(base: SidePaneElementProps) {
-        return {
-            ...initialState,
-            ...base,
-        };
-    }
-}
+import BuildInPluginState, { BuildInPluginProps, UrlPlaceholder } from '../../BuildInPluginState';
```
JiuqingSong, line 1: Please check why the whole file is shown as changed, this makes it hard to see what is the real change

JiuqingSong:
Please add some description to explain the bug and how did you fix it

packages/roosterjs-editor-api/lib/format/createLink.ts
```diff
@@ -67,6 +67,7 @@ export default function createLink(
 
         editor.addUndoSnapshot(() => {
             let range = editor.getSelectionRange();
+
```
JiuqingSong, line 70: no need to add empty line here

packages/roosterjs-editor-api/lib/format/createLink.ts
```diff
@@ -86,12 +87,21 @@ export default function createLink(
                 // the selection is not collapsed, use browser execCommand
                 editor.getDocument().execCommand(DocumentCommand.CreateLink, false, normalizedUrl);
                 const traverser = editor.getSelectionTraverser();
+
                 let currentInline = traverser.getNextInlineElement();
+
+                // list for removing unwanted lines
+                let deletionInlineList = [];
```
JiuqingSong, line 94: need type declaration here. What is the type of this array?

packages/roosterjs-editor-api/lib/format/createLink.ts
```diff
@@ -86,12 +87,21 @@ export default function createLink(
...
                 while (currentInline) {
-                    const a = currentInline.getContainerNode();
-                    a.parentElement.removeChild(a);
+                    deletionInlineList.push(currentInline.getContainerNode());
                     currentInline = traverser.getNextInlineElement();
                 }
+
+                while (deletionInlineList.length) {
+                    editor.deleteNode(deletionInlineList.pop());
+                }
```
JiuqingSong, line 103: you can do deletionInlinelist.forEach(node => editor.deleteNode(node));

packages/roosterjs-editor-api/lib/format/createLink.ts
Andres-CT98: Can I ask what is the criteria for when an emty line is allowed and when it is not?

demo/scripts/controls/sidePane/editorOptions/EditorOptionsPlugin.ts
Andres-CT98: I do not know how to fix this, I only changed the string 'Ctrl+Click to follow the link:' to 'Ctrl+Click to follow the link: '.

BryanValverdeU: I think this usually happens if you have autocrlf disabled. Try using this config: 
git config --global core.autocrlf true

JiuqingSong (approved):

PR #1058, opened 2022-06-27
Improve table shading menu

JiuqingSong:
1. Add new context menu item "Shading" for table
2. Refactor color picker code for better reusing
3. Remove original shading button since it is not used
4. Remove original table edit/alignment buttons from demo site since we have table context menu now
5. Remove demo site code SamplePicker since it is not used any more.

image


JiuqingSong:
#1060

juliaroldi (approved):

PR #1068, opened 2022-06-29
Remove comment related styles

BryanValverdeU:
There are some comment related styles that are causing some issues in the content when pasting them from Office Apps. 
To fix this we can remove these styles on paste


packages/roosterjs-editor-plugins/lib/plugins/Paste/wordConverter/commentsRemoval.ts
```diff
@@ -83,4 +84,11 @@ export default function commentsRemoval(elementCallbacks: ElementCallbackMap) {
         }
         return true;
     });
+
+    /**
+     * Remove comment related styles
+     */
+    chainSanitizerCallback(styleCallbacks, MSO_COMMENT_REFERENCE, () => false);
+    chainSanitizerCallback(styleCallbacks, MSO_COMMENT_DATE, () => false);
+    chainSanitizerCallback(styleCallbacks, MSO_COMMENT_PARENT, () => false);
```
JiuqingSong, line 93: What are the issues if we don't remove them?

BryanValverdeU: If we do not remove this styles, when we send the email and a Win32 user receives the email, they see some style on the word that contain the comment.

For example, this is how a win32 user see a message:
image

vs how we sent it:
image

JiuqingSong: got it, thanks. Better add some comment

JiuqingSong (approved):

PR #1069, opened 2022-06-29
Fix #1067 Need to trigger ContentChangedEvent after resize image

### LINKED ISSUE #1067: ContentChangedEvent should be fired after resize image
**Describe the bug**

**To Reproduce**
Insert image, resize/crop/rotate it, there should be a ContentChangedEvent fired

**Expected behavior**

**Screenshots**

**Device Information**

**Additional context**

JiuqingSong:
We had code to check if image is not changed after resize, no undo snapshot is added and no event is triggered. However the code to check change is not right. So I removed it.

Now every time we finish editing image (even image is not changed), we will trigger a ContentChangedEvent since user did add image editing wrapper and removed it. We call the API to add undo snapshot but snapshot will only be added when content is really changed.


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -291,18 +291,16 @@ export default class ImageEdit implements EditorPlugin {
             this.clearDndHelpers();
 
             // Apply the changes, and add undo snapshot if necessary
-            if (
-                applyChange(this.editor, this.image, this.editInfo, this.lastSrc, this.wasResized)
-            ) {
-                this.editor.addUndoSnapshot(() => this.image, ChangeSource.ImageResize);
-            }
+            applyChange(this.editor, this.image, this.editInfo, this.lastSrc, this.wasResized);
```
juliaroldi, line 294: why do we need to trigger ContentChange when we set and remove the wrapper? When we use insertEntity, we do trigger ContentChangeEvent. Do will need to this for another wrappers?

JiuqingSong: Actually this is not for the wrapper. We trigger the event here because it is possible image got resized before this point. So content is really changed.

juliaroldi (approved):

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
JiuqingSong: If you ask if that is "faster", the answer is yes.
But doing this is not for performance, but just want to make it simpler. We can remember the image size before resize and check it. But that will introduce additional complexity. And even user click the image but didn't really resize it, we can still trigger this event and nothing harmful.

PR #1081, opened 2022-07-13
Modified shouldReplace to add editor as optional parameter 

Andres-CT98:
Modified the shouldReplace function to add a new optional parameter to pass the editor that contains the string to replace. Also modified the plugin event to always pass the editor, effectively making it a default parameter but still supporting replacement rules that do not use it.

This allows for Custom replacement rules that can use the full editor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "roosterjs",
-    "version": "8.27.0",
+    "version": "8.27.1",
     "description": "Framework-independent javascript editor",
     "repository": {
         "type": "git",
@@ -18,13 +18,13 @@
         "builddemo": "node tools/build.js builddemo",
         "builddoc": "node tools/build.js builddoc",
         "build": "node tools/build.js clean checkdep normalize tslint buildcommonjs dts packprod builddemo",
-        "build:ci": "node tools/build.js --noProgressBar clean checkdep normalize tslint buildcommonjs buildamd dts pack packprod builddemo builddoc",
+        "build:ci": "node tools/build.js --noProgressBar clean checkdep normalize tslint buildcommonjs buildamd buildmjs dts pack packprod builddemo builddoc",
         "start": "node tools/start.js",
         "test": "node tools/build.js normalize & karma start",
         "test:chrome": "node tools/build.js normalize & karma start --chrome",
         "test:debug": "node tools/build.js normalize & karma start --no-single-run",
         "test:coverage": "node tools/build.js normalize & karma start --coverage",
-        "publish": "node tools/build.js clean normalize tslint buildcommonjs buildamd dts pack packprod builddemo builddoc publish"
+        "publish": "node tools/build.js clean normalize tslint buildcommonjs buildamd buildmjs dts pack packprod builddemo builddoc publish"
     },
     "devDependencies": {
         "@microsoft/loader-load-themed-styles": "1.8.11",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "roosterjs",
-    "version": "8.26.2",
+    "version": "8.27.0",
     "description": "Framework-independent javascript editor",
     "repository": {
         "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "roosterjs",
-    "version": "8.26.1",
+    "version": "8.26.2",
     "description": "Framework-independent javascript editor",
     "repository": {
         "type": "git",
--- a/packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
@@ -11,7 +11,11 @@ const makeReplacement = (
     sourceString: string,
     replacementHTML: string,
     matchSourceCaseSensitive: boolean,
-    shouldReplace?: (replacement: CustomReplacement, content: string, contentTag: string) => boolean
+    shouldReplace?: (
+        replacement: CustomReplacement,
+        content: string,
+        contentTag?: string
+    ) => boolean
 ): CustomReplacement => ({
     sourceString,
     replacementHTML,
@@ -99,7 +103,7 @@ export default class CustomReplacePlugin implements EditorPlugin {
 
         const searcher = this.editor.getContentSearcherOfCursor(event);
         const stringToSearch = searcher.getSubStringBefore(this.longestReplacementLength);
-        const sourceTag = range.commonAncestorContainer.parentElement.tagName;
+        const sourceTag = this.editor.getElementAtCursor().tagName;
 
         const replacement = this.getMatchingReplacement(stringToSearch);
         if (replacement == null) {
@@ -147,6 +151,7 @@ export default class CustomReplacePlugin implements EditorPlugin {
             const [sourceMatch, replacementMatch] = replacement.matchSourceCaseSensitive
                 ? [originalStringToSearch, replacement.sourceString]
                 : [lowerCaseStringToSearch, replacement.sourceString.toLocaleLowerCase()];
+
             if (
                 sourceMatch.substring(sourceMatch.length - replacementMatch.length) ==
                 replacementMatch
--- a/packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
@@ -11,7 +11,7 @@ const makeReplacement = (
     sourceString: string,
     replacementHTML: string,
     matchSourceCaseSensitive: boolean,
-    shouldReplace?: (replacement: CustomReplacement, content: string) => boolean
+    shouldReplace?: (replacement: CustomReplacement, content: string, contentTag: string) => boolean
 ): CustomReplacement => ({
     sourceString,
     replacementHTML,
@@ -99,6 +99,7 @@ export default class CustomReplacePlugin implements EditorPlugin {
 
         const searcher = this.editor.getContentSearcherOfCursor(event);
         const stringToSearch = searcher.getSubStringBefore(this.longestReplacementLength);
+        const sourceTag = range.commonAncestorContainer.parentElement.tagName;
 
         const replacement = this.getMatchingReplacement(stringToSearch);
         if (replacement == null) {
@@ -107,7 +108,7 @@ export default class CustomReplacePlugin implements EditorPlugin {
 
         if (
             replacement.shouldReplace &&
-            !replacement.shouldReplace(replacement, searcher.getWordBefore())
+            !replacement.shouldReplace(replacement, searcher.getWordBefore(), sourceTag)
         ) {
             return;
         }
@@ -146,7 +147,6 @@ export default class CustomReplacePlugin implements EditorPlugin {
             const [sourceMatch, replacementMatch] = replacement.matchSourceCaseSensitive
                 ? [originalStringToSearch, replacement.sourceString]
                 : [lowerCaseStringToSearch, replacement.sourceString.toLocaleLowerCase()];
-
             if (
                 sourceMatch.substring(sourceMatch.length - replacementMatch.length) ==
                 replacementMatch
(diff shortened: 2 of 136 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "roosterjs",
-    "version": "8.26.1",
+    "version": "8.26.2",
```
juliaroldi, line 3: Change the package.json triggers a new version of roosterJS and we only do this in every 2 weeks. 
We usually open a PR just to bump the version.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -11,7 +11,7 @@ const makeReplacement = (
     sourceString: string,
     replacementHTML: string,
     matchSourceCaseSensitive: boolean,
-    shouldReplace?: (replacement: CustomReplacement, content: string) => boolean
+    shouldReplace?: (replacement: CustomReplacement, content: string, contentTag: string) => boolean
```
juliaroldi, line 14: This is a public parameter, add it as a required parameter will cause a breaking change. Please, check if add it is really necessary, if is, add as optional.

packages/roosterjs-editor-types/lib/interface/CustomReplacement.ts
```diff
@@ -21,7 +21,12 @@ export default interface CustomReplacement {
      * A callback to check if the string should be replaced
      * @param content the content where the string is
      * @param sourceString string to be replaced
+     * @param contentTag format of the string to be replaced
```
juliaroldi, line 24: I think you can change this comment to "contentTag  tagName of the html element that wraps contains the string"

package.json
Andres-CT98: I am following the debugging guide and it said to do that, what should I do then?

JiuqingSong: As Julia said, we normally do this every two weeks, unless this is an urgent fix and want to be integrated immediately. You don't need to touch this file in this change.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -99,6 +99,7 @@ export default class CustomReplacePlugin implements EditorPlugin {
 
         const searcher = this.editor.getContentSearcherOfCursor(event);
         const stringToSearch = searcher.getSubStringBefore(this.longestReplacementLength);
+        const sourceTag = range.commonAncestorContainer.parentElement.tagName;
```
JiuqingSong, line 102: commonAncestorContainer can be different kind of elements, and getting its parent element makes this even more unpredictable. Why we need to do this?

JiuqingSong (commented):
Please add more description and link to an issue to explain why need this change, what is the issue you'd like to solve.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -146,7 +147,6 @@ export default class CustomReplacePlugin implements EditorPlugin {
             const [sourceMatch, replacementMatch] = replacement.matchSourceCaseSensitive
                 ? [originalStringToSearch, replacement.sourceString]
                 : [lowerCaseStringToSearch, replacement.sourceString.toLocaleLowerCase()];
-
```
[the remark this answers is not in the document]
Andres-CT98: It keeps it with the same style as the other for in the file.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
Andres-CT98: I need to get the format of the current string, specifically if it is superscripted.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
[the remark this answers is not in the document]
Andres-CT98: Is there an official style guide I can follow to prevent issues like this?

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
Andres-CT98: Can a case like that happen in roosterjs?

JiuqingSong: why not?

JiuqingSong: user can paste anything, so any kind of those html can happen

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
JiuqingSong: contentTag?: string

Andres-CT98: Does not that leave the parameter as undefined instead of optional?

JiuqingSong: https://www.typescripttutorial.net/typescript-tutorial/typescript-optional-parameters/

That means the parameter is optional. Caller can decide whether pass it or not. And when it is not passed, its value will be undefined. So in your function you need to check if it is undefined before use.

Andres-CT98: I did not know typescrpt had that notation. I was confusing it with the ternary operator because they use the same notation. Good to know now.

juliaroldi (approved):

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -99,6 +103,7 @@ export default class CustomReplacePlugin implements EditorPlugin {
 
         const searcher = this.editor.getContentSearcherOfCursor(event);
         const stringToSearch = searcher.getSubStringBefore(this.longestReplacementLength);
+        const sourceTag = this.editor.getElementAtCursor().tagName;
```
JiuqingSong, line 106: Same question, how to handle `<sub><span>****</span></sub>`

Andres-CT98: It does not, it only gets the first html tag the cursor is in. This function is intended for replacing a string to another string while someone is typing. If someone inserts their own html CustomReplace does not make replacements.

juliaroldi: I think what Andres is trying to do is get the tag of cursor position, then compare if the tag html that is being in inserted, then the should replace will do the logic to replace or not

juliaroldi: It is just one more information to use in shouldReplace callback. Is that it @Andres-CT98?

JiuqingSong: My point is, we should not rely on the direct element tag. If you want to check whether user is inputting under a special tag, you need to find it all the way above until the root of editor. For example:
```html
const sub = editor.getElementAtCursor('sub');
if (sub) {
 ... 
}
```
In that way, no matter what is inside SUB, it can always find it.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
```diff
@@ -107,7 +112,7 @@ export default class CustomReplacePlugin implements EditorPlugin {
 
         if (
             replacement.shouldReplace &&
-            !replacement.shouldReplace(replacement, searcher.getWordBefore())
+            !replacement.shouldReplace(replacement, searcher.getWordBefore(), sourceTag)
```
JiuqingSong, line 115: And passing in a sourceTag is not general enough. What if next time we have an requirement to check something else, such as if a given element has a special attribute or style but not a tag name? We should not keep adding new parameters here. Each new parameter should be general enough.

Instead passing in the tag name, you can pass in the whole editor (as IEditor) reference then you can do any check inside the callback.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
JiuqingSong: And always remember user can input content inside the content that is pasted from other places. So we should never assume something can't exist in the content

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
Andres-CT98: I thought passing the whole editor would be too extreme because I do not know how big it can get, It certainly would make for the most complete replacement rules.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
Andres-CT98: I understand users can insert anything, but replacement rules only apply while typing.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
JiuqingSong: It will just pass a reference (in 32bit system, just 4 bytes). So no matter how big editor is, the cost is same.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
JiuqingSong: User can paste something, then put cursor inside the pasted content, then type. In that case, we still need to understand what kind of HTML structure it is.

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
Andres-CT98: Ok, I will pass the editor

packages/roosterjs-editor-plugins/lib/plugins/CustomReplace/CustomReplace.ts
Andres-CT98: That is not the task of this function, that is for whatever replacement rules users write to define the behaviour. Passing the editor will give them more tools though, so I think it is the optimal implementation.

JiuqingSong: No matter what function it is, we should never rely on the tag name of current direct container element since HTML structure can be variant. User can simply press Ctrl+B to add a `<b>` tag around, then what you get will be "B" rather than the one you want to check.

JiuqingSong: And as a publish API, once we add a new parameter, we can't change or remove it until we bump the major version otherwise it will become a breaking change. So that we need to be super careful when we add new parameter to public API, and try to make it general. If we found tagName is not enough someday and need to pass in more info, how can we do it? Then we still need to add editor parameter at that time. Then it will result in two parameter: tagName and editor. But which one should we use in a callback? What if they are conflict? All these makes it into a bad design.

juliaroldi: Now I understand, the user can use CustomReplace inside a pasted content, then the tag name is not reliable anymore.

Andres-CT98:
What should I do in cases like this, where the title and description of the PR says something that is inaccurate to the implementation at the end? Do I close this PR and make a new one?

juliaroldi:
> What should I do in cases like this, where the title and description of the PR says something that is inaccurate to the implementation at the end? Do I close this PR and make a new one?
You can edit the PR name and description

JiuqingSong (approved):

PR #1086, opened 2022-07-13
Fix empty range

FrancisMengx:


packages/roosterjs-editor-core/lib/corePlugins/TypeInContainerPlugin.ts
```diff
@@ -30,6 +30,13 @@ export default class TypeInContainerPlugin implements EditorPlugin {
         this.editor = null;
     }
 
+    isRangeEmpty(range: Range) {
+        if (range.collapsed && !range.startContainer.firstChild) {
```
JiuqingSong, line 34: If startContainer is text node, it will always return true

Better check if it is an element, and check the tag name is DIV

packages/roosterjs-editor-core/lib/corePlugins/TypeInContainerPlugin.ts
```diff
@@ -30,6 +30,19 @@ export default class TypeInContainerPlugin implements EditorPlugin {
         this.editor = null;
     }
 
+    isRangeEmpty(range: Range) {
```
JiuqingSong, line 33: make this function private since we don't call it from outside

packages/roosterjs-editor-core/lib/corePlugins/TypeInContainerPlugin.ts
```diff
@@ -46,10 +59,11 @@ export default class TypeInContainerPlugin implements EditorPlugin {
             let range = this.editor.getSelectionRange();
 
             if (
-                !range ||
-                this.editor.contains(
-                    findClosestElementAncestor(range.startContainer, null /* root */, '[style]')
-                )
+                !this.isRangeEmpty(range) &&
```
JiuqingSong, line 62: You can move this check after line 63 so that we know range must not be null then no need to check again in line 35.

When enable strict mode, current code will fail the build because line 33 requires a range object but here it could benull

JiuqingSong (approved):

Issue #1093, opened 2022-07-14
Support Content Model


JiuqingSong:
Support Content Model in roosterjs.

https://microsoft.sharepoint.com/:w:/t/outlookwebteam/EQ_5svQWahlCpe0q2kvgu3MB2WoxwIz6V_YzPhOZ5epf5Q?e=3C7pdE&isSPOFile=1

evilbs:
> 

hi JiuqingSong，I want to know about this disign file, can you share it?
<img width="832" alt="image" src="https://user-images.githubusercontent.com/3305041/196479925-404ca0a8-437c-4334-a17f-88dbf5069805.png">

PR #1096, opened 2022-07-14
Content Model Step 2: Add general processor to handle all elements in a general way

JiuqingSong:
Add a general to handle all elements in a general way.
This will read all HTML element and store it in ContentModel General Block/Segment, then later we can write it back. Once this is done, we can add more specific processors to handle known elements and styles.


JiuqingSong:
#1093

PR #1098, opened 2022-07-15
Content Model Step 2: Add general processor to handle all elements in a general way

JiuqingSong:
#1093 

Add a general to handle all elements in a general way.
This will read all HTML element and store it in ContentModel General Block/Segment, then later we can write it back. Once this is done, we can add more specific processors to handle known elements and styles.


packages/roosterjs-content-model/lib/domToModel/processors/containerProcessor.ts
```diff
@@ -0,0 +1,28 @@
...
+import { generalSegmentProcessor } from './generalSegmentProcessor';
+import { isBlockElement } from 'roosterjs-editor-dom';
+import { isNodeOfType } from '../../domUtils/isNodeOfType';
+import { NodeType } from 'roosterjs-editor-types';
+import { textProcessor } from './textProcessor';
+
+/**
+ * @internal
+ */
+export function containerProcessor(group: ContentModelBlockGroup, parent: ParentNode) {
```
BryanValverdeU, line 12: Do we use the ElementProcessor Type in this function and the other function Element processors or it is not needed?

BryanValverdeU (approved):

packages/roosterjs-content-model/lib/domToModel/processors/containerProcessor.ts
JiuqingSong: Yes, it should be. Let me solve it later.

PR #1099, opened 2022-07-15
Content Model Step 3: Create DOM tree from Content Model

JiuqingSong:
#1093 
Add support to create DOM tree from Content Model.


packages/roosterjs-content-model/lib/modelToDom/optimizers/mergeNode.ts
```diff
@@ -0,0 +1,47 @@
...
+function hasSameAttributes(element1: HTMLElement, element2: HTMLElement) {
+    const attr1 = element1.attributes;
+    const attr2 = element2.attributes;
+
+    if (attr1.length != attr2.length) {
+        return false;
+    }
+
+    for (let i = 0; i < attr1.length; i++) {
+        if (attr1[i].name != attr2[i].name || attr1[i].value != attr2[i].value) {
```
flyingbee2012, line 41: Could it happen that element1 and element1 have the same attributes and values but with different order of attributes?

flyingbee2012 (approved):

packages/roosterjs-content-model/lib/modelToDom/optimizers/mergeNode.ts
JiuqingSong: For elements generated from Content Model, that should not happen. Because all attributes are generated from ContentModel code. If the orders are different, it must because some of the attributes are different.

Issue #1103, opened 2022-07-18
Table still exist after deleting all cells in it


StephenArk30:
**Describe the bug**
After deleting all <td> in a table, the <table> element still exist and get odd reaction.

**To Reproduce**
1. Create a 1X1 table
2. Right-click on the only cell
3. Choose "Delete", click "Delete Column" (or "Delete row")
4. Hover the empty line (with Table Resize Plugin enabled)

**Expected behavior**
Delete the whole table when it's empty.

**Screenshots**
![screenshot](https://user-images.githubusercontent.com/20739682/179448180-08ba2d4d-cc78-45b4-8242-661175376e32.gif)

**Device Information**
 - OS: Mac OS
 - Browser chrome
 - Version 8.27.0

StephenArk30:
fix in #1106

PR #1104, opened 2022-07-18
Fix(#1103): Delete the whole table if the table is empty after deletion

StephenArk30:
After Deleting rows or columns, see if the table is empty. If so, set cells to null.


packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -288,6 +291,9 @@ export default class VTable {
                     });
                     deletedColumns++;
                 }
+                if (this.cells[0]?.length === 0) {
```
JiuqingSong, line 294: better check cells in all rows and make sure all of them are 0

StephenArk30:
Seems like I messed up my commit history misusing rebase. I will open another pr.

PR #1105, opened 2022-07-18
Content Model Table Step 1: Add initial support to table

JiuqingSong:
#1093 

- Add new block type: Table
- Add new block group type: TableCell
- Add processor for table to create ContentModelTable from DOM
- Add handler for table to create DOM from ContentModelTable
- Add initial version of ProcessorMap


flyingbee2012 (approved):

packages/roosterjs-content-model/lib/domToModel/processors/tableProcessor.ts
```diff
@@ -0,0 +1,48 @@
...
+export const tableProcessor: ElementProcessor = (group, element) => {
+    const tableElement = element as HTMLTableElement;
+    const table = createTable(tableElement.rows.length);
+
+    addBlock(group, table);
+
+    for (let row = 0; row < tableElement.rows.length; row++) {
+        const tr = tableElement.rows[row];
+        for (let sourceCol = 0, targetCol = 0; sourceCol < tr.cells.length; sourceCol++) {
+            for (; table.cells[row][targetCol]; targetCol++) {}
```
flyingbee2012, line 30: nit: can do: while (table.cells[row][targetCol++])

JiuqingSong: Should be same

BryanValverdeU (approved):

PR #1107, opened 2022-07-19
Fix: Horizontal table inserter not disposing

StephenArk30:
In TableResizePlugin, the horizontal TableInserter should be disposed when the mouse leaves the leftmost cells, but it doesn't.

https://user-images.githubusercontent.com/20739682/179770217-74d3531d-3c27-41a1-85b4-206244d8e022.mov

This is because in this [line](https://github.com/microsoft/roosterjs/blob/cc152f0e4282b51c39070282c032712b6e450a29/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts#L143) we passed ``isHorizontal = undefined`` to ``setInserterTd ``. So in [here](https://github.com/microsoft/roosterjs/blob/cc152f0e4282b51c39070282c032712b6e450a29/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts#L187) the inserter will always be the ``verticalInserter``.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
@@ -140,7 +140,8 @@ export default class TableEditor {
                             true /*isHorizontal*/
                         );
                     } else {
-                        this.setInserterTd(null);
+                        this.setInserterTd(null, true);
+                        this.setInserterTd(null, false);
                     }
 
                     this.setResizingTd(td);


packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
```diff
@@ -140,7 +140,8 @@ export default class TableEditor {
                             true /*isHorizontal*/
                         );
                     } else {
-                        this.setInserterTd(null);
+                        this.setInserterTd(null, true);
```
JiuqingSong, line 143: Thanks for the fix. 

Can we fix inside setInsertTd() instead, when td is null, we always hide both inerters?

StephenArk30: It's ok but better describe this default behavior in comment

JiuqingSong: yes, feel free to add comment

JiuqingSong (approved):

PR #1110, opened 2022-07-20
Add buildmjs to export esmodules under lib-mjs

SudharsanSukumar:
This change adds the buildmjs step which builds the esmodule version of rooster and puts the output in lib-mjs.


SudharsanSukumar:
Note that I didn't make any updates to pack or dts since I don't think it is useful for esm. I can add that if reviewers want.

JiuqingSong (approved):

PR #1119, opened 2022-07-21
Add module field to package.json

SudharsanSukumar:
This will help with esm usage/tree shaking. In Planner's usage, this change gave us another ~100kb stat size savings in our bundle.

In the future, we also need to add exports similar to how FluentUI does it: https://github.com/microsoft/fluentui/blob/master/packages/react/package.json

More info on exports: https://webpack.js.org/guides/package-exports/

### DIFF
--- a/packages-ui/roosterjs-react/package.json
+++ b/packages-ui/roosterjs-react/package.json
@@ -15,5 +15,6 @@
         "react-dom": ">=16.0.0",
         "@fluentui/react": ">=8.0.0"
     },
-    "main": "./lib/index.ts"
+    "main": "./lib/index.ts",
+    "module": "./lib-mjs/index.ts"
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "roosterjs",
-    "version": "8.27.1",
+    "version": "8.27.2",
     "description": "Framework-independent javascript editor",
     "repository": {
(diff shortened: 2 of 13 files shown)


packages-ui/roosterjs-react/package.json
```diff
@@ -15,5 +15,6 @@
         "react-dom": ">=16.0.0",
         "@fluentui/react": ">=8.0.0"
     },
-    "main": "./lib/index.ts"
+    "main": "./lib/index.ts",
+    "module": "./lib-mjs/index.ts"
```
JiuqingSong, line 19: No need to change these lines. Let normalize.js handle it for you.

SudharsanSukumar: These changes just add a comma to line 18 and then add module (line 19). I think it's probably good form to have module in here as well instead of inserting via normalize. wdyt?

SudharsanSukumar:
We tested this change in OWA and it caused a build break with const enum usage. We may consider this addition in roosterjs v9 but for now I'll abandon this PR

JiuqingSong:
As discussed, let's keep this PR and later we can fully switch to esModule when we start roosterjs v9

JiuqingSong:
Already handled in https://github.com/microsoft/roosterjs/pull/2197

PR #1122, opened 2022-07-21
Disallow editing of readonly entity children

ianeli1:
Add checks for both inline styling and element deletion to disallow any changes to *children* of readonly entity elements.

Currently, children elements of readonly elements can be styled by changing the color and font size, as well as using `Ctrl + X` to cut text. 

Given that both `deleteSelectedContent` and `applyInlineStyle` work by calling callbacks on an array of HTML elements, we can verify if those elements/nodes are ~~children of entities~~ editable before passing them to the callback, disallowing said changes to these readonly entities.

An additional check was added to prevent nodes inside a readonly container to be merged together.


packages/roosterjs-editor-api/lib/utils/applyInlineStyle.ts
```diff
@@ -27,7 +42,7 @@ export default function applyInlineStyle(
                 (getTagOfNode(node.firstChild) == 'BR' && !node.firstChild.nextSibling));
         if (isEmptySpan) {
             editor.addUndoSnapshot();
-            callback(node as HTMLElement);
+            entitySafeCallback(node as HTMLElement);
```
JiuqingSong, line 45: I think we only need to check if this node is read only (node.isContentEditable), no need to check entity.

ianeli1: Do you mean this for all the cases, or just the empty `<span>`?

JiuqingSong: for all the cases

ianeli1: Yup, you were right 😁

JiuqingSong (approved):

PR #1123, opened 2022-07-22
Table Edit and Format test

juliaroldi:


packages/roosterjs-editor-api/test/table/formatTableTest.ts
```diff
@@ -0,0 +1,78 @@
...
+        const transformToDarkColorSpy = spyOn(editor, 'transformToDarkColor');
+        formatTable(editor, format);
+        expect(editor.addUndoSnapshot).not.toHaveBeenCalled();
+        expect(focusSpy).not.toHaveBeenCalled();
+        expect(transformToDarkColorSpy).not.toHaveBeenCalled();
+    }
+
+    it('formatTable | default', () => {
+        runTest(DEFAULT_FORMAT);
+    });
```
JiuqingSong, line 73: Can we also add test case for some complex scenario, such as:
1. Single row/column table vs multiple row/column table
2. Table with header
3. Apply format on a table that already has format
4. Expand a table or merge/split table cell that already has format

JiuqingSong (approved):

PR #1132, opened 2022-07-26
Disable multiline linking

Andres-CT98:
### Description:
Disable the link button if the user's selection is multiline.

### Fix

Compared the start and end blocks of the selection to determine if they are the same. If they are, linking is disabled. Changed the plugin event that updated the ribbon from MouseDown to MouseUp to update it once a selection has finished, not started.

### Cases tested:

> This is line 1, select me alone and you can make me a link
> This is line 2, select me and line 1 and you can not create a link
> https://exampleurl.com

**Before:** Multiline selections would be made into one line, disappearing new lines.
**After:** Making a multiline selection disables the link button.

### Warnings and implications

- Triple clicking a line selects the whole line, however this selection is treated as if it was multiline. Replicable on Chrome and Edge but not on Firefox.


packages/roosterjs-editor-api/lib/format/getFormatState.ts
```diff
@@ -20,13 +20,43 @@ export function getElementBasedFormatState(
     event?: PluginEvent
 ): ElementBasedFormatState {
     let listTag = getTagOfNode(editor.getElementAtCursor('OL,UL', null /*startFrom*/, event));
+
+    let uuu = editor.getBlockTraverser();
```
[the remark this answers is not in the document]
JiuqingSong: @juliaroldi  why inline element can help here?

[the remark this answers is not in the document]
juliaroldi: I think he is trying to check if it is a multiline link, right? He can check if the selection has more that one node and if use check if the elements are inline

packages-ui/roosterjs-react/lib/ribbon/component/getButtons.ts
```diff
@@ -34,6 +34,7 @@ import { underline } from './buttons/underline';
 import { undo } from './buttons/undo';
 
 const KnownRibbonButtons: { [key in KnownRibbonButtonKey]: RibbonButton<AllButtonStringKeys> } = {
+    [KnownRibbonButtonKey.InsertLink]: insertLink,
```
JiuqingSong, line 37: any reason to move this button to the first?

packages/roosterjs-editor-api/lib/format/getFormatState.ts
```diff
@@ -20,13 +20,24 @@ export function getElementBasedFormatState(
...
     let listTag = getTagOfNode(editor.getElementAtCursor('OL,UL', null /*startFrom*/, event));
+    let range = editor.getSelectionRange();
+
+    // Check if selection is multiline, spans more than one block
+    let multiline = false;
+    if (range) {
+        let startingBlock = editor.getBlockElementAtNode(range.startContainer);
+        let endingBlock = editor.getBlockElementAtNode(range.endContainer);
+        multiline = endingBlock.equals(startingBlock);
+    }
```
JiuqingSong, line 31: As an optimization, you can first check if range is collapsed (range.collapsed), when it is true, we know it must not be multiple line, and no need to create block element. This is a the most common case and it is fast enough.

packages/roosterjs-editor-api/lib/format/getFormatState.ts
```diff
@@ -20,13 +20,24 @@ export function getElementBasedFormatState(
...
+
     let headerTag = getTagOfNode(
         editor.getElementAtCursor('H1,H2,H3,H4,H5,H6', null /*startFrom*/, event)
     );
 
     return {
         isBullet: listTag == 'UL',
         isNumbering: listTag == 'OL',
+        isMultiline: multiline,
```
JiuqingSong, line 40: How about name it "isMultiLineSelection"?

packages-ui/roosterjs-react/lib/ribbon/component/getButtons.ts
Andres-CT98: For testing, I will revert it back when this PR is ready for review.

JiuqingSong (approved):

PR #1137, opened 2022-07-27
InsertNode with option InsertInNewLine is inserting the new node incorrectly when target is a table cell

BryanValverdeU:
**Describe the bug**
Using InsertNode Editor API inside of a table and setting the Insert Option, InsertInNewLine to true, it is inserting the new element incorrectly.

**Code Change**
If inside of a cell add a Div so we can insert the content in a new line, instead of inserting the content next to the cell

**Before**
image

**After**
InsertNode1

### DIFF
--- a/packages/roosterjs-editor-core/lib/coreApi/insertNode.ts
+++ b/packages/roosterjs-editor-core/lib/coreApi/insertNode.ts
@@ -156,6 +156,7 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
                     } else {
                         pos = adjustInsertPosition(contentDiv, node, pos, range);
                     }
+
                     let nodeForCursor =
                         node.nodeType == NodeType.DocumentFragment ? node.lastChild : node;
                     range = createRange(pos);
--- a/packages/roosterjs-editor-core/lib/coreApi/insertNode.ts
+++ b/packages/roosterjs-editor-core/lib/coreApi/insertNode.ts
@@ -152,11 +152,10 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
                         option.insertOnNewLine &&
                         (blockElement = getBlockElementAtNode(contentDiv, pos.normalize().node))
                     ) {
-                        pos = new Position(blockElement.getEndNode(), PositionType.After);
+                        pos = adjustInsertPositionNewLine(blockElement, core, pos);
                     } else {
                         pos = adjustInsertPosition(contentDiv, node, pos, range);
                     }
-
                     let nodeForCursor =
                         node.nodeType == NodeType.DocumentFragment ? node.lastChild : node;
                     range = createRange(pos);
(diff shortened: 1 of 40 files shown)


packages/roosterjs-editor-core/lib/coreApi/insertNode.ts
```diff
@@ -179,3 +178,23 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
...
+    let tempPos = new Position(blockElement.getEndNode(), PositionType.After);
+    if (safeInstanceOf(tempPos.node, 'HTMLTableRowElement')) {
+        const div = core.contentDiv.ownerDocument.createElement('div');
+        div.id = 'newLineInTd';
+        insertNode(core, div, {
+            position: ContentPosition.SelectionStart,
+            insertOnNewLine: false,
+            updateCursor: true,
+            replaceSelection: true,
+        });
```
JiuqingSong, line 191: Inside insertNode() function, can we avoid calling it recursively? Can we just directly insert the node using some DOM API such as insertBefore or appendChild?

packages/roosterjs-editor-core/lib/coreApi/insertNode.ts
```diff
@@ -179,3 +178,23 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
 
     return true;
 };
+function adjustInsertPositionNewLine(blockElement: BlockElement, core: EditorCore, pos: Position) {
+    let tempPos = new Position(blockElement.getEndNode(), PositionType.After);
+    if (safeInstanceOf(tempPos.node, 'HTMLTableRowElement')) {
+        const div = core.contentDiv.ownerDocument.createElement('div');
+        div.id = 'newLineInTd';
```
JiuqingSong, line 185: No need to use an Id, just hold the reference of DIV and it won't change.

packages/roosterjs-editor-core/lib/coreApi/insertNode.ts
BryanValverdeU: Sounds good will use the Range.insertNode Api to do this.

packages/roosterjs-editor-core/lib/coreApi/insertNode.ts
```diff
@@ -179,3 +179,13 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
 
     return true;
 };
+function adjustInsertPositionNewLine(blockElement: BlockElement, core: EditorCore, pos: Position) {
+    let tempPos = new Position(blockElement.getEndNode(), PositionType.After);
+    if (safeInstanceOf(tempPos.node, 'HTMLTableRowElement')) {
+        const div = core.contentDiv.ownerDocument.createElement('div');
+        const range = createRange(pos);
+        range.insertNode(div);
```
JiuqingSong, line 187: Wait, is this going to insert a DIV after a TR?

Can you give me an example including:
- HTML before insert
- The position
- HTML after insert

BryanValverdeU: No, it is going to insert a Div inside of a TD. To treat it as a new Line inside of a table cell

BryanValverdeU: Before Insert 
image

Position
image

After Insert
image

JiuqingSong (approved):

PR #1138, opened 2022-07-27
Fix cut action when the selection is a table

juliaroldi:
Bug: Select a table or some table columns   and press CRTL + X. 
Expected: Table os removed from the editor
Actual Behavior: Cut action behaves as copy. 

Fix: Check if the table has a selection, if yes creates a VTable and use DeleteColumns operation to delete the selection, if the selection includes whole columns. 

RemovedCutTable
 
cutPasteTableSafari


BryanValverdeU (approved):

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -188,20 +188,27 @@ export default class TableCellSelection implements EditorPlugin {
...
-                        }
-                    });
+                    this.deleteTableColumns(selection.table, selection.coordinates);
                 }
             }
         }
     }
 
+    private deleteTableColumns(table: HTMLTableElement, selection: TableSelection) {
+        const isWholeColumnSelection = table.rows.length - 1 === selection.lastCell.y;
```
JiuqingSong, line 201: do you also need to check selection.firstCell.y? What if I select from a middle row to the last row?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -185,6 +185,11 @@ export default class VTable {
      * @param operation Table operation
      */
     edit(operation: TableOperation | CompatibleTableOperation) {
+        if (this.selection && (this.row === undefined || this.col == undefined)) {
+            this.col = 0;
+            this.row = 0;
+        }
```
JiuqingSong, line 191: hmm, so we sometimes read col and row, sometimes read selection?

I think we should always read selection, and deprecate col and row

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -188,20 +188,27 @@ export default class TableCellSelection implements EditorPlugin {
...
+
+        if (isWholeColumnSelection) {
```
JiuqingSong, line 203: What if whole row is selected? we should handle both cases

juliaroldi: The word online just handles whole columns or whole table. If a whole row is or not whole columns is selected, the editor does dot remove the cells.

JiuqingSong:
With this change, if I select a whole column and cut, it removes the column, which is fine.
But if I select a whole row, or select some cells or a row (or a column) then cut, it will act as copy, which is wrong.

juliaroldi:
> With this change, if I select a whole column and cut, it removes the column, which is fine. But if I select a whole row, or select some cells or a row (or a column) then cut, it will act as copy, which is wrong.

That is the same behavior of office online.

copyPasteTableWord

JiuqingSong:
For the table size issue, I think you just need to remove table width/height in TableCellSelection plugin when handle the copy/cut event and remove unselected cells

PR #1142, opened 2022-07-28
Fix numbering list triggers

juliaroldi:
Ensure list will be trigger only by the right combination of characters. 
listBugs

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getAutoNumberingListStyle.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getAutoNumberingListStyle.ts
@@ -19,14 +19,14 @@ const characters: Record<string, number> = {
     '.': Character.Dot,
     '-': Character.Dash,
     ')': Character.Parenthesis,
-    '(': Character.DoubleParenthesis,
 };
 
-const identifyCharacter = (text: string, secondSeparator?: string) => {
-    const charType = characters[text];
-    return charType === Character.DoubleParenthesis && secondSeparator !== ')'
-        ? undefined
-        : charType;
+const numberingTriggers: Record<string, number> = {
+    '1': NumberingTypes.Decimal,
+    i: NumberingTypes.LowerRoman,
+    I: NumberingTypes.UpperRoman,
+    a: NumberingTypes.LowerAlpha,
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getAutoNumberingListStyle.ts
```diff
@@ -109,9 +112,17 @@ const identifyNumberingListType = (numbering: string): NumberingListType | null
  * @param textBeforeCursor The trigger character
  * @returns The style of a numbering list triggered by a string
  */
-export default function getAutoNumberingListStyle(textBeforeCursor: string): NumberingListType {
+export default function getAutoNumberingListStyle(
+    textBeforeCursor: string,
+    num?: number
```
JiuqingSong, line 117: add comment for this parameter

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getAutoNumberingListStyle.ts
```diff
@@ -90,16 +90,19 @@ const DecimalsTypes: Record<number, number> = {
...
+    const separatorCharacter = isDoubleParenthesis
+        ? Character.DoubleParenthesis
+        : characters[numbering[1]];
     // if separator is not valid, no need to check if the number is valid.
-    if (char) {
+    if (separatorCharacter) {
         const number = numbering.length === 3 ? numbering[1] : numbering[0];
-        const numberingType = identifyNumberingType(number);
-        return numberingType ? numberingListTypes[numberingType](char) : null;
+        const numberingType = num ? identifyNumberingType(number) : numberingTriggers[number];
```
JiuqingSong, line 104: How to do you want to handle num == 0?

juliaroldi: 0 can treated as undefined, if num === 0 it means the list must start from beginning, so we should only check for a, A, i, I and 1, them numberingTrigger is enough

JiuqingSong: ok

JiuqingSong (approved):

PR #1158, opened 2022-08-05
Watermarkv2 plugin

### LINKED ISSUE #363: Add option to keep watermark until text is entered
**Describe the bug**
We would like the watermark text to work like placeholder text. When the field is clicked upon, the text stays visible until the user inputs text. This could be optional.

**To Reproduce**
1. Implement a roosterJs field with watermark text
2. Click on the text box

**Expected behavior**
The placeholder text should be visible still

**Screenshots**

**Additional context**
This is a PM ask for the new Project

StephenArk30:
Psuedo Watermark, fix #363


JiuqingSong:
Thanks for making this PR. I'll take a look once I got a chance.

packages/roosterjs-editor-plugins/lib/plugins/WatermarkV2/WatermarkV2.ts
```diff
@@ -0,0 +1,119 @@
...
+} from 'roosterjs-editor-types';
+import { kebabCase } from './utils';
+
+const ENTITY_TYPE = 'WATERMARK_WRAPPER';
+const WATERMARK_ATTR = 'data-rooster-watermark';
+const STYLE_ID = 'watermarkStyle';
+
+const DEFAULT_STYLE = {
+    color: 'rgba(0,0,0,0.3)',
+    position: 'absolute',
```
JiuqingSong, line 16: So this means the editor DIV must have position either to be absolute or relative. But this is not always true. I need to think about if we can always add a position:relative if there is not.

packages/roosterjs-editor-plugins/lib/plugins/WatermarkV2/WatermarkV2.ts
```diff
@@ -0,0 +1,119 @@
...
+    Entity,
+    IEditor,
+    PluginEvent,
+    PluginEventType,
+} from 'roosterjs-editor-types';
+import { kebabCase } from './utils';
+
+const ENTITY_TYPE = 'WATERMARK_WRAPPER';
+const WATERMARK_ATTR = 'data-rooster-watermark';
+const STYLE_ID = 'watermarkStyle';
```
JiuqingSong, line 12: There can be issue if there are more than 1 editor instances in the same page and one of them is disposed, then this style tag will be removed.

packages/roosterjs-editor-plugins/lib/plugins/WatermarkV2/WatermarkV2.ts
```diff
@@ -0,0 +1,119 @@
...
+        let styleElement = doc.getElementById(STYLE_ID) as HTMLStyleElement;
+        if (!styleElement) {
+            styleElement = doc.createElement('style');
+            doc.head.appendChild(styleElement);
+            styleElement.id = STYLE_ID;
+        }
+        const cssText = Object.keys(this.style)
+            .map(key => `${kebabCase(key)}:${this.style[key as any]}`)
+            .join(';');
+        const className = this.editor.getEditorDomAttribute('class').split(' ')[0];
```
JiuqingSong, line 63: If editor DIV is using a shared class name with other elements, this may have some issue. 

We can leverage the existing unique ID strategy that we are using for table selection (and the code to manage style tags as well) to make this to be more robust.

packages/roosterjs-editor-plugins/lib/plugins/WatermarkV2/WatermarkV2.ts
```diff
@@ -0,0 +1,119 @@
...
+            ) ||
+            (event.eventType == PluginEventType.ContentChanged &&
+                (<Entity>event.data)?.type != ENTITY_TYPE)
+        ) {
+            this.showHideWatermark();
+        }
+    }
+
+    private showHideWatermark = () => {
+        this.editor.setEditorDomAttribute(
```
JiuqingSong, line 107: this is nice. Only concern is, with multi-language support, if user changes language during this editing session, the watermark text can't be auto switched to new language. Maybe we can make the watermark parameter to be a callback function

packages/roosterjs-editor-plugins/lib/plugins/WatermarkV2/WatermarkV2.ts
```diff
@@ -0,0 +1,119 @@
...
+            WATERMARK_ATTR,
+            isEmpty(this.editor) ? this.watermark : null
+        );
+    };
+}
+
+function isEmpty(editor: IEditor) {
+    if (editor.isEmpty()) {
```
JiuqingSong, line 115: This is the top concern of the whole approach. This check will happen on each input event. So if the content is super long, and user is typing fast, there might be perf bottleneck since this check is relatively expensive. I think we may need to cache the result and refresh it in a proper way to save most of the check.

Let me think about that.

packages/roosterjs-editor-plugins/lib/plugins/WatermarkV2/WatermarkV2.ts
StephenArk30: Maybe use [ensureUniqueId](https://github.com/microsoft/roosterjs/blob/2e598554cbd848ebb1fcd823656f7aa4d0aa47b2/packages/roosterjs-editor-core/lib/coreApi/selectTable.ts#L195)? Can we make this exported?

packages/roosterjs-editor-plugins/lib/plugins/WatermarkV2/WatermarkV2.ts
StephenArk30: How about using a debounce?

PR #1161, opened 2022-08-05
Fix #1115: Use pending format SPAN for style-based format change when selection is collapsed

JiuqingSong:
Today when do a format change such as font family, font size, text color, background color, and selection is collapsed, we will insert a ZeroWidthSpace inside a SPAN into editor to hold the new format. This approach has the following issue:
1. An invisible space is inserted so it may confuse announcer (#1115)
2. When move cursor, it will need an extra key press to move cursor over the ZeroWidthSpace

In this change I added a new experimental feature "PendingStyleBasedFormat". With this feature enabled, we will use an in-memory SPAN instead of in-content SPAN to hold the pending format change, and only insert it into editor content when user types something, or it will be discarded if cursor position is changed. So that there will not be any ZeroWidthSpace remain in editor.

Also fix related code such as format state, increase/decrease font size, and demo site.


packages/roosterjs-editor-core/lib/coreApi/getStyleBasedFormatState.ts
```diff
@@ -17,22 +17,41 @@ export const getStyleBasedFormatState: GetStyleBasedFormatState = (
...
     const ogTextColorNode =
-        isDarkMode && findClosestElementAncestor(node, root, ORIGINAL_STYLE_COLOR_SELECTOR);
+        isDarkMode &&
+        (override[2]
+            ? pendableFormatSpan
+            : findClosestElementAncestor(node, root, ORIGINAL_STYLE_COLOR_SELECTOR));
     const ogBackgroundColorNode =
-        isDarkMode && findClosestElementAncestor(node, root, ORIGINAL_STYLE_BACK_COLOR_SELECTOR);
+        isDarkMode &&
+        (override[2]
```
BryanValverdeU, line 43: Should this be override[3]?

JiuqingSong: good catch!

BryanValverdeU (approved):

PR #1167, opened 2022-08-09
Add a line between tables when inserting a table

BryanValverdeU:
When inserting a table after another table, the user is not able to put the cursor between the two tables. in order to fix this, we add a new line when inserting a new table right after an already existing table in the editor.

#Before Change:
image

#After change:
image


packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -281,6 +282,42 @@ function adjustInsertPositionForNotEditableNode(
...
+    range: Range
+): NodePosition {
+    if (
+        (nodeToInsert.childNodes.length == 1 &&
+            getTagOfNode(nodeToInsert.childNodes[0]) == 'TABLE') ||
+        getTagOfNode(nodeToInsert) == 'TABLE'
+    ) {
+        const { element, offset } = position;
+        const previousElementSibling = element.previousElementSibling;
+        const prevElement =
```
JiuqingSong, line 301: Suggest use ContentTraverser to find prevous element, which is more accurate than just directly access DOM tree

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -281,6 +282,42 @@ function adjustInsertPositionForNotEditableNode(
...
+            offset - 1 >= 0
+                ? element.childNodes.item(offset - 1)
+                : previousElementSibling?.childNodes.item(
+                      previousElementSibling.childElementCount - 1
+                  );
+        if (prevElement && getTagOfNode(prevElement) == 'TABLE' && element?.isContentEditable) {
```
JiuqingSong, line 307: If feel even element is not editable, we should also add this empty line. Any reason we check content editable here?

BryanValverdeU: No reason at all, will remove it.

JiuqingSong (approved):

PR #1172, opened 2022-08-12
fix: imageEditPlugin cause content bounce and original image size change

JingweiXiong:
## Question 1
When start image edit cause content  bounce because change the content verticalAlign.

https://user-images.githubusercontent.com/19570679/184290934-88e7b5c3-f2fd-45fc-ac67-093296402189.mov

## Question 2
When click image start edit, but don't resize. After blur, the original image may be change its size because set initial when not wasResized

https://user-images.githubusercontent.com/19570679/184291139-4e856509-aebe-4f4a-a95f-b7d45a46d897.mov

## Fixed
Fixed in my code, thanks.


JiuqingSong (approved):

JiuqingSong:
I found a regression from this change:
1. insert an image
2. select it
3. rotate it
4. unselect
result: the image shape is changed incorrectly

I have made a change and it is included in this PR https://github.com/microsoft/roosterjs/pull/1175

JingweiXiong:
> I found a regression from this change:
> 
> 1. insert an image
> 2. select it
> 3. rotate it
> 4. unselect
>    result: the image shape is changed incorrectly
> 
> I have made a change and it is included in this PR #1175

ok, thanks.

PR #1174, opened 2022-08-12
Add effective background color value.

marcinkiewicz:
Add `effectiveBackgroundColor` which is representing actual calculated background visible to the user.

When we click in the middle of element which has `bold`, `italic` and `background color` activated, `backgroundColor` value shows `rgba(0,0,0,0)` which is accurate from DOM, but not from user point of view. User should be able to see the actual background value obtained by traversing through the ancestors in the DOM tree and effectively show `rgb(0, 85, 0)`:
image

### DIFF
--- a/packages/roosterjs-editor-dom/lib/utils/getEffectiveBackgroundColor.ts
+++ b/packages/roosterjs-editor-dom/lib/utils/getEffectiveBackgroundColor.ts
@@ -1,0 +1,19 @@
+import findClosestElementAncestor from './findClosestElementAncestor';
+import getComputedStyles from './getComputedStyles';
+
+/**
+ * Get effective background value based on parent nodes
+ * @param node The node to get computed styles from
+ * @param rootNode Top most ancestor node to stop searching for inherited background value
+ * @returns An array of the computed styles
+ */
+export default function getEffectiveBackgroundColor(node: Node, rootNode: Node): string {
+    const value = getComputedStyles(node, 'background-color')[0];
+    if (value === 'rgba(0, 0, 0, 0)') {
+        const parentNode = findClosestElementAncestor(node.parentNode, rootNode);
+        if (parentNode) {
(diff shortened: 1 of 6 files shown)


packages/roosterjs-editor-dom/lib/utils/getEffectiveBackgroundColor.ts
```diff
@@ -0,0 +1,19 @@
+import findClosestElementAncestor from './findClosestElementAncestor';
+import getComputedStyles from './getComputedStyles';
+
+/**
+ * Get effective background value based on parent nodes
+ * @param node The node to get computed styles from
+ * @param rootNode Top most ancestor node to stop searching for inherited background value
+ * @returns An array of the computed styles
+ */
+export default function getEffectiveBackgroundColor(node: Node, rootNode: Node): string {
```
JiuqingSong, line 10: A little bit concern about the performance here since it can recursively call until it get a valid color or hit the root of editor. This function is frequently called especially when user is typing. So we want to make this result to be fast enough, even sometimes it is not very accurate.

We are working on a new ContentModel approach which can solve this kind of issue and return a more accurate format state.

marcinkiewicz: It happens always if element is nested in another one. Do you have a reference to the release plan for the new ContentModel?

JiuqingSong: Content Model is under development now.

If you are blocked by this issue, you can create a plugin to handle BeforeCopy/BeforeCut event to manually assign the color to copied content

JiuqingSong:
This is now included in Content Model API [getFormatState](https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs-content-model/lib/publicApi/format/getFormatState.ts) 

We will add a ContentModel editor class to allow using this API soon. Close this one.

PR #1200, opened 2022-08-25
Keep start property and marker (if supported) when copying list from WordOnline

BryanValverdeU:
When copying lists from Word Online, sometimes the marker is not saved and the start of the list is reset:

For example, this content in Word Online
image

| Gets pasted like this in Rooster:  | With this change, we add support to keep the list start property and also the marker type if it is supported and mantain the start property of the lists|
| ------------- | ------------- |
|image  | image |

Some of the Word Online markers in the clipboard are unsupported. If I tried to use them the marker would look bad, so only can support some of the Word Online Markers

| Word Online  | RoosterJS |
| ------------- | ------------- |
|Screenshot 2022-08-24 152244 | Screenshot 2022-08-24 152400  | 

Why some of the markers are not supported:
The clipboard provided from WordOnline does not contain the data-leveltext hence we cannot use it to change the list marker
image

### DIFF
--- a/packages/roosterjs-editor-dom/lib/list/VListItem.ts
+++ b/packages/roosterjs-editor-dom/lib/list/VListItem.ts
@@ -207,13 +207,8 @@ export default class VListItem {
      */
     outdent(preventItemRemoval: boolean = false) {
         const expectedLength = preventItemRemoval ? 2 : 1;
-        const listTypesLength = this.listTypes.length;
-        if (listTypesLength > expectedLength) {
+        if (this.listTypes.length > expectedLength) {
             this.listTypes.pop();
-        }
-
-        if (this.listTypes[listTypesLength - 1] == ListType.Ordered) {
-            this.node.style.listStyleType = '';
         }
     }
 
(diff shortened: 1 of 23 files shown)


packages/roosterjs-editor-dom/lib/list/VListItem.ts
```diff
@@ -207,9 +207,14 @@ export default class VListItem {
...
     outdent(preventItemRemoval: boolean = false) {
         const expectedLength = preventItemRemoval ? 2 : 1;
-        if (this.listTypes.length > expectedLength) {
+        const listTypesLength = this.listTypes.length;
+        if (listTypesLength > expectedLength) {
             this.listTypes.pop();
         }
+
+        if (this.listTypes[listTypesLength - 1] == ListType.Ordered) {
+            this.node.style.listStyleType = '';
```
JiuqingSong, line 216: why set to empty? What if the parent level list also have a list style type, should we also follow?

BryanValverdeU: Removed this part of code as the scope of this PR is only going to be pasting from Word Online.
However, I found that if a list have a 'custom' list style type and we indent or outdent, Vlist is going to remove the list style and will have a different list style. Adding a gif to explain better:

WordPast2

JiuqingSong: So this should be a separate issue, right?
Does it repro without this change?

BryanValverdeU: Yes, separate issue.

JiuqingSong: ok. please open issue to track.

JiuqingSong (approved):

Issue #1201, opened 2022-08-25
Select table failed when <th> and <td> in the same row


StephenArk30:
**Describe the bug**
If <th> and <td> in the same <tr>, table selection will fail.

**To Reproduce**
1. Copy table [here](https://codepen.io/stephenark30/pen/bGvJLpO)
2. Paste table in Rooster editor
3. Try to select the table

**Expected behavior**
Correctly select the table

**Screenshots**

https://user-images.githubusercontent.com/20739682/186564385-283ded5e-a34b-4580-a263-23467c46808f.mov

**Device Information**
 - OS: MacOS
 - Browser chrome
 - Version 8.29.4

**Additional context**
Table code

```html
<table style="border-collapse: collapse">
  <tbody>
    <tr>
      <th style="border: 1px solid; background: yellow">th</th>
      <td style="border: 1px solid">td</td>
    </tr>
    <tr>
      <th style="border: 1px solid; background: yellow">th</th>
      <td style="border: 1px solid">td</td>
    </tr>
  </tbody>
</table>
```

PR #1202, opened 2022-08-25
Fix(#1201): Table selection when <th> and <td> in the same row

StephenArk30:
Check [nth-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child) selector. No need to consider the tag name when using this selector.


JiuqingSong:
@BryanValverdeU would you please help review this change?

@StephenArk30  Thanks for make the fix. Would you please also add test case for the fix?

BryanValverdeU (approved):
Looks good to me. Thanks for the fix.

StephenArk30:
> Looks good to me. Thanks for the fix. Please add a unit test for this scenario

ok

StephenArk30:
Done. Is that enough?

JiuqingSong (approved):

PR #1209, opened 2022-08-26
Fix table resize after merge cells. 

juliaroldi:
Recalculate cell height after cells are merged. 
cellHeight


packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -488,6 +499,20 @@ export default class VTable {
         }
     }
 
+    private isAEmptyCell(td: HTMLTableCellElement) {
```
JiuqingSong, line 502: What does "A" mean in  the function name?

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -148,6 +150,15 @@ export default class VTable {
         }
     }
 
+    private recalculateCellHeight(td: HTMLTableCellElement) {
+        if (this.isAEmptyCell(td) && td.rowSpan && td.rowSpan > 1) {
```
JiuqingSong, line 154: no need to do null check for rowSpan. It must always be there

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -148,6 +150,15 @@ export default class VTable {
         }
     }
 
+    private recalculateCellHeight(td: HTMLTableCellElement) {
+        if (this.isAEmptyCell(td) && td.rowSpan && td.rowSpan > 1) {
+            for (let i = 1; i < td.rowSpan; i++) {
+                const br = document.createElement('br');
+                td.appendChild(br);
```
JiuqingSong, line 157: why do this?

JiuqingSong: hmm, ok. But I feel maybe the better way is to set the height of cell, but not use BR to fill it.

We can go with this for now.

JiuqingSong (approved):

PR #1221, opened 2022-08-30
Fix crash when enabling Watermark plugin

ianeli1:
Steps to repro:
1. [Open Rooster demo](https://microsoft.github.io/roosterjs/index.html#options), don't type anything
2. Open side panel
3. Plugins
4. Toggle Watermark Plugin

[domToContentModel](https://github.com/microsoft/roosterjs/blob/130459ede719255c5ad6685a14041d2d12bbcdf0/packages/roosterjs-content-model/lib/publicApi/domToContentModel.ts) gets called with an `undefined` root, causing a crash. 
image
Instead of using the experimental editor's contentDiv, we can access the editor core's directly, which seems to fix the problem.

### DIFF
--- a/demo/scripts/controls/editor/ExperimentalContentModelEditor.ts
+++ b/demo/scripts/controls/editor/ExperimentalContentModelEditor.ts
@@ -23,7 +23,7 @@ export default class ExperimentalContentModelEditor extends Editor
      * @param contentDiv The DIV HTML element which will be the container element of editor
      * @param options An optional options object to customize the editor
      */
-    constructor(contentDiv: HTMLDivElement, options?: EditorOptions) {
+    constructor(private contentDiv: HTMLDivElement, options?: EditorOptions) {
         super(contentDiv, options);
         this.getDarkColor = options?.getDarkColor;
     }
@@ -35,7 +35,7 @@ export default class ExperimentalContentModelEditor extends Editor
         return {
             isDarkMode: this.isDarkMode(),
             zoomScale: this.getZoomScale(),
-            isRightToLeft: getComputedStyles(this.core.contentDiv, 'direction')[0] == 'rtl',
+            isRightToLeft: getComputedStyles(this.contentDiv, 'direction')[0] == 'rtl',
             getDarkColor: this.getDarkColor,
         };
     }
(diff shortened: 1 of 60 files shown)


JiuqingSong:
Do you know why the original way not work?

JiuqingSong:
Although the change may fix the issue, but I'm afraid it also hides the real root cause.

ianeli1:
> Do you know why the original way not work?

Fun debugging session 😄 
It's a timing issue, we're assuming the properties of the experimental editor get assigned immediately
```ts
    constructor(private contentDiv: HTMLDivElement, options?: EditorOptions) { //this.contentDiv gets assigned
        super(contentDiv, options); //Editor gets initialized
        this.getDarkColor = options?.getDarkColor; //then this, all built
    }
```

In reality the `super` gets called first, which means the plugins get initialized, ending in the Lifecycle plugin emitting `PluginEventType.EditorReady`, which then triggers the Watermark plugin to show the watermark. 
Showing the watermark changes the content and makes the editor create a new content model, which won't be possible as `this.contentDiv` hasn't been assigned yet (this all happens within the `super` call). 

Conveniently, none of this happens if we have content in the editor: 
```ts
} else if (!hasFocus && !isShowing && this.editor.isEmpty()) {
            insertEntity(
```

JiuqingSong:
> > Do you know why the original way not work?
> 
> Fun debugging session 😄 It's a timing issue, we're assuming the properties of the experimental editor get assigned immediately
> 
> ```ts
>     constructor(private contentDiv: HTMLDivElement, options?: EditorOptions) { //this.contentDiv gets assigned
>         super(contentDiv, options); //Editor gets initialized
>         this.getDarkColor = options?.getDarkColor; //then this, all built
>     }
> ```
> 
> In reality the `super` gets called first, which means the plugins get initialized, ending in the Lifecycle plugin emitting `PluginEventType.EditorReady`, which then triggers the Watermark plugin to show the watermark. Showing the watermark changes the content and makes the editor create a new content model, which won't be possible as `this.contentDiv` hasn't been assigned yet (this all happens within the `super` call).
> 
> Conveniently, none of this happens if we have content in the editor:
> 
> ```ts
> } else if (!hasFocus && !isShowing && this.editor.isEmpty()) {
>             insertEntity(
> ```

I tried and I can repro. So it seems the problem is in ContentModel side pane plugin only. In that case, we can just to a null check in that plugin and if content DIV is not ready yet, just skip creating content model. 

We'd like to hide the detail of editor as much as possible.

JiuqingSong (approved):

PR #1222, opened 2022-08-30
fix: maybe loss of background color when copy

JingweiXiong:
## Question

https://user-images.githubusercontent.com/19570679/187338983-e2e2c316-01e4-4228-89fd-88d099d9efef.mov

When copy, loss of background-color because the focus node is inside, background-color property is set in outside node.

## After Resolve

https://user-images.githubusercontent.com/19570679/187339239-aa2397e5-7d4c-459a-9c4f-2a38785c0328.mov

Thanks.


JiuqingSong:
For this issue I'd like to leave it there until our content model feature is ready enough so that can be easily handled by content model

JingweiXiong:
> 

ok, I'm looking forward to its completion.

JiuqingSong:
Content Model copy is now checked in. So close this one.

PR #1224, opened 2022-08-30
fix: pendable format state isBold is incorrect

JingweiXiong:
## Question
Xnip2022-08-30_15-26-17
This case is inside h2，isBold should be true, but get false

## After solved
Xnip2022-08-30_15-20-35

Thanks.


JiuqingSong:
The problem is, if we show it as bold for headers, if user press bold button, it will not "unbold" the text, so it is still confusing.

JiuqingSong (approved):

PR #1225, opened 2022-08-30
Fix Cut/Copy tables

juliaroldi:
CopyPaste plugin should handle table selection. 
When the BeforeCopyCut event is trigger, CopyPaste plugin will check if the selection is a table and if it is, the plugin will use the selection data to copy the cell and erase the cells, if the event is cut. 
CutCopyTables


packages/roosterjs-editor-dom/lib/table/forEachSelectedCell.ts
```diff
@@ -0,0 +1,21 @@
+import VTable from './VTable';
+import { VCell } from 'roosterjs-editor-types';
+
+/**
+ * Executes an action to all the cells within the selection range.
+ * @param callback action to apply on each selected cell
+ * @returns the amount of cells modified
+ */
+export default function forEachSelectedCell(vTable: VTable, callback: (cell: VCell) => void): void {
```
JiuqingSong, line 9: Where is this function used? If it only used in CopyPastePlugin but nowhere else, you can put it inside that file and no need to export. Let's try to avoid exporting unnecessary function.

packages/roosterjs-editor-dom/lib/table/removeCellsOutsideSelection.ts
```diff
@@ -0,0 +1,36 @@
+import isWholeTableSelected from './isWholeTableSelected';
+import VTable from './VTable';
+import { VCell } from 'roosterjs-editor-types';
+
+/**
+ * Remove the cells outside of the selection.
+ * @param vTable VTable to remove selection
+ */
+export default function removeCellsOutsideSelection(vTable: VTable) {
```
JiuqingSong, line 9: same here

packages/roosterjs-editor-core/lib/corePlugins/CopyPastePlugin.ts
```diff
@@ -86,10 +92,20 @@ export default class CopyPastePlugin implements PluginWithState<CopyPastePluginS
                 html,
                 this.editor.getTrustedHTMLHandler()
             );
-            const newRange =
-                metadata?.type === SelectionRangeTypes.Normal
-                    ? createRange(tempDiv, metadata.start, metadata.end)
-                    : null;
+            let newRange;
```
JiuqingSong, line 95: need type declaration

packages/roosterjs-editor-core/lib/corePlugins/CopyPastePlugin.ts
```diff
@@ -86,10 +92,20 @@ export default class CopyPastePlugin implements PluginWithState<CopyPastePluginS
...
+                }
+            } else {
+                newRange =
+                    metadata?.type === SelectionRangeTypes.Normal
+                        ? createRange(tempDiv, metadata.start, metadata.end)
+                        : null;
+            }
 
             this.editor.triggerPluginEvent(PluginEventType.BeforeCutCopy, {
                 clonedRoot: tempDiv,
```
JiuqingSong, line 111: From line 117, better use the event (returned from triggerPluginEvent from line 110) range property:
```ts
const event = this.editor.triggerPluginEvent(...);
if (event.range) {
 ...
}
```

JiuqingSong: Because we allow plugin to replace the range object

JiuqingSong (approved):

PR #1226, opened 2022-08-30
Content Model: fix table size when zoom

JiuqingSong:
#1093 

Found another bug in Content Model table API, when calculate table cell size, zoom scale is not included.

Fix it and add test case.


JiuqingSong:
#1227

flyingbee2012 (approved):

PR #1229, opened 2022-08-31
Enable Strict Mode on rooster-editor-core/lib/editor

ianeli1:
In an effort to enable strict mode project wide, this PR enables strict mode within `Editor.ts` and includes all the required fixes and changes on related packages, along with some small public API changes


packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -712,7 +732,7 @@ export default class Editor implements IEditor {
      */
     public getContentSearcherOfCursor(event?: PluginEvent): IPositionContentSearcher {
         return cacheGetEventData(event, 'ContentSearcher', () => {
-            let range = this.getSelectionRange();
+            let range = this.getSelectionRange() ?? new Range();
```
ianeli1, line 735: Do we rather change the API for all content -traverser-returning methods to returning nullish values OR return empty content traverser (aka with a collapsed range)? If the latter, do we create an empty range on the content div?

ianeli1: I'm thinking of creating an empty range on the firstChild of contentDiv. Can I assume there will always be one? 
```ts
let range = this.getSelectionRange() ?? createRange(this.core.contentDiv.firstChild!);
```

JiuqingSong: No. If getSelectionRange() returns null, it means editor has never got focused before. So we don't know where should we focus to (and user doesn't know as well). So we should let it return null.

packages/roosterjs-content-model/lib/publicApi/table/setTableCellShade.ts
```diff
@@ -19,7 +19,7 @@ export default function setTableCellShade(editor: IExperimentalContentModelEdito
         editor.addUndoSnapshot(
             () => {
                 editor.focus();
-                editor.setContentModel(model, fragment => editor.replaceNode(table, fragment));
+                editor.setContentModel(model!, fragment => editor.replaceNode(table!, fragment));
```
JiuqingSong, line 22: same

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -131,7 +135,7 @@ export default class Editor implements IEditor {
             this.core.plugins[i].dispose();
         }
 
-        this.core = null;
+        this.core = null!;
```
JiuqingSong, line 138: Actually, this.core can really be null, so we should allow it when declare the type.

Then in each place we use it, we can check and throw:
```ts
const Disposed = 'Editor is already disposed';

...

if (!this.core) { throw newError(Disposed); }

... // now you can use this.core
}
```

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -189,13 +193,19 @@ export default class Editor implements IEditor {
...
+                this.core.api.transformColor(
+                    this.core,
+                    toNode,
+                    true /*includeSelf*/,
+                    nodeReplaceCallback,
+                    ColorTransformDirection.LightToDark
+                );
+            } else {
+                nodeReplaceCallback();
+            }
```
JiuqingSong, line 208: Does this change the logic? In original code even transformColorForDarkMode is false, we still call transformColor()

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -236,7 +246,7 @@ export default class Editor implements IEditor {
                 );
             });
         } else {
-            return queryElements(this.core.contentDiv, selector, callback, scope, null);
+            return queryElements(this.core.contentDiv, selector, callback, scope, undefined);
```
JiuqingSong, line 249: add an inline comment to "undefined". e.g.

undefined /*some comment, normally the parameter name */

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -528,7 +551,7 @@ export default class Editor implements IEditor {
         nameOrMap: string | Record<string, DOMEventHandler>,
         handler?: DOMEventHandler
     ): () => void {
-        const eventsToMap = typeof nameOrMap == 'string' ? { [nameOrMap]: handler } : nameOrMap;
+        const eventsToMap = typeof nameOrMap == 'string' ? { [nameOrMap]: handler! } : nameOrMap;
```
JiuqingSong, line 554: Can handler really be null or undefined?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -712,7 +737,7 @@ export default class Editor implements IEditor {
      */
     public getContentSearcherOfCursor(event?: PluginEvent): IPositionContentSearcher {
         return cacheGetEventData(event, 'ContentSearcher', () => {
-            let range = this.getSelectionRange();
+            let range = this.getSelectionRange() ?? new Range();
```
JiuqingSong, line 740: Do not new Range

packages/roosterjs-editor-core/lib/editor/Editor.ts
ianeli1: As far as I can tell if the second parameter is null, transformColor only calls the callback and then returns. Would you like me to update the types in transformColor instead?

packages/roosterjs-editor-core/lib/editor/Editor.ts
ianeli1: According to Typescript, no, as the overload either doesnt have a string `nameOrMap` or both parameters are required. However one might still introduce a nullish handler by using ts-ignore or regular JS. I can add a runtime check just in case.

packages/roosterjs-editor-core/lib/editor/Editor.ts
ianeli1: Oops!

packages/roosterjs-editor-types/lib/interface/EditorCore.ts
```diff
@@ -157,7 +157,7 @@ export type GetSelectionRangeEx = (core: EditorCore) => SelectionRangeEx;
  * @param core The EditorCore objects
  * @param node The node to get style from
  */
-export type GetStyleBasedFormatState = (core: EditorCore, node: Node) => StyleBasedFormatState;
+export type GetStyleBasedFormatState = (core: EditorCore, node?: Node) => StyleBasedFormatState;
```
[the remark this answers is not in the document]
ianeli1: It doesn't, falls back to `{}`
https://github.com/microsoft/roosterjs/blob/b37276eb1bddcfbd6245c6cdad75ff8a030a4f7c/packages/roosterjs-editor-core/lib/coreApi/getStyleBasedFormatState.ts#L15-L19

packages/roosterjs-editor-core/lib/editor/Editor.ts
ianeli1: Hm, checking on each method may be a little too WET
image
I'm wondering if I should introduce a private `getCore` and run this check there, replacing `this.core` reads with `this.getCore()`

JiuqingSong: A private getCore() sounds good to me

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -127,11 +131,11 @@ export default class Editor implements IEditor {
      * Dispose this editor, dispose all plugins and custom data
      */
     public dispose(): void {
-        for (let i = this.core.plugins.length - 1; i >= 0; i--) {
-            this.core.plugins[i].dispose();
+        for (let i = this.getCore().plugins.length - 1; i >= 0; i--) {
```
JiuqingSong, line 134: For those cases where core is used multiple places in one function, you can save it to a local var first, then no need to core getCore() many times:

```ts
const core = this.getCore();
...
```

Same for other functions.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -93,12 +94,15 @@ export default class Editor implements IEditor {
...
-                arrayPush(plugins, options.plugins);
+                if (options.plugins) {
+                    arrayPush(plugins, options.plugins);
+                }
             } else {
                 plugins.push(corePlugins[name]);
             }
         });
 
+        options.zoomScale = options.zoomScale ?? 1;
```
JiuqingSong, line 105: Better not change value of the passed in object. We don't know if it will throw exception when set a value, or if it reused by the caller. So better save the value to a local variable, e.g. do it in line 106

JiuqingSong (approved):

PR #1231, opened 2022-08-31
Fix Drag and drop tests

Andres-CT98:
Changed use of exact pixels for movement tests

Made some changes to the helper to verify that the correct mouse type is used. Now this test checks two things, If the helper uses the correct controls and if they correctly respond to their assigned events


packages/roosterjs-editor-plugins/test/pluginUtils/DragAndDropHelperTest.ts
```diff
@@ -60,7 +60,7 @@ describe('DragAndDropHelper |', () => {
         dndHelper.dispose();
     });
 
-    xit('mouse movement', () => {
+    it('mouse movement', () => {
         // Arrange
         const target = document.getElementById(id);
         creteDnD(target, false);
```
JiuqingSong, line 66: typo: create

packages/roosterjs-editor-plugins/test/pluginUtils/DragAndDropHelperTest.ts
```diff
@@ -83,7 +83,8 @@ describe('DragAndDropHelper |', () => {
         simulateMouseEvent('mouseup', targetEnd);
 
         // Assert
-        expect(target?.style.top).toBe('50px');
+        let divMoved = target?.style.top != '0px';
```
JiuqingSong, line 86: I feel this check is too weak now. If the value is "0", or null or undefined, it can also pass the check.

And from line 68, you explicitly assign '50px' to targetEnd.style.top, so the value is not assigned by the code we are testing.

Andres-CT98: I can not think of a scenario where it is any of those things, CSS will force a value on it.

The value is assigned in simulateMouseEvent, by the movement.

JiuqingSong: The problem is, you have already set style.top to be 50px in line 68. So divMoved will always be true, even dev code is not working. That means the test code is not really testing what we want to test.

I tried to comment out from line 71 to 83, then divMoved is still true. But since I have commented out all act code, the test case should fail.

Andres-CT98: I forgot that I had not set in beforeEach the top and left to be 0px as a default.

Andres-CT98: When assigning targetEnd, it is just a reference to target. Do I still attempt to test movement or just testing if the handler attaches correctly?

JiuqingSong: targetEnd is same with target, any change to targetEnd will also change target, they are the same.

It depends on what you'd like to test. If you want to check whether a function is called, you can spyOn that function and check if it hasBeenCalled(); if you want to check whether some dev code get right result, you can check that result value. But here you just verify what you have assigned directly, it is not actually verifying the dev code.

Andres-CT98: Made some changes to the helper to verify that the correct mouse type is used. Now this test checks two things, If the helper uses the correct controls and if they correctly respond to their assigned events

JiuqingSong (approved):

PR #1244, opened 2022-09-05
[Enable Strict Mode Core Plugin Part 1] - Fix CopyPaste plugin to enable strict mode

juliaroldi:
Fix CopyPaste plugin to enable strict mode of the core plugin package.


packages/roosterjs-editor-core/lib/corePlugins/CopyPastePlugin.ts
```diff
@@ -60,7 +60,7 @@ export default class CopyPastePlugin implements PluginWithState<CopyPastePluginS
     initialize(editor: IEditor) {
         this.editor = editor;
         this.disposer = this.editor.addDomEventHandler({
-            paste: this.onPaste,
+            paste: e => this.onPaste(this.editor!, e),
```
JiuqingSong, line 63: hmm, although this.editor is not likely to be null when this function is called, but better still have a check. And you can do it inside this.onPaste

JiuqingSong (approved):

PR #1245, opened 2022-09-06
Fix Merged Table Selection.

BryanValverdeU:
When selecting Merged cells in the table, some cells are not selected. This PR adds functionality to select these cells:

Eg:
#1
Before
image

After:
image

#2
Before
image

After
image


JiuqingSong:
For  the second case, I tried in Word desktop, and it allows me to select a non-rectangle selection:

image


And I think Word's behavior is better, otherwise I have no way to only select cell 0 and 1 with out 2 and 3.

BryanValverdeU:
Seems like Word behavior is dependant on what cell is the firstSelected and which cell is the last selected.

For example:
If selection end is a merged cell does this:
MergeSel1

If selection starts from merged:
MergeSel2

Refactoring to this fix this.

packages/roosterjs-content-model/lib/modelToDom/context/ModelToDomContext.ts
```diff
@@ -58,6 +58,11 @@ export interface TableSelection {
      * Coordinate of last selected cell
      */
     lastCell: Coordinates;
+
+    /**
+     * Path of selection done by the user.
+     */
+    selectionPath: Coordinates[];
```
JiuqingSong, line 65: why need this here?

packages/roosterjs-editor-types/lib/interface/TableSelection.ts
```diff
@@ -13,4 +13,9 @@ export default interface TableSelection {
      * Last Cell selection x and y position in the table
      */
     lastCell: Coordinates;
+
+    /**
+     * Path of selection done by the user.
+     */
+    selectionPath: Coordinates[];
```
JiuqingSong, line 20: We already have firstCell/lastCell, why we still need selection path?

BryanValverdeU: firstCell and lastCell are not always the actual cell selected in the table since we normalize the x,y. This is the case when selection starts from top right and finishes in the left bottom, like this:

image

Normalized selection would be:
firstCell: 1,1
lastCell: 8,7.

But the selection was from
firstCell: 8,1
lastCell: 1,7

I was having some issues with that, so I added the selection Path so we can get the first and last cell that the user selected.  

Any feedback for a better implementation?

JiuqingSong: Can you help my understand why you need to know the first/last cell user selected?

BryanValverdeU: For some scenarios like this:
> For the second case, I tried in Word desktop, and it allows me to select a non-rectangle selection:
> 
> image
> 
> And I think Word's behavior is better, otherwise I have no way to only select cell 0 and 1 with out 2 and 3.

Word behavior depends on where the selection started as mentioned here: 

 
> For example: If selection end is a merged cell does this: MergeSel1 
> 
> If selection starts from merged: MergeSel2

PR #1250, opened 2022-09-08
Move normalizeBlockquote from normalizePlugin to Format API

BryanValverdeU:
Introduced a change in c515ffdbaf6a67584f36f60594e0fb1361afb31b to normalize blockquotes to allow user to indent when text is centered to the Normalize Plugin. Moving this feature to a better place in the setIndentation and setAlignment APIs


packages/roosterjs-editor-api/lib/format/setAlignment.ts
```diff
@@ -91,7 +92,15 @@ function alignText(editor: IEditor, alignment: Alignment | CompatibleAlignment)
         align = 'right';
     }
     execCommand(editor, command);
-    editor.queryElements('[align]', QueryScope.OnSelection, node => (node.style.textAlign = align));
+    const elements = editor.queryElements('[align]', QueryScope.OnSelection, node => {
+        node.style.textAlign = align;
+        normalizeBlockquote(node);
+    });
+
+    if (elements.length == 0) {
```
JiuqingSong, line 100: In what case this can happen?

BryanValverdeU: When selection is collapsed.

JiuqingSong (approved):

PR #1254, opened 2022-09-12
Add getVisibleViewport function and hide TableSelector when is not inside of the Visible Rect

BryanValverdeU:
Add getVisibleViewport function, add to the Editor Options a way to override the function.
Use the new Editor API to determine whether to show the Table Selector.


JiuqingSong:
Why we need this option? Is there a bug related?

And if we really need this option, can we just pass in some flag value, but no need to pass in a callback?

And when we know one kind of table editor is not enabled, can we just not create it, but no need to let the table editor itself knows too much about this detail? For example:

tableResizer = isTableResizeAllowed? new TableResizer() : null;

So that inside TableResizer, all elements must not be null.

packages/roosterjs-editor-types/lib/interface/EditorOptions.ts
```diff
@@ -127,4 +128,9 @@ export default interface EditorOptions {
      * @deprecated Use zoomScale instead
      */
     sizeTransformer?: SizeTransformer;
+
+    /**
+     * Retrieves the visible viewport of the Editor. The default viewport is the Rect of the scrollContainer.
+     */
+    getVisibleViewport?: () => Rect;
```
JiuqingSong, line 135: I feel with this function, we can deprecate IEditor.getRelativeDistanceToEditor() now since this one is more common and powerful.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableSelector.ts
```diff
@@ -84,3 +86,18 @@ function setSelectorDivPosition(context: DragAndDropContext, trigger: HTMLElemen
...
+function isSelectorInsideEditor(
+    editor: IEditor,
+    rect: Rect | null,
+    contentDiv?: EventTarget
+): boolean {
+    const visibleViewport = editor.getVisibleViewport();
+    if (contentDiv && safeInstanceOf(contentDiv, 'HTMLElement') && visibleViewport && rect) {
+        const containerRect = normalizeRect(contentDiv.getBoundingClientRect());
+
+        return containerRect.top <= rect.top && visibleViewport.top <= rect.top;
```
JiuqingSong, line 99: Do we need to check bottom?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -115,6 +117,9 @@ export default class Editor implements IEditor {
             trustedHTMLHandler: options.trustedHTMLHandler || ((html: string) => html),
             zoomScale: zoomScale,
             sizeTransformer: options.sizeTransformer || ((size: number) => size / zoomScale),
+            getVisibleViewport:
+                options.getVisibleViewport ||
+                (() => normalizeRect(this.getScrollContainer().getBoundingClientRect())!),
```
JiuqingSong, line 122: If scroll container is not the editor DIV, we need to calculate both size and get the intersected area.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableSelector.ts
BryanValverdeU: I think for Table Selector we only need to check if the top of the table is visible, renamed the function name from `isSelectorInsideEditor `to `isTableTopVisible ` since the name was confusing.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -115,6 +117,27 @@ export default class Editor implements IEditor {
             trustedHTMLHandler: options.trustedHTMLHandler || ((html: string) => html),
             zoomScale: zoomScale,
             sizeTransformer: options.sizeTransformer || ((size: number) => size / zoomScale),
+            getVisibleViewport:
+                options.getVisibleViewport ||
+                (() => {
```
JiuqingSong, line 122: Suggest put this function into roosterjs-editor-dom and export it, then when you override it, you can still call sthis function and make minor change to the result, but no need to override the whole function.

You can pass an array of elements and calculate the intersect area for all of them.

Question: if there is no intersect area, do we want to return an invalid react (e.g. top is larger than bottom), or just return null?

BryanValverdeU: Added the new function, I also added an additional parameter for Rects array, I think this would be useful in case we need to calculate or modify rect outside of the function.

JiuqingSong (approved):

Issue #1256, opened 2022-09-14
Incorrect marker on nested list


tshibley:
**Describe the bug**
The `list-style-type` attribute leads to incorrect markers when toggling list type of nested bullets. 

**To Reproduce**
1. Go to the [roosterjs dev site](https://microsoft.github.io/roosterjs/index.html)
2. Create an unordered list with a nested bullet. 
3. With your cursor at the beginning of an empty, nested bullet select "Numbered list" from the top bar
4. Observe that the marker type does not change. 

**Expected behavior**
The marker of the nested bullet should be changed to `a.`

**Screenshots**

**Device Information**
 - OS: mac
 - Browser: chrome

**Additional context**
If you inspect that html you can see that the list type of the inner list is set correctly, but the `list-style-type` attribute is over-riding the default behavior.

tshibley:
In `applyListStyles` which runs during the `writeBack` phase of list editing there is a check that only uses the outer most list type to set the type of the current bullet: https://github.com/microsoft/roosterjs/blob/0e9b92f7a0293064bce24049f08190dfeba42526/packages/roosterjs-editor-dom/lib/list/VListItem.ts#L264

I believe that should be changed to instead use the type of the inner most list relative to this list item

tshibley:
Potential fix: https://github.com/microsoft/roosterjs/pull/1258

PR #1258, opened 2022-09-14
fix: toggling inner list type

tshibley:
fixing a bug where list style type was applied to inner lists even when the type was toggled

https://github.com/microsoft/roosterjs/issues/1256

### DIFF
--- a/packages/roosterjs-editor-dom/lib/list/VListItem.ts
+++ b/packages/roosterjs-editor-dom/lib/list/VListItem.ts
@@ -318,7 +318,10 @@ export default class VListItem {
             //apply the styles of the current parent list to the new list
             if (this.getDeepChildIndex(originalRoot) == stackLength) {
                 const listStyleType = this.node.parentElement?.style.listStyleType;
-                if (listStyleType) {
+                if (
+                    listStyleType &&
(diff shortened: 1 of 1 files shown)


BryanValverdeU:
Thank you for the fix.
It seems like the issue is only happening when toggling the list, so I think we can add a new member property 
VListItem 'wasListTypeToggled' and update this member property to true in changeListType function.
https://github.com/microsoft/roosterjs/blob/0e9b92f7a0293064bce24049f08190dfeba42526/packages/roosterjs-editor-dom/lib/list/VListItem.ts#L223-L237

Then we can use the new property and if it is true, do not modify the list style type of the new list.

@JiuqingSong do you think adding this new member for this scenario is okay?

tshibley:
@BryanValverdeU Thanks for taking a look! 

I'm not sure I totally follow why we would need to create a special case here for the toggling vs. checking to make sure the tags match any time we writeback. Is there a time when we still want to override the list type with marker of a different type? (Such that an `ol` has a bullet marker for instance)

BryanValverdeU:
Got it, you are right. 
I was able to test the functionality and it is working fine.

Thanks for the fix

BryanValverdeU (approved):

tshibley:
Great, thanks!

PR #1260, opened 2022-09-14
Remove Table Selection on Delete Row/Columns

BryanValverdeU:
Remove Table selection when deleting rows or columns

Before
collapseSelB4

After
collapseSel


packages/roosterjs-content-model/lib/modelApi/selection/collapseTableSelection.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * @internal
+ */
+export function collapseTableSelection(
+    cells: ContentModelTableCell[][],
+    selection: TableSelectionCoordinates
+) {
+    const { firstCol, firstRow, lastCol, lastRow } = selection;
+
+    for (let row = 0; row < cells.length; row++) {
```
JiuqingSong, line 15: You just need to add a selection marker to the first selected cell if exist:

const cell =cells[firstRow]?.[firstCol];

if (cell) { addSegment(...}

BryanValverdeU: Ah yes, good point. 😅

JiuqingSong:
Please also add test case to make sure the selection result is correct.

JiuqingSong (approved):

PR #1262, opened 2022-09-15
SelectImage API

juliaroldi:
Add a new selectImage api. With this change is possible to select an image using editor select api, also it is possible to detect when an image is selected using the getSelectionRangeEx api. 
Also an interface playground was add to make it easier to debug selection inside the editor. 

apiPlay


packages/roosterjs-editor-types/lib/interface/SelectionRangeEx.ts
```diff
@@ -41,6 +41,23 @@ export interface TableSelectionRange
...
+        SelectionRangeTypes.ImageSelection | CompatibleSelectionRangeTypes.ImageSelection
+    > {
+    /**
+     * Selected Image
+     */
+    image: HTMLImageElement;
+    /**
+     * Id of the selected Image
+     */
+    imageId: string | undefined;
```
JiuqingSong, line 58: Do we need imageId here? I think we only need one of them

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -483,6 +498,27 @@ export default class Editor implements IEditor {
         return !!range && this.contains(range) && core.api.selectRange(core, range);
     }
 
+    private getImageSelection(
+        core: EditorCore,
+        arg: Range | NodePosition | Node | SelectionPath | HTMLTableElement | null
+    ) {
+        if (safeInstanceOf(arg, 'HTMLImageElement')) {
```
JiuqingSong, line 505: Also need to check if arg2 is undefined.

Because if I call select(image, PositionType.Before), it is still not an image selection

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -483,6 +498,27 @@ export default class Editor implements IEditor {
...
 
+    private getImageSelection(
+        core: EditorCore,
+        arg: Range | NodePosition | Node | SelectionPath | HTMLTableElement | null
+    ) {
+        if (safeInstanceOf(arg, 'HTMLImageElement')) {
+            const selection = core.api.selectImage(core, arg);
+            return selection;
+        }
+        if (arg && safeInstanceOf(arg, 'HTMLSpanElement')) {
```
JiuqingSong, line 509: This is specially for image edit plugin, right?

I feel we can change the way how image edit plugin show the border, then we can get rid of this part.

And once we remove this part, no need to have this private function, but just put everything inside select function

packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
```diff
@@ -152,6 +153,11 @@ export default class DOMEventPlugin implements PluginWithState<DOMEventPluginSta
             this.editor.select(this.state.selectionRange);
         }
 
+        const { image } = this.state.imageSelectionRange || {};
```
JiuqingSong, line 156: From line 150 to 160, we only need to execute one of these two.

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Select a image and save data of the selected range
+ * @param core The EditorCore object
+ * @param image Image to select
+ * @param wrapper Selected image wrapper
+ * @returns Selected image information
+ */
+export const selectImage: SelectImage = (
+    core: EditorCore,
+    image: HTMLImageElement | null,
+    wrapper?: HTMLSpanElement
```
JiuqingSong, line 15: same here

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
```diff
@@ -0,0 +1,36 @@
...
+) => {
+    const selectedImage = image
```
JiuqingSong, line 17: I think you need to ensure image has a unique id here

JiuqingSong: Similar with selectTable.
Also ensure content DIV have id. Then we can use id and CSS to show image border.

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
```diff
@@ -0,0 +1,36 @@
...
+        ? image
+        : wrapper
+        ? (document.querySelector(`img[id$="${wrapper.id}"]`) as HTMLImageElement)
+        : null;
+    if (selectedImage) {
+        const range = wrapper ? createRange(wrapper) : createRange(selectedImage);
+        core.api.selectRange(core, range);
```
JiuqingSong, line 24: hmm, for image selection, I think we don't really need to select the range, but just let the editor core remember image is selected which you have already done. For UI layer, we can use some CSS to show a selection border (so no need to do it from image edit plugin). We can discuss this part later.

packages/roosterjs-editor-core/lib/editor/Editor.ts
JiuqingSong: I think one problem here is how we can distinguish if we want an image selection, or a normal selection with just a IMG tag? Both of them have some scenario to use.

Maybe we need some additional parameter. Or, let this API support ImageSelectionRange (as well as TableSelectionRange) so that we can check the object type

demo/scripts/controls/sidePane/apiPlayground/getSelection/getSelectionPane.tsx
```diff
@@ -0,0 +1,245 @@
...
+        };
+    }
+
+    onPluginEvent(e: PluginEvent) {
+        if (e.eventType == PluginEventType.MouseUp && !this.state.manualSelect) {
+            this.getSelection();
+        }
+    }
+
+    private getSelection = () => {
```
JiuqingSong, line 49: name it "updateSelection"?

demo/scripts/controls/sidePane/apiPlayground/getSelection/getSelectionPane.tsx
```diff
@@ -0,0 +1,245 @@
...
+interface SelectionPaneState {
+    selection: SelectionRangeEx;
+    selectionMessage: string;
+    isImageSelectionOption: boolean;
+    manualSelect: boolean;
+}
+
+const styles = require('./getSelectionPane.scss');
+
+export default class GetSelectionPane extends React.Component<ApiPaneProps, SelectionPaneState> {
```
JiuqingSong, line 20: this is nice!

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
juliaroldi: The id is incremental and reseted after dispose the editor? Why do we need use the id in the contentDiv?

packages/roosterjs-editor-core/lib/editor/Editor.ts
juliaroldi: If we select something else and image it will be recognize as normalSelection, because the arg will be the parent node the image and something else.

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
juliaroldi: I think we should add the selection, because it can helps with operations such copy/paste images.

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
BryanValverdeU: An example could be when re-adding the content with an image that already contains the ID.
Let's say you select an image, add the ID with number 0, if you copy and paste or set the content with that image to another editor where the incremental is still 0 there would be a duplicated ID

PR #1265, opened 2022-09-15
Table Pasting

Andres-CT98:
### Description:
When pasting a table inside another table, it now merges them both using the place the cursor is located as a pivot.

### Fix

Used VTable to modify the existing table to its new expected size and then inserted the new content of the cells. Merged cells from the existing table are respected and the new cells format is inherited from the existing table.

Changed the insertTable function to insert new tables as HTMLTables instead of DocumentFragments.

### Cases tested:

Paste table B on the cell containing the 9 in table A.

Table A:
<table><tr><td>1</td><td>2</td><td>3</td></tr><tr><td>4</td><td>5</td><td>6</td></tr><tr><td>7</td><td>8</td><td>9</td></tr></table>

Table B
<table><tr><td>A</td><td>B</td><td>C</td></tr><tr><td>D</td><td>E</td><td>F</td></tr><tr><td>G</td><td>H</td><td>I</td></tr></table>

**Before:** Pasting a table inside a table resulted in a nested table.

<table><tr><td>1</td><td>2</td><td>3</td></tr><tr><td>4</td><td>5</td><td>6</td></tr><tr><td>7</td><td>8</td><td>9<table><tr><td>A</td><td>B</td><td>C</td></tr><tr><td>D</td><td>E</td><td>F</td></tr><tr><td>G</td><td>H</td><td>I</td></tr></table></td></tr></table>

**After:** Pasting a table inside a table results in a new merged table.

<table><tr><td>1</td><td>2</td><td>3</td><td></td><td></td></tr><tr><td>4</td><td>5</td><td>6</td><td></td><td></td></tr><tr><td>7</td><td>8</td><td>A</td><td>B</td><td>C</td></tr><tr><td></td><td></td><td>D</td><td>E</td><td>F</td></tr><tr><td></td><td></td><td>G</td><td>H</td><td>I</td></tr></table>

### Warnings and implications

- Pasting nested tables is still possible, it requires copying a table along an empty line or any other content besides the table. It is also possible to insert a new one from the ribbon.
- Making a table selection will not move the pasting pivot point, pasting will occur where the cursor is located.
- There are some issues with data loss if pasting over merged cells.


packages/roosterjs-editor-core/lib/coreApi/insertNode.ts
```diff
@@ -157,21 +157,26 @@ export const insertNode: InsertNode = (core: EditorCore, node: Node, option: Ins
...
-                    let nodeForCursor =
-                        node.nodeType == NodeType.DocumentFragment ? node.lastChild : node;
-                    range = createRange(pos);
-                    range.insertNode(node);
-                    if (option.updateCursor && nodeForCursor) {
-                        rangeToRestore = createRange(
-                            new Position(nodeForCursor, PositionType.After).normalize()
-                        );
-                    }
+                    if (pos) {
```
JiuqingSong, line 160: can you help me understand why we need to change this file?

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -161,11 +160,19 @@ function adjustInsertPositionForStructuredNode(
             );
         }
     } else if (tag == 'TABLE' && trNode) {
+        pasteTable(root, nodeToInsert, position, range);
+        console.log('#');
+        debugger;
+        position = null;
         // When inserting a table into a table, if these tables have the same column count, and
```
JiuqingSong, line 167: With this change, do we still need this behavior?

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -161,11 +160,19 @@ function adjustInsertPositionForStructuredNode(
             );
         }
     } else if (tag == 'TABLE' && trNode) {
+        pasteTable(root, nodeToInsert, position, range);
```
JiuqingSong, line 163: Instead of calling pasteTable() here, let's just add a new "adjustInsertPosition" function, say "adjustInsertPositionForTable".

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -298,40 +306,34 @@ function adjustInsertPositionForTable(
     position: NodePosition,
     range: Range
 ): NodePosition {
-    if (
```
JiuqingSong, line 301: why remove?

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,71 @@
...
+ *
+ */
+export default function pasteTable(
+    root: HTMLElement,
+    nodeToInsert: Node,
+    position: NodePosition,
+    range: Range
+) {
+    if (
+        (nodeToInsert.childNodes.length >= 1 &&
```
JiuqingSong, line 19: If length > 1, it means there will be other content than table to be pasted. 

I tried in Word in that case, it will directly insert whatever we copied, to have a nested table. I think we can follow this behavior.

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,71 @@
...
+            }
+        }
+
+        currentcell.writeBack();
+
+        return null as Node;
+    }
+}
+
+void function addEmpty(tr: HTMLTableRowElement, prevtr: HTMLTableRowElement) {
```
JiuqingSong, line 69: lets follow typescript syntax

function funcName(..): void

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,74 @@
...
+import toArray from '../jsUtils/toArray';
+import VTable from '../table/VTable';
+import { NodePosition, NodeType, TableOperation } from 'roosterjs-editor-types';
+
+/**
+ * @internal
+ * @param root The table to paste into
+ * @param nodeToInsert A Node containing the table to be inserted
+ * @param position The position to paste the table
+ */
```
juliaroldi, line 12: Please create test cases for this function

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -160,28 +159,11 @@ function adjustInsertPositionForStructuredNode(
...
-                getTagOfNode(rootNodeToInsert!.firstChild) == 'TBODY' &&
-                !rootNodeToInsert!.firstChild?.nextSibling
-            ) {
-                unwrap(rootNodeToInsert!.firstChild!);
-            }
-            unwrap(rootNodeToInsert!);
-            position = new Position(trNode, PositionType.After);
-        }
+    if (tag == 'TABLE' && trNode) {
+        pasteTable(root, nodeToInsert, position, range);
```
JiuqingSong, line 165: you can pass in rootNodeToInsert so no need to check if it is TABLE again in pasteTable()

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,74 @@
...
+            getTagOfNode(nodeToInsert.childNodes[1]) == 'BR')
+    ) {
+        let rootNodeToInsert: Node | null = nodeToInsert;
+
+        if (rootNodeToInsert.nodeType == NodeType.DocumentFragment) {
+            let rootNodes = toArray(rootNodeToInsert.childNodes).filter(
+                (n: ChildNode) => getTagOfNode(n) != 'BR'
+            );
+            rootNodeToInsert = rootNodes.length == 1 ? rootNodes[0] : null;
+        }
```
JiuqingSong, line 33: Already handled

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,74 @@
...
+
+        let tdNode = findClosestElementAncestor(position.node, root, 'TD,TH');
```
JiuqingSong, line 35: this is for TD in existing document, right? Let's rename it to make it easy to understand

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,74 @@
...
+
+        // This is the table on the clipboard
+        let newTable = new VTable(<HTMLTableElement>rootNodeToInsert);
```
JiuqingSong, line 38: As I said, if we only accept HTMLTableElement in this function, no need to do type cast here

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,74 @@
...
+        // This table is already on the editor
+        let currentCell = new VTable(<HTMLTableCellElement>tdNode);
```
JiuqingSong, line 40: This will give you a TABLE but not just a Cell, so rename it.

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,74 @@
...
+
+        let currentRow = currentCell.row!;
+        let currentCol = currentCell.col!;
+
+        let rows = currentRow + newTable.cells?.length!;
+        let columns = currentCol + newTable.cells?.[0].length!;
```
JiuqingSong, line 46: Better use ".length ?? 0" to make sure it is a number

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        for (let i = currentRow; i < rows; i++) {
+            for (let j = currentCol; j < columns; j++) {
+                let cell = currentCell.getCell(i, j);
+                let newcell = newTable.getTd(i - currentRow, j - currentCol);
+                if (!cell.spanAbove && !cell.spanLeft) {
+                    if (newcell!.hasChildNodes()) {
+                        // Prevent double <br>
+                        cell.td?.removeChild(cell.td.lastChild!);
```
JiuqingSong, line 63: How can you guarantee the last child is a BR?

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,74 @@
...
+
+        while (currentCell.cells![0].length < columns) {
+            currentCell.edit(TableOperation.InsertRight);
+        }
+
+        for (let i = currentRow; i < rows; i++) {
+            for (let j = currentCol; j < columns; j++) {
+                let cell = currentCell.getCell(i, j);
+                let newcell = newTable.getTd(i - currentRow, j - currentCol);
+                if (!cell.spanAbove && !cell.spanLeft) {
```
JiuqingSong, line 60: Why we check spanAbove and spanLeft here? I think we can still do the replacement even if it is spanLeft or spanAbove?

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,74 @@
...
+                    if (newcell!.hasChildNodes()) {
```
JiuqingSong, line 61: why do this check? And what if newCell is really null?

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,74 @@
...
+                        // Prevent double <br>
+                        cell.td?.removeChild(cell.td.lastChild!);
+                        newcell!.childNodes.forEach(child => {
```
JiuqingSong, line 64: I think an easier way is just replace the whole VCell object

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
```diff
@@ -0,0 +1,74 @@
...
+    nodeToInsert: Node,
+    position: NodePosition,
+    range: Range
+) {
+    if (
+        (nodeToInsert.childNodes.length == 1 &&
+            getTagOfNode(nodeToInsert.childNodes[0]) == 'TABLE') ||
+        (nodeToInsert.childNodes.length == 2 &&
+            getTagOfNode(nodeToInsert.childNodes[0]) == 'TABLE' &&
+            getTagOfNode(nodeToInsert.childNodes[1]) == 'BR')
```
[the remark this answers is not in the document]
Andres-CT98: Where is this handled exactly?

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
Andres-CT98: This creates formatting problems

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -160,28 +159,11 @@ function adjustInsertPositionForStructuredNode(
...
+        return null;
```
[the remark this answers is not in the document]
Andres-CT98: The behaviour is insertNode is kept consistent. The two options I see is add a check for non null position before inserting or add a check for table inside table insertion before inserting. I do not see how the second one is more efficient.

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
Andres-CT98: To skip the replacement as it can be more costly.

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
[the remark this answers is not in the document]
JiuqingSong: As a general function, insertNode() should know as few details of the adjusting result as possible. So if we always return a valid position, it doesn't need to check whether it is null and always do the same handling. Then we keep all those details in a smaller scope, such as in what case insertNode() need to insert.

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
JiuqingSong: How cost it is?

A simple loop here won't be a problem.

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
JiuqingSong: Can you elaborate more about the formatting problems?

Andres-CT98: Currently the cell has the correct format it needs to have. By replacing the whole object I would be overwriting the format. The only thing that needs to change is the content.

JiuqingSong: In that case, you can call moveChildNodes() here

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
JiuqingSong: And what if the source table cell has spanLeft/spanAbove but the target cell doesn't have?

Andres-CT98: Based on Word online, the merged cells of the source table are not kept.

JiuqingSong: Right, so we need to follow this behavior, and do not skip cells that has spanLeft/spanAbove since that pasted table cell may not be in that case

Andres-CT98: I have to skip them if they have no content. I can just check for .td to exist instead of the spans, but I have to skip VCells that are empty.

JiuqingSong: In those cases, source table cell won't have TD, but target table will have TD. If you skip the cells that don't have TD, the content from target table will be lost.

Since we are replacing everything from target table to source table, you don't need to check if a table cell is empty, but just move whatever we have from target table to source table. If it is empty, just move empty to source table.

Andres-CT98: I may be confused on the terms of source and target, but it is fine if content you intend to paste gets lost due to merged cells.

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
JiuqingSong: what do you mean by "checking twice"?

You just need to simply call moveChildNodes() to move anything we have from target TD to source TD, and replace anything we have there.

Andres-CT98: I get a *Type 'null' is not assignable to type 'Node | undefined'.* warning if I do not add some check or add ! at the end of each node.

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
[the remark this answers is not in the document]
Andres-CT98: How do I keep them on a smaller scope without affecting the whole behaviour? I still do not understand why it is more efficient.

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
JiuqingSong: Because the source TD can really be null. So that you must not add "!" otherwise there will be runtime exception.

We should only add "!" when we are pretty sure the object we are using MUST NOT be null and the compiler just not smart enough to detect that. Otherwise we should do null check.

And in this case, you need to create a TD element since TD doesn't exist in merged cells. And that is exactly the case strict mode can help us to prevent.

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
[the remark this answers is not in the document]
JiuqingSong: Let's see two examples:

example 1:
```ts
function foo(): Position | null {
  if (...) {
     return {...};
  } else {
     return null;
  }
}

const result = foo();
if (result) {
  // do something with the result
} else {
  // result is null, do something else
}
```

Example 2:
```ts
function foo(): Position {
  if (...) {
     return {...};
  } else {
     return { ... } ; // Still return a valid position, so when call the function, no need to know the detail that when return object, and when return null
  }
}

const result = foo();
// Here we know result must not be null, so we can always do the samething to result. function foo() hides the implementation details so caller doesn't need to care it.
```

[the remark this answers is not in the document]
Andres-CT98: Sure I understand that part, and I thank you for the explanation, but what I do not understand is why the alternative is better or even how you are suggesting I implement the alternative.

If I do return a valid position there will be an unwanted insertion, I need to stop that insert or modify what is inserted to be nothing.

packages/roosterjs-editor-dom/lib/edit/pasteTable.ts
Andres-CT98: What I am asking is moveChildNodes() makes sure that the nodes are not null and do have children, so even if they are it is already handled there.

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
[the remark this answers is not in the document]
JiuqingSong: When the node to insert is an empty DocumentFragment, the insert operation will be an no-op, so nothing special need to be done there.

About "why the alternative is better", it is because if a function can always return a valid value and can be processed used by one single way, we need to do so. Otherwise each place this function is called, we always need to handle the result in "if ... else". Sometimes that necessary, but in this case, since we can handle it in one single way, why still give a "null" result and let the caller handle it?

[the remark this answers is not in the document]
JiuqingSong: If you are familiar with C++ (or any other OO language), you may be familiar with virtual function. A virtual function can have different implementation in each sub class. As a caller, we don't need to care what implementation it is, but just always call it. That means, with the same interface, we can always do the same handling. Here it is the same.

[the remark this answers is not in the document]
Andres-CT98: This tread started with the suggestion to return a valid position and return an empty fragment to insert. I understand why you made that suggestion and agree it would be better, but I still do not know how it is possible to make the fragment empty, with the current way insertNode works, in a manner than does not use position. As position is the only thing that adjustInsertPosition returns and there is where the pasting is handled.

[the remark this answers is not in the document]
Andres-CT98: I used moveChildNodes again to remove the children of the node to insert. I honestly did not know that nodeToInsert was a reference and not a copy to the node to insert.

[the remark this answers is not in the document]
JiuqingSong: To make fragment empty, you just need to call moveChildNodes() to clear it. And actually you need to do that even you return null in order to keep consistent with other cases. After insert, we always expect the fragment to be empty

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -25,7 +26,6 @@ import {
     PositionType,
     QueryScope,
 } from 'roosterjs-editor-types';
-
```
JiuqingSong, line 28: do not remove this empty line

packages/roosterjs-editor-dom/lib/index.ts
```diff
@@ -60,6 +60,7 @@ export { default as VListChain } from './list/VListChain';
 export { default as setListItemStyle } from './list/setListItemStyle';
 export { getTableFormatInfo } from './table/tableFormatInfo';
 export { saveTableCellMetadata } from './table/tableCellInfo';
+export { default as pasteTable } from './table/pasteTable';
```
JiuqingSong, line 63: Is is used anywhere outside this package? If not, no need to export it

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -354,7 +337,9 @@ export default function adjustInsertPositionBySteps(
     range: Range
 ): NodePosition {
     adjustSteps.forEach(handler => {
-        position = handler(root, nodeToInsert, position, range);
+        if (position) {
```
JiuqingSong, line 340: no need to check here

packages/roosterjs-editor-dom/lib/table/pasteTable.ts
```diff
@@ -0,0 +1,57 @@
...
+    rootNodeToInsert: HTMLTableElement,
+    position: NodePosition,
+    range: Range
+) {
+    let currentTd = findClosestElementAncestor(position.node, root, 'TD,TH');
+
+    // This is the table on the clipboard
+    let newTable = new VTable(rootNodeToInsert);
+    // This table is already on the editor
+    let currentTable = new VTable(<HTMLTableCellElement>currentTd);
```
JiuqingSong, line 23: do we need to do null check here?

I think there is already code outside this function to get the TD node, you can just pass it in so that no need to do null check and type cast here

packages/roosterjs-editor-dom/lib/table/pasteTable.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    // Create final table
+    for (let i = cursorRow; i < rows; i++) {
+        for (let j = cursorCol; j < columns; j++) {
+            let cell = currentTable.getCell(i, j);
+            let newCell = newTable.getTd(i - cursorRow, j - cursorCol);
+            if (cell.td) {
+                moveChildNodes(cell.td, newCell!);
+            }
```
JiuqingSong, line 52: else?

If there is no TD, we should create one (or just bring the one from new table)

packages/roosterjs-editor-dom/lib/index.ts
Andres-CT98: Is it not standard to export everything?

packages/roosterjs-editor-dom/lib/table/pasteTable.ts
Andres-CT98: Why? There are cases where the new table does not have empty td to spare.

packages/roosterjs-editor-dom/lib/index.ts
JiuqingSong: Anything exported from index means it is expected to be used externally. But if some function is only used internally, and we don't expect it is directly used outside roosterjs, we don't need to export it.

Andres-CT98: So do I add @internal to the function?

JiuqingSong: Correct

packages/roosterjs-editor-dom/lib/table/pasteTable.ts
JiuqingSong: If it returns null, then you should not create VTable, right?

But actually it is already check in adjustInsertPosition.ts, Line 132, so you just need to pass the result from there into this function, so no need to call findClosestElementAncestor again and no need to do null check.

JiuqingSong: @Andres-CT98 any update to this comment?

Andres-CT98: I removed the repeated call to findClosestElementAncestor and just pass the firsts result.

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -160,28 +162,12 @@ function adjustInsertPositionForStructuredNode(
...
-            ) {
-                unwrap(rootNodeToInsert!.firstChild!);
-            }
-            unwrap(rootNodeToInsert!);
-            position = new Position(trNode, PositionType.After);
-        }
+    if (isFragment && tag == 'TABLE' && tdNode) {
+        pasteTable(tdNode, <HTMLTableElement>rootNodeToInsert, position, range);
+        moveChildNodes(nodeToInsert);
+        return position;
```
JiuqingSong, line 170: This line can be removed

packages/roosterjs-editor-dom/lib/table/pasteTable.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Pastes a table inside another, modifying the original to create a merged one
+ * @param currentTd The cell where the cursor is in the table to paste into
+ * @param rootNodeToInsert A Node containing the table to be inserted
+ * @param position The position to paste the table
+ * @param range The selected range of the table
+ *
+ * Position and range are here for when table selection allows to move pivot point
+ */
+export default function pasteTable(
+    currentTd: HTMLElement,
```
JiuqingSong, line 16: Change the parameter type to be HTMLTableCellElement so you can avoid type casting in line 24

packages/roosterjs-editor-dom/lib/edit/adjustInsertPosition.ts
```diff
@@ -160,28 +162,12 @@ function adjustInsertPositionForStructuredNode(
...
-            isPositionAtBeginningOf(position, tdNode)
-        ) {
-            if (
-                getTagOfNode(rootNodeToInsert!.firstChild) == 'TBODY' &&
-                !rootNodeToInsert!.firstChild?.nextSibling
-            ) {
-                unwrap(rootNodeToInsert!.firstChild!);
-            }
-            unwrap(rootNodeToInsert!);
-            position = new Position(trNode, PositionType.After);
```
JiuqingSong, line 183: I think we still need these logic if isFragment is false, otherwise it will regress the existing behavior of insert table

JiuqingSong: Try merge it into the new code you added to avoid duplicated code.

Andres-CT98: Can you explain how it regresses insert table? It passes al tests. The removed code is already covered with the new paste table.

JiuqingSong: Ok, for now it may not be a problem. But when doing part 2 and change insertTable to not using a fragment, it will be a problem. I'm ok to fix it together in part 2

JiuqingSong: But what if I insert another table that has less column count than the existing one?

packages/roosterjs-editor-api/lib/table/insertTable.ts
```diff
@@ -46,7 +44,7 @@ export default function insertTable(
             let vtable = new VTable(table);
             vtable.applyFormat(format);
             vtable.writeBack();
-            editor.insertNode(fragment);
+            editor.insertNode(table);
```
JiuqingSong, line 47: now with this change, the insert table behavior is really regressed.

Repro:
1. insert a 3*4 table
    
image
2. Go to row 3, colomn 1 (the first cell of the last row)
4. insert another 3*4 table

Original behavior (expected):
The 3*4 table is expanded to be a 6*4 table
image


New behavior:
I got a 3*4 table nested inside the cell

image

JiuqingSong: Discussed offline and we can accept this new behavior.

JiuqingSong (approved):

Issue #1270, opened 2022-09-20
Toggling type of nested list creates multiple root lists


tshibley:
**Describe the bug**
When toggling the list type for a nested list, we end up with a messy state in the dom + multiple root nodes. 

**To Reproduce**
See the video below for an example of how to reproduce: 

https://user-images.githubusercontent.com/15166410/191145215-4a490d03-091c-4b31-b4d1-c084f97a7cef.mov

As you can see the list gets split into two, with one `ol` root and one `ul` root. This also creates a display issue in the DOM, since we add a small vertical separation between the lists. 

**Expected behavior**
I would expect theses lists to have one root node after toggling the type of the whole list.

**Screenshots**

**Device Information**
 - OS: Mac
 - Browser Chrome

tshibley:
I believe the issue is caused by the way the core `VListItem` type works - which will potentially make it tricky to fix.

Since we store the list types of an item as an array, when we call `changeListType` on an item we only replace the closest list type to the item. So then when we go to `writeBack` those changes into the DOM, the earlier list types in our array no longer match the previous bullets, so we create new lists. 

When it comes to nesting lists, I don't believe the type of the previous levels matters at all for display? So it might be the case that we could ignore mismatches in the types of lists that aren't the one directly surrounding the bullet when we call `writeBack`? 

I think this would generally leave us with a cleaner/more correct DOM as an end user would expect it, but I'm a little worried that it would create other side effects, since it would make `writeBack` a potentially lossy operation - especially on lists that had been pasted into the composer for instanct.

JiuqingSong:
This is an interesting finding. For now I don't know if there is any side effect, but I think we'd better put this change under an ExperimentalFeature since it changes the original behavior. Then you can enable the feature in your project. After some time if we didn't see any issue, we can turn this Experimental feature always on.

PR #1275, opened 2022-09-20
Use elements from list stack even if they don't match 

tshibley:
Fixes: https://github.com/microsoft/roosterjs/issues/1270

Previously, we were only using elements from the list stack if they matched the type from the listTypes array. This PR changes the approach to only ensuring that the list element surrounding this list item is correct. 

This approach creates much cleaner HTML in the DOM - where previously we would end up with split lists because a grandparent bullet type didn't match, we know have a much more "semantically normal" list. It works as expected from a user standpoint still, since the only list element that matters from a display standpoint is the one immediately surrounding each item - the rest are just needed for defining the correct indentation.


tshibley:
Curious if anyone has thoughts as to this approach! It creates _much_ cleaner html for lists, and fixes some bugs like that splitting issue I've attached. It seems better than the current approach to me, but I'm curious if anyone knows of a reason we would want to strictly maintain the types from the list types array? 

If this approach seems reasonable I can fix the test failures here before merging, but wanted to know if anyone had thoughts before doing that

tshibley:
Per discussion in https://github.com/microsoft/roosterjs/issues/1270 I moved this change behind a flag (and made the default false, so the current behavior will continue).

packages/roosterjs-editor-api/lib/utils/commitListChains.ts
```diff
@@ -11,7 +11,9 @@ export default function commitListChains(editor: IEditor, chains: VListChain[])
         const range = editor.getSelectionRange();
         const start = range && Position.getStart(range);
         const end = range && Position.getEnd(range);
-        chains.forEach(chain => chain.commit());
+        chains.forEach(chain =>
+            chain.commit(editor.isFeatureEnabled(ExperimentalFeatures.ReuseAllAncestorListElements))
```
JiuqingSong, line 15: let's move the call to `editor.isFeatureEnabled(ExperimentalFeatures.ReuseAllAncestorListElements)` out of the loop

JiuqingSong (approved):

PR #1276, opened 2022-09-20
Content Model List step 1: add types and demo site views

JiuqingSong:
This is the first step of Content Model List.

Add list related types (listItem, quote, formats)
Add list and quote related View components in demo site.


BryanValverdeU (approved):
As a side note, I noticed that the Content Model Ribbon is not opening anymore. But it is not related to this PR. As I am able to repro in the live demo site.

PR #1282, opened 2022-09-21
Fix invisible caret in dark mode for Safari

cyanzhong:
This PR fixes the issue where caret is not visible in dark mode running on Safari, which occurs when copying and pasting text.

This is a WebKit specific bug, it can only be reproduced on WebKit powered browsers, such as Safari and WKWebView on Mac. The root cause is that when copy content in contentEditable, WebKit always sets caret-color to its text color, this is not roosterjs related.

However, in roosterjs' `onCutCopy` callback, we create a light mode only temp div to do the copy paste:

This makes the caret-color black (since it's light mode), unfortunately, when we transform colors to dark mode, we don't take caret-color into account, hence the caret becomes invisible.
 
I think there're two ways to solve this, we can either support caret-color in `transformColor` (a bit overkill), or strip caret-color in `createPasteFragment` (what this PR does).

Besides, we can also override caret-color globally, but it's not preferred because it changes the default behavior, the caret color should align with the text instead of either black or white, i.e. red text shows red caret.

### DIFF
--- a/packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
+++ b/packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
@@ -1,6 +1,7 @@
 import {
     applyFormat,
     applyTextStyle,
+    Browser,
     createDefaultHtmlSanitizerOptions,
     getInheritableStyles,
     getTagOfNode,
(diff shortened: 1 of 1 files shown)


JiuqingSong:
Thanks to make the fix. To make the fix simpler and more robust, I have opened #1283.

PR #1284, opened 2022-09-21
[FHL] Part 1: Image Selection - selectImage core api

juliaroldi:
Add a new selectImage api experimental feature. With this change is possible to select an image using editor select api, also it is possible to detect when an image is selected using the getSelectionRangeEx api.
Also an interface playground was add to make it easier to debug selection inside the editor.


packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -174,7 +174,6 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
 
             case PluginEventType.KeyDown:
                 this.eventHandledOnKeyDown = false;
-
```
JiuqingSong, line 177: no need to change here

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -452,22 +452,42 @@ export default class Editor implements IEditor {
...
         arg2?: NodePosition | number | PositionType | TableSelection,
         arg3?: Node,
         arg4?: number | PositionType
     ): boolean {
         const core = this.getCore();
+
         if (arg1 && 'rows' in arg1) {
             const selection = core.api.selectTable(core, arg1, <TableSelection>arg2);
             core.domEvent.tableSelectionRange = selection;
-
```
JiuqingSong, line 464: no need to change here

JiuqingSong (approved):

PR #1286, opened 2022-09-21
[FHL] Part 2 - Image Selection - Image Selection Plugin

juliaroldi:
Add Image Selection Plugin, this plugin is responsible for detect if a click was made above an image. If an image is selected, the selecImage core api will be triggered and will add border to the selected image. 

selectImagePlugin


packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -0,0 +1,73 @@
...
+    PluginEventType,
+    SelectionRangeTypes,
+} from 'roosterjs-editor-types';
+
+/**
+ * Detect image selection and help highlight the image
+ */
+export default class ImageSelection implements EditorPlugin {
+    private editor: IEditor | null = null;
+    private image: HTMLImageElement | null = null;
```
JiuqingSong, line 15: do we need to cache the image here? Can we get it (its id) using getSelectionRangeEx()?

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
```diff
@@ -21,3 +30,38 @@ export const selectImage: SelectImage = (image: HTMLImageElement | null) => {
 
     return null;
 };
+
+const select = (core: EditorCore, image: HTMLImageElement) => {
```
JiuqingSong, line 34: Is there any duplicated code with table selection? we can move them out into dom package under style folder

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
```diff
@@ -21,3 +30,38 @@ export const selectImage: SelectImage = (image: HTMLImageElement | null) => {
...
+};
+
+const buildBorderCSS = (core: EditorCore, imageId: string): string => {
+    const borderColor = core.imageSelectionBorderColor || '#DB626C';
+    return (
+        '#' +
+        core.contentDiv.id +
+        ' #' +
+        imageId +
+        ' { margin: -2px; border: 2px solid' +
```
JiuqingSong, line 55: question: do we need to add "!important" to the styles in case the image already has some conflict style?

(same for table selection)

packages/roosterjs-editor-core/lib/coreApi/ensureUniqueId.ts
```diff
@@ -0,0 +1,27 @@
...
+        //Ensure that there are no elements with the same ID
+        let element = getElement();
+        while (element) {
+            cont++;
+            element = getElement();
+        }
+
+        el.id = idPrefix + cont;
+    } else {
+        const elements = doc.querySelectorAll(`#${el.id}`);
```
JiuqingSong, line 21: I know this is moved from table selection, but let's refine it. It may not be a good idea to call the function recursively. So we can do:

```
if (has id) {
  if (id is not unique) {
     remove id;
  }
}
if (not have id) {
   apply an id
}
```

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
juliaroldi: I noticed table selection already has "!important". I also added to image selection.

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
juliaroldi: I added this duplicated code as function under utils folder of the core, it add styles, but only for selection and it uses editor core object.

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
BryanValverdeU: If the inline style contains !important will have higher priority than the head style, I think in table selection I added a function to remove the !important in case it is present in the inline style.

packages/roosterjs-editor-core/lib/coreApi/ensureUniqueId.ts
juliaroldi: I fixed and changed the name to addUniqueId, since the function is responsible to add the id and not only ensure that is unique.

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
juliaroldi: Instead of saving the image, I'm saving only the id.

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
juliaroldi: I abstract the code for removeImportant function and made a new one so we can use it in both selectImage and selectTable.

PR #1287, opened 2022-09-22
Enable strict mode for "roosterjs-editor-core/lib/corePlugins"

ianeli1:
Enables strict mode for all files under `roosterjs-editor-core/lib/corePlugins/` and fixes all required changes


packages/roosterjs-editor-core/lib/coreApi/transformColor.ts
```diff
@@ -58,7 +58,7 @@ export const transformColor: TransformColor = (
     if (direction == ColorTransformDirection.DarkToLight) {
         transformToLightMode(elements);
     } else if (core.lifecycle.onExternalContentTransform) {
-        elements.forEach(element => core.lifecycle.onExternalContentTransform(element));
+        elements.forEach(element => core.lifecycle.onExternalContentTransform?.(element));
```
JiuqingSong, line 61: We already checked null for core.lifecycle.onExternalContentTransform on line 60, so we can add "!" here if compiler still complains

packages/roosterjs-editor-core/lib/corePlugins/CopyPastePlugin.ts
```diff
@@ -150,22 +152,25 @@ export default class CopyPastePlugin implements PluginWithState<CopyPastePluginS
...
 
     private onPaste = (event: Event) => {
-        let range: Range;
+        let range: Range | null = null;
         if (this.editor) {
+            const editor = this.editor;
             extractClipboardEvent(
                 event as ClipboardEvent,
-                clipboardData => this.editor.paste(clipboardData),
+                clipboardData => this.editor?.paste(clipboardData),
```
JiuqingSong, line 160: use "editor" instead

packages/roosterjs-editor-types/lib/interface/CorePlugins.ts
```diff
@@ -44,12 +44,6 @@ export default interface CorePlugins {
      */
     readonly mouseUp: EditorPlugin;
 
-    /**
-     * @deprecated after Firefox update
```
JiuqingSong, line 48: don't remove this one. We will remove it when we do major version bump.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -740,7 +740,7 @@ export default class Editor implements IEditor {
      * @returns Default format object of this editor
      */
     public getDefaultFormat(): DefaultFormat {
-        return this.getCore().lifecycle.defaultFormat;
+        return this.getCore().lifecycle.defaultFormat!;
```
JiuqingSong, line 743: If defaultFormat can really be null, we should allow this function returns null, or return an empty object "{}"

packages/roosterjs-editor-core/lib/corePlugins/createCorePlugins.ts
```diff
@@ -35,7 +35,6 @@ export default function createCorePlugins(
         edit: map.edit || new EditPlugin(),
         pendingFormatState: map.pendingFormatState || new PendingFormatStatePlugin(),
         _placeholder: null,
-        typeAfterLink: null, //deprecated after firefox update
```
JiuqingSong, line 38: do not remove

packages/roosterjs-editor-core/lib/corePlugins/TypeInContainerPlugin.ts
```diff
@@ -57,12 +57,15 @@ export default class TypeInContainerPlugin implements EditorPlugin {
...
+                findClosestElementAncestor(range.startContainer, undefined /* root */, '[style]');
+
             if (
                 !range ||
                 (!this.isRangeEmpty(range) &&
-                    this.editor.contains(
-                        findClosestElementAncestor(range.startContainer, null /* root */, '[style]')
-                    ))
+                    styledAncestor &&
+                    this.editor.contains(styledAncestor))
```
JiuqingSong, line 68: I think we should allow passing in null for contains(), so no need to check it every where.

Same for editor.select()

When pass in null, just do nothing/return false.

packages/roosterjs-editor-core/lib/corePlugins/LifecyclePlugin.ts
```diff
@@ -91,10 +91,10 @@ export default class LifecyclePlugin implements PluginWithState<LifecyclePluginS
 
         this.state = {
             customData: {},
-            defaultFormat: options.defaultFormat || null,
+            defaultFormat: options.defaultFormat ?? null,
             isDarkMode: !!options.inDarkMode,
-            getDarkColor: options.getDarkColor || ((color: string) => color),
-            onExternalContentTransform: options.onExternalContentTransform,
+            getDarkColor: options.getDarkColor ?? ((color: string | null) => color ?? '#000'),
```
JiuqingSong, line 96: why use #000 as default value?

JiuqingSong: @ianeli1  can you provide some update for this comment? Thanks.

ianeli1: Not sure what the solution is here
If given no color (`null`), it previously returned `null`, which doesn't match the typings given
Should this behavior be kept and return `string | null` instead, or should it return `""` as default?
I can only see this method being used here, which removes the styling attribute if `null` is returned
https://github.com/microsoft/roosterjs/blob/457a8e4cdad3c17d861ff2f4b5da969d892b6d39/packages/roosterjs-editor-core/lib/coreApi/transformColor.ts#L63
https://github.com/microsoft/roosterjs/blob/457a8e4cdad3c17d861ff2f4b5da969d892b6d39/packages/roosterjs-editor-core/lib/coreApi/transformColor.ts#L131-L141

JiuqingSong: @ianeli1

ianeli1: Do you mean having the definition only accept strings
`(color: string) => string`
or allowing it to return null?
`(color: string | null) => string | null`
We can do either I believe

JiuqingSong: I prefer the first one

JiuqingSong (approved):

PR #1288, opened 2022-09-22
[FHL] Part 3 - Image Selection - Selection Event

juliaroldi:
Added a new PluginEventType called SelectionChanged that must be trigger when the user selection change. The data of this event, SelectionChangedEvent, stores the selected element, the selection type, the coordinates (if is a table selection) and the keyboard key that triggered the selection changed, if it is a keyboard event that triggered the change. 
Also added: 

1. a new entry for the getSelection playground api 
2. Improvements in ImageSelection Plugin to handle keydown event in a selectedImage
3. RemoveImportantStyleRule to remove !important from css attributes.


packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -47,6 +54,11 @@ export default class ImageSelection implements EditorPlugin {
                     const selection = this.editor.getSelectionRangeEx();
                     if (selection.type == SelectionRangeTypes.ImageSelection) {
                         this.editor.select(selection.image);
+                        this.triggerSelectionEvent(
```
JiuqingSong, line 57: Let editor core API to trigger this event instead. So that any other plugin can also call that API to trigger the same event

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -57,6 +69,11 @@ export default class ImageSelection implements EditorPlugin {
                         if (images.length == 1) {
                             const image = images[0];
                             this.editor.select(image);
+                            this.triggerSelectionEvent(
```
JiuqingSong, line 72: same

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -57,6 +69,11 @@ export default class ImageSelection implements EditorPlugin {
...
+                                this.editor,
+                                SelectionRangeTypes.ImageSelection,
+                                image
+                            );
                         }
                         this.imageId = null;
```
JiuqingSong, line 78: Instead of update imageId here, let's handle the Selection Changed event in this plugin as well, and check if it is a Image Selection so everything fall into the same code path

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -65,10 +82,63 @@ export default class ImageSelection implements EditorPlugin {
                     const target = event.rawEvent.target;
                     if (safeInstanceOf(target, 'HTMLImageElement')) {
                         this.editor.select(target);
+                        this.triggerSelectionEvent(
```
JiuqingSong, line 85: same here

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -65,10 +82,63 @@ export default class ImageSelection implements EditorPlugin {
...
+                            this.editor,
+                            SelectionRangeTypes.ImageSelection,
+                            target
+                        );
                         this.imageId = target.id;
+                    } else {
+                        const mousedownSelection = this.editor.getSelectionRangeEx();
+                        this.triggerSelectionEvent(this.editor, mousedownSelection.type);
```
JiuqingSong, line 93: same

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -65,10 +82,63 @@ export default class ImageSelection implements EditorPlugin {
...
+                        } else if (
+                            key === SHIFT_KEYCODE ||
+                            key === CTRL_KEYCODE ||
+                            key === ALT_KEYCODE ||
+                            event.rawEvent.metaKey
+                        ) {
+                            this.editor.select(keyDownSelection.ranges[0]);
+                        }
+                        const newSelection = this.editor.getSelectionRangeEx();
+                        this.triggerSelectionEvent(
```
JiuqingSong, line 119: same

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -65,10 +82,63 @@ export default class ImageSelection implements EditorPlugin {
                     const target = event.rawEvent.target;
                     if (safeInstanceOf(target, 'HTMLImageElement')) {
                         this.editor.select(target);
+                        this.triggerSelectionEvent(
+                            this.editor,
+                            SelectionRangeTypes.ImageSelection,
+                            target
+                        );
                         this.imageId = target.id;
```
JiuqingSong, line 90: same here

packages/roosterjs-editor-types/lib/event/SelectionChangeEvent.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * If is a table selection, the coordinates of the selection
+     */
+    coordinates?: TableSelection;
+
+    /**
+     * If the selection changed was caused by a keyboard event, the key that changed selection
+     */
+    keyboardKey?: number | string;
```
JiuqingSong, line 35: I feel we don't need to pass the key info here. We should just know selection is changed, but no need to know the reason

packages/roosterjs-editor-types/lib/event/SelectionChangeEvent.ts
```diff
@@ -0,0 +1,50 @@
...
+import TableSelection from '../interface/TableSelection';
+import { PluginEventType } from '../enum/PluginEventType';
+import { SelectionRangeTypes } from '../enum/SelectionRangeTypes';
+import type { CompatibleSelectionRangeTypes } from '../compatibleEnum/SelectionRangeTypes';
+import type { CompatiblePluginEventType } from '../compatibleEnum/PluginEventType';
+
+/**
+ * Data of SelectionChangedEvent
+ */
+export interface SelectionChangedEventData {
```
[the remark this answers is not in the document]
JiuqingSong: Actually I think we just need to pass selection type for performance. since in most case plugin doesn't really need the detailed selection info. If we really need, we can add later.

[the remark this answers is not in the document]
juliaroldi: Should we use the whole SelectionRangeEx or just pass the selection type?

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
juliaroldi: do you mean the selectImage Api?

packages/roosterjs-editor-types/lib/event/SelectionChangeEvent.ts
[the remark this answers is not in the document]
juliaroldi: in ImageEdit we will consume the this plugin data, is better to call getSelectionRange to retrieve the selected image or store it in this event?

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
JiuqingSong: yes

JiuqingSong:
I think we also need to trigger the event when table selection is changed.

And, when document selection is changed, should we also trigger this event? Or it can be in another PR since this will happen very frequently.

juliaroldi:
> I think we also need to trigger the event when table selection is changed.
> 
> And, when document selection is changed, should we also trigger this event? Or it can be in another PR since this will happen very frequently.

I think we need to do another PR for trigger table selection changed and another one for trigger normal selection change. 
I tested do the same changed I did for selectImage api in selectTable and it broke some test.

JiuqingSong:
> > I think we also need to trigger the event when table selection is changed.
> > And, when document selection is changed, should we also trigger this event? Or it can be in another PR since this will happen very frequently.
> 
> I think we need to do another PR for trigger table selection changed and another one for trigger normal selection change. I tested do the same changed I did for selectImage api in selectTable and it broke some test.

Sure, let's do that in another PR.

packages/roosterjs-editor-types/lib/event/SelectionChangeEvent.ts
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * Information of the selection
+     */
+    selectionRangeEx: SelectionRangeEx;
```
JiuqingSong, line 13: How do you want to handle the case when unselect an image or a table?

In theory that will cause a normal selection to be created. But since that will be handled in separate PR, what will we pass for selectionRangeEx for now?

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -37,35 +45,47 @@ export default class ImageSelection implements EditorPlugin {
...
+                            this.editor.getSelectionRange().collapse();
+                        } else if (
+                            key === SHIFT ||
+                            key === CTRL ||
+                            key === Alt ||
+                            event.rawEvent.metaKey
+                        ) {
+                            this.editor.select(keyDownSelection.ranges[0]);
+                        } else if (key === Backspace || key === Enter) {
+                            event.rawEvent.preventDefault();
```
JiuqingSong, line 83: Hmm, for Backspace and Enter (as well as Delete), we should remove the selected image (and add a new line for Enter key). So better do the same handling with Ctrl/Shift/Alt to select the image then let browser handle the deleting for us.

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -37,35 +45,47 @@ export default class ImageSelection implements EditorPlugin {
...
+                            event.rawEvent.stopPropagation();
+                        } else {
+                            const cursorPosition = this.editor.getSelectionRange();
```
JiuqingSong, line 86: For other input types, such as typing something, we should also delete the image and replace with the typed key. So, the same, select the image and let browser handle it.

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -37,35 +45,47 @@ export default class ImageSelection implements EditorPlugin {
...
+                            this.editor.select(cursorPosition);
+                        }
                     }
```
JiuqingSong, line 89: There are some other keys that need to be handled, such as Home, End, PageUp, Page Down. For those cases, just unselect the image, and put focus before the image, then let browser handle it.

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -37,35 +45,47 @@ export default class ImageSelection implements EditorPlugin {
...
+                        if (key === Escape) {
+                            this.editor.select(keyDownSelection.image, PositionType.Before);
+                            this.editor.getSelectionRange().collapse();
+                        } else if (
+                            key === SHIFT ||
+                            key === CTRL ||
+                            key === Alt ||
+                            event.rawEvent.metaKey
+                        ) {
+                            this.editor.select(keyDownSelection.ranges[0]);
```
JiuqingSong, line 81: For these keys, we should select the image using normal selection instead, then browser can handle the rest for us.

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
```diff
@@ -20,19 +28,34 @@ export const selectImage: SelectImage = (core: EditorCore, image: HTMLImageEleme
...
         addUniqueId(core.contentDiv, CONTENT_DIV_ID);
 
+        core.api.selectRange(core, createRange(new Position(image, PositionType.After)));
+
         select(core, image);
-        return {
+
+        const selection: ImageSelectionRange = {
             type: SelectionRangeTypes.ImageSelection,
             ranges: [range],
```
JiuqingSong, line 37: Better pass the range with the whole image selected, but not just after it.

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
```diff
@@ -20,19 +28,34 @@ export const selectImage: SelectImage = (core: EditorCore, image: HTMLImageEleme
...
+            core,
+            {
+                eventType: PluginEventType.SelectionChanged,
+                selectionRangeEx: selection as SelectionRangeEx,
+            },
+            true /** broadcast */
+        );
     }
 
     return null;
```
JiuqingSong, line 54: If image is null, it means we are unselecting. So the "selection" is also "changed", we should trigger event as well

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
```diff
@@ -20,19 +28,34 @@ export const selectImage: SelectImage = (core: EditorCore, image: HTMLImageEleme
...
-        return {
+
+        const selection: ImageSelectionRange = {
             type: SelectionRangeTypes.ImageSelection,
             ranges: [range],
             image: image,
             areAllCollapsed: range.collapsed,
         };
+
+        core.domEvent.imageSelectionRange = selection;
```
JiuqingSong, line 42: also need to update imageSelectionRange when image is null for "unselect" scenario.

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
juliaroldi: I was copying the before word online that nothing happen for theses keys. Backspace is a type, is space only. I'll change.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -483,11 +484,10 @@ export default class Editor implements IEditor {
             !arg2
         ) {
             const selection = core.api.selectImage(core, arg1);
-            core.domEvent.imageSelectionRange = selection;
+            this.triggerSelectionChanged(selection as SelectionRangeEx);
```
JiuqingSong, line 487: same

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -37,35 +49,52 @@ export default class ImageSelection implements EditorPlugin {
...
+                    }
+                    break;
+                case PluginEventType.KeyDown:
+                    const key = event.rawEvent.key;
+                    const keyDownSelection = this.editor.getSelectionRangeEx();
+                    if (keyDownSelection.type === SelectionRangeTypes.ImageSelection) {
+                        if (
+                            key === Escape ||
+                            key === PageDown ||
+                            key === PageUp ||
```
JiuqingSong, line 78: also need to handle up, down, left, right

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -37,35 +49,52 @@ export default class ImageSelection implements EditorPlugin {
...
+                            key === CTRL ||
+                            key === Alt ||
+                            key === Enter ||
+                            key == Space ||
+                            event.rawEvent.metaKey
+                        ) {
+                            this.editor.select(keyDownSelection.ranges[0]);
+                        } else {
+                            keyDownSelection.ranges[0].deleteContents();
+                            const cursorPosition = this.editor.getSelectionRange();
```
JiuqingSong, line 95: Actually, can we just allow the behavior we have in ImageEdit plugin here https://github.com/microsoft/roosterjs/blob/fc0408cf3ce3165fc9625b4a1ae1beb3026f3a78/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts#L205 to handle keyboard event?

JiuqingSong: In most cases (actually almost all), we just need to select the image with normal selection, then let browser handle the rest.

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
juliaroldi: For up, down, left, right, the range of the image is selected and the browser handles the action

JiuqingSong: where is this logic?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -470,7 +471,7 @@ export default class Editor implements IEditor {
         if (arg1 && 'rows' in arg1) {
             const selection = core.api.selectTable(core, arg1, <TableSelection>arg2);
             core.domEvent.tableSelectionRange = selection;
-
+            this.triggerSelectionChanged(selection as SelectionRangeEx);
```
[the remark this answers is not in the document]
JiuqingSong: Do you still need to do type cast now?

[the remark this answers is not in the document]
juliaroldi: Yes, because it returns TableSelectionRange for tables and ImageSelection for images.

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
juliaroldi: In the else case, the ImageSelectionRange store the range of the image, so when we select the range inside this object we do normal selection. 
arrowKeys

packages/roosterjs-editor-plugins/lib/plugins/ImageSelection/ImageSelection.ts
```diff
@@ -1,18 +1,25 @@
 import { safeInstanceOf } from 'roosterjs-editor-dom';
```
BryanValverdeU, line 1: Should we add this plugin to the EditorOptions Pane plugin list?

juliaroldi: I'll add it

JiuqingSong (approved):

PR #1301, opened 2022-09-29
Remove Content Edit Feature

juliaroldi:
When dispose the ContentEdit Plugin remove the added features.


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/ContentEdit.ts
```diff
@@ -59,13 +66,26 @@ export default class ContentEdit implements EditorPlugin {
...
-            .forEach(feature => editor.addContentEditFeature(feature));
+        return features.concat(this.additionalFeatures || []);
+    }
+
+    private addFeatures(editor: IEditor) {
+        const features = this.getFeatures();
+        features.forEach(feature => editor.addContentEditFeature(feature));
+    }
+
+    private disposeFeatures() {
```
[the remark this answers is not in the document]
juliaroldi: The purpose of this function is remove features added by the ContentEdit Plugin. I think the features directly added by editor.addContentEditFeature can be removed by editor.removeContentEditFeature.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -876,6 +876,17 @@ export default class Editor implements IEditor {
...
     }
 
+    /**
+     * Remove a Content Edit feature.
+     * @param feature The feature to remove
+     */
+    public removeContentEditFeature(feature: GenericContentEditFeature<PluginEvent>) {
+        const core = this.getCore();
+        feature?.keys.forEach(key => {
+            core.edit.features[key] = [];
```
JiuqingSong, line 886: no, do not set to empty array because there may be other features in this array.

You need to find the index of the given feature and use splice() function to remove it.

Then if it end up with an empty array, delete this key from core.editor.features.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/ContentEdit.ts
```diff
@@ -43,6 +45,11 @@ export default class ContentEdit implements EditorPlugin {
      * @param editor The editor instance
      */
     initialize(editor: IEditor): void {
+        this.editor = editor;
+        this.addFeatures(editor);
```
JiuqingSong, line 49: suggest cache the feature array in this plugin class. Then when dispose, just remove them one by one from this array, no need to get them again.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/ContentEdit.ts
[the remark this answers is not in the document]
BryanValverdeU: I see,
Just curious in what scenarios we only want to remove the content edit features added by this plugin and keep the rest?

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -895,6 +895,18 @@ export default class Editor implements IEditor {
...
 
+    /**
+     * Remove a Content Edit feature.
+     * @param feature The feature to remove
+     */
+    public removeContentEditFeature(feature: GenericContentEditFeature<PluginEvent>) {
+        const core = this.getCore();
+        feature?.keys.forEach(key => {
+            const index = core.edit.features[key].indexOf(feature);
+            core.edit.features[key].splice(index, 1);
```
JiuqingSong, line 906: If core.edit.features[key] is empty after this change, we should delete it:
```ts
delete core.edit.features[key]
```

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/ContentEdit.ts
```diff
@@ -55,17 +56,25 @@ export default class ContentEdit implements EditorPlugin {
...
                 (!hasSettingForKey && !feature.defaultDisabled)
             ) {
-                features.push(feature);
+                this.features.push(feature);
             }
         });
-
-        features
+        this.features
             .concat(this.additionalFeatures || [])
```
JiuqingSong, line 63: Need to save the result of concat() to the plugin, so we can remove it when dispose as well

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/ContentEdit.ts
```diff
@@ -55,17 +56,25 @@ export default class ContentEdit implements EditorPlugin {
...
-            .forEach(feature => editor.addContentEditFeature(feature));
+            .forEach(feature => this.editor.addContentEditFeature(feature));
+    }
+
+    private disposeFeatures() {
+        if (this.editor) {
+            this.features.forEach(feature => this.editor.removeContentEditFeature(feature));
```
JiuqingSong, line 69: Reset this.features to be empty array

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/ContentEdit.ts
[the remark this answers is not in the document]
juliaroldi: Hey Bryan, sorry for the late replay. Sometimes we need to update and dispose this plugin without update the whole editor, then we just need remove the feature added by this plugin.

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -895,6 +895,23 @@ export default class Editor implements IEditor {
...
 
+    /**
+     * Remove a Content Edit feature.
+     * @param feature The feature to remove
+     */
+    public removeContentEditFeature(feature: GenericContentEditFeature<PluginEvent>) {
+        const core = this.getCore();
+        feature?.keys.forEach(key => {
+            const index = core.edit.features[key]?.indexOf(feature);
+            if (index) {
```
JiuqingSong, line 906: what if index is 0?

juliaroldi: Added a check

packages/roosterjs-editor-core/lib/editor/Editor.ts
```diff
@@ -895,6 +895,23 @@ export default class Editor implements IEditor {
...
 
+    /**
+     * Remove a Content Edit feature.
+     * @param feature The feature to remove
+     */
+    public removeContentEditFeature(feature: GenericContentEditFeature<PluginEvent>) {
+        const core = this.getCore();
+        feature?.keys.forEach(key => {
+            const index = core.edit.features[key]?.indexOf(feature);
+            if (!isNaN(index)) {
```
JiuqingSong, line 906: when key doesn't exist in features, index will be undefined. So better do:
```ts
const featureSet = core.edit.features[key];

if (featureSet && featureSet.indexOf(feature) >= 0) {
...
```

JiuqingSong (approved):

PR #1302, opened 2022-09-29
Fix list chain

juliaroldi:
If a list already exist on the editor and the triggers for start a new list (1), 1-, a., ...) is typed, there is no need to check if a list already exist. Just start a new list.


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getAutoNumberingListStyle.ts
```diff
@@ -116,7 +116,7 @@ export default function getAutoNumberingListStyle(
     //The index is always the character before the last character
     const listIndex = trigger[trigger.length - 2];
 
-    if (previousListChain) {
+    if (previousListChain && parseInt(listIndex) > 1) {
```
JiuqingSong, line 119: parseInt(listIndex) is also used in line 124, so save the result to a local var.

JiuqingSong (approved):

PR #1309, opened 2022-10-05
Fix context menu not working issue

JiuqingSong:
This is a regression from the strict mode change of roosterjs-react package. It has a check to context object and if it is not passed, context menu onclick handler is not triggered.

Fix by skipping checking context and allow undefined context to be passed since that is the default scenario.


JiuqingSong:
#1310

ianeli1 (approved):
lgtm

PR #1316, opened 2022-10-07
Allow insert entity on region root

JiuqingSong:
When insert entity, we now allow force insert the entity to be at the root of current region. e.g.

Before insert:
```html
<div>
  <div>text before[Cursor]text after</div>
</div>
```

When insert a new DIV at "[Cursor]", before this change, it will be:
```html
<div>
  <div>text beforetext after</div>
  <div>(New DIV)</div>      
</div>
```

After this change, when pass the parameter `insertToRegionRoot` to be true, it will be:

```html
<div>
  <div>text before</div>
</div>
<div>(New DIV)</div>      
</div>
  <div>text after</div>
</div>
```

Now the new div's parent must be region root.


packages/roosterjs-editor-core/lib/coreApi/insertNode.ts
```diff
@@ -156,7 +161,9 @@ export const insertNode: InsertNode = (
                     let pos: NodePosition = Position.getStart(range);
                     let blockElement: BlockElement | null;
 
-                    if (
+                    if (option.insertOnNewLine && option.insertToRegionRoot) {
+                        pos = adjustInertPositionRegionRoot(core, range, pos);
```
BryanValverdeU, line 165: Should it be adjustInsertPositionRegionRoot?

BryanValverdeU (approved):

PR #1319, opened 2022-10-07
Enable strict mode for color utils

JiuqingSong:
1. Enable strict mode for color utils package
2. Prepare strict mode for api package
3. Add a parameter for getDarkColor() to allow pass in a customized dark base value
4. Bump color utils package version to 1.1.0 (this will not publish new version until we integrate this change to release branch)


JiuqingSong:
#886 
#890

BryanValverdeU (approved):

PR #1326, opened 2022-10-10
Fix markdown plugin for copy/paste code

juliaroldi:
When a text is pasted, it might create new inlineElement inside the current element, then to ensure the text must be converted or not, check text of the parentBlock. 

markDownFeat


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
```diff
@@ -54,8 +54,11 @@ function cacheGetRangeForMarkdownOperation(
                 return false;
             }
 
+            const parentBlockText = textInlineElement.getParentBlock().getTextContent();
```
JiuqingSong, line 57: need to add comment to explain why we do this, better with some example

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
```diff
@@ -54,6 +54,8 @@ function cacheGetRangeForMarkdownOperation(
                 return false;
             }
 
+            //if the text is pasted, it might create a inner element inside the text element,
```
JiuqingSong, line 57: Is it possible to have two or more layers of DIV?

juliaroldi: if we have to or more layers, the markdown plugin will not work. It only work for inline elements.

JiuqingSong (approved):

PR #1327, opened 2022-10-10
Return the Table Operation in the editTable API and store it in the Content Edit Event Data

BryanValverdeU:
Store the content edit data when editing a table, so this information can be used during the content changed event.

### DIFF
--- a/demo/scripts/controls/sidePane/eventViewer/EventViewPane.tsx
+++ b/demo/scripts/controls/sidePane/eventViewer/EventViewPane.tsx
@@ -174,6 +174,8 @@ export default class EventViewPane extends React.Component<
                         Source=
                         {event.source}, Data=
                         {event.data && event.data.toString && event.data.toString()}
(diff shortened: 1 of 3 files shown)


JiuqingSong:
is this for logging?

BryanValverdeU:
Yes it can be used for logging, but the mail reason I am adding it is to use it in the Content Changed event so we can announce to the user that a new row has been added.

JiuqingSong:
> Yes it can be used for logging, but the main reason I am adding it is to use it in the Content Changed event so we can announce to the user that a new row has been added.

I see. I think it is ok to do this for now. But we more and more kind of data added to ContentChangedEvent, and they are not strong-typed, it will be hard to manage in the future. I think later we may need to design a unified way to let plugins know what kind of change it happened.

JiuqingSong (approved):

BryanValverdeU:
Marking this as Draft for now.

PR #1329, opened 2022-10-11
Add tests to HyperLink plugin

ianeli1:
This test suite covers:
- Hovering over nothing displaying no title
- Hovering over a link displaying a message
- Ctrl+Clicking the link opening it in a new tab
- Clicking the link doing nothing


packages/roosterjs-editor-plugins/test/HyperLink/HyperLinkTest.ts
```diff
@@ -0,0 +1,81 @@
...
+        div = editorDiv;
+    });
+
+    afterEach(() => {
+        editor.dispose();
+        div.remove();
+    });
+
+    it('Removes title attribute if nothing is being hovered', () => {
+        div.dispatchEvent(new MouseEvent('mouseover'));
```
JiuqingSong, line 33: instead of really dispatch an event, we prefer to mock a PluginEvent and call plugin.onPluginEvent() directly since we don't really need to test browser's code, but just test our plugin code only.

JiuqingSong: Same for the others

packages/roosterjs-editor-plugins/test/HyperLink/HyperLinkTest.ts
```diff
@@ -0,0 +1,81 @@
...
+                bubbles: true,
+            })
+        );
+        expect(div.getAttribute('title')).toEqual(HREF);
+    });
+
+    it('Opens the link whenever Ctrl+Click', () => {
+        const openSpy = spyOn(window, 'open');
+        div.appendChild(anchor);
+        plugin.onPluginEvent({
```
JiuqingSong, line 51: this is nice

packages/roosterjs-editor-plugins/test/HyperLink/HyperLinkTest.ts
ianeli1: this is handled by mouseover and mouseout events, not `onPluginEvent` 

Should I `(<any>plugin).onMouse` and just test them directly?

JiuqingSong: @ianeli1 I think you can mock/spy editor.addDomEventHandler, so that you can easily get the onMouse callback in an elegant way.

JiuqingSong (approved):

PR #1336, opened 2022-10-15
Handle var color for dark mode

JiuqingSong:
when use color with variable such as var(--my-color, #123456), we need to parse the color value and retrieve its fallback value if any, and use its fallback value to calculate dark mode color.


JiuqingSong:
Created a better fix #1339 , so close this one

PR #1285, opened 2022-09-21
Enable strict mode for "packages-ui"

ianeli1:
Enables strict mode for all files under `packages-ui/`

PR #1340, opened 2022-10-18
Fix broken paste option plugin in demo app

cyanzhong:
This PR fixes a regression introduced in #1285, we never expected strings in `showPasteOptionPane` to be required.

To fix it we should either:
- Make the parameter optional and remove the check, or
- Provide localized strings when creating the pane

This PR goes with option 1, feel free to close if you think there's a better option.

> It should be OK in production since we will likely provide localized strings.


JiuqingSong:
@ianeli1 would you please review this change

Thanks.

ianeli1 (approved):
Looks good

PR #1343, opened 2022-10-18
Fix clear format not handling heading and blockquote wrappers

ianeli1:
Selects all `<hX>` and `<blockquote>` elements under selection and unwraps them as they are not removed by `removeFormat` command.
https://dvcs.w3.org/hg/editing/raw-file/tip/editing.html#the-removeformat-command

Initial
image
Current clearFormat impl
image
Fixed clearFormat impl
image

Additionally set a default font weight so partial clear format under `<h1>` tags isn't bold
image


JiuqingSong:
Please fix the build

ianeli1:
https://github.com/microsoft/roosterjs/pull/1360/files

PR #1360, opened 2022-10-24
Fix clear format not handling heading and blockquote wrappers - v2

ianeli1:
Selects all `<hX>` and `<blockquote>` elements under selection and unwraps them as they are not removed by `removeFormat` command.
https://dvcs.w3.org/hg/editing/raw-file/tip/editing.html#the-removeformat-command

Initial
image
Current clearFormat impl
image
Fixed clearFormat impl
image

Additionally set a default font weight so partial clear format under `<h1>` tags isn't bold
image

### DIFF
--- a/packages/roosterjs-editor-api/lib/format/clearFormat.ts
+++ b/packages/roosterjs-editor-api/lib/format/clearFormat.ts
@@ -25,6 +25,7 @@ import {
     isNodeInRegion,
     isVoidHtmlElement,
     PartialInlineElement,
+    Position,
     safeInstanceOf,
     setStyles,
     splitBalancedNodeRange,
@@ -40,6 +41,7 @@ const TAGS_TO_UNWRAP = 'B,I,U,STRONG,EM,SUB,SUP,STRIKE,FONT,CENTER,H1,H2,H3,H4,H
 );
 const ATTRIBUTES_TO_PRESERVE = ['href', 'src', 'cellpadding', 'cellspacing'];
 const TAGS_TO_STOP_UNWRAP = ['TD', 'TH', 'TR', 'TABLE', 'TBODY', 'THEAD'];
+const WRAPPER_SELECTOR = 'h1, h2, h3, h4, h5, h6, blockquote';
 
 /**
  * @param editor The editor instance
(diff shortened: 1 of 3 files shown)


packages/roosterjs-editor-api/lib/format/setFontWeight.ts
```diff
@@ -0,0 +1,24 @@
...
+ * @param editor The editor instance
+ * @param fontWeight The fontWight string, should be a valid CSS font-weight style.
+ * Currently there's no validation to the string, if the passed string is invalid, it won't take affect
+ */
+export default function setFontWeight(editor: IEditor, fontWeight: string) {
+    applyListItemStyleWrap(
+        editor,
+        'font-weight',
+        element => {
+            let computedFontWeight = getComputedStyle(element, 'font-weight');
```
juliaroldi, line 17: Why not always set the font-weight? If we always set the font-wight, we will not need to call getComputedStyles and the function will be cheaper to call.

ianeli1: True, this function is only called by clearFormat anyway

packages/roosterjs-editor-api/lib/format/setFontWeight.ts
```diff
@@ -0,0 +1,18 @@
+import applyListItemStyleWrap from '../utils/applyListItemWrap';
```
JiuqingSong, line 1: suggest move this function to where it is used and do not export it for now. We can export it later when we really need.

It is weird that it can pass the build without marking it is @internal.

packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -195,6 +198,15 @@ function clearBlockFormat(editor: IEditor) {
     );
 }
 
+function isRangeContainingAll(range: Range) {
+    return (
+        range.startContainer === range.startContainer.parentElement.firstChild &&
```
JiuqingSong, line 203: you can check
```
!range.startContainer.previousSibling
```

Same for endContainer

ianeli1:
No longer needed

PR #1364, opened 2022-10-26
Content Model: Provide default settings in EditorContext

JiuqingSong:
Provide default settings in context so that after override default settings, we can still call default processors/handlers/format handlers to perform default actions.


JiuqingSong:
> _ No description provided. _

PR #1367, opened 2022-10-27
Apply format to word where cursor is located

Andres-CT98:
### Description:
If the cursor is located inside a word, using the format buttons on the content model ribbon will apply it to the word.

Applies to these formats: 

* Bold
* Italic
* Underline
* Subscript
* Superscript
* Strikethrough
* Change Case
* Font size
* Font face
* Font colour
* Blackground colour
* Format painter
* Clear formatting

### Fix

Traverse where the cursor is placed in the word, first to the left and then to the right, to find the bounds of the word as determined by a space or punctuation.

### Cases tested:

Using the string: Word1 Word2 <i>Word3</i>

1. Put cursor after the W on Word1 and press the Bold button
2. Put cursor before the r on Word2 and press the Strikethrough button
3. Put cursor before the 3 on Word3 and press the Italic button

**Before:** The string will not change

**After:** The string will look like this: <b>Word1</b> <s>Word2</s> Word3

### Warnings and implications

- Does not work with shortcuts or non Content Model buttons and functions


JiuqingSong:
Do we decide to do this without content model?

Andres-CT98:
> Do we decide to do this without content model?

Looking at the roadmap. Port format API would work for all these scenarios, but I do not know when it would be implemented. I tried to use Paragraphs but it is not clear how they work.

JiuqingSong:
> > Do we decide to do this without content model?
> 
> Looking at the roadmap. Port format API would work for all these scenarios, but I do not know when it would be implemented. I tried to use Paragraphs but it is not clear how they work.

Format API work will be started in coming 1-2 weeks. Would you like to hold on and wait for that change happen?

packages/roosterjs-content-model/lib/modelApi/selection/iterateSelections.ts
```diff
@@ -103,9 +105,63 @@ export function iterateSelections(
...
+                                }
+                            } else {
+                                break;
+                            }
+                        }
+                        markerSelectionIndex = block.segments.indexOf(segments[0]);
+                        for (let i = markerSelectionIndex + 1; i < block.segments.length; i++) {
+                            if (block.segments[i].segmentType == 'Text') {
+                                const found = findDelimiter(
+                                    block.segments[i] as ContentModelText,
```
Andres-CT98, line 140: I feel I am doing a lot of casting to ContentModelText

packages/roosterjs-content-model/lib/modelApi/selection/iterateSelections.ts
```diff
@@ -103,9 +105,63 @@ export function iterateSelections(
...
+                                if (found != null) {
+                                    if (found == 0) {
+                                        break;
+                                    }
+                                    splitTextSegment(
+                                        block.segments as ContentModelText[],
+                                        i,
+                                        found
+                                    );
+                                    //block.segments[i].isSelected = true;
```
Andres-CT98, line 152: This and the other comment are here to decide if we also add the functionality to allow this function to select the word not just apply format.

packages/roosterjs-content-model/lib/publicTypes/compatibleEnum/BlockGroupType.ts
```diff
@@ -0,0 +1,19 @@
+/**
```
JiuqingSong, line 1: compatibleEnum folder is not needed any more, please remove it

packages/roosterjs-content-model/lib/modelApi/selection/iterateSelections.ts
```diff
@@ -103,9 +105,63 @@ export function iterateSelections(
                 });
 
                 if (segments.length > 0) {
+                    if (segments.length == 1 && segments[0].segmentType == 'SelectionMarker') {
```
JiuqingSong, line 108: iterateSelections() is a common function, it should not do any feature-specified logic, instead it will just invoke the callback function with raw selection info.

You can move this code out to be a util function.

I expect you finally export a function like:

funciton adjustSelectionForWord(model: ContentModelDocument) {
 // Here you can first get selections using iterateSelections() or some other util function, check if selection is collapsed, then adjust it by splitting existing text segment if any, like you have already done.
// Then after running this function, the selections are changed to the splitted segments. Then we just need to run regular format API to do formatting as if those words was selected by user.
}

packages/roosterjs-content-model/lib/modelApi/selection/wordSelection.ts
```diff
@@ -0,0 +1,47 @@
...
+                offset = i + 1;
+                break;
+            }
+        }
+    }
+    return offset;
+}
+
+export function splitTextSegment(segments: ContentModelText[], index: number, found: number) {
+    const segment1 = createText(segments[index].text.substring(0, found));
```
JiuqingSong, line 29: The new segment need to inherit format from the original one. Just pass oldSegment.format as the second parameter

JiuqingSong: And you can change text of the original segment so that no need to create a new one twice.

packages/roosterjs-content-model/lib/modelApi/selection/iterateSelections.ts
JiuqingSong: to avoid type casting, you can do:
```ts
const segment = block.segments[i];

if (segment.segmentType == 'Text') {
   // Now you can use segment as if it is a Text without any type casting because you have already check its type. This is supported by typescript
}
```

packages/roosterjs-content-model/lib/publicApi/utils/formatSegmentWithContentModel.ts
```diff
@@ -33,19 +40,7 @@ export function formatSegmentWithContentModel(
 
             formats.forEach(format => toggleStyleCallback(format, !isTurningOff));
 
-            const isCollapsedSelection =
-                segments.length == 1 && segments[0].segmentType == 'SelectionMarker';
-
-            if (!pendingFormat && isCollapsedSelection) {
```
JiuqingSong, line 39: Actually, after selectWord(), selection is still possible collapsed. That can happen when cursor is on an empty line, ore even the whole doc is empty. So we need to keep this part.

packages/roosterjs-content-model/lib/publicApi/utils/formatSegmentWithContentModel.ts
```diff
@@ -21,8 +21,15 @@ export function formatSegmentWithContentModel(
...
         apiName,
         model => {
-            const segments = getSelectedSegments(model, !!includingFormatHolder);
+            let segments = getSelectedSegments(model, !!includingFormatHolder);
             const pendingFormat = editor.getPendingFormat();
+            const isCollapsedSelection =
+                segments.length == 1 && segments[0].segmentType == 'SelectionMarker';
+
+            if (isCollapsedSelection) {
+                segments = selectWord([model], segments[0]);
```
JiuqingSong, line 30: [model] does not always work. When selection is under a list, or a table, or a quote, "path" parameter need to contain the whole block group path. However, we don't have this info here. So suggest we just pass the whole model object here, then let the function selectWord() to find the right start point.

packages/roosterjs-content-model/lib/publicApi/utils/formatSegmentWithContentModel.ts
Andres-CT98: The selection is still collapsed but what changes? Do we support adding pending states to empty lines?

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
```diff
@@ -0,0 +1,103 @@
...
+        return [];
+    }
+}
+
+const PUNCTUATION_REGEX = /[.,:!?()\[\]\\/]/gu;
+const SPACES_REGEX = /[\u00A0\u1680​\u180e\u2000\u2009\u200a​\u200b​\u202f\u205f​\u3000\s\t\r\n]/gm;
+
+function findDelimiter(segment: ContentModelText, moveRightward: boolean): number | null {
+    const word = segment.text;
+    let offset = null;
```
JiuqingSong, line 66: You can return -1 as "not found" to follow the same pattern with other array searching functions

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
```diff
@@ -0,0 +1,103 @@
...
+        const segments: ContentModelSegment[] = [];
+        let markerSelectionIndex = block.segments.indexOf(marker);
+        for (let i = markerSelectionIndex - 1; i >= 0; i--) {
+            const currentSegment = block.segments[i];
+            if (currentSegment.segmentType == 'Text') {
+                const found = findDelimiter(currentSegment, false);
+                if (found != null) {
+                    if (found == currentSegment.text.length) {
+                        break;
+                    }
```
JiuqingSong, line 23: Make it return -1 as "not found", so only need to check once here

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
```diff
@@ -0,0 +1,103 @@
...
+                    splitTextSegment(block.segments as ContentModelText[], i, found);
+                    //block.segments[i + 1].isSelected = true;
```
JiuqingSong, line 25: remove this unneeded comment

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
```diff
@@ -0,0 +1,103 @@
...
+        let markerSelectionIndex = block.segments.indexOf(marker);
+        for (let i = markerSelectionIndex - 1; i >= 0; i--) {
+            const currentSegment = block.segments[i];
+            if (currentSegment.segmentType == 'Text') {
+                const found = findDelimiter(currentSegment, false);
+                if (found != null) {
+                    if (found == currentSegment.text.length) {
+                        break;
+                    }
+                    splitTextSegment(block.segments as ContentModelText[], i, found);
```
JiuqingSong, line 24: Although this work, but we should not do this kind of type casting, because not all segments are Text node here.

Instead you can pass the array and the found text node rather than the index, so no need to do type casting any more:

```ts
function splitTextSegments(segments: ContentModelSegment[], textToSplit: ContentModelText, ....) {
   const index = segments.indexOf(textToSplit);
   if (index >= 0) {
      ... 
   }
}
```

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
```diff
@@ -0,0 +1,103 @@
...
+import { ContentModelSegment } from '../../publicTypes/segment/ContentModelSegment';
+import { ContentModelText } from '../../publicTypes/segment/ContentModelText';
+import { createText } from '../../modelApi/creators/createText';
+
+export function selectWord(
+    path: ContentModelBlockGroup[],
+    marker: ContentModelSegment
+): ContentModelSegment[] {
+    const parent = path[0];
+    const block = parent.blocks[0];
```
JiuqingSong, line 11: use iterateSelections() to find out the selections. Because parent.blocks[0] may not always be the right block.

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
Andres-CT98: It is needed as I outlined in my previous comment. I would prefer if it was addressed before dismissing it.

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
Andres-CT98: We still need to check twice. The first check is to see if a delimiter was found and the second to dismiss it if the delimiter was found at the very end of the string.

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
Andres-CT98: Why do you recommend I do indexOf when I already have the index?

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
JiuqingSong: What is the difference between found in middle and found at the end? My understanding is, as long as found, we just need to split after it, no matter where it is.

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
JiuqingSong: Understand. This is just to avoid a type casting.

Do indexOf() twice will cost more time, but it is not that bad. This operation it self is super fast. But unnecessary type casting can hide bugs later when we change some type.

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
Andres-CT98: Here we are searching right to left, so if the first thing you find is the delimiter character, it means that segment does not need to be split because it already is.

So for example: "this is a short sentence". If you put the selection marker right before the s in short, the segment is already split. The 3 segments are: "this is a ", SelectionMarker,"short sentence". We only need to split the right one so we can stop checking the left.

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
Andres-CT98: Which selections do I need to find?

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
JiuqingSong: I see. That makes sense to me.

Although without the second check, I believe it can still work even we have an extra split. (It is ok to have an empty text node in content model because later it will be removed when normalize), but I think it is ok to have this optimization here as you suggested.

packages/roosterjs-content-model/lib/publicApi/utils/selectWord.ts
JiuqingSong: iterateSelection() need a callback as parameter. In that callback it will give you all info for a selection that it find. So you just need to check the first selection with a valid paragraph and a selection marker, then return true in the callback to stop it searching:

```ts
iterateSelection([model], (path, tableContext, block, segments) => {
  if (block?.blockType == 'Paragraph' && segments?.length == 1 && segments[0].segmentType == 'SelectionMarker') {
    // Now you have found the selection, block is the parent paragraph, segments[0] is the marker, then you can do your split here
  }

    return true; // this will stop the searching. You can always stop it since when run to here, it must only one single selection
});
```

packages/roosterjs-content-model/lib/modelApi/selection/adjustWordSelection.ts
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        return segments;
+    } else {
+        return [];
+    }
+}
+
+const PUNCTUATION_REGEX = /[.,:?!"()\[\]\\/]/gu;
+const SPACES_REGEX = /[\u00A0\u1680​\u180e\u2000\u2009\u200a​\u200b​\u202f\u205f​\u3000\s\t\r\n]/gm;
```
JiuqingSong, line 83: Suggest add some comment to explain each character in this Regex

packages/roosterjs-content-model/lib/modelApi/selection/adjustWordSelection.ts
```diff
@@ -0,0 +1,129 @@
...
+import { ContentModelText } from '../../publicTypes/segment/ContentModelText';
+import { createText } from '../creators/createText';
+import { iterateSelections } from '../../modelApi/selection/iterateSelections';
+
+export function adjustWordSelection(
+    model: ContentModelDocument,
+    marker: ContentModelSegment
+): ContentModelSegment[] {
+    const path = [model];
+    let markerBlock: ContentModelBlock = path[0].blocks[0];
```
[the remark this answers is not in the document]
Andres-CT98: I tried doing that but it causes the type to be never after iterateSelections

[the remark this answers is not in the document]
JiuqingSong: I see. Then try use undefined as init value instead, it should work.

packages/roosterjs-content-model/lib/publicApi/utils/formatSegmentWithContentModel.ts
```diff
@@ -48,19 +58,13 @@ export function formatSegmentWithContentModel(
...
             );
 
-            const isCollapsedSelection =
-                segments.length == 1 && segments[0].segmentType == 'SelectionMarker';
-
             if (!pendingFormat && isCollapsedSelection) {
                 editor.setPendingFormat(segments[0].format);
-            }
-
-            if (isCollapsedSelection) {
```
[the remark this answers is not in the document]
Andres-CT98: Can I get a specific function that should return? Pending state was causing a lot of formatting not to work.

packages/roosterjs-content-model/lib/modelApi/selection/adjustWordSelection.ts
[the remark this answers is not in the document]
Andres-CT98: Same behaviour with undefined. Creating an empty ContentModelParagraph seems to work.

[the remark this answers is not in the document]
JiuqingSong: image

This works fine to me.

packages/roosterjs-content-model/lib/publicApi/utils/formatSegmentWithContentModel.ts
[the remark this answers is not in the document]
JiuqingSong: When we have changed something to the content model, return true. Otherwise return false.

For example:
1. when there is an expanded selection, we will change its format, so return true
2. when there is only collapsed selection, but we successfully found a word to format, so we will format the word, so still return true
3. when there is only collapsed selection, and no word is found, so we will update a pending format, but we didn't do any change to the content model itself, so return false.

packages/roosterjs-content-model/test/modelApi/selection/adjustWordSelectionTest.ts
```diff
@@ -0,0 +1,280 @@
...
+const defaultMarker: ContentModelSegment = {
+    segmentType: 'SelectionMarker',
+    format: {},
+    isSelected: true,
+};
+
+describe('adjustWordSelection', () => {
+    function runTest(
+        model: ContentModelDocument,
+        result: ContentModelSegment[],
```
JiuqingSong, line 14: Suggest we also verify the model object since in your code you will split some segments.

packages/roosterjs-content-model/test/modelApi/selection/adjustWordSelectionTest.ts
```diff
@@ -0,0 +1,280 @@
...
+        markerIndex: number
+    ) {
+        const marker: ContentModelSegment =
+            model.blocks[0].blockType == 'Paragraph'
+                ? model.blocks[0].segments[markerIndex]
+                : defaultMarker;
```
JiuqingSong, line 20: why does defaultMarker mean here?

packages/roosterjs-content-model/test/modelApi/selection/adjustWordSelectionTest.ts
```diff
@@ -0,0 +1,280 @@
...
+
+describe('adjustWordSelection', () => {
+    function runTest(
+        model: ContentModelDocument,
+        result: ContentModelSegment[],
+        markerIndex: number
+    ) {
+        const marker: ContentModelSegment =
+            model.blocks[0].blockType == 'Paragraph'
+                ? model.blocks[0].segments[markerIndex]
```
JiuqingSong, line 19: So in this test we will only cover the case when selection marker is under the first block? Do you also want to test if it is under the second block, or inside a list or a table?

packages/roosterjs-content-model/lib/modelApi/selection/adjustWordSelection.ts
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        // Marker is at end of word
+        if (segments.length <= 2) {
+            return [marker];
+        }
+
+        return segments;
+    } else {
+        return [];
```
JiuqingSong, line 80: For the case that we didn't find a block, suggest we just return [marker] for no change was happened.

packages/roosterjs-content-model/lib/modelApi/selection/adjustWordSelection.ts
```diff
@@ -0,0 +1,145 @@
...
+    }
+}
+
+const PUNCTUATION_REGEX = /[.,:?!"()\[\]\\/]/gu;
+const SPACES_REGEX = /[\u1680\u2000\u2009\u200a​\u200b​\u202f\u205f​\u3000\s\t\r\n]/gm;
+/*
+// These are unicode characters from the Category Space Separator (Zs)
```
JiuqingSong, line 87: Let's more the comment before line 85.

packages/roosterjs-content-model/lib/modelApi/selection/adjustWordSelection.ts
```diff
@@ -0,0 +1,145 @@
...
+\u2000 = EN QUAD
+\u2009 = THIN SPACE
+\u200a = HAIR SPACE
+​\u200b = ZERO WIDTH SPACE
+​\u202f = NARROW NO-BREAK SPACE
+\u205f​ = MEDIUM MATHEMATICAL SPACE
+\u3000 = IDEOGRAPHIC SPACE
+\u1680 = OGHAM SPACE MARK
+
+\u180e = MONGOLIAN VOWEL SEPARATOR
```
JiuqingSong, line 98: If there is a source doc for this list, please also add the url in comment to help other people check.

packages/roosterjs-content-model/test/modelApi/selection/adjustWordSelectionTest.ts
Andres-CT98: Which content model object?

JiuqingSong: function runTest(
  model,  <================ model will be changed during the test, so we should also verify the changing result is expected.
  result,
)

Andres-CT98: That feels like a test for splitTextSegment not for adjustWordSelection.

JiuqingSong: Yes, you can say that. But is there a dedicate test for splitTextSegment? If not, you can also cover it here, given that splitTextSegment() is not a public function.

JiuqingSong (approved):

PR #1377, opened 2022-10-31
Modify Image wrapper to insert image wrapper outside the editor

juliaroldi:
After this change, when in editing the image will be cloned and it's clone will be added to a wrapper outside the editor. 
Every change made to cloned image will be reflected to image when the quite editing, the editing process will happen outside the editor and DOM structure of the editor will not change while editing. 

imageWrapper


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -329,17 +343,24 @@ export default class ImageEdit implements EditorPlugin {
      * Create editing wrapper for the image
      */
     private createWrapper(operation: ImageEditOperation | CompatibleImageEditOperation) {
-        // Wrap the image with an entity so that we can easily retrieve it later
+        //Clone the image and insert the clone in a entity
+        const clone = this.image.cloneNode();
+        const wrappedImage = wrap(clone, KnownCreateElementDataIndex.ImageEditWrapper);
+
+        this.clonedImage = wrappedImage.firstElementChild as HTMLImageElement;
```
JiuqingSong, line 350: Why not just use "clone" instead of "wrappedImage.firstElementChild as HTMLImageElement" here?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -329,17 +343,24 @@ export default class ImageEdit implements EditorPlugin {
...
+
+        // Wrap the cloned image with an entity so that we can easily retrieve it later
         const { wrapper } = insertEntity(
```
JiuqingSong, line 353: Since the editing wrapper is not part of editor now, I don't think we still need to "insertEntity"

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -352,9 +373,11 @@ export default class ImageEdit implements EditorPlugin {
         this.lastSrc = this.image.getAttribute('src');
 
         // Set image src to original src to help show editing UI, also it will be used when regenerate image dataURL after editing
+        this.image.style.visibility = 'hidden';
```
JiuqingSong, line 376: So this will still modify the editor content.

Can we add this visibility rule to image selection's global CSS so that editor content is not impacted?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -352,9 +373,11 @@ export default class ImageEdit implements EditorPlugin {
...
         this.image.src = this.editInfo.src;
```
JiuqingSong, line 377: Do we still need to change original image's src?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -399,45 +431,31 @@ export default class ImageEdit implements EditorPlugin {
...
      * Get image wrapper from image
-     * @param image The image to get wrapper from
      */
-    private getImageWrapper(image: HTMLImageElement): HTMLElement {
-        // Get the image wrapper from image using Entity API
-        const entity = getEntityFromElement(image?.parentNode?.parentNode as HTMLElement);
-
-        return entity?.type == IMAGE_EDIT_WRAPPER_ENTITY_TYPE ? entity.wrapper : null;
+    private getImageWrapper(): HTMLElement {
+        return document.getElementById(this.imageId);
```
JiuqingSong, line 436: So the cloned image will have the same id with the original one, how can we guarantee this will return the cloned one?

And if there are 2 editors in the same document, it is possible two editors both have image with the same id. Better change the image id to combine editor id and image id to make it unique

JiuqingSong: And instead of querying by image id, since there will be only one image editing per editor, you can just hold the reference to the wrapper element in this plugin. So it will be easier to get it

JiuqingSong:
Generally, the behavior is good. Only one issue, if editor is zoomed, the editing image is not zoomed together, so when select and edit the image, its size changes.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
juliaroldi: Yes, because when we applyFormat we compare the image edited with the previous src, so to make sure apply format works as expected we also need to change this src.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -344,36 +320,25 @@ export default class ImageEdit implements EditorPlugin {
...
-
-        wrapper.id = this.imageId;
-        wrapper.style.maxWidth = '100%';
-        wrapper.style.position = 'fixed';
+        this.clonedImage = this.image.cloneNode(true) as HTMLImageElement;
+        const wrappedImage = wrap(this.clonedImage, KnownCreateElementDataIndex.ImageEditWrapper);
+        this.wrapper = this.editor.getDocument().createElement('span');
+        this.wrapper.appendChild(wrappedImage);
+        this.wrapper.style.maxWidth = '100%';
+        this.wrapper.style.position = 'fixed';
```
JiuqingSong, line 328: Can this be added into KnownCreateElementDataIndex.ImageEditWrapper so that one single wrap() call should make all of them working?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -344,36 +320,25 @@ export default class ImageEdit implements EditorPlugin {
...
 
-        wrapper.style.display = Browser.isSafari ? 'inline-block' : 'inline-flex';
+        this.wrapper.style.display = Browser.isSafari ? 'inline-block' : 'inline-flex';
 
         // Cache current src so that we can compare it after edit see if src is changed
         this.lastSrc = this.image.getAttribute('src');
 
         // Set image src to original src to help show editing UI, also it will be used when regenerate image dataURL after editing
-        this.image.style.visibility = 'hidden';
+        this.image.className = STYLE_IMAGE;
```
JiuqingSong, line 341: Actually the image should already have id. If not, you add some API in editor to ensure it. So no need to change its className

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -407,11 +372,31 @@ export default class ImageEdit implements EditorPlugin {
...
+        const doc = this.editor.getDocument();
+        const styleTag = doc.getElementById(STYLE_IMAGE) as HTMLStyleElement;
+        if (styleTag) {
+            doc.head.removeChild(styleTag);
+        }
+    }
+
+    private addVisibilityCssTag() {
+        const doc = this.editor.getDocument();
+        let styleTag = doc.getElementById(STYLE_IMAGE) as HTMLStyleElement;
```
JiuqingSong, line 392: Why reuse the same string for style tag and image class?

Also, if there are multiple editor in the same page, there will be duplicated id for style tags then hard to distinguish.

We already have code in editor core to generate STYLE tag and add style rule. Consider reuse that code by adding some editor API or parameter to existing API. The reuse part can be in separate PR, but at least make sure the style element id is unique. You can use editor DIV id.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -407,11 +372,31 @@ export default class ImageEdit implements EditorPlugin {
...
+        if (!styleTag) {
+            const cssRule = `.${STYLE_IMAGE} {visibility: hidden}`;
```
JiuqingSong, line 394: Use editor id and image id, just like how we do selection border for image

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -407,11 +372,31 @@ export default class ImageEdit implements EditorPlugin {
...
+            styleTag = doc.createElement('style');
+            styleTag.id = STYLE_IMAGE;
```
JiuqingSong, line 396: Actually I think you can hold a reference to this style element so no id is needed at all.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -174,6 +168,7 @@ export default class ImageEdit implements EditorPlugin {
      * Dispose this plugin
      */
     dispose() {
+        this.removeVisibilityCssTag();
```
JiuqingSong, line 171: We also need to remove the editing image DOM element here, and the call to removeVisibilityCssTag() can be part of it.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -428,36 +413,32 @@ export default class ImageEdit implements EditorPlugin {
...
     };
 
     /**
      * Update image edit elements to reflect current editing result
      * @param context
      */
     private updateWrapper = (context?: DragAndDropContext) => {
-        const wrapper = this.getImageWrapper();
-        if (wrapper) {
+        if (this.wrapper) {
```
JiuqingSong, line 433: consider do
```ts
const wrapper = this.wrapper;

if (wrapper) { 
...
```

So no need to do use "this.wrapper" everywhere to save some bytes

packages/roosterjs-editor-core/lib/coreApi/selectImage.ts
```diff
@@ -49,7 +53,7 @@ export const selectImage: SelectImage = (core: EditorCore, image: HTMLImageEleme
 const select = (core: EditorCore, image: HTMLImageElement) => {
     removeImportantStyleRule(image, ['border', 'margin']);
     const borderCSS = buildBorderCSS(core, image.id);
-    addSelectionStyle(core, borderCSS, STYLE_ID);
+    setGlobalCssStyles(core.contentDiv.ownerDocument, borderCSS, STYLE_ID + core.contentDiv.id);
```
JiuqingSong, line 56: Use removeGlobalCssStyle() in line 67.
Same for selectTable()

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -329,32 +315,27 @@ export default class ImageEdit implements EditorPlugin {
...
-            this.editor,
-            IMAGE_EDIT_WRAPPER_ENTITY_TYPE,
-            wrap(this.image, KnownCreateElementDataIndex.ImageEditWrapper),
-            false /*isBlock*/,
-            true /*isReadonly*/
-        );
+        //Clone the image and insert the clone in a entity
+        this.clonedImage = this.image.cloneNode(true) as HTMLImageElement;
+        const wrappedImage = wrap(this.clonedImage, KnownCreateElementDataIndex.ImageEditWrapper);
+        this.wrapper = wrap(wrappedImage, KnownCreateElementDataIndex.ImageEditContainer);
```
JiuqingSong, line 321: Actually, I mean merge these two into one. Because wrappedImage is only used here. So we can do
this.wrapper = wrap(this.clonedImage, KnownCreateElementDataIndex.ImageEditWrapper);

Then you can modify the definition of ImageEditWrapper to make it has two layers of wrapper.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -329,32 +315,27 @@ export default class ImageEdit implements EditorPlugin {
...
         }
 
-        wrapper.style.display = Browser.isSafari ? 'inline-block' : 'inline-flex';
+        this.wrapper.style.display = Browser.isSafari ? 'inline-block' : 'inline-flex';
 
         // Cache current src so that we can compare it after edit see if src is changed
         this.lastSrc = this.image.getAttribute('src');
 
         // Set image src to original src to help show editing UI, also it will be used when regenerate image dataURL after editing
         this.image.src = this.editInfo.src;
```
JiuqingSong, line 335: I still have concern here. If you need to compare, can you compare with the cloned image?

My primary goal here is not to do any change to editor content when edit.

packages/roosterjs-editor-dom/lib/style/removeGlobalCssStyle.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Remove a css rule style from a style sheet
+ * @param doc The document object
+ * @param styleId the ID of the style tag
+ */
+
+export default function removeGlobalCssStyle(doc: Document, styleId: string) {
+    const styleTag = doc.getElementById(styleId) as HTMLStyleElement;
+    if (styleTag) {
+        doc.head.removeChild(styleTag);
```
JiuqingSong, line 10: Use `styleTag.parentNode?.removeChild(styleTag)` instead

Although accoding to our code, style tag should always be under HEAD. But if we changed that behavior, there will be no build time error or test case to remind us change here. So better make it more robust

JiuqingSong (approved):

PR #1379, opened 2022-11-03
[Hot fix] Add null check to list metadata

juliaroldi:
When the list does not have metadata, return previous list style as null.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
@@ -23,6 +23,7 @@ import {
     createObjectDefinition,
     createNumberDefinition,
     getMetadata,
+    findClosestElementAncestor,
 } from 'roosterjs-editor-dom';
 import {
     BuildInEditFeature,
@@ -310,7 +311,7 @@ const AutoNumberingList: BuildInEditFeature<PluginKeyboardEvent> = {
                         ? 1
                         : parseInt(textBeforeCursor);
 
-                    const isLi = getPreviousList(editor, textRange);
+                    const isLi = getPreviousListItem(editor, textRange);
                     const listStyle = getAutoNumberingListStyle(textBeforeCursor);
                     prepareAutoBullet(editor, textRange);
                     toggleNumbering(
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -340,7 +340,9 @@ const getPreviousListType = (editor: IEditor, textRange: Range, listType: ListTy
     const type = listType === ListType.Ordered ? 'orderedStyleType' : 'unorderedStyleType';
     const previousNode = getPreviousList(editor, textRange);
 
-    return previousNode && getTagOfNode(previousNode) === 'LI'
+    return previousNode &&
+        getTagOfNode(previousNode) === 'LI' &&
+        getMetadata(previousNode.parentElement, ListStyleDefinitionMetadata)
```
JiuqingSong, line 345: Do not call getMetadata twice. It is heavy

BryanValverdeU, line 345: parentElement can be null, can we add a null check?

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -339,10 +339,9 @@ const getPreviousList = (editor: IEditor, textRange: Range) => {
 const getPreviousListType = (editor: IEditor, textRange: Range, listType: ListType) => {
     const type = listType === ListType.Ordered ? 'orderedStyleType' : 'unorderedStyleType';
     const previousNode = getPreviousList(editor, textRange);
+    const metadata = getMetadata(previousNode.parentElement, ListStyleDefinitionMetadata);
```
JiuqingSong, line 342: If previousNode or previousNode.parentElement is null, this will still be a problem.

BTW, what is the parent element here?

juliaroldi: It should to be the UL or OL element, but I noticed that the list structure can different to just <ol><li></li><ol>, then I changed the code to use findClosestElementAntecesor to find the ul or ol element that contain the metadata.

BryanValverdeU (approved):

JiuqingSong (approved):

PR #1383, opened 2022-11-03
Fix image zoom scale while editing

juliaroldi:
Need https://github.com/microsoft/roosterjs/pull/1377 to be merged.
Adjust zoom in image wrapper when editor zoom changes. 
ZoomWrapper


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -124,6 +124,16 @@ export default class ImageEdit implements EditorPlugin {
      */
     private wasResized: boolean;
 
+    /**
+     * Editor zoom scale
+     */
+    private zoomScale: number;
```
JiuqingSong, line 130: no need to cache this value, you can always get it using editor.getZoomScale()

JiuqingSong (approved):

PR #1388, opened 2022-11-07
Content Model: Better support dataset and metadata step 1 of 5: New dataset and metadata API, and apply to image

JiuqingSong:
Content Model: Better support dataset and metadata step 1: New dataset and metadata API, and apply to image

To better support content fidelity and reduce data loss when convert HTML to content model, we need to preserve the values from element's dataset. And since roosterjs metadata is also built on top of dataset, we can provide a set of API to better preserve these info. This includes
1. Read dataset from HTML element and put into content model
2. Do not parse metadata until we really need it, for better performance
3. Write values of dataset back to HTML element

In this change I also port ContentModelImage to use new API for metadata. Later I'll publish another PR to port other elements.


packages/roosterjs-content-model/lib/publicTypes/format/ContentModelWithDataset.ts
```diff
@@ -0,0 +1,11 @@
+import { DatasetFormat } from './formatParts/DatasetFormat';
+
+/**
+ * Represents base format of an element that supports dataset and/or metadata
+ */
+export interface ContentModelWithDataset<T> {
```
BryanValverdeU, line 6: Question, is the generic type going to be used later?

BryanValverdeU (approved):

packages/roosterjs-content-model/lib/publicTypes/format/ContentModelWithDataset.ts
JiuqingSong: Yes, when parse metadata, we want this type to help us prevent unwanted type mismatch.

PR #1405, opened 2022-11-14
Disable the logic in ImageEdit plugin to set image editing when users click images

haven2world:
We extended this plugin for mobile user and used a custom way to activate image editing. So we need an option to disable the default one.


JiuqingSong:
If you extended this class, I think you don't need to add such parameter. Instead, you just need to override onPluginEvent, and don't call super.onPluginEvent() if the event is SelectionChanged event.

JiuqingSong:
```ts
class MyImageEdit extends ImageEdit {
  onPluginEvent(e) {
    if (e.type != PluginEventType.SelectionChanged) {
       super.onPluginEvent(e);
   }
 }
}
```

haven2world:
That's our current temporary workaround actually.  But the point is, on the one hand, we want to leverage the implementation of ImageEdit as more as possible, we just want to avoid this plugin handling users' click to activate the editing, but we still want to keep other events handling logic implemented in this plugin. On the other hand, before the recent change, we had some workaround, but it was broken due to the latest upgrade. So we'd like to add this parameter to let the people working on this plugin know that they should make sure this option always works. Or we have to patch our workaround again and again.

JiuqingSong:
> That's our current temporary workaround actually. But the point is, on the one hand, we want to leverage the implementation of ImageEdit as more as possible, we just want to avoid this plugin handling users' click to activate the editing, but we still want to keep other events handling logic implemented in this plugin. On the other hand, before the recent change, we had some workaround, but it was broken due to the latest upgrade. So we'd like to add this parameter to let the people working on this plugin know that they should make sure this option always works. Or we have to patch our workaround again and again.

Ok in that case, we need a better way. You may noticed there are two states of image editing:
1. resize/rotate
2. crop
current the behavior for onclick is always to resize/rotate. You can add a parameter for default behavior when image is selected, and then you pass in ImageEditOperation.None so that it will do nothing.

```ts
interface ImageEditOptions {
  ...
  onSelectState: ImageEditOperation; /* default: ImageEditOperation.Resize */
}
```

haven2world:
Updated

packages/roosterjs-editor-types/lib/interface/ImageEditOptions.ts
```diff
@@ -1,4 +1,5 @@
 import ModeIndependentColor from './ModeIndependentColor';
+import { ImageEditOperation } from '../enum';
```
JiuqingSong, line 2: Please fix this import, And also import CompatibleImageEditOperation using **type import**:

```ts
import { ImageEditOperation } from '../enum/ImageEditOperation';
import type { CompatibleImageEditOperation } from '../compatibleEnum/ImageEditOperation';
```

packages/roosterjs-editor-types/lib/interface/ImageEditOptions.ts
```diff
@@ -63,4 +64,10 @@ export default interface ImageEditOptions {
      * Whether image crop is disabled. @default false
      */
     disableCrop?: boolean;
+
+    /**
+     * Which operations will be executed when image is selected
+     * @default ImageEditOperation.ResizeAndRotate
+     */
+    onSelectState?: ImageEditOperation;
```
JiuqingSong, line 72: Use both ImageEditOperation and CompatibleImageEditOperation here:
```ts
onSelectState?: ImageEditOperation | CompatibleImageEditOperation;
```

packages/roosterjs-editor-types/lib/interface/ImageEditOptions.ts
haven2world: Fixed

packages/roosterjs-editor-types/lib/interface/ImageEditOptions.ts
haven2world: Added

JiuqingSong (approved):

haven2world:
Hey @JiuqingSong , could you help merge this PR? I don't have the permission

PR #1406, opened 2022-11-14
Restore Ctrl+Click behavior on Firefox

ianeli1:
Ctrl+Clicking an anchor with the Hyperlink plugin enabled no longer opens a new tab on Firefox. Restore behavior by removing the Firefox specific code found in Hyperlink plugin and calling `window.open` always.

image

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
@@ -1,4 +1,4 @@
-import { Browser, isCharacterValue, isCtrlOrMetaPressed, matchLink } from 'roosterjs-editor-dom';
+import { isCharacterValue, isCtrlOrMetaPressed, matchLink } from 'roosterjs-editor-dom';
 import {
     DOMEventHandler,
     EditorPlugin,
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-plugins/lib/plugins/HyperLink/HyperLink.ts
```diff
@@ -141,7 +141,6 @@ export default class HyperLink implements EditorPlugin {
 
                 let href: string | null;
                 if (
-                    !Browser.isFirefox &&
                     (href = this.tryGetHref(anchor)) &&
                     isCtrlOrMetaPressed(event.rawEvent) &&
                     event.rawEvent.button === 0
```
BryanValverdeU, line 146: If this behavior changed overtime and Firefox is no longer opening the links by default, should we prevent the default behavior of the event if browser is Firefox in case a user with an older version and tries to open a link with Ctrl + Click?

ianeli1: True, I don't think this would have any side effects so might as well preventDefault it just in case

JiuqingSong (approved):

PR #1407, opened 2022-11-14
ClearFormat API improvements

ianeli1:
1. Prevents block format being applied when `AutoDetect`ing clear format and an entire element is being selected (eg. an entire table cell, an entire bold word) by identifying if the node completely selected has any siblings (or its parents)
image
Wrong result:
image
Fix:
image

Should be noted that this introduces a behavior change when clearing the format of the content of table cells:
Selection:
image
Previous:
image
New:
image

2. Addresses partial formats of `<h1>` blocks appearing bolder by setting a standard fontWeight when defaultFormat is defined.
Previous:
image
Fix:
image

### DIFF
--- a/packages/roosterjs-editor-api/lib/format/clearFormat.ts
+++ b/packages/roosterjs-editor-api/lib/format/clearFormat.ts
@@ -24,6 +24,7 @@ import {
     isBlockElement,
     isNodeInRegion,
     isVoidHtmlElement,
+    NodeInlineElement,
     PartialInlineElement,
     safeInstanceOf,
     setStyles,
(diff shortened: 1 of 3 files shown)


packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -1,307 +1,339 @@
...
-            clearInlineFormat(editor);
-            break;
-        case ClearFormatMode.Block:
-            clearBlockFormat(editor);
-            break;
-        default:
-            clearAutoDetectFormat(editor);
-    }
-}
+import blockFormat from '../utils/blockFormat';
```
JiuqingSong, line 1: try not change the line end so we can do better review

ianeli1: welp, https://github.com/microsoft/roosterjs/pull/1415

PR #1415, opened 2022-11-17
ClearFormat API improvements - v2

ianeli1:
1. Prevents block format being applied when `AutoDetect`ing clear format and an entire element is being selected (eg. an entire table cell, an entire bold word) by identifying if the node completely selected has any siblings (or its parents)
image
Wrong result:
image
Fix:
image

Should be noted that this introduces a behavior change when clearing the format of the content of table cells:
Selection:
image
Previous:
image
New:
image

2. Addresses partial formats of `<h1>` blocks appearing bolder by setting a standard fontWeight when defaultFormat is defined.
Previous:
image
Fix:
image


packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -140,6 +142,22 @@ function removeNotTableDefaultStyles(element: HTMLTableElement) {
...
+/**
+ * Verifies recursively if a node and its parents have any siblings
+ * Ignoring the children of contentDiv and returning true if any parent is LI
+ * @returns `true` if this node, and its parents (minus the children of the contentDiv) have no siblings
+ */
+function isNodeWholeBlock(node: Node, editor: IEditor) {
+    let currentNode = node;
+    while (currentNode && editor.contains(currentNode.parentNode)) {
+        if (currentNode.nextSibling || currentNode.previousSibling) {
+            return safeInstanceOf(currentNode, 'HTMLLIElement');
```
JiuqingSong, line 154: The comment says "return true if any parent is LI", but here it checks the node itself.

packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -140,6 +142,22 @@ function removeNotTableDefaultStyles(element: HTMLTableElement) {
     });
 }
 
+/**
+ * Verifies recursively if a node and its parents have any siblings
+ * Ignoring the children of contentDiv and returning true if any parent is LI
```
JiuqingSong, line 147: Why we have special handling for LI here?

ianeli1: lists and its items are always inside a single block (meaning it's gonna find siblings) and would break the logic for clear formatting otherwise as the `<li>` items should be removed

ianeli1: image

JiuqingSong (approved):

PR #1425, opened 2022-11-21
Valid numbering

juliaroldi:
Add proper validation to list index to avoid list being triggering by special characters. 
Ex: 1:1. was triggering list because the first character is a number and the last is a proper marker character. 
To avoid that behavior change we check if the string before the marker is number or string without number and special characters

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getAutoNumberingListStyle.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getAutoNumberingListStyle.ts
@@ -159,6 +159,12 @@ export default function getAutoNumberingListStyle(
     return numberingType;
 }
 
+/**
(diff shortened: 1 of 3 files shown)


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getAutoNumberingListStyle.ts
```diff
@@ -153,10 +153,15 @@ export default function getAutoNumberingListStyle(
...
-        isDoubleParenthesis,
-        previousListStyle
-    );
+    const numberingType = isValidNumbering(trigger, isDoubleParenthesis)
+        ? identifyNumberingListType(trigger, isDoubleParenthesis, previousListStyle)
+        : null;
     return numberingType;
 }
+
+function isValidNumbering(textBeforeCursor: string, isDoubleParenthesis: boolean) {
```
JiuqingSong, line 162: Need some comment to help understand, and test cases

JiuqingSong (approved):

PR #1427, opened 2022-11-21
Add link to a image

juliaroldi:
Before this change, if an image was selected and a link was inserted, a new link was added next to the image. 
To fix this, when check if a image is selected and if it is, we add a image inside the anchor tag. 

imageLink

### DIFF
--- a/packages/roosterjs-editor-api/lib/format/createLink.ts
+++ b/packages/roosterjs-editor-api/lib/format/createLink.ts
@@ -1,4 +1,4 @@
-import { HtmlSanitizer, matchLink } from 'roosterjs-editor-dom';
+import { HtmlSanitizer, matchLink, wrap } from 'roosterjs-editor-dom';
 import {
     ChangeSource,
     DocumentCommand,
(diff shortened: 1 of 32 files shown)


packages/roosterjs-editor-api/lib/format/createLink.ts
```diff
@@ -66,42 +72,54 @@ export default function createLink(
...
-                updateAnchorDisplayText(anchor, displayText);
+                    anchor = getAnchorNodeAtCursor(editor);
+                    updateAnchorDisplayText(anchor, displayText);
+                }
+            } else if (selection.type === SelectionRangeTypes.ImageSelection) {
+                selection.ranges[0].deleteContents();
+                anchor = editor.getDocument().createElement('A') as HTMLAnchorElement;
+                anchor.appendChild(selection.image);
+                anchor.href = normalizedUrl;
+                editor.insertNode(anchor);
```
JiuqingSong, line 120: You just need to wrap the image with a new A tag
const a = wrap(selection.imag, 'a');
a.href = ....

juliaroldi: True! I'll refactor!

JiuqingSong (approved):

PR #1431, opened 2022-11-22
Copy table cell styles

juliaroldi:
When paste a table within a table, also copy the cell style. 
copyTableCellStyle

### DIFF
--- a/packages/roosterjs-editor-dom/lib/table/pasteTable.ts
+++ b/packages/roosterjs-editor-dom/lib/table/pasteTable.ts
@@ -48,12 +48,21 @@ export default function pasteTable(
         for (let j = cursorCol; j < columns; j++) {
             let cell = currentTable.getCell(i, j);
             let newCell = newTable.getTd(i - cursorRow, j - cursorCol);
-            if (cell.td) {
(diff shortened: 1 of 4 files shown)


packages/roosterjs-editor-dom/lib/table/pasteTable.ts
```diff
@@ -48,12 +48,21 @@ export default function pasteTable(
...
+                moveChildNodes(cell.td, newCell);
+                copyCellStyle(cell.td, newCell);
             } else {
                 cell.td = document.createElement('td');
             }
         }
     }
-    currentTable.writeBack();
+
+    currentTable.writeBack(true /* skipApplyFormat*/);
```
JiuqingSong, line 60: What if I apply format later, say I pasted a table into a table, then I insert a row or column, what will happen?

juliaroldi: I'm using saveMetadata with  bgColorOverride: true, then write applyFormat will not apply format to this cells.

JiuqingSong (approved):

PR #1432, opened 2022-11-23
Fix markdown text replacement 

juliaroldi:
When calculate the size of the text to replace use the selected range plus the next typed character. 
Before: 
markdownBug

After: 
markdownFix


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
```diff
@@ -106,13 +106,12 @@ function handleMarkdownEvent(
                     textContentRange.startOffset + 1
                 );
 
-                // set the removal range to include the typed last character.
-                const lastIndex: number = (range.endContainer as Text).length;
-                range.setEnd(range.endContainer, lastIndex);
+                const text = textContentRange.extractContents().textContent;
+                const textNode = editor.getDocument().createTextNode(text);
```
BryanValverdeU, line 110: `text` can be null, should we add a null check?

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
```diff
@@ -106,13 +106,12 @@ function handleMarkdownEvent(
...
 
                 // extract content and put it into a new element.
-                const elementToWrap = editor.getDocument().createElement(elementTag);
-                elementToWrap.appendChild(textContentRange.extractContents());
+                const elementToWrap = wrap(textNode, elementTag);
+                range.setEnd(textContentRange.endContainer, textContentRange.endOffset + 1);
```
BryanValverdeU, line 114: I think `textContentRange.endOffset + 1` could throw exceptions if the current endOffset is the last child index and endOffset + 1 does not exist, not sure if this is a possibility for the markdown features.

But I think we could verify this using `new Position(range.endContainer, PositionType.End)` and verify that `textContentRange + 1` is lower than the position offset

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
juliaroldi: this range is calculated `cacheGetRangeForMarkdownOperation`, then if the text is null, the range is null and the event will not be handled.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/markdownFeatures.ts
juliaroldi: That is interesting, I'll add this check.

BryanValverdeU (approved):

PR #1437, opened 2022-11-28
Get biggest z index of element

juliaroldi:
Add `getBiggestZIndex` api that search from an elements to it's root for the biggest z-index value. 
In ImageEdit use `getBiggestZIndex` api to find the biggest index above the editor and add it plus one to the image wrapper. 

zIndex


packages/roosterjs-editor-dom/lib/style/getBiggestZIndex.ts
```diff
@@ -0,0 +1,22 @@
+import getTagOfNode from '../utils/getTagOfNode';
+
+/**
+ * Search through from element to it's root for the biggest z-index value
+ * @param element the parent element
+ * @returns
+ */
+export default function getBiggestZIndex(element: HTMLElement) {
```
JiuqingSong, line 8: This function is now only used by image edit plugin, so let's put it there and no need to export, to reduce the size of dom package

packages/roosterjs-editor-dom/lib/style/getBiggestZIndex.ts
```diff
@@ -0,0 +1,22 @@
...
+    let zIndex = element.style.zIndex ? parseInt(element.style.zIndex) : 0;
+    let child: HTMLElement | null = element;
+    while (child && getTagOfNode(child) !== 'BODY') {
+        const parent: HTMLElement | null = child?.parentElement;
+        if (parent) {
+            const parentZIndex = parent.style?.zIndex ? parseInt(parent.style.zIndex) : 0;
+            if (parentZIndex > zIndex) {
```
JiuqingSong, line 15: no, what we need is not the largest z index, but the last z index. e.g.
```html
<body>
 <div zIndex = 1>
   <div zIndex = 100>
    ...
```
Although 1 < 100, but since 1 is the last zindex, we use it instead of 100

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -407,11 +408,23 @@ export default class ImageEdit implements EditorPlugin {
...
+        wrapper: HTMLSpanElement,
+        scale: number
+    ) {
         this.zoomWrapper = this.copyImageSize(image, this.createZoomWrapper(wrapper, scale));
+        const editorDiv = this.getEditorDiv(editor);
+        this.zoomWrapper.style.zIndex = `${getBiggestZIndex(editorDiv) + 1}`;
         this.editor.getDocument().body.appendChild(this.zoomWrapper);
     }
 
+    private getEditorDiv(editor: IEditor) {
```
JiuqingSong, line 423: Let's try to avoid getting editor div directly. Plugin should not know editor DIV at all.

You can use editor.getScrollContainer() as the first element.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/getLastZIndex.ts
```diff
@@ -0,0 +1,22 @@
+import { getTagOfNode } from 'roosterjs-editor-dom';
+
+/**
```
JiuqingSong, line 3: Please add "@internal" in comment

I need to check why  the build tool doesn't fail since "@internal" is required here.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/getLastZIndex.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Search through from editor div to it's root for the latest z-index value
+ * @param editor the editor object
+ * @returns
+ */
+export default function getLatestZIndex(editorDiv: HTMLElement) {
+    let child: HTMLElement | null = editorDiv;
+    let zIndex = child.style.zIndex ? parseInt(child.style.zIndex) : 0;
```
JiuqingSong, line 10: you can simplify this to something like:
```ts
let zIndex = 0;
while (child && getTagOfNode(child) != 'BODY') {
  if (child.style.zIndex) {
     zIndex = parseInt(child.style.zIndex);
  }
  child = child.parentElement;
}
```

PR #1444, opened 2022-11-30
Allow Shift-Alt-<Left|Right> to indent/outdent list

ianeli1:
Merges the following ContentEdit features from `listFeatures.ts`:
 - `indentWhenTab`
 - `outdentWhenShiftTab`

Into a new `outdentIndentListItem` feature which also allows Shift-Alt-Right to perform the action of indenting and Shift-Alt-Left for outdenting


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -556,8 +554,7 @@ export const ListFeatures: Record<
     BuildInEditFeature<PluginKeyboardEvent>
 > = {
     autoBullet: AutoBullet,
-    indentWhenTab: IndentWhenTab,
-    outdentWhenShiftTab: OutdentWhenShiftTab,
```
JiuqingSong, line 560: This is already a public interface, remove item from it will cause a breaking change

ianeli1: Ok, I'll add a new feature

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -68,30 +68,45 @@ const ListStyleDefinitionMetadata = createObjectDefinition<ListStyleMetadata>(
...
+};
+
 /**
  * IndentWhenTab edit feature, provides the ability to indent current list when user press TAB
  */
 const IndentWhenTab: BuildInEditFeature<PluginKeyboardEvent> = {
-    keys: [Keys.TAB],
-    shouldHandleEvent: (event, editor) =>
-        !event.rawEvent.shiftKey && cacheGetListElement(event, editor),
+    keys: [Keys.TAB, Keys.RIGHT],
```
JiuqingSong, line 90: What if RTL?

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -68,30 +69,50 @@ const ListStyleDefinitionMetadata = createObjectDefinition<ListStyleMetadata>(
...
+    );
+};
+
+const handleIndentationEvent = (indenting: boolean) => (
+    event: PluginKeyboardEvent,
+    editor: IEditor
+) => {
+    const isRTL =
+        event.rawEvent.keyCode !== Keys.TAB &&
+        getComputedStyle(editor.getElementAtCursor(), 'direction') == 'rtl';
```
JiuqingSong, line 93: getComputedStyle() is kind of heavy. So we should only call it when we really need to.

In this case, it means when we know that it is LEFT/RIGHT key with SHIFT key pressed, and there is list, then finally we can check if it is RTL. So that this function is only called when really need.

ianeli1: Shift key check is handled by `shouldHandleIndentationEvent`, getComputedStyle doesn't get called if keycode is TAB

JiuqingSong (approved):

PR #1447, opened 2022-12-01
Allow styled table header cells to be clear formatted

ianeli1:
Whenever content in a table cell is completely selected, a `<br>` element remains unselected, which disrupts the clear format logic. Fix this by considering empty nodes with no text content as valid siblings for block formatting.

Initial:
image

Current:
image

Fix: 
image


packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -143,15 +143,25 @@ function removeNotTableDefaultStyles(element: HTMLTableElement) {
...
     while (currentNode && editor.contains(currentNode.parentNode)) {
         if (currentNode.nextSibling || currentNode.previousSibling) {
-            return safeInstanceOf(currentNode, 'HTMLLIElement');
+            if (safeInstanceOf(currentNode, 'HTMLLIElement')) {
+                return true;
+            }
+            let isOnlySiblingWithContent = true;
+            currentNode.parentNode?.childNodes.forEach(
+                node =>
+                    !node.isSameNode(currentNode) &&
```
JiuqingSong, line 160: I think you just need to check node != currentNode.

isSameNode() will return true if they have same attributes but not the same reference

packages/roosterjs-editor-api/lib/format/clearFormat.ts
```diff
@@ -143,15 +143,25 @@ function removeNotTableDefaultStyles(element: HTMLTableElement) {
...
+                    node.textContent?.length &&
+                    (isOnlySiblingWithContent = false)
```
JiuqingSong, line 162: Please use a normal if () {... } block to make it easier to read.

JiuqingSong (approved):

PR #1457, opened 2022-12-06
Call set font and toggle bold

huahuahu:
Draft pr for a bug report repo


huahuahu:
@JiuqingSong When would Content Model be available?

JiuqingSong:
> @JiuqingSong When would Content Model be available?

We are actively working on it. The first public version should be done in 2 months.

PR #1463, opened 2022-12-09
Clear local file paths and remove link

juliaroldi:
Only non local files must be pasted, then when paste, sanitize the link to remove local paths.


packages/roosterjs-editor-plugins/lib/plugins/Paste/sanitizeLinks/sanitizeLinks.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * @internal
+ * Clear local paths and remove link
+ * @param sanitizingOption the sanitizingOption of BeforePasteEvent
+ * */
+export default function sanitizeLinks(sanitizingOption: Required<HtmlSanitizerOptions>) {
+    ['href'].forEach(property => {
+        chainSanitizerCallback(
+            sanitizingOption.attributeCallbacks,
+            property,
```
BryanValverdeU, line 16: I think we can just add 'href' property here since we are only sanitizing a single property

BryanValverdeU (approved):

packages/roosterjs-editor-plugins/lib/plugins/Paste/sanitizeLinks/sanitizeLinks.ts
juliaroldi: Oh my god!  I copied a code I did in the past and didn't noticed that. Thanks.

PR #1472, opened 2022-12-22
Selection metadata

juliaroldi:
Instead of using id attribute, use the unique-id metadata to store selection data id. Then the original element id will be not replaced.


JiuqingSong:
If there is already id on image (and if it is unique in editor), we should just reuse it and no need to apply a new id, so that it won't be overwritten.

juliaroldi:
> If there is already id on image (and if it is unique in editor), we should just reuse it and no need to apply a new id, so that it won't be overwritten.

The user need to keep the original id. If the user inserts two images with the same id, they want the id keep the same from I what I understood.

JiuqingSong:
> > If there is already id on image (and if it is unique in editor), we should just reuse it and no need to apply a new id, so that it won't be overwritten.
> 
> The user need to keep the original id. If the user inserts two images with the same id, they want the id keep the same from I what I understood. If we do not need keep duplicated id, I think this fix is enough: #1480

User need to guarantee inserted content has unique id. If there is duplicated id, it is ok to change it.

PR #1474, opened 2022-12-27
Add #1469 again

BryanValverdeU:
Add #1469 again without the position change in ensureTypeInContainer

### DIFF
--- a/packages/roosterjs-editor-core/lib/coreApi/ensureTypeInContainer.ts
+++ b/packages/roosterjs-editor-core/lib/coreApi/ensureTypeInContainer.ts
@@ -90,7 +90,7 @@ export const ensureTypeInContainer: EnsureTypeInContainer = (
 
     // If this is triggered by a keyboard event, let's select the new position
     if (keyboardEvent) {
-        core.api.selectRange(core, createRange(position));
+        core.api.selectRange(core, createRange(new Position(position)));
     }
 };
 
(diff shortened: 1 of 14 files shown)


packages/roosterjs-editor-core/lib/coreApi/ensureTypeInContainer.ts
```diff
@@ -90,7 +90,7 @@ export const ensureTypeInContainer: EnsureTypeInContainer = (
 
     // If this is triggered by a keyboard event, let's select the new position
     if (keyboardEvent) {
-        core.api.selectRange(core, createRange(position));
+        core.api.selectRange(core, createRange(new Position(position)));
```
juliaroldi, line 93: What is the difference in position to `new Position(position)`?

BryanValverdeU: The difference is that since with the ExpFeature Enabled, we are invalidating the Position, due to wrapping the text node:

Example with flight off:

Position input:

```
<Position> {
 element: div,
 node: #TextNode
}
```

HTML

```
<div> 
#TextNode
</div>
```

Position After:
```
<Position> {
 element: div,
 node: #TextNode
}
```
This is correct.

But since we wrap the text node when the exp Feature is On, the Position above is invalid:

Position Before:
```
<Position> {
 element: div,
 node: #TextNode
}
```

HTML after wrapping:
```
<div> 
  <span ...defaultFormat>
       #TextNode
  </span>
</div>
```

Position After wrapping, which is invalid as the `element` Prop should be the span wrapping the Text Node:
```
<Position> {
 element: div,
 node: #TextNode
}
```

When we call new Position again, the Position is fixed and returns:

```
<Position> {
 element: span,
 node: #TextNode
}
```

juliaroldi: I see. You are updating the position object.

juliaroldi (approved):

PR #1482, opened 2023-01-05
Add a span inside of div when backspacing to previous line.

BryanValverdeU:
Add a span inside of div when backspacing to previous line, in order to keep the default format in the State


packages/roosterjs-editor-core/lib/corePlugins/TypeInContainerPlugin.ts
```diff
@@ -84,5 +101,28 @@ export default class TypeInContainerPlugin implements EditorPlugin {
...
+            const element = this.editor?.getElementAtCursor();
+            const block =
+                element &&
+                this.editor?.getBlockElementAtNode(element)?.getStartNode().previousSibling;
+
+            if (block) {
+                this.editor?.runAsync(editor => {
+                    const position = editor.getFocusedPosition();
+                    if (position && block == position.element) {
+                        editor.ensureTypeInContainer(position, event.rawEvent);
```
JiuqingSong, line 122: If user changed to another format, will this go back to default format?

BryanValverdeU: Yes, it will go back to default format, that seem to be the same behavior when the Experimental Feature is OFF.

JiuqingSong (approved):

PR #1485, opened 2023-01-06
Add base64 encoding  to metadata

juliaroldi:
Encode the metadata information using base 64 to hide the metadata implementation and avoid special characters in the HTML Elements. 

Screenshot 2023-01-06 at 13 33 14


JiuqingSong:
Why need  this fix?

PR #1494, opened 2023-01-09
Check if trigger character is inside a paragraph

juliaroldi:
Fix auto format list to never trigger inside a text paragraph. Compare the inline text and check if the only text type is the trigger character, if yes trigger the list, if not do not trigger. Also add convert alpha to decimals to check if alpha list can keep the previous list. 

listTextP


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -476,7 +476,9 @@ function shouldTriggerList(
 ) {
     const searcher = editor.getContentSearcherOfCursor(event);
     const textBeforeCursor = searcher.getSubStringBefore(4);
-    const itHasSpace = /\s/g.test(textBeforeCursor);
+    const traverser = editor.getBlockTraverser();
+    const text = traverser.currentBlockElement.getTextContent();
```
BryanValverdeU, line 480: traverser and currentBlockElement can be null

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/utils/getAutoNumberingListStyle.ts
```diff
@@ -140,7 +141,12 @@ export default function getAutoNumberingListStyle(
     //Only the staring items ['1', 'a', 'A', 'I', 'i'] must trigger a new list. All the other triggers is used to keep the list chain.
     //The index is always the character before the last character
     const listIndex = trigger[trigger.length - 2];
-    const index = parseInt(listIndex);
+    const indexNumber = parseInt(listIndex);
+    let index = !isNaN(indexNumber) ? indexNumber : convertAlphaToDecimals(listIndex);
+
+    if (!index || index < 1) {
+        return null;
```
BryanValverdeU, line 148: Should we also add null to the return type of this function?

JiuqingSong (approved):

PR #1500, opened 2023-01-11
fix image plugin z-index calc

shaipetel:
Fixed the z-index calculation for image edit plugin.
1. childZIndex was not being used when assigning the value to zIndex.
2. in situations where the parent might have a lower z-index than a child element that was containing the rooster editor, the lower number was returned. Added a check to return the higher value using Math.max

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/getLastZIndex.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/getLastZIndex.ts
@@ -11,7 +11,7 @@ export default function getLatestZIndex(editorDiv: HTMLElement) {
     let zIndex = 0;
     while (child && getTagOfNode(child) !== 'BODY') {
         const childZIndex = parseInt(child.style.zIndex || getComputedStyle(child).zIndex, 10);
-        if (childZIndex && !isNaN(childZIndex)) {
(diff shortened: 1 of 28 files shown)


JiuqingSong:
@juliaroldi please take a look. Thanks

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/getLastZIndex.ts
```diff
@@ -10,9 +10,9 @@ export default function getLatestZIndex(editorDiv: HTMLElement) {
     let child: HTMLElement | null = editorDiv;
     let zIndex = 0;
     while (child && getTagOfNode(child) !== 'BODY') {
-        const childZIndex = child.style.zIndex || getComputedStyle(child).zIndex;
-        if (childZIndex) {
-            zIndex = parseInt(child.style.zIndex);
+        const childZIndex = parseInt(child.style.zIndex || getComputedStyle(child).zIndex, 10);
+        if (childZIndex && !isNaN(childZIndex)) {
```
juliaroldi, line 14: If childZIndex is NaN, the condition is already false, so no need `&& !isNaN(childZIndex)`

shaipetel:
Updated as per comment

juliaroldi (approved):

PR #1510, opened 2023-01-17
Fix for image selection

juliaroldi:
Bug: if ImageEditPlugin and TableCellSelection are disable, image selection does work properly. 
After an image is selected, is not possible unselect it. 

Fix: verify if a element was click and if the a image is selected, then if the condition is true, select the element selected. 

imageEditBUGTEST

### DIFF
--- a/packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
@@ -62,17 +62,17 @@ export default class ImageSelection implements EditorPlugin {
                         } else if (event.rawEvent.button === mouseLeftButton) {
                             this.editor.select(target);
                         }
-                    } else if (safeInstanceOf(target, 'HTMLElement')) {
-                        const mouseSelection = this.editor.getSelectionRangeEx();
-                        const range = this.editor.getSelectionRange();
-                        if (
-                            range &&
-                            mouseSelection &&
-                            mouseSelection.type === SelectionRangeTypes.ImageSelection
-                        ) {
-                            this.editor.select(range);
-                            event.rawEvent.stopPropagation();
-                        }
+                    }
+                    break;
+                case PluginEventType.MouseDown:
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
```diff
@@ -62,6 +62,17 @@ export default class ImageSelection implements EditorPlugin {
                         } else if (event.rawEvent.button === mouseLeftButton) {
                             this.editor.select(target);
                         }
+                    } else if (safeInstanceOf(target, 'HTMLElement')) {
+                        const mouseSelection = this.editor.getSelectionRangeEx();
```
JiuqingSong, line 66: Can we just let the browser handle the mouse selection?

I think you can check this in MouseDown event instead, and if target element is not current image, just unselect it, then browser should be able to handle the rest.

juliaroldi: I understand. I also think we need to check if the last selection is a image to not cause problems with table cell selection.

packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
```diff
@@ -64,6 +64,17 @@ export default class ImageSelection implements EditorPlugin {
                         }
                     }
                     break;
+                case PluginEventType.MouseDown:
+                    const mouseTarget = event.rawEvent.target;
+                    const mouseSelection = this.editor.getSelectionRangeEx();
+                    if (
+                        !safeInstanceOf(mouseTarget, 'HTMLImageElement') &&
+                        mouseSelection &&
+                        mouseSelection.type === SelectionRangeTypes.ImageSelection
```
JiuqingSong, line 73: You just need to check if the target element is same with current selected image:

if (mouseSelection.type != Image || mouseSelection.image != event.rawEvent.target) {
  select(null)

JiuqingSong (approved):

Issue #1503, opened 2023-01-12
[Image Resize] It exits the image editing mode automatically on scrolling


milliezhang-qinlan:
**Describe the bug**
During the process of resizing an image, if scrolling the editor, it will exit the image editing mode automatically and the resize handle disappears.

I find it is because it triggers `this.setEditingImage(null);` in `onPluginEvent` of `ImageEdit`
<img width="953" alt="Screenshot 2023-01-12 at 18 25 44" src="https://user-images.githubusercontent.com/63031370/212042538-fb937488-74f1-4145-bb7c-2efcac62863f.png">

For the previous versions of roosterjs, it will not exit the resize mode on scorlling. And when resizing an image on mobile, it would scroll a lot during the process of resizing an image. 

Correct me if I am wrong. I guess the reason to do so is to ensure the image displayed at a right position on scroll? 🤔 If so I am wondering if there is a way to update the image wrapper position instead of exiting the image editing mode on scrolling.

And there is also another issue, after exiting the resize mode on scrolling, the image has a red border:
<img width="1290" alt="image" src="https://user-images.githubusercontent.com/63031370/212044502-ef8685d1-67ee-40ea-a025-7c9fdfdd9a33.png">

<img width="279" alt="image" src="https://user-images.githubusercontent.com/63031370/212044799-1aea966c-a192-4ca6-922f-e5484f7c6ef8.png">

**To Reproduce**
1. Go to roosterjs sample site
2. Insert an image and click on it and resize
3. Do not exit resize mode and scroll the editor

**Expected behavior**
The image should be in resize mode with resize handle displayed and in a correct position after scroll.

**Screenshots**

**Device Information**
 - Browser: Chrome

JiuqingSong:
Thanks for reporting.

@juliaroldi  is working on this issue.

PR #1520, opened 2023-01-19
U/juliaroldi/shadow dom image wrapper

juliaroldi:


JiuqingSong:
1. We only need to disable drag when in editing state. If image is not in editing (editing wrapper not shown), drag and drop can should still work.
2. 
image

Create a long rectangle image, insert it. Then rotate, I think the size of wrapper SPAN is wrong (or incorrectly rotated?) That causes the image float higher than other text. Maybe not a new issue, but better also fix it.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -163,7 +157,10 @@ export default class ImageEdit implements EditorPlugin {
      */
     initialize(editor: IEditor) {
         this.editor = editor;
-        this.disposer = editor.addDomEventHandler('blur', this.onBlur);
+        this.disposer = editor.addDomEventHandler({
+            blur: () => this.onBlur,
+            drop: e => e.preventDefault(),
```
JiuqingSong, line 162: I think we should disable onDrag instead, and only disable when image is in editing.

JiuqingSong:
#1503

JiuqingSong (approved):

PR #1528, opened 2023-01-25
End of word format fix

Andres-CT98:
Fix scenario where format was wrongly applied where the cursor was located at the end of a word


packages/roosterjs-content-model/lib/modelApi/selection/adjustWordSelection.ts
```diff
@@ -71,7 +71,7 @@ export function adjustWordSelection(
         }
 
         // Marker is at end of word
-        if (segments.length <= 2) {
+        if (segments[segments.length - 1] == marker) {
```
JiuqingSong, line 74: Let's add a test case to cover this case

JiuqingSong (approved):

PR #1533, opened 2023-01-26
Fix #1532: Support isCode in FormatState

### LINKED ISSUE #1532: `isCodeBlock` missing from `FormatState`
**Description**
The roosterjs-editor-api provides a [`toggleCodeBlock`](https://microsoft.github.io/roosterjs/docs/modules/roosterjs_editor_api.html#togglecodeblock), but there is no corresponding `isCodeBlock` in [`FormatState`](https://microsoft.github.io/roosterjs/docs/interfaces/roosterjs_editor_types.formatstate.html).

**Expected behavior**
I expect there to be an `isCodeBlock` just like there is a [`isBlockQuote`](https://microsoft.github.io/roosterjs/docs/interfaces/roosterjs_editor_types.formatstate.html#isblockquote) for [`toggleBlockQuote`](https://microsoft.github.io/roosterjs/docs/modules/roosterjs_editor_api.html#toggleblockquote).

JiuqingSong:
Support isCode in FormatState.

This change doesn't include the format state change for Content Model. It will be covered later.

### DIFF
--- a/packages/roosterjs-editor-api/lib/format/getFormatState.ts
+++ b/packages/roosterjs-editor-api/lib/format/getFormatState.ts
@@ -49,6 +49,7 @@ export function getElementBasedFormatState(
         canUnlink: !!editor.queryElements('a[href]', QueryScope.OnSelection)[0],
         canAddImageAltText: !!editor.queryElements('img', QueryScope.OnSelection)[0],
         isBlockQuote: !!editor.queryElements('blockquote', QueryScope.OnSelection)[0],
+        isCode: !!editor.queryElements('code', QueryScope.OnSelection)[0],
         isInTable: !!table,
         tableFormat: tableFormat,
         tableHasHeader: hasHeader,
--- a/packages/roosterjs-editor-api/lib/format/getFormatState.ts
+++ b/packages/roosterjs-editor-api/lib/format/getFormatState.ts
@@ -49,7 +49,7 @@ export function getElementBasedFormatState(
         canUnlink: !!editor.queryElements('a[href]', QueryScope.OnSelection)[0],
         canAddImageAltText: !!editor.queryElements('img', QueryScope.OnSelection)[0],
         isBlockQuote: !!editor.queryElements('blockquote', QueryScope.OnSelection)[0],
-        isCode: !!editor.queryElements('code', QueryScope.OnSelection)[0],
+        isCodeBlock: !!editor.queryElements('pre>code', QueryScope.OnSelection)[0],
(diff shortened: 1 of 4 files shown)


packages/roosterjs-editor-types/lib/interface/FormatState.ts
```diff
@@ -59,6 +59,11 @@ export interface ElementBasedFormatState {
      */
     isBlockQuote?: boolean;
 
+    /**
+     * Whether the text is in Code block
+     */
+    isCode?: boolean;
```
stonebk, line 65: Should it be is**CodeBlock** for congruency with toggle**CodeBlock**?

Not sure if you would want to reserve the shorter `isCode` for inline code:

This is `inline code` formatting vs...

```
code
block
formatting
````

File: packages/roosterjs-editor-api/lib/format/getFormatState.ts [hunk 2]
```diff
@@ -49,6 +49,7 @@ export function getElementBasedFormatState(
         canUnlink: !!editor.queryElements('a[href]', QueryScope.OnSelection)[0],
         canAddImageAltText: !!editor.queryElements('img', QueryScope.OnSelection)[0],
         isBlockQuote: !!editor.queryElements('blockquote', QueryScope.OnSelection)[0],
+        isCode: !!editor.queryElements('code', QueryScope.OnSelection)[0],
```
stonebk, line 52: Should this be testing for the `pre` element rather than `code`? In other RTEs I've seen `pre` used for code blocks and `code` used for inline code.

JiuqingSong: It depends on if you want to treat a pure PRE tag as code. It can happen if you copy from some content that contains PRE tag and paste, so there will be PRE but the content may not be code. But if you toggle code block from roosterjs, there will always be both PRE and CODE.

I personally fill both CODE and PRE is fine to me, so just depends on your requirement.

packages/roosterjs-editor-types/lib/interface/FormatState.ts
JiuqingSong: I'm ok with isCodeBlock.

packages/roosterjs-editor-api/lib/format/getFormatState.ts
JiuqingSong: I changed to check `PRE > CODE` instead

juliaroldi (approved):

PR #1540, opened 2023-01-30
Fix list trigger before a paragraph

juliaroldi:
If a trigger character is added at the beginning of the sentence, trigger the list. 
This change should verify if the 4 first characters of a text is equal to a trigger character.  
fixListTrigger

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
@@ -479,7 +479,7 @@ function shouldTriggerList(
     const traverser = editor.getBlockTraverser();
     const text =
         traverser && traverser.currentBlockElement
-            ? traverser.currentBlockElement.getTextContent()
+            ? traverser.currentBlockElement.getTextContent().slice(0, textBeforeCursor.length)
             : null;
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -479,7 +479,7 @@ function shouldTriggerList(
     const traverser = editor.getBlockTraverser();
     const text =
         traverser && traverser.currentBlockElement
-            ? traverser.currentBlockElement.getTextContent()
+            ? traverser.currentBlockElement.getTextContent().slice(0, textBeforeCursor.length)
```
JiuqingSong, line 482: I remember we were using PositionContentSearcher to help find text before cursor before. Any reason we don't use it now?

juliaroldi: We still use PositionContentSearcher to find the text before cursor, the code I added is to check if the trigger character is the same text on the begging of the paragraph. The aim of this code is to avoid the list being trigger in unexpected positions, since we have more characters that triggers lists.

JiuqingSong (approved):

PR #1543, opened 2023-01-31
Add support for paragraph line spacing for content model

ianeli1:
image

Creates a new ribbon button for setting paragraph spacing in content model editor
Adds a lineHeight CSS property to FormatState
Creates new content model api for setting line height

image

### DIFF
--- a/packages-ui/roosterjs-react/lib/ribbon/type/RibbonButtonStringKeys.ts
+++ b/packages-ui/roosterjs-react/lib/ribbon/type/RibbonButtonStringKeys.ts
@@ -187,6 +187,11 @@ export type UndoButtonStringKey = 'buttonNameUndo';
 export type CellShadeButtonStringKey = 'buttonNameCellShade' | BackgroundColorKeys;
 
 /**
+ * Key of localized strings of Spacing button
+ */
+export type SpacingButtonStringKey = 'buttonNameSpacing';
+
+/**
  * A public type for localized string keys of all buttons
  */
 export type AllButtonStringKeys =
--- a/packages-ui/roosterjs-react/lib/ribbon/type/RibbonButtonStringKeys.ts
+++ b/packages-ui/roosterjs-react/lib/ribbon/type/RibbonButtonStringKeys.ts
@@ -187,11 +187,6 @@ export type UndoButtonStringKey = 'buttonNameUndo';
 export type CellShadeButtonStringKey = 'buttonNameCellShade' | BackgroundColorKeys;
 
 /**
- * Key of localized strings of Spacing button
- */
-export type SpacingButtonStringKey = 'buttonNameSpacing';
-
-/**
  * A public type for localized string keys of all buttons
  */
 export type AllButtonStringKeys =
--- a/packages/roosterjs-content-model/lib/formatHandlers/block/lineHeightFormatHandler.ts
+++ b/packages/roosterjs-content-model/lib/formatHandlers/block/lineHeightFormatHandler.ts
@@ -6,7 +6,27 @@ import { LineHeightFormat } from '../../publicTypes/format/formatParts/LineHeigh
  */
 export const lineHeightFormatHandler: FormatHandler<LineHeightFormat> = {
     parse: (format, element, context, defaultStyle) => {
-        const lineHeight = element.style.lineHeight || defaultStyle.lineHeight;
+        let lineHeight = element.style.lineHeight || defaultStyle.lineHeight;
+        if (
+            !lineHeight &&
+            element.childNodes.length &&
+            element.childNodes.length === element.childElementCount
+        ) {
+            let childLineHeight = (<HTMLElement | null>(
+                element.children.item(0)
+            ))?.style?.getPropertyValue('line-height');
+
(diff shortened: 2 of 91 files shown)


packages-ui/roosterjs-react/lib/ribbon/type/RibbonButtonStringKeys.ts
```diff
@@ -186,6 +186,11 @@ export type UndoButtonStringKey = 'buttonNameUndo';
  */
 export type CellShadeButtonStringKey = 'buttonNameCellShade' | BackgroundColorKeys;
 
+/**
+ * Key of localized strings of Spacing button
+ */
+export type SpacingButtonStringKey = 'buttonNameSpacing';
```
JiuqingSong, line 192: If we don't add new ribbon button to roosterjs-react, no need to add string key here. You can hard code the string in demo code

packages/roosterjs-content-model/lib/formatHandlers/block/lineHeightFormatHandler.ts
```diff
@@ -6,7 +6,27 @@ import { LineHeightFormat } from '../../publicTypes/format/formatParts/LineHeigh
...
     parse: (format, element, context, defaultStyle) => {
-        const lineHeight = element.style.lineHeight || defaultStyle.lineHeight;
+        let lineHeight = element.style.lineHeight || defaultStyle.lineHeight;
+        if (
+            !lineHeight &&
+            element.childNodes.length &&
+            element.childNodes.length === element.childElementCount
+        ) {
+            let childLineHeight = (<HTMLElement | null>(
+                element.children.item(0)
```
JiuqingSong, line 16: Normally, you don't need to handle style from child node. They should be handled when element processor processes child node. If you see HTML like
```html
<div>
  <div style="line-height: ....">test</div>
</div>
```
Outer DIV doesn't have line-height, but that is ok, processor will create an empty paragraph for it but later it will be removed. Line height will be handled when inner DIV is being processed.

packages/roosterjs-content-model/lib/publicApi/block/setSpacing.ts
```diff
@@ -0,0 +1,11 @@
+import { IExperimentalContentModelEditor } from '../../publicTypes/IExperimentalContentModelEditor';
+import { formatParagraphWithContentModel } from '../utils/formatParagraphWithContentModel';
+
+export default function setSpacing(
+    editor: IExperimentalContentModelEditor,
+    spacing: number | string
+) {
+    formatParagraphWithContentModel(editor, 'setSpacing', paragraph => {
+        paragraph.format.lineHeight = spacing.toString();
```
JiuqingSong, line 9: Do we need to add unit here? Such as 'px' or 'pt'?

packages/roosterjs-content-model/lib/formatHandlers/block/lineHeightFormatHandler.ts
JiuqingSong: And if you added some logic here, please also add unit test to cover it.

packages/roosterjs-content-model/lib/publicApi/block/setSpacing.ts
```diff
@@ -0,0 +1,11 @@
+import { IExperimentalContentModelEditor } from '../../publicTypes/IExperimentalContentModelEditor';
+import { formatParagraphWithContentModel } from '../utils/formatParagraphWithContentModel';
+
+export default function setSpacing(
```
JiuqingSong, line 4: Need unit test

packages/roosterjs-content-model/lib/formatHandlers/block/lineHeightFormatHandler.ts
ianeli1: Oh ok, if div only has one child with line-height, no need to process? I see
What about 
```html
<div style="...">
   <div style="line-height: ..."></div>
</div>
```
?

Usually clipboard html from word processors include the line height in a span under p, with both having styles

packages/roosterjs-content-model/lib/publicApi/block/setSpacing.ts
ianeli1: unitless line-height is [the preferred way](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height)
we can still set px or pt if we want through spacing: string

JiuqingSong: sounds good.

packages/roosterjs-content-model/lib/formatHandlers/block/lineHeightFormatHandler.ts
JiuqingSong: I see. The line-height inside SPAN is a problem. Let me think about it.

ianeli1: I decided to allow segments to hold a line-height, seems more straight forward to do here
Opinions on normalizing format after model is generated?

JiuqingSong (approved):

PR #1549, opened 2023-02-01
Add data to the ContentChanged event after inserting in PickerPlugin

BryanValverdeU:
Add HTMLNode and the Word replaced on when triggering the content changed event in the PickerPlugin, so we use that data on the changed event event in other plugins

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
@@ -104,12 +104,18 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
                         this.editor.insertNode(htmlNode);
                     }
                     this.setIsSuggesting(false);
+
+                    return {
+                        wordToReplace,
+                        htmlNode,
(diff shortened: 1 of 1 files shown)


packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -104,6 +104,11 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
                         this.editor.insertNode(htmlNode);
                     }
                     this.setIsSuggesting(false);
+
+                    return {
```
juliaroldi, line 108: How we are going to use this info?

BryanValverdeU: I am planning to use this info to announce to a screen reader the element inserted with PickerPlugin after auto-correcting a emoji.

BryanValverdeU: I think I found another way to do it. Without this change. I will close this now.

PR #1552, opened 2023-02-01
Disable list indentation on mac

ianeli1:
Shift+Option+Left/Right is used in Mac to jump between words, this edit feature is preventing that behavior due to Shift+Alt+L/R matching Shift+Option+L/R on Mac. Disable it there


JiuqingSong (approved):

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -75,6 +75,7 @@ const shouldHandleIndentationEvent = (indenting: boolean) => (
 ) => {
     const { keyCode, altKey, shiftKey, ctrlKey, metaKey } = event.rawEvent;
     return (
+        !Browser.isMac &&
```
BryanValverdeU, line 78: Maybe we can add this check to the defaultDisabled parameter of the ContentEdit Feature instead, Like in `OutdentWhenEnterOnEmptyLine`

BryanValverdeU (approved):

PR #1560, opened 2023-02-06
Add filter to client rects

juliaroldi:
Change `getIntersectedRect` to also filter `additionalRects ` to avoid undefined values.


packages/roosterjs-editor-dom/lib/utils/getIntersectedRect.ts
```diff
@@ -32,14 +32,13 @@ export default function getIntersectedRect(
 ): Rect | null {
     const rects = elements
         .map(element => normalizeRect(element.getBoundingClientRect()))
-        .filter(element => !!element)
         .concat(additionalRects) as Rect[];
```
BryanValverdeU, line 35: I think you can add a single filter after the concat function

BryanValverdeU (approved):

JiuqingSong (approved):

PR #1563, opened 2023-02-07
Content Model: Support insert image with src

JiuqingSong:
Support insert image with src using insertImage API in content model, to match the original API's behavior.


juliaroldi (approved):

juliaroldi:
Should we also do this change to ChangeImage Api?

PR #1565, opened 2023-02-07
Add support for adding/removing space before/after paragraphs in content model

ianeli1:
Adds the ability to call `toggleMargins` public api, which, verifies if a value already exists in the format of paragraph, setting it if not, and removing it otherwise.

Adds two new properties to `FormatState`: `marginTop` & `marginBottom`

image


juliaroldi (approved):

packages/roosterjs-content-model/test/modelApi/common/retrieveModelFormatStateTest.ts
```diff
@@ -443,6 +447,8 @@ describe('retrieveModelFormatState', () => {
             isUnderline: undefined,
             isStrikeThrough: undefined,
             lineHeight: undefined,
+            marginTop: undefined,
+            marginBottom: undefined,
```
JiuqingSong, line 451: Please also add a test case for margin top/bottom has valid value

packages/roosterjs-content-model/lib/publicApi/block/toggleMargins.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Toggles the current block(s) margin properties.
+ * @param editor The editor to operate on
+ * @param marginFormat Object containing margin props to be changed.
+ * If a margin property is not in this object, it won't be changed.
+ */
+export default function toggleMargins(editor: IContentModelEditor, marginFormat: MarginFormat) {
+    formatParagraphWithContentModel(editor, 'toggleMargins', para => {
+        if (!para.decorator || para.decorator.tagName !== 'p') {
+            para.decorator = createParagraphDecorator('p');
```
JiuqingSong, line 16: How do you want to handle header? Because tagName can also be h1-h6. What is the expected behavior when toggle margin on a header?

packages/roosterjs-content-model/lib/publicApi/block/toggleMargins.ts
```diff
@@ -0,0 +1,29 @@
...
+import { IContentModelEditor } from '../../publicTypes/IContentModelEditor';
+import { formatParagraphWithContentModel } from '../utils/formatParagraphWithContentModel';
+
+/**
+ * Toggles the current block(s) margin properties.
+ * @param editor The editor to operate on
+ * @param marginFormat Object containing margin props to be changed.
+ * If a margin property is not in this object, it won't be changed.
+ */
+export default function toggleMargins(editor: IContentModelEditor, marginFormat: MarginFormat) {
```
JiuqingSong, line 13: Do we want to support margin left and right here? If not, you can make the parameter for margin top and bottom only

packages/roosterjs-content-model/lib/publicApi/block/toggleMargins.ts
```diff
@@ -0,0 +1,29 @@
...
+    formatParagraphWithContentModel(editor, 'toggleMargins', para => {
+        if (!para.decorator || para.decorator.tagName !== 'p') {
+            para.decorator = createParagraphDecorator('p');
+        }
+        getObjectKeys(marginFormat).forEach(key => {
```
JiuqingSong, line 18: what if marginFormat only contain marginTop but existing format contains marginBottom? Do we want to delete it?

getObjectKeys() can only return keys that exists in marginFormat

packages/roosterjs-content-model/lib/publicApi/block/toggleMargins.ts
```diff
@@ -0,0 +1,29 @@
...
+            if (
+                para.format[key] &&
+                (para.format[key] === marginFormat[key] || parseInt(para.format[key]!) > 0)
```
JiuqingSong, line 21: why delete para.format[key] when format[key] == marginFormat[key]?

packages/roosterjs-content-model/lib/publicApi/block/toggleMargins.ts
JiuqingSong, line 13: I think name it "setMargins" is better

JiuqingSong (changes requested):

packages/roosterjs-content-model/lib/publicApi/block/toggleMargins.ts
ianeli1: To answer this and "I think name it "setMargins" is better", if we make it delete the margin format when format is already set, we can toggle the value, no need to get formatState to apply margin. Also removes a positive margin (coming from copy paste for example)

packages/roosterjs-content-model/lib/publicApi/block/toggleMargins.ts
ianeli1: I was going for a function that let us toggle marginTop and/or marginBottom in one implementation. So if a key isn't passed, it will be ignored

packages/roosterjs-content-model/lib/publicApi/block/toggleMargins.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Toggles the current block(s) margin properties.
+ * @param editor The editor to operate on
+ * @param marginFormat Object containing margin props to be changed.
+ * If a margin property is not in this object, it won't be changed.
+ */
+export default function toggleMargins(
+    editor: IContentModelEditor,
+    marginFormat: Pick<MarginFormat, 'marginTop' | 'marginBottom'>
```
JiuqingSong, line 17: I think the easiest way to organize the parameters is to have 2 separate parameters for marginTop and margin Bottom.

Both of the are allowed to be undefined to keep current value.

packages/roosterjs-content-model/lib/publicApi/block/toggleMargins.ts
```diff
@@ -0,0 +1,34 @@
...
+) {
+    formatParagraphWithContentModel(editor, 'toggleMargins', para => {
+        if (!para.decorator || tagNamesSupportingMargins.indexOf(para.decorator.tagName) < 0) {
+            para.decorator = createParagraphDecorator('p');
```
JiuqingSong, line 21: Although today we don't support other tags then the listed ones, but it may be a problem to recreate the whole decorator to replace the existing one. Can we just check if decorator is null, then create one with P, otherwise just leave it there?

packages/roosterjs-content-model/lib/publicApi/block/toggleMargins.ts
JiuqingSong: Let' just have separate parameters and handle them one by one.

Using getObjectKeys() may accidentally overwrite properties that we don't want to if user pass in a larger object then expected parameter type as sub class.

packages/roosterjs-content-model/lib/publicApi/block/toggleMargins.ts
JiuqingSong: Hmm, I think that make cause some confusion to user. If user doesn't want to change the margin, they need to be very careful to make sure not passing in a value that is the same with existing value. So user first need to get the existing value to check if it is the same with the one they want to check.

When a function is named "toggle", it means user doesn't need to know the original value, but just call it to turn on or off.

So still suggest:
1. name the function "setMargin" or "setParagraphMargin"
2. have two separate parameters for top and bottom
3. when pass in a valide value for top/bottom, apply it, no matter if it is the same with existing value
4. when pass in undefined, don't change
5. when pass in null or 0, delete the margin
so that is easier to understand and use.

JiuqingSong: We have an API "toggleHeader" in original roosterjs code which is a bad example. In content model I have renamed it to be "setHeaderLevel"

ianeli1: Got it, went ahead with all of your fixes
Had to add support for passing formatState in onClick callback for the spaceBeforeAfter buttons to remain simple

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
```diff
@@ -92,10 +92,10 @@ class RibbonPluginImpl implements RibbonPlugin {
         key: T,
         strings?: LocalizedStrings<T>
     ) {
-        if (this.editor && this.uiUtilities) {
+        if (this.editor && this.uiUtilities && this.formatState) {
```
JiuqingSong, line 95: If this.formatState is null (which is possible in some cases), it will cause button click not work. We should not add this parameter.

If a button needs to know current format state, it can call getFormatState(editor), plugin doesn't need to provide it.

demo/scripts/controls/ribbonButtons/contentModel/spaceBeforeAfterButtons.ts
```diff
@@ -0,0 +1,47 @@
...
+ * "Add space before" button on the format ribbon
+ */
+export const spaceBeforeButton: RibbonButton<typeof spaceBeforeButtonKey> = {
+    key: spaceBeforeButtonKey,
+    unlocalizedText: 'Add space before',
+    iconName: 'CaretUp8',
+    isChecked: formatState => parseInt(formatState.marginTop) > 0,
+    onClick: (editor, _, _1, _2, { marginTop }) => {
+        if (isContentModelEditor(editor)) {
+            setParagraphMargin(
```
JiuqingSong, line 39: If you want to do a toggle, you can get format state using getFormatState() here. Do not let plugin pass the format to you. Or if you really want, make it an optional parameter so that we know it can be null.

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
JiuqingSong: button onClick should still be called even formatState is null

JiuqingSong: Here is my repro
buttonbug

JiuqingSong: Here is my repro
buttonbug

JiuqingSong (approved):

PR #1568, opened 2023-02-08
Check if there is rotate  handle, before update the rotate handles

juliaroldi:
For some editing operations, the rotate handle are not rendered. Then we need to check if they exist, before we update them.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
@@ -499,13 +499,15 @@ export default class ImageEdit implements EditorPlugin {
                     this.updateWrapper();
                 }
 
-                updateRotateHandlePosition(
-                    this.editInfo,
-                    this.editor.getVisibleViewport(),
-                    marginVertical,
-                    rotateCenter,
-                    rotateHandle
-                );
+                if (rotateHandle && rotateCenter) {
+                    updateRotateHandlePosition(
+                        this.editInfo,
+                        this.editor.getVisibleViewport(),
+                        marginVertical,
+                        rotateCenter,
+                        rotateHandle
+                    );
+                }
 
                 updateHandleCursor(resizeHandles, angleRad);
             }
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/imageEditors/Rotator.ts
```diff
@@ -51,19 +51,22 @@ export function updateRotateHandlePosition(
...
     rotateHandle: HTMLElement
 ) {
-    const top = rotateHandle.getBoundingClientRect()?.top - editorRect?.top;
-    const { angleRad, heightPx } = editInfo;
-    const cosAngle = Math.cos(angleRad);
-    const adjustedDistance =
-        cosAngle <= 0
-            ? Number.MAX_SAFE_INTEGER
-            : (top + heightPx / 2 + marginVertical) / cosAngle - heightPx / 2;
+    const rotateHandleRect = rotateHandle.getBoundingClientRect();
```
BryanValverdeU, line 54: Can we also add a null check in rotate handle, before calling getBoundingClientRect? I think the callstack is saying that rotateHandle could be null
```
Uncaught TypeError: Cannot read properties of undefined (reading 'getBoundingClientRect')

File: packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts [hunk 2]
```diff
@@ -499,13 +499,15 @@ export default class ImageEdit implements EditorPlugin {
...
                 }
 
-                updateRotateHandlePosition(
-                    this.editInfo,
-                    this.editor.getVisibleViewport(),
-                    marginVertical,
-                    rotateCenter,
-                    rotateHandle
-                );
+                if (rotateHandle && rotateCenter) {
```
juliaroldi, line 502: @BryanValverdeU I'm doing the check for the rotateHandle and rotateCenter here. If they do exist the `updateRotateHandlePosition` will not be called.

BryanValverdeU: Oh I see!
Can we also add to the null check this.editor.getVisibleViewport result? Since the ReturnType is `Rect | null`

juliaroldi: That is a good observation. I'll add it 👍🏻

BryanValverdeU (approved):

PR #1576, opened 2023-02-15
Content Model: Set alignment

juliaroldi:
Add setModelAlignment api that supports align lists and tables. 
alignment


packages/roosterjs-content-model/lib/publicTypes/format/formatParts/DirectionFormat.ts
```diff
@@ -16,4 +16,9 @@ export type DirectionFormat = {
      * Whether the text align value comes from attribute rather than CSS
      */
     isTextAlignFromAttr?: boolean;
+
+    /**
+     * Horizontal alignment of an element
+     */
+    alignSelf?: string;
```
JiuqingSong, line 23: Instead of adding this raw CSS here, can we just respect the single `textAlign` property?

Content Model property doesn't need to 1:1 map to CSS property, but the goal is to be easier to do format, that means, when change alignment, we just need to change single property, then let content processor/handler to handle different CSS properties when create Content Model or write back

packages/roosterjs-content-model/lib/publicTypes/format/formatParts/DisplayFormat.ts
```diff
@@ -6,4 +6,9 @@ export type DisplayFormat = {
      * Display of this element
      */
     display?: string;
+
+    /**
+     * If display flex, the flex direction
+     */
+    flexDirection?: string;
```
JiuqingSong, line 13: same here

packages/roosterjs-content-model/lib/publicTypes/format/ContentModelSegmentFormat.ts
```diff
@@ -13,6 +14,7 @@ import { UnderlineFormat } from './formatParts/UnderlineFormat';
  * The format object for a segment in Content Model
  */
 export type ContentModelSegmentFormat = TextColorFormat &
+    DirectionFormat &
```
JiuqingSong, line 17: Why need to add DirectionFormat to SegmentFormat?

packages/roosterjs-content-model/lib/modelApi/block/setModelAlignment.ts
```diff
@@ -0,0 +1,80 @@
...
+ * @internal
+ */
+export function setModelAlignment(
+    model: ContentModelDocument,
+    alignment: 'left' | 'center' | 'right'
+) {
+    const paragraphOrListItemOrTable = getOperationalBlocks<ContentModelListItem>(
+        model,
+        ['ListItem'],
+        ['TableCell']
```
JiuqingSong, line 35: Do we really need to add TableCell here?

If full table is selected, I think `getOperationalBlocks` should give you the whole table block directly.

packages/roosterjs-content-model/lib/modelApi/block/setModelAlignment.ts
```diff
@@ -0,0 +1,80 @@
...
+    );
+
+    paragraphOrListItemOrTable.forEach(({ block }) => {
+        if (isBlockGroupOfType<ContentModelListItem>(block, 'ListItem')) {
+            block.formatHolder.format.alignSelf =
+                ResultMap[alignment][block.format.direction == 'rtl' ? 'rtl' : 'ltr'];
+            block.levels.forEach(level => {
+                level.flexDirection = 'column';
```
JiuqingSong, line 43: Maybe we only need to change format of the last level

packages/roosterjs-content-model/lib/modelApi/block/setModelAlignment.ts
```diff
@@ -0,0 +1,80 @@
...
+                level.display = 'flex';
+            });
+        } else if (block.blockType === 'Table') {
+            const isWholeTableSelected = block?.cells.every(row =>
```
JiuqingSong, line 47: I think when table is returned, it must be full table selected, so no need to check here.

You can verify.

packages/roosterjs-content-model/lib/modelApi/block/setModelAlignment.ts
```diff
@@ -0,0 +1,80 @@
...
+        } else if (block) {
+            const { format } = block;
+            format.textAlign = ResultMap[alignment][format.direction == 'rtl' ? 'rtl' : 'ltr'];
+        }
+    });
+
+    return paragraphOrListItemOrTable.length > 0;
+}
+
+function alignTable(
```
JiuqingSong, line 62: We already have a `alignTable` function, you can just call to it https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs-content-model/lib/modelApi/table/alignTable.ts

packages/roosterjs-content-model/lib/formatHandlers/block/displayFormatHandler.ts
```diff
@@ -7,14 +7,23 @@ import { FormatHandler } from '../FormatHandler';
 export const displayFormatHandler: FormatHandler<DisplayFormat> = {
     parse: (format, element) => {
         const display = element.style.display;
+        const flexDirection = element.style.flexDirection;
 
         if (display) {
             format.display = display;
         }
+
+        if (flexDirection) {
```
JiuqingSong, line 16: Try merge this into direction handler. if it has flexDirection and display is flex, use it, otherwise fallback to textAlign.

packages/roosterjs-content-model/lib/formatHandlers/block/directionFormatHandler.ts
```diff
@@ -51,6 +52,10 @@ export const directionFormatHandler: FormatHandler<DirectionFormat> = {
         if (alignFromAttr && !element.style.textAlign) {
             format.isTextAlignFromAttr = true;
         }
+
+        if (alignSelf) {
+            format.alignSelf = alignSelf;
```
JiuqingSong, line 57: Same here, try just use one single property for direction.

packages/roosterjs-content-model/lib/publicTypes/format/ContentModelSegmentFormat.ts
juliaroldi: We need it so it can use to set alignment in the formatHolder

JiuqingSong: In that case, this format should not be part of the formatHolder since formatHolder only handles inline format. (I realized that using a selectionMarker as format holder may not be a good idea, let me change this in separate change).

Actually ListItem.format is already ContentModelBlockFormat which contains DirectionFormat

packages/roosterjs-content-model/lib/publicTypes/format/ContentModelListItemLevelFormat.ts
```diff
@@ -9,4 +10,5 @@ import { ListTypeFormat } from './formatParts/ListTypeFormat';
 export type ContentModelListItemLevelFormat = ListTypeFormat &
     ListThreadFormat &
     ListMetadataFormat &
+    DisplayFormat &
```
JiuqingSong, line 13: I think you don't need to explicitly add DisplayFormat here.

in directionFormatHandler you can just read/write display from the element and translate it to textAlign in Content Model

packages/roosterjs-content-model/lib/publicTypes/format/ContentModelBlockFormat.ts
```diff
@@ -15,4 +16,5 @@ export type ContentModelBlockFormat = BackgroundColorFormat &
     PaddingFormat &
     LineHeightFormat &
     WhiteSpaceFormat &
+    DisplayFormat &
```
JiuqingSong, line 19: Same here

packages/roosterjs-content-model/lib/modelApi/block/setModelAlignment.ts
```diff
@@ -0,0 +1,77 @@
...
+ * @internal
+ */
+export function setModelAlignment(
+    model: ContentModelDocument,
+    alignment: 'left' | 'center' | 'right'
+) {
+    const paragraphOrListItemOrTable = getOperationalBlocks<ContentModelListItem>(
+        model,
+        ['ListItem'],
+        []
```
JiuqingSong, line 57: Sorry I misunderstood. I think it is ok to put TableCell here since we should not search for list item across table cell.

packages/roosterjs-content-model/lib/modelApi/block/setModelAlignment.ts
```diff
@@ -0,0 +1,77 @@
...
+    );
+
+    paragraphOrListItemOrTable.forEach(({ block }) => {
+        if (isBlockGroupOfType<ContentModelListItem>(block, 'ListItem')) {
```
JiuqingSong, line 61: save the result of `ResultMap[alignment][format.direction == 'rtl' ? 'rtl' : 'ltr']` to a local var since it is used multiple times.

I think the handling for table can also be merged together. But that can be done in separate PR.

packages/roosterjs-content-model/lib/formatHandlers/defaultFormatHandlers.ts
```diff
@@ -88,14 +88,15 @@ const blockFormatHandlers: (keyof FormatHandlerTypeMap)[] = [
...
+    'display',
 ];
 
 const defaultFormatKeysPerCategory: {
     [key in keyof ContentModelFormatMap]: (keyof FormatHandlerTypeMap)[];
 } = {
     block: blockFormatHandlers,
     listItem: ['listItemThread', 'listItemMetadata'],
-    listLevel: ['listType', 'listLevelThread', 'listLevelMetadata', 'direction'],
+    listLevel: ['listType', 'listLevelThread', 'listLevelMetadata', 'direction', 'display'],
```
JiuqingSong, line 99: same here, no need to read display

packages/roosterjs-content-model/lib/formatHandlers/defaultFormatHandlers.ts
```diff
@@ -107,6 +108,7 @@ const defaultFormatKeysPerCategory: {
         'textColor',
         'backgroundColor',
         'lineHeight',
+        'direction',
```
JiuqingSong, line 111: no need to add this

packages/roosterjs-content-model/lib/formatHandlers/block/displayFormatHandler.ts
```diff
@@ -16,5 +16,9 @@ export const displayFormatHandler: FormatHandler<DisplayFormat> = {
         if (format.display) {
             element.style.display = format.display;
         }
+
+        if (format.display === 'flex' && (element.tagName === 'UL' || element.tagName === 'OL')) {
+            element.style.flexDirection = 'column';
+        }
```
JiuqingSong, line 22: Let's handle this in directionFormatHandler

packages/roosterjs-content-model/lib/formatHandlers/block/directionFormatHandler.ts
```diff
@@ -23,7 +23,12 @@ export const directionFormatHandler: FormatHandler<DirectionFormat> = {
...
         const alignFromAttr = element.getAttribute('align');
-        const align = element.style.textAlign || alignFromAttr || defaultStyle.textAlign;
+        const textAlign = element.style.textAlign || alignFromAttr || defaultStyle.textAlign;
+        const alignSelf = element.style.alignSelf;
+        const isLI = element.tagName === 'li';
+        const shouldApplyAlignSelf =
+            isLI && element.style.display === 'flex' && element.style.flexDirection === 'column';
+        const align = shouldApplyAlignSelf ? alignSelf : textAlign;
 
         if (dir) {
```
JiuqingSong, line 33: Also read display and align-self here for `LI` to determine direction.

packages/roosterjs-content-model/lib/formatHandlers/block/directionFormatHandler.ts
```diff
@@ -62,6 +67,8 @@ export const directionFormatHandler: FormatHandler<DirectionFormat> = {
 
             if (format.isTextAlignFromAttr) {
                 element.setAttribute('align', value);
+            } else if (element.tagName === 'LI') {
+                element.style.alignSelf = format.textAlign;
```
JiuqingSong, line 71: write element.style.display and element.style.flexDirection here as well.

juliaroldi: The LI element is does not have flexDirection and display, only the UL/OL elements that have

packages/roosterjs-content-model/lib/modelApi/block/setModelAlignment.ts
```diff
@@ -0,0 +1,79 @@
...
+    const paragraphOrListItemOrTable = getOperationalBlocks<ContentModelListItem>(
+        model,
+        ['ListItem'],
+        ['TableCell']
+    );
+
+    paragraphOrListItemOrTable.forEach(({ block }) => {
+        const newAligment = ResultMap[alignment][block.format.direction == 'rtl' ? 'rtl' : 'ltr'];
+        if (isBlockGroupOfType<ContentModelListItem>(block, 'ListItem')) {
+            block.listItemFormat.format.textColor = 'red';
```
JiuqingSong, line 63: This is for test only?

packages/roosterjs-content-model/lib/formatHandlers/block/directionFormatHandler.ts
```diff
@@ -62,9 +67,15 @@ export const directionFormatHandler: FormatHandler<DirectionFormat> = {
...
             } else {
                 element.style.textAlign = value;
             }
         }
+
+        if (element.tagName == 'OL' || element.tagName == 'UL') {
+            element.style.flexDirection = 'column';
```
JiuqingSong, line 78: do you also want to set display to flex here?

PR #1580, opened 2023-02-16
Step 1: Add HTMLSpanElements around ReadOnly Inline entities to prevent cursor issues,

BryanValverdeU:
### The problem
When the cursor is next to readonly inline entity the cursor under some scenarios it hidden and user cannot see it due to browser not handling it correctly

To attemp to fix this, add two editable inline entities surrounding the readonly entity wit Zero Width Space,

In next PRs:
1. Add arrow keys handle to prevent the cursor to be stuck in the ZeroWidthSpaces and the Read only entity
2. Prevent the user to type inside of the readonly entities that are going to be added around the Read only entity.
3. Logic to remove the Delimiter and the Readonly Entity


packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/isDelimiter.ts
```diff
@@ -0,0 +1,24 @@
...
+    ent: Element | Entity | null | undefined
+): [DelimiterType, Entity] | null {
+    if (!ent) {
+        return null;
+    }
+    let entity: Entity | null = (<Entity>ent).wrapper
+        ? <Entity>ent
+        : getEntityFromElement(<HTMLElement>ent);
+    if (entity) {
+        const id = entity.type;
```
JiuqingSong, line 15: You can use entity type to check if this is delimiter.

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
```diff
@@ -0,0 +1,195 @@
...
+                        const idBefore = delimiterId + DELIMITER_BEFORE;
+                        const idAfter = delimiterId + DELIMITER_AFTER;
+                        commitEntity(elementBefore, idBefore, false, idAfter);
+                        commitEntity(elementAfter, idAfter, false, idBefore);
+
+                        editor.triggerContentChangedEvent(ChangeSource.InsertEntity, elementBefore);
+                        editor.triggerContentChangedEvent(ChangeSource.InsertEntity, elementAfter);
+
+                        editor.runAsync(() => {
+                            editor.select(new Position(new Position(wrapper, PositionType.After)));
```
JiuqingSong, line 58: Why 2 new Position?

BryanValverdeU: Good catch, probably a typo. Will keep only one.

JiuqingSong:
Got exception when insert entity from demo site:

EntityPlugin.ts:351 Uncaught TypeError: Cannot read properties of undefined (reading 'shadowRoot')
    at ./packages/roosterjs-editor-core/lib/corePlugins/EntityPlugin.ts.EntityPlugin.handleNewEntity (EntityPlugin.ts:351:28)
    at EntityPlugin.ts:257:18
    at Array.forEach (<anonymous>)
    at ./packages/roosterjs-editor-core/lib/corePlugins/EntityPlugin.ts.EntityPlugin.handleContentChangedEvent (EntityPlugin.ts:252:21)
    at ./packages/roosterjs-editor-core/lib/corePlugins/EntityPlugin.ts.EntityPlugin.onPluginEvent (EntityPlugin.ts:141:22)
    at triggerEvent.ts:36:24
    at Array.forEach (<anonymous>)
    at Object.triggerEvent (triggerEvent.ts:34:22)
    at ./packages/roosterjs-editor-core/lib/editor/Editor.ts.Editor.triggerPluginEvent (Editor.ts:681:18)
    at ./packages/roosterjs-editor-core/lib/editor/Editor.ts.Editor.triggerContentChangedEvent (Editor.ts:695:14)

JiuqingSong:
I'm rethinking about this entity-based solution. In general, this introduces 2 additional entities to a readonly entity. Although there is code to check for readonly in NewEntity event that can avoid adding more and more space entities to the spacing entity, but they are still entities. Is it possible to just add the 2 space SPANs but not make them entites?

I know you'd like to leverage the existing entity-deletion code to know if the space SPANs are deleted, but since this new space SPAN should be part of entity core, maybe we should also improve the code of deletion detection to let it know user is about to delete the space SPAN. Because deleting space SPAN should be one of the reason to trigger deleting entity and related events, but not a result of deleting entity.

We can discuss more about the design.

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
```diff
@@ -0,0 +1,195 @@
...
+
+export function inlineEntityOnPluginEvent(event: PluginEvent, editor: IEditor) {
+    switch (event.eventType) {
+        case PluginEventType.EntityOperation:
+            const { wrapper } = event.entity;
+            switch (event.operation) {
+                case EntityOperation.NewEntity:
+                    // If the new entity is Read Only and inline, add delimiters to it
+                    if (isReadOnly(event.entity)) {
+                        const element = wrapper as HTMLElement;
```
JiuqingSong, line 33: I feel this should be part of the insertEntity() API, so that it directly inserts a complete entity (an entity with two space SPANs). NewEntity event is used for other plugins to know an entity is added.

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
```diff
@@ -0,0 +1,195 @@
...
+                        editor.triggerContentChangedEvent(ChangeSource.InsertEntity, elementAfter);
+
+                        editor.runAsync(() => {
+                            editor.select(new Position(wrapper, PositionType.After));
+                        });
+                    }
+                    break;
+                case EntityOperation.RemoveFromStart:
+                case EntityOperation.RemoveFromEnd:
+                case EntityOperation.Overwrite:
```
JiuqingSong, line 64: Similarly, this can be part of the code where we detect and trigger these events.

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
```diff
@@ -0,0 +1,195 @@
...
+
+                case EntityOperation.Click:
+                    const range = editor.getSelectionRange();
+                    // If the Entity is ReadOnly inline is clicked modify the selection to be the delimiter after. So we can see the cursor.
+                    if (isReadOnly(event.entity) && range?.collapsed) {
+                        const [_, delimiterAfter] = getDelimitersFromReadOnlyEntity(event.entity);
+
+                        if (delimiterAfter) {
+                            editor.select(createRange(delimiterAfter.wrapper, 1));
+                        }
```
JiuqingSong, line 117: What if the readonly entity was designed to accept click event? In that case we should not change the selection.

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/isDelimiter.ts
```diff
@@ -0,0 +1,23 @@
...
+    ent: Element | Entity | null | undefined
+): [DelimiterType, Entity] | null {
+    if (!ent) {
+        return null;
+    }
+    let entity: Entity | null = (<Entity>ent).wrapper
+        ? <Entity>ent
+        : getEntityFromElement(<HTMLElement>ent);
+    if (entity) {
+        if (entity.type.indexOf(DELIMITER_AFTER) > -1) {
```
JiuqingSong, line 15: Although I suggest we don't make the delimiters entity, but if we still do this, I suggest we have two separate entity type names, so that we just need to compare their type name directly

BryanValverdeU:
> I'm rethinking about this entity-based solution. In general, this introduces 2 additional entities to a readonly entity. Although there is code to check for readonly in NewEntity event that can avoid adding more and more space entities to the spacing entity, but they are still entities. Is it possible to just add the 2 space SPANs but not make them entites?
> 
> I know you'd like to leverage the existing entity-deletion code to know if the space SPANs are deleted, but since this new space SPAN should be part of entity core, maybe we should also improve the code of deletion detection to let it know user is about to delete the space SPAN. Because deleting space SPAN should be one of the reason to trigger deleting entity and related events, but not a result of deleting entity.
> 
> We can discuss more about the design.

Sounds good. I think we can stop adding the Delimiters as Entities and trigger the Removal events in Core Plugin. Will try to do the changes

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
BryanValverdeU: I think we should just stop making the Empty spans Entities because of your comment above. So, I think we can still add the empty spans EntityOperation.NewEntity, 
That way we keep all the logic in Core Plugin, is that okay?

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
BryanValverdeU: Good point, will remove this part

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
JiuqingSong: Since all plugins are equivalent, when we add the SPANs inside a plugin, that means it is possible some other plugin runs earlier than here so they will get an entity with out these SPANs. In most cases that should be fine, but I still feel we should make the entity "fully ready" when other plugins got it, in that case plugin is not a good place to handle it.

Since all entities need to be inserted using insertEntity(), so that function should be the best place to insert the SPAN.

And there is a scenario that we will have entity but insertEntity() is not called: load existing content that contains entities. There are 2 options for that case:
1. Just do not handle it and let it be.
2. Add the SPANs here. If we go with 2, we can move the code of inserting SPANs to roosterjs-editor-dom package and wrap as a public function, then call it from both here and insertEntity()

For now, I think 1 should be fine.

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/isDelimiter.ts
```diff
@@ -0,0 +1,15 @@
+import { DELIMITER_AFTER, DELIMITER_BEFORE, DelimiterType } from './constants';
+
+export function isDelimiter(element: Element | null | undefined): DelimiterType | null {
+    if (!element) {
+        return null;
+    }
+
+    if (element.id.indexOf(DELIMITER_AFTER) > -1) {
```
JiuqingSong, line 8: Instead of using ID, I prefer using className, so that no need to make it unique. And to find a delimiter of a given entity,  we just need to use previousSibling and nextSibling of entity element, and check its class.

JiuqingSong:
We also need to make sure the delimiter is preserved in Content Model. There are two options:

1. Let Content Model recognize the delimiter, so it will be shown in Content Model and when write back it is also written. This need to create a new model type for delimiter
2. Ignore all delimiter when create Content Model. So inside Content Model there will be no delimiter. When write back, we recreate delimiter for each inline readonly entity.

I personally prefer 2.

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/isDelimiter.ts
BryanValverdeU: Sounds good. started using classname

BryanValverdeU:
> We also need to make sure the delimiter is preserved in Content Model. There are two options:
> 
> 1. Let Content Model recognize the delimiter, so it will be shown in Content Model and when write back it is also written. This need to create a new model type for delimiter
> 2. Ignore all delimiter when create Content Model. So inside Content Model there will be no delimiter. When write back, we recreate delimiter for each inline readonly entity.
> 
> I personally prefer 2.

Good point, will handle it in a future step

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const DELIMITER_SELECTOR =
+    '.' + DelimiterClasses.DELIMITER_AFTER + ',.' + DelimiterClasses.DELIMITER_BEFORE;
+
+export function inlineEntityOnPluginEvent(event: PluginEvent) {
+    switch (event.eventType) {
+        case PluginEventType.ExtractContentWithDom:
+        case PluginEventType.BeforeCutCopy:
+            event.clonedRoot.querySelectorAll(DELIMITER_SELECTOR).forEach(node => {
+                node.parentElement?.removeChild(node);
```
JiuqingSong, line 13: Since we remove delimiter when copy or save, we also need to add them back when load content.

JiuqingSong (approved):

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
BryanValverdeU: Good point. Will handle it in a future step.

PR #1581, opened 2023-02-16
Step 2: Prevent user from typing inside of readonly entities around the Readonly Inline Entity

BryanValverdeU:
Prevent user typing inside of Delimiters so we can handle events such as Delete and Arrow Keys cursor movement in next steps.
When user types inside of a delimiter, move the cursor outside of the Delimiter, either before of after depending of the type of delimiter. 
Delimiter


packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
```diff
@@ -24,11 +38,68 @@ export function inlineEntityOnPluginEvent(event: PluginEvent) {
...
                     if (isReadOnly(entity)) {
-                        removeDelimiters(wrapper);
+                        removeDelimiters(entity.wrapper);
                     }
                     break;
             }
             break;
+
+        case PluginEventType.KeyDown:
+        case PluginEventType.KeyUp:
```
JiuqingSong, line 48: I think just handling KeyDown event should be enough

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
```diff
@@ -24,11 +38,68 @@ export function inlineEntityOnPluginEvent(event: PluginEvent) {
...
+            if (isCharacterValue(event.rawEvent)) {
+                const position = editor.getFocusedPosition();
```
JiuqingSong, line 50: What do we want to do if selection is not collapsed?

getFocusedPosition() will always give you a single position event selection is collapsed.

And what do we want to do for table selection and image selection?

JiuqingSong:
Is it possible that cursor is just BETWEEN delimiter and entity, but not inside delimiter?

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
BryanValverdeU: Good point, in this case we should only handle it when the selection is collapsed.

BryanValverdeU:
> Is it possible that cursor is just BETWEEN delimiter and entity, but not inside delimiter?

Right now yes, using arrow keys. in the next step I am going to add a content edit feature to prevent this from happenning.

packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts
BryanValverdeU: Removed KeyUp as suggested

JiuqingSong (approved):

PR #1582, opened 2023-02-16
Step 3: Add Content Edit Features to prevent cursor to be stuck between ZeroWidthSpace & ReadOnly Entity

BryanValverdeU:
Add 3 new Content Edit Features
1. When Right Arrow Key in the before delimiter, move the cursor to the delimiter After
2. When Left Arrow Key in the After Delimiter, move the cursor to the delimiter Before
3. When Backspace or Delete is pressed in the delimiter, Remove the Entity

Delimiter2


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
```diff
@@ -191,6 +196,216 @@ function cacheGetNeighborEntityElement(
     return element;
 }
 
+/**
+ * Content edit feature to move the cursor from Delimiter After Entity to the delimiter Before
+ */
+const MoveBeforeDelimiterFeature: BuildInEditFeature<PluginKeyboardEvent> = {
+    keys: [Keys.LEFT],
```
JiuqingSong, line 203: Do we need special handling for RTL?

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
```diff
@@ -191,6 +196,216 @@ function cacheGetNeighborEntityElement(
...
+    shouldHandleEvent: (event: PluginKeyboardEvent, editor: IEditor) => {
+        return getIsDelimiterAtCursor(event, editor, false /* checkBefore */);
+    },
+    handleEvent(event: PluginKeyboardEvent, editor: IEditor) {
+        const delimiter = cacheDelimiter(event, false /* checkBefore */);
+        if (!delimiter) {
+            return;
+        }
+        const delimiterBefore = delimiter.previousElementSibling?.previousElementSibling;
```
JiuqingSong, line 212: Do we also need to check previous sibling is entity?

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
```diff
@@ -191,6 +196,216 @@ function cacheGetNeighborEntityElement(
...
+
+            if (selection) {
+                event.rawEvent.preventDefault();
+                editor.runAsync(() => {
+                    const position = new Position(delimiterBefore, PositionType.Before);
+                    if (event.rawEvent.shiftKey) {
+                        selection.extend(position.element, position.offset);
+                    } else {
+                        selection.setPosition(position.element, position.offset);
+                    }
```
JiuqingSong, line 225: You can do editor.select(delimiterBefore, PositionType.Before)

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
```diff
@@ -191,6 +196,216 @@ function cacheGetNeighborEntityElement(
...
+    checkBefore: boolean,
+    entity?: Element | null,
+    operation?: EntityOperation
+) {
+    const element = cacheGetEventData(event, 'entity_delimiter_cache_key_' + checkBefore, () =>
+        editor.getElementAtCursor(getEntitySelector(), entity)
+    );
+
+    if (element && operation !== undefined) {
+        editor.triggerPluginEvent(PluginEventType.EntityOperation, {
```
JiuqingSong, line 399: Do we expect plugin to help delete the delimiters?

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
```diff
@@ -191,6 +196,216 @@ function cacheGetNeighborEntityElement(
...
+};
+
+/**
+ * Content edit Feature to trigger a Delete Entity Operation when one of the Delimiter is about to be removed with DELETE or Backspace
+ */
+const RemoveEntityBetweenDelimitersFeature: BuildInEditFeature<PluginKeyboardEvent> = {
+    keys: [Keys.BACKSPACE, Keys.DELETE],
+    shouldHandleEvent(event: PluginKeyboardEvent, editor: IEditor) {
+        const range = editor.getSelectionRange();
+        if (!range.collapsed) {
```
JiuqingSong, line 274: If press SHIFT+Left/Right, I think we should also handle the selection. For example, cursor is after delimiter, press SHIFT+LEFT, need to select the whole entity as well as two delimiters.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
BryanValverdeU: Ah I see, yes we need to do that.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
BryanValverdeU: I added some logic in previous PR to delete the delimeters when a entity is deleted [here](https://github.com/microsoft/roosterjs/blob/0341fa88ecb018f2f09964ee87249f6848c94a31/packages/roosterjs-editor-core/lib/corePlugins/utils/InlineEntityHandlers/inlineEntityOnPluginEvent.ts#L41). Not sure if that is okay or we should move the logic elsewhere

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
BryanValverdeU: Should be fixed in latest commit
Delimiter2

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
BryanValverdeU: Good point, added a check too.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
BryanValverdeU: Added in the latest commit:
Delimiter3

packages/roosterjs-editor-types/lib/interface/ContentEditFeatureSettings.ts
```diff
@@ -62,6 +62,16 @@ export interface EntityFeatureSettings {
      * press DELETE right after an entity
      */
     deleteBeforeEntity: boolean;
+
+    /**
+     * Content edit feature to move the cursor from Delimiters around Entities when using Right or Left Arrow Keys
+     */
+    MoveBetweenDelimitersFeature: boolean;
```
JiuqingSong, line 69: start with lower case

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
```diff
@@ -191,6 +198,227 @@ function cacheGetNeighborEntityElement(
...
+                shouldCheckBefore = true;
+            }
+        }
+        if (event.rawEvent.which === Keys.LEFT) {
+            if (isRTL) {
+                shouldCheckBefore = true;
+            } else {
+                shouldCheckBefore = false;
+            }
+        }
```
JiuqingSong, line 223: How about: 
const shouldCheckBefore = isRTL == (event.rawEvent.which === Keys.LEFT);

JiuqingSong (approved):

PR #1583, opened 2023-02-16
Delete table with backspace, if whole table is selected.

juliaroldi:
If the whole table is selected, delete the table if backspaced was pressed. 
deleteTableBackspace

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/tableFeatures.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/tableFeatures.ts
@@ -12,6 +12,7 @@ import {
     SelectionRangeTypes,
     TableSelectionRange,
     Indentation,
+    ExperimentalFeatures,
 } from 'roosterjs-editor-types';
 import {
     Browser,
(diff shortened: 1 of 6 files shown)


packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/tableFeatures.ts
```diff
@@ -151,6 +152,23 @@ const UpDownInTable: BuildInEditFeature<PluginKeyboardEvent> = {
...
 
+/**
+ * Requires @see ExperimentalFeatures.DeleteTableWithBackspace
+ * Delete a table selected with the table selector pressing Backspace key
+ */
+const DeleteTableWithBackspace: BuildInEditFeature<PluginKeyboardEvent> = {
+    keys: [Keys.BACKSPACE],
+    shouldHandleEvent: (event: PluginKeyboardEvent, editor: IEditor) =>
+        cacheIsWholeTableSelected(event, editor) &&
+        editor.isFeatureEnabled(ExperimentalFeatures.DeleteTableWithBackspace),
```
BryanValverdeU, line 163: Should we check `editor.isFeatureEnabled(ExperimentalFeatures.DeleteTableWithBackspace),` first? Since it could take less time that trying to create a VTable and check if it is selected

JiuqingSong (approved):

PR #1590, opened 2023-02-20
Add `dir` and `textAlign` in `FormatState`

Echo-az:
Add dir and text-align info in FormatState. They will be used to calculate the alignment (left | center | right) of text at cursor. This state will be displayed to the users through the icons of the button.
image

### DIFF
--- a/packages/roosterjs-editor-types/lib/interface/FormatState.ts
+++ b/packages/roosterjs-editor-types/lib/interface/FormatState.ts
@@ -153,6 +153,16 @@ export interface StyleBasedFormatState {
      * Margin Bottom
      */
     marginBottom?: string;
+
+    /**
+     * Text Align
+     */
+    textAlign?: string;
+
+    /**
+     * Direction of the element ('ltr' or 'rtl')
+     */
+    direction?: string;
 }
 
 /**
--- a/packages/roosterjs-content-model/lib/modelApi/common/retrieveModelFormatState.ts
+++ b/packages/roosterjs-content-model/lib/modelApi/common/retrieveModelFormatState.ts
@@ -123,6 +123,8 @@ function retrieveParagraphFormat(
     mergeValue(result, 'marginBottom', paragraph.format.marginBottom, isFirst);
     mergeValue(result, 'marginTop', paragraph.format.marginTop, isFirst);
     mergeValue(result, 'headerLevel', validHeaderLevel, isFirst);
+    mergeValue(result, 'textAlign', paragraph.format.textAlign, isFirst);
+    mergeValue(result, 'direction', paragraph.format.direction, isFirst);
 }
 
 function retrieveStructureFormat(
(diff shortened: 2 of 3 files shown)


Echo-az:
@microsoft-github-policy-service agree company="Microsoft"

packages/roosterjs-editor-types/lib/interface/FormatState.ts
```diff
@@ -103,6 +103,11 @@ export interface ElementBasedFormatState {
      * Whether we can execute table cell merge operation
      */
     canMergeTableCell?: boolean;
+
+    /**
+     * Dir of the element ('ltr' or 'rtl')
+     */
+    dir?: string;
```
JiuqingSong, line 110: This should be part of StyleBasedFormatState

packages/roosterjs-editor-api/lib/format/getFormatState.ts
```diff
@@ -42,6 +42,13 @@ export function getElementBasedFormatState(
         ? toArray(table.rows[0].cells).every(cell => getTagOfNode(cell) == 'TH')
         : undefined;
 
+    let dir = undefined;
+    let element = editor.getElementAtCursor();
+    while (!dir && element) {
+        dir = element.getAttribute('dir');
```
JiuqingSong, line 48: dir can also be specified using CSS "direction".

So do this in getStyleBasedFormatState is much easier.

JiuqingSong:
Can you also modify https://github.com/microsoft/roosterjs/blob/1884e38d621cd9612f6a37a3dcec26b837e8274e/packages/roosterjs-content-model/lib/modelApi/common/retrieveModelFormatState.ts#L88 to let the new retrieveModelFormatState function also return correct value of your new attributes? This is a nice-to-have.

packages/roosterjs-editor-api/lib/format/getFormatState.ts
Echo-az: Thanks for suggestion! Moved and renamed it.

JiuqingSong (approved):

JiuqingSong:
@Echo-az  There are test cases fail, would you please fix them as well?
You can run `yarn test` to run test locally.

Echo-az:
> @Echo-az There are test cases fail, would you please fix them as well? You can run `yarn test` to run test locally.

Hi @JiuqingSong, I ran `yarn test` both before and after my change. There're some failed test cases even without any change. But the count of failed test cases didn't increase, I'm not sure my change is ok. Do you have any idea about this?

> Chrome 110.0.0.0 (Mac OS 10.15.7): Executed 3455 of 3689 (12 FAILED) (skipped 234) (2.828 secs / 2.188 secs)
> Something in shortcutFeatureTest.ts, undoPluginTest.ts, HyperLinkTest.ts

JiuqingSong (approved):

PR #1592, opened 2023-02-23
[Don't Review] demo for picker bug

huahuahu:
The query string wasn't delete after commit.

https://user-images.githubusercontent.com/5757671/220813436-14c2617c-7bd4-42c9-a379-c5e432ebd110.MP4


JiuqingSong:
Close this one since
1. We don't plan to check it in
2. We can still see the PR from its url linked from #1593

PR #1596, opened 2023-02-23
Enabled Strict Mode on Image Edit Plugin

juliaroldi:
Enabled Strict Mode on Image Edit Plugin to avoid unexpected errors.


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/api/canRegenerateImage.ts
```diff
@@ -14,8 +14,11 @@ export default function canRegenerateImage(img: HTMLImageElement): boolean {
...
         canvas.height = 10;
         const context = canvas.getContext('2d');
-        context.drawImage(img, 0, 0);
-        context.getImageData(0, 0, 1, 1);
+        if (context) {
+            context.drawImage(img, 0, 0);
+            context.getImageData(0, 0, 1, 1);
+        }
+
         return true;
```
JiuqingSong, line 22: Move this line above into the block of "if(context)", and for all other cases, return false

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/editInfo.ts
```diff
@@ -35,7 +35,9 @@ export function deleteEditInfo(image: HTMLImageElement) {
  */
 export function getEditInfoFromImage(image: HTMLImageElement): ImageEditInfo {
     const obj = getMetadata<ImageEditInfo>(image);
-    return checkEditInfoState(obj) == ImageEditInfoState.Invalid ? getInitialEditInfo(image) : obj;
+    return !obj || (obj && checkEditInfoState(obj) == ImageEditInfoState.Invalid)
```
JiuqingSong, line 38: This can be simplified to

!obj || checkEditInfoState(obj) == ImageEditInfoState.Invalid

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -554,25 +581,25 @@ export default class ImageEdit implements EditorPlugin {
...
+    width: number | undefined,
+    height: number | undefined
 ) {
-    element.style.left = getPx(left);
-    element.style.top = getPx(top);
-    element.style.right = getPx(right);
-    element.style.bottom = getPx(bottom);
-    element.style.width = getPx(width);
-    element.style.height = getPx(height);
+    element.style.left = left ? getPx(left) : element.style.left;
```
JiuqingSong, line 597: what if left is 0?

Same for others.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -611,12 +638,15 @@ function handleRadIndexCalculator(angleRad: number): number {
...
 }
 
-function rotateHandles(element: HTMLElement, angleRad: number): string {
+function rotateHandles(element: HTMLElement, angleRad: number): string | null {
     const radIndex = handleRadIndexCalculator(angleRad);
-    const originalDirection = element.dataset.y + element.dataset.x;
-    const originalIndex = DirectionOrder.indexOf(originalDirection);
-    const rotatedIndex = originalIndex >= 0 && originalIndex + radIndex;
-    return DirectionOrder[rotatedIndex % DIRECTIONS];
+    if (element.dataset.y && element.dataset.x) {
```
JiuqingSong, line 643: In what case x or y does not exist? And what if their value is 0?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -414,7 +437,14 @@ export default class ImageEdit implements EditorPlugin {
...
     private updateWrapper = (context?: DragAndDropContext) => {
         const wrapper = this.wrapper;
-        if (wrapper) {
+        if (
+            wrapper &&
+            this.editInfo &&
+            this.image &&
+            this.clonedImage &&
+            this.options &&
+            this.shadowSpan?.parentElement
```
JiuqingSong, line 446: In theory these info should either all exist or none exists.

Can we create an interface for them and create the instance together, so that we only need to check if that instance exist.

For example:
```ts
interface EditingImageInfo {
   image: HtmlImageElement;
   clonedImage: HtmlImageElement;
   ...
}

class ImageEdit {
   private edtingImageInfo: EditingImageInfo | null = null;
   ..


  this.editingImageInfo = {
   ...
  }


  ...


  if (this.editingImageInfo) {
     ...
 }
 
...

this.editingImageInfo = null;
}
```

juliaroldi: How do we add the elements in this object? Some of them are created in different parts in the code.

JiuqingSong: This needs some kind of refactoring.

If that is too complex, feel free to do it in separate PR.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -334,74 +347,84 @@ export default class ImageEdit implements EditorPlugin {
      * quit editing mode when editor lose focus
      */
     private onBlur = () => {
-        this.setEditingImage(null, true);
+        //this.setEditingImage(null, true);
```
JiuqingSong, line 350: Uncomment this

JiuqingSong (approved):

PR #1607, opened 2023-02-28
Use keypress instead of keydown in AutoFormat plugin

juliaroldi:
AutoFormat Plugin does not need to observe key like SHIFT,CTRL, ALT, ESC, then switch keydown to keypress. 
autoFormatKeyPress


JiuqingSong:
Generally this is ok. But if I remember correctly, KeyPress event does not work in Android device. So this may break some Android if any one is using this plugin in Android.

JiuqingSong (approved):

juliaroldi:
I can make a fix in a separate PR for a workaround in Android.

PR #1608, opened 2023-03-01
Delimiters in Content Model

BryanValverdeU:
When converting Dom to Content Model, remove delimiters.
When converting Content Model to DOM, add delimiters around Inline Readonly Entities

### DIFF
--- a/packages/roosterjs-content-model/lib/domToModel/processors/childProcessor.ts
+++ b/packages/roosterjs-content-model/lib/domToModel/processors/childProcessor.ts
@@ -17,6 +17,10 @@ export const childProcessor: ElementProcessor<ParentNode> = (
     const [nodeStartOffset, nodeEndOffset] = getRegularSelectionOffsets(context, parent);
     let index = 0;
 
+    if (context.shouldSkipProcessElement?.child?.(parent, context)) {
+        return;
+    }
+
     for (let child = parent.firstChild; child; child = child.nextSibling) {
         handleRegularSelection(index, context, group, nodeStartOffset, nodeEndOffset);
 
--- a/packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
+++ b/packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
@@ -1,16 +1,20 @@
 import contentModelToDom from '../modelToDom/contentModelToDom';
 import domToContentModel from '../domToModel/domToContentModel';
 import { ContentModelDocument } from '../publicTypes/group/ContentModelDocument';
+import { DelimiterClasses, SelectionRangeTypes } from 'roosterjs-editor-types';
+import { DomToModelContext } from '../publicTypes/context/DomToModelContext';
 import { Editor } from 'roosterjs-editor-core';
 import { EditorContext } from '../publicTypes/context/EditorContext';
 import { Position, restoreContentWithEntityPlaceholder } from 'roosterjs-editor-dom';
-import { SelectionRangeTypes } from 'roosterjs-editor-types';
-
+import { safeInstanceOf } from 'roosterjs-editor-dom';
+import { ShouldSkipProcessElementMap } from '../publicTypes/context/DomToModelSettings';
 import {
     DomToModelOption,
     IContentModelEditor,
     ModelToDomOption,
 } from '../publicTypes/IContentModelEditor';
+
+const ZERO_WIDTH_SPACE = '\u200B';
 
 /**
  * Editor for Content Model.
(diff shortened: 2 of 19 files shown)


juliaroldi:
What the delimiters do?

BryanValverdeU:
> What the delimiters do?

Are these new elements that we add around the Readonly Inline Entities:

image

We are adding them to prevent some bugs when the cursor is after them.

juliaroldi (approved):

packages/roosterjs-content-model/lib/modelToDom/handlers/handleEntity.ts
```diff
@@ -40,14 +46,32 @@ export const handleEntity: ContentModelHandler<ContentModelEntity> = (
...
         context.entities[entity.id] = wrapper;
     }
 };
+function addDelimiterElements(
+    wrapper: HTMLElement,
+    isReadonly: boolean,
+    context: ModelToDomContext
+) {
+    if (
+        context.isFeatureEnabled?.(ExperimentalFeatures.InlineEntityReadOnlyDelimiters) &&
```
JiuqingSong, line 71: For the content model code, I'd like it can be independent from editor code. So for this check, I prefer to use a simple variable inside `EditorContext` rather than a isFeatureEnabled() function:
```ts
interface EditorContext{
  ...
  addDelimiterForEntity: boolean
}
```

When create this context we can read its value from isFeatureEnabled() in ContentModelEditor. Later we just need to read it from EditorContext.

packages/roosterjs-content-model/lib/domToModel/processors/childProcessor.ts
```diff
@@ -17,6 +17,10 @@ export const childProcessor: ElementProcessor<ParentNode> = (
     const [nodeStartOffset, nodeEndOffset] = getRegularSelectionOffsets(context, parent);
     let index = 0;
 
+    if (context.shouldSkipProcessElement?.child?.(parent, context)) {
+        return;
+    }
```
JiuqingSong, line 22: Actually we can't just skip delimiter, because there may be selection at the delimiter, then we still need to handle selection.

We can let knownElementProcessor to keep process it, but if we see it is a delimiter, just don't create any model for it.

packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
```diff
@@ -28,9 +32,17 @@ export default class ContentModelEditor extends Editor implements IContentModelE
             selectionRange: this.getSelectionRangeEx(),
             alwaysNormalizeTable: true,
             ...(option || {}),
+            shouldSkipProcessElement: this.createShouldHandleElement(),
         });
     }
 
+    private createShouldHandleElement(): Partial<ShouldSkipProcessElementMap> | undefined {
```
JiuqingSong, line 39: at least for now we haven't see other requirements to check if we should skip element. So I think we can just do this check simply when process element in `knownElementProcessor` and directly call isDelimiter() there.

JiuqingSong: It should be safe enough to just skip delimiter since we checked its class and its content. There is nothing visible different from user perspective with or without it.

packages/roosterjs-content-model/lib/publicTypes/IContentModelEditor.ts
```diff
@@ -53,6 +54,11 @@ export interface DomToModelOption {
      * @default false
      */
     alwaysNormalizeTable?: boolean;
+
+    /**
+     * Checks whether elements needs to be skipped from being processed
+     */
+    shouldSkipProcessElement?: Partial<ShouldSkipProcessElementMap>;
```
JiuqingSong, line 61: no need to add this. Just check delimiter directly.

JiuqingSong: Or, another way, you can add a delimiterProcessor, and choose this processor in https://github.com/microsoft/roosterjs/blob/fbf328176e345f697d8514029ef9d3b9a9e48bfa/packages/roosterjs-content-model/lib/domToModel/processors/elementProcessor.ts#L12, then do nothing in this processor, that could be an elegant way to skip delimiter.

packages/roosterjs-content-model/lib/modelToDom/handlers/handleEntity.ts
BryanValverdeU: Remove isFeatureEnabled and added the new property

packages/roosterjs-content-model/lib/domToModel/processors/childProcessor.ts
BryanValverdeU: Added Delimiter processor to check the selection as suggested

JiuqingSong (approved):

PR #1613, opened 2023-03-02
Content Model: Fix a bug in normalization

JiuqingSong:
When call normalizeContentModel and recreate DOM,
Case 1:
input: 
```html
<div>text<br></div>
```
Output:
```html
<div>text</div>
```
This is correct

Case 2:
input: 
```html
<div>text<br><br></div>
```
Output:
```html
<div>text<br></div>
```
This is wrong. Because we need the two BR at the end to make sure there is an empty line below "text".

Fix: check there must be only one BR at the end then we can remove it.

### DIFF
--- a/packages/roosterjs-content-model/lib/modelApi/common/normalizeContentModel.ts
+++ b/packages/roosterjs-content-model/lib/modelApi/common/normalizeContentModel.ts
@@ -48,14 +48,17 @@ function normalizeParagraph(block: ContentModelParagraph) {
 
         if (segments.length == 1 && segments[0].segmentType == 'SelectionMarker') {
             segments.push(createBr(segments[0].format));
-        } else if (
-            segments.length > 1 &&
-            segments[segments.length - 1].segmentType == 'Br' &&
-            segments.some(
-                segment => segment.segmentType != 'SelectionMarker' && segment.segmentType != 'Br'
-            )
-        ) {
-            segments.pop();
+        } else if (segments.length > 1 && segments[segments.length - 1].segmentType == 'Br') {
+            const noMarkerSegments = segments.filter(x => x.segmentType != 'SelectionMarker');
+
+            // When there is content with a <BR> tag at the end, we can remove the BR.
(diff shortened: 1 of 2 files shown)


packages/roosterjs-content-model/lib/modelApi/common/normalizeContentModel.ts
```diff
@@ -48,14 +48,17 @@ function normalizeParagraph(block: ContentModelParagraph) {
...
-        } else if (
-            segments.length > 1 &&
-            segments[segments.length - 1].segmentType == 'Br' &&
-            segments.some(
-                segment => segment.segmentType != 'SelectionMarker' && segment.segmentType != 'Br'
-            )
-        ) {
-            segments.pop();
+        } else if (segments.length > 1 && segments[segments.length - 1].segmentType == 'Br') {
+            const noMarkerSegments = segments.filter(x => x.segmentType != 'SelectionMarker');
```
juliaroldi, line 52: Is image a segment that we need br after?

JiuqingSong: Do you mean something like:
```html
<div><img><br></div>
```
In that case, no. You can try to remove the BR and you will get the same rendering result.

juliaroldi: Should we check the segmentType to not add the br to this kind of element or this would make a difference?

juliaroldi (approved):

packages/roosterjs-content-model/lib/modelApi/common/normalizeContentModel.ts
JiuqingSong: We don't add BR here. Do you mean the code at line 50? It will only add BR when there is only one segment and it is SelectionMarker. So if there is Image, it will not add BR.

juliaroldi: got it.

PR #1622, opened 2023-03-07
Handle Enter in a Delimiter

BryanValverdeU:
When pressing Enter on a delimiter, usually when the entity is at the end of the block or is the only element in the block, sometimes the delimiter element is moved away from the Entity or the element after Enter also contains the Delimiter class. This PR adds functionality to prevent this from happening.

Before
DelimiterEnterBefore

After
DelimiterEnterAfter


JiuqingSong:
1. In demo site, prepare insert entity pane to be ready to insert a readonly inline entity
2. type something
3. Insert entity
4. Keep type something

Use debugger, I can see the delimiter after entity is gone. This doesn't happen without this change.

entitybug1

JiuqingSong:
Restore this snapshot:
```html
<span style="font-family: &quot;Segoe UI&quot;, &quot;Segoe UI Web (West European)&quot;, &quot;Segoe UI&quot;, -apple-system, BlinkMacSystemFont, Roboto, &quot;Helvetica Neue&quot;, sans-serif; font-size: 10.5pt; color: rgb(0, 0, 0); background-color: rgb(255, 255, 255);">aaaaa<span class="entityDelimiterBefore">​</span><span class="_Entity _EType_bbb _EId_bbb _EReadonly_1" contenteditable="false" style="display: inline-block;"><span data-hydrated-html=""><b>bbbb</b></span></span><span class="entityDelimiterAfter">​</span>ccccc</span><!--{"type":0,"isDarkMode":false,"start":[0,4],"end":[0,4]}-->
```
Then press Enter, press Backspace, then I need to press Backspace twice to delete the entity.

But this seems also happen without this change.

BryanValverdeU:
> 1. In demo site, prepare insert entity pane to be ready to insert a readonly inline entity
> 2. type something
> 3. Insert entity
> 4. Keep type something
> 
> Use debugger, I can see the delimiter after entity is gone. This doesn't happen without this change.
> 
> entitybug1 [ entitybug1 ](https://user-images.githubusercontent.com/23065085/223650092-0e97d6d7-c00c-4066-8e6c-b6d7aaa977ac.gif) [ ](https://user-images.githubusercontent.com/23065085/223650092-0e97d6d7-c00c-4066-8e6c-b6d7aaa977ac.gif)

I see good catch.

BryanValverdeU:
> Restore this snapshot:
> 
> ```
> <span style="font-family: &quot;Segoe UI&quot;, &quot;Segoe UI Web (West European)&quot;, &quot;Segoe UI&quot;, -apple-system, BlinkMacSystemFont, Roboto, &quot;Helvetica Neue&quot;, sans-serif; font-size: 10.5pt; color: rgb(0, 0, 0); background-color: rgb(255, 255, 255);">aaaaa<span class="entityDelimiterBefore">​</span><span class="_Entity _EType_bbb _EId_bbb _EReadonly_1" contenteditable="false" style="display: inline-block;"><span data-hydrated-html=""><b>bbbb</b></span></span><span class="entityDelimiterAfter">​</span>ccccc</span>
> ```
> 
> Then press Enter, press Backspace, then I need to press Backspace twice to delete the entity.
> 
> But this seems also happen without this change.

We are still triggering the Entity Operations to Delete the Entity. But since there is no Plugin to delete the entity in this scenario we are not deleting it. 
In this case should we still delete the entity when pressing Del or Backspace on the delimiter? 

I will have a PR later too to add Normalization of the delimiters on the ContentChangedEvent as we discussed last week. Since in Compose Beautiful links can prevent the default event and we would have multiple incorrect delimiters

JiuqingSong:
> > Restore this snapshot:
> > ```
> > <span style="font-family: &quot;Segoe UI&quot;, &quot;Segoe UI Web (West European)&quot;, &quot;Segoe UI&quot;, -apple-system, BlinkMacSystemFont, Roboto, &quot;Helvetica Neue&quot;, sans-serif; font-size: 10.5pt; color: rgb(0, 0, 0); background-color: rgb(255, 255, 255);">aaaaa<span class="entityDelimiterBefore">​</span><span class="_Entity _EType_bbb _EId_bbb _EReadonly_1" contenteditable="false" style="display: inline-block;"><span data-hydrated-html=""><b>bbbb</b></span></span><span class="entityDelimiterAfter">​</span>ccccc</span>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then press Enter, press Backspace, then I need to press Backspace twice to delete the entity.
> > But this seems also happen without this change.
> 
> We are still triggering the Entity Operations to Delete the Entity. But since there is no Plugin to delete the entity in this scenario we are not deleting it. In this case should we still delete the entity when pressing Del or Backspace on the delimiter?
> 
> I will have a PR later too to add Normalization of the delimiters on the ContentChangedEvent as we discussed last week. Since in Compose Beautiful links can prevent the default event and we would have multiple incorrect delimiters

This is a good question. We should delete the entity, but when do it here it may cause some issue since originally the entity was deleted by browser after this event is handled. So if other plugins require the entity is still in editor at this point, there may be some problem.

Can we try this:
When receive delete/backspace and we know we are about to delete entity, first we delete the delimiter, then let other plugins to handle the event. then we use runAsync() to check if the entity is still there. If so, we can add back the delimiter, otherwise clear the other delimiter as well.

JiuqingSong (approved):

PR #1630, opened 2023-03-10
Improve deletion of entities

BryanValverdeU:
Improve the Content Edit Feature to delete delimiters.
 If After the Content Edit was handled:
1. If the Entity still exists, add the delimiter removed.
2. If the entity was removed in the ContentChanged Event, remove the Delimiters.
3. If the event was not default prevented, select the delimiters and entity and let the browser handle the deletion.

Also, in EntityPlugin if the known entity removed is a inline entity, normalize delimiters

### DIFF
--- a/packages/roosterjs-editor-core/lib/corePlugins/EntityPlugin.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/EntityPlugin.ts
@@ -15,7 +15,7 @@ import {
     createRange,
     moveChildNodes,
     getObjectKeys,
-    getTagOfNode,
+    isBlockElement,
 } from 'roosterjs-editor-dom';
 import {
     ChangeSource,
@@ -245,7 +245,7 @@ export default class EntityPlugin implements PluginWithState<EntityPluginState> 
                 if (
                     !shouldNormalizeDelimiters &&
                     !element.isContentEditable &&
-                    getTagOfNode(element) === 'SPAN'
+                    !isBlockElement(element)
                 ) {
                     shouldNormalizeDelimiters = true;
                 }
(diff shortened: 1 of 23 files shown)


packages/roosterjs-editor-core/lib/corePlugins/EntityPlugin.ts
```diff
@@ -236,6 +241,14 @@ export default class EntityPlugin implements PluginWithState<EntityPluginState>
                 if (element.shadowRoot) {
                     this.triggerEvent(element, EntityOperation.RemoveShadowRoot);
                 }
+
+                if (
+                    !shouldNormalizeDelimiters &&
+                    !element.isContentEditable &&
+                    getTagOfNode(element) === 'SPAN'
```
JiuqingSong, line 248: Why check SPAN?

packages/roosterjs-editor-dom/lib/index.ts
```diff
@@ -4,11 +4,7 @@ export { default as getFirstLastBlockElement } from './blockElements/getFirstLas
 export { default as ContentTraverser } from './contentTraverser/ContentTraverser';
 export { default as PositionContentSearcher } from './contentTraverser/PositionContentSearcher';
 
-export {
-    default as addDelimiters,
-    addDelimiterAfter,
```
JiuqingSong, line 9: This is a 
breaking change?

packages/roosterjs-editor-core/lib/corePlugins/EntityPlugin.ts
BryanValverdeU: To only handle the normalization if the deleted entity is readonly and inline

packages/roosterjs-editor-dom/lib/index.ts
BryanValverdeU: Hmm, fixing that

packages/roosterjs-editor-core/lib/corePlugins/EntityPlugin.ts
BryanValverdeU: Agree, that might be better.

JiuqingSong (approved):

PR #1633, opened 2023-03-10
Rotate and Flip Image

juliaroldi:
Add rotateImage and flipImage to Image Edit plugin. 

flipAndRotateImage


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/types/ImageEditInfo.ts
```diff
@@ -62,11 +62,22 @@ export interface RotateInfo {
...
 
+/**
+ * @internal
+ * Flip info for inline image rotate
+ */
+export interface FlipInfo {
+    /**
+     * If true, the image was flipped.
+     */
+    flippedImage?: boolean;
```
JiuqingSong, line 73: How do we know if it is flipped vertically or horizontally?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -343,6 +344,35 @@ export default class ImageEdit implements EditorPlugin {
...
+     * Flip the image.
+     * @param image The image to be flipped
+     */
+    public flipImage(image: HTMLImageElement) {
+        this.image = image;
+        this.editInfo = getEditInfoFromImage(image);
+        this.editInfo.flippedImage = !this.editInfo.flippedImage;
+        this.createWrapper(ImageEditOperation.Rotate);
+        this.updateWrapper();
+        this.setEditingImage(null);
```
JiuqingSong, line 357: I'm curious why we need to create wrapper than set editing image to null which will dispose the wrapper? Can we just not create wrapper at all?

packages-ui/roosterjs-react/lib/contextMenu/menus/createImageEditMenuProvider.tsx
```diff
@@ -82,6 +82,41 @@ const ImageResizeMenuItem: ContextMenuItem<ImageEditMenuItemStringKey, ImageEdit
     },
 };
 
+const ImageRotateMenuItem: ContextMenuItem<ImageEditMenuItemStringKey, ImageEdit> = {
+    key: 'menuNameImageRotate',
+    unlocalizedText: 'Rotate image',
+    subItems: {
+        menuNameImageRotateLeft: 'Left',
+        menuNameImageRotateRight: 'Right',
+        menuNameImageRotateUpsidedown: 'Flip vertically',
```
JiuqingSong, line 91: Shouldn't this be something like "Rotate 180 degree"?

I tried it and I got 
image

I think for "flip vertically" it means show a up side down mirrored image, but not rotated.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
juliaroldi: createWarpper creates the clonedImage that we use in the updateWrapper, we use updateWrapper to proper calculate the image size if rotated

JiuqingSong: I see.

JiuqingSong (approved):

PR #1640, opened 2023-03-15
Fix Movement Between Entity Delimiter issue

BryanValverdeU:
When trying to move between the delimiters after using the Format API without Content Model and a delimiter element in the selection, The cursor is stuck in the ZWS so use the arrow key twice is needed.

The root cause of the issue is that we are using (Next/Previous)ElementSibling and since the delimiter element is wrapped in a new element it is returning null
To fix this intead use ContentTraverser to get the next sibling.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
@@ -1,3 +1,4 @@
+import { ContentTraverser } from 'roosterjs-editor-dom';
 import {
     addDelimiters,
     cacheGetEventData,
@@ -216,8 +217,12 @@ function cacheGetNeighborEntityElement(
  */
 const MoveBetweenDelimitersFeature: BuildInEditFeature<PluginKeyboardEvent> = {
     keys: [Keys.RIGHT, Keys.LEFT],
+    allowFunctionKeys: true,
     shouldHandleEvent: (event: PluginKeyboardEvent, editor: IEditor) => {
-        if (!editor.isFeatureEnabled(ExperimentalFeatures.InlineEntityReadOnlyDelimiters)) {
+        if (
+            event.rawEvent.altKey &&
+            !editor.isFeatureEnabled(ExperimentalFeatures.InlineEntityReadOnlyDelimiters)
+        ) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


JiuqingSong:
Restore the snapshot below, press LEFT.

Expect: move cursor to the right side of entity
Actual: move cursor to the left side of entity

```html
<div><span>aaaaaaaaabbb<span class="entityDelimiterBefore">​</span><span class="_Entity _EType_aaa _EId_aaa _EReadonly_1" contenteditable="false" style="display: inline-block;"><span data-hydrated-html=""><b>bbbb</b></span></span><span class="entityDelimiterAfter">​</span></span></div><div><span>ddsaf</span></div><!--{"type":0,"isDarkMode":false,"start":[1,0,0,0],"end":[1,0,0,0]}-->
```

BryanValverdeU:
> Restore the snapshot below, press LEFT.
> 
> Expect: move cursor to the right side of entity Actual: move cursor to the left side of entity
> 
> ```
> <div><span>aaaaaaaaabbb<span class="entityDelimiterBefore">​</span><span class="_Entity _EType_aaa _EId_aaa _EReadonly_1" contenteditable="false" style="display: inline-block;"><span data-hydrated-html=""><b>bbbb</b></span></span><span class="entityDelimiterAfter">​</span></span></div><div><span>ddsaf</span></div>
> ```

Hmm, interesting. 

> Restore the snapshot below, press LEFT.
> 
> Expect: move cursor to the right side of entity Actual: move cursor to the left side of entity
> 
> ```
> <div><span>aaaaaaaaabbb<span class="entityDelimiterBefore">​</span><span class="_Entity _EType_aaa _EId_aaa _EReadonly_1" contenteditable="false" style="display: inline-block;"><span data-hydrated-html=""><b>bbbb</b></span></span><span class="entityDelimiterAfter">​</span></span></div><div><span>ddsaf</span></div>
> ```

Good catch

BryanValverdeU:
> Use the snapshot below,
> 
> * press RIGHT, need to press twice to go across entity.
> * press DOWN, expect cursor move to the next line, but it actually go to the entity in the current line
> 
> ```
> <div><span><span>aaaa<span class="entityDelimiterBefore">​</span><span class="_Entity _EType_aaa _EId_aaa _EReadonly_1" style="display:inline-block" contenteditable="false"><span data-hydrated-html=""><b>bbb<b></b></b></span></span><span class="entityDelimiterAfter">​</span>cccc</span></span><br></div><span></span><div><span>aaa<span class="entityDelimiterBefore">​</span><span class="_Entity _EType_ccc _EId_ccc _EReadonly_1" contenteditable="false" style="display: inline-block;"><span data-hydrated-html=""><b>ddd</b></span></span><span class="entityDelimiterAfter">​</span>dasf</span></div>
> ```

Might need another fix for UP and DOWN key weird behavior,

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/entityFeatures.ts
```diff
@@ -214,8 +217,12 @@ function cacheGetNeighborEntityElement(
  */
 const MoveBetweenDelimitersFeature: BuildInEditFeature<PluginKeyboardEvent> = {
     keys: [Keys.RIGHT, Keys.LEFT],
+    allowFunctionKeys: true,
     shouldHandleEvent: (event: PluginKeyboardEvent, editor: IEditor) => {
-        if (!editor.isFeatureEnabled(ExperimentalFeatures.InlineEntityReadOnlyDelimiters)) {
+        if (
+            event.rawEvent.altKey &&
```
JiuqingSong, line 223: why "&&"?

BryanValverdeU: Changed to ||

JiuqingSong (approved):

PR #1650, opened 2023-03-17
Content Model: Set alignment API

juliaroldi:


packages/roosterjs-content-model/lib/domToModel/processors/listItemProcessor.ts
```diff
@@ -23,12 +23,14 @@ export const listItemProcessor: ElementProcessor<HTMLLIElement> = (group, elemen
...
-                const listItem = createListItem(listFormat.levels, context.segmentFormat);
+                const listItem = createListItem(listFormat.levels, context.segmentFormat, element);
+                parseFormat(element, context.formatParsers.listItem, listItem.format, context);
+
                 listFormat.listParent!.blocks.push(listItem);
 
                 parseFormat(
                     element,
-                    context.formatParsers.listItem,
+                    context.formatParsers.listLevel,
```
JiuqingSong, line 33: don't change this. listItem and listLevel are doing different job.

packages/roosterjs-content-model/lib/modelApi/block/setModelAlignment.ts
```diff
@@ -0,0 +1,76 @@
...
+        if (isBlockGroupOfType<ContentModelListItem>(block, 'ListItem')) {
+            block.format.textAlign = newAligment;
+        } else if (block.blockType === 'Table') {
+            alignTable(
+                block,
+                TableAlignMap[alignment][block.format.direction == 'rtl' ? 'rtl' : 'ltr']
+            );
+        } else if (block) {
+            const { format } = block;
+            format.textAlign = newAligment;
```
JiuqingSong, line 71: Actually this part is the same with line 63. We can merge them

packages/roosterjs-content-model/test/domToModel/processors/listItemProcessorTest.ts
```diff
@@ -54,7 +54,7 @@ describe('listItemProcessor', () => {
                     blockType: 'BlockGroup',
                     blockGroupType: 'ListItem',
                     blocks: [],
-                    levels: [{ listType: 'UL', displayForDummyItem: 'block' }],
```
JiuqingSong, line 57: This shows the code has regression. displayForDummyItem needs to be 'block' here

JiuqingSong (changes requested):

packages/roosterjs-content-model/lib/modelApi/creators/createListItem.ts
```diff
@@ -8,14 +8,19 @@ import { createSelectionMarker } from './createSelectionMarker';
  */
 export function createListItem(
     levels: ContentModelListItemLevelFormat[],
-    format?: ContentModelSegmentFormat
+    format?: ContentModelSegmentFormat,
+    textAlign?: string
```
JiuqingSong, line 12: As I commented in another place, this should be done by format handler, but not through a parameter of creator.

JiuqingSong: One of the reason is, here you allow to pass a string, then later you convert the type to "start|end|center". What if people pass in some other string? We can't detect it and there is no build time error. So that will most likely cause bug. To solve it, people need to convert their string to the type we need. But that is already done in format handler, so it is a dup.

packages/roosterjs-content-model/lib/domToModel/processors/listItemProcessor.ts
```diff
@@ -23,7 +23,18 @@ export const listItemProcessor: ElementProcessor<HTMLLIElement> = (group, elemen
                     context
                 );
 
-                const listItem = createListItem(listFormat.levels, context.segmentFormat);
+                const listItem = createListItem(
+                    listFormat.levels,
+                    context.segmentFormat,
+                    element.style.alignSelf
```
JiuqingSong, line 29: same here, leave it to format handler.

packages/roosterjs-content-model/lib/formatHandlers/list/listItemElementFormatHandler.ts
```diff
@@ -0,0 +1,19 @@
+import { ContentModelListItemFormat } from '../../publicTypes/format/ContentModelListItemFormat';
+import { FormatHandler } from '../FormatHandler';
+
+/**
+ * @internal
+ */
+export const listItemElementFormatHandler: FormatHandler<ContentModelListItemFormat> = {
+    parse: (format, element) => {
+        const alignment = element.style.alignSelf;
+        if (alignment && ['end', 'start', 'center'].indexOf(alignment) > -1) {
```
JiuqingSong, line 10: I think all the logic here is already included in directionFormatHandler, right? Then we don't need this handler at all.

packages/roosterjs-content-model/lib/publicTypes/format/ContentModelFormatMap.ts
```diff
@@ -48,7 +49,7 @@ export interface ContentModelFormatMap {
      */
     tableCell: ContentModelTableCellFormat;
 
-    /**
+    /*
```
JiuqingSong, line 52: why change here?

juliaroldi: It happened when I was solving conflicts. I'll fix.

JiuqingSong (approved):

PR #1654, opened 2023-03-20
Remove space after link

juliaroldi:
Remove space when create link. 
linkWithSpaceBefore


JiuqingSong:
This causes the spaces are removed from document, that is not what the bug want to do. We should keep the space, but just not include it in the link.

The right fix should be shrinking the selection in adjustLinkSelection.ts before add link.

PR #1656, opened 2023-03-21
Add Image Format to getFormatState

juliaroldi:
Add Image Format that includes the image border and box shadow styles to Format State.

### DIFF
--- a/packages/roosterjs-content-model/lib/modelApi/common/retrieveModelFormatState.ts
+++ b/packages/roosterjs-content-model/lib/modelApi/common/retrieveModelFormatState.ts
@@ -1,10 +1,12 @@
 import { ContentModelBlock } from '../../publicTypes/block/ContentModelBlock';
 import { ContentModelBlockGroup } from '../../publicTypes/group/ContentModelBlockGroup';
 import { ContentModelDocument } from '../../publicTypes/group/ContentModelDocument';
+import { ContentModelImage } from 'roosterjs-content-model/lib/publicTypes';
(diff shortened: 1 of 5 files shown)


packages/roosterjs-content-model/lib/modelApi/common/retrieveModelFormatState.ts
```diff
@@ -59,6 +62,11 @@ export function retrieveModelFormatState(
                         segments.some(segment => segment.segmentType == 'Image');
 
                     isFirst = false;
+
+                    if (segment.segmentType === 'Image' && isFirstImage) {
```
JiuqingSong, line 66: A question to PM: for segments format, if the selection contains multiple different format, such as different font names, we will return undefined for that format so that user know that there are different fonts selected. Do we want to do the same for image?

packages/roosterjs-editor-types/lib/interface/FormatState.ts
```diff
@@ -84,6 +85,11 @@ export interface ElementBasedFormatState {
      */
     canAddImageAltText?: boolean;
 
+    /**
+     * Format of image, if there is table at cursor position
+     */
+    imageFormat?: ImageFormat;
```
JiuqingSong, line 91: For other format sub types, they are all supported by both Content model and original API. But this one is only supported by Content Model. In that case, I think we can have a sub type of FormatState in Content Model package to hold image format. So when people use original API, they will only see original formats.

What do you think?

JiuqingSong (approved):

PR #1657, opened 2023-03-22
Fixed plus sign position on table handle

Andres-CT98:
### Description:

The plus sign in the table handle used to add more rows and columns is misaligned with the circle.

### Fix

Modified the value on line-height to rise by 2 pixels the position of the plus sign on the table handler. This was done to better center its position, especially when the handle is on a horizontal position.

### Cases tested:

Observed the handle in vertical and horizontal position.

**Before:**

image

**After:**

image

### Warnings and implications

- There is still some misalignment of the overall handle top and stick.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableInserter.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableInserter.ts
@@ -108,7 +108,7 @@ function getInsertElementData(
     backgroundColor: string
 ): CreateElementData {
     const inserterColor = isDark ? INSERTER_COLOR_DARK_MODE : INSERTER_COLOR;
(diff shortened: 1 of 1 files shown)


JiuqingSong:
Please add some description and/or screenshots to show the difference before and after your change?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableInserter.ts
```diff
@@ -108,7 +108,7 @@ function getInsertElementData(
     backgroundColor: string
 ): CreateElementData {
     const inserterColor = isDark ? INSERTER_COLOR_DARK_MODE : INSERTER_COLOR;
-    const outerDivStyle = `position: fixed; width: ${INSERTER_SIDE_LENGTH}px; height: ${INSERTER_SIDE_LENGTH}px; font-size: 16px; color: ${inserterColor}; line-height: 10px; vertical-align: middle; text-align: center; cursor: pointer; border: solid ${INSERTER_BORDER_SIZE}px ${inserterColor}; border-radius: 50%; background-color: ${backgroundColor}`;
```
juliaroldi, line 111: is it possible to use css like align: center to style the plus sign, so we do not need to relay on hard coded number?

Andres-CT98: The way the handle is constructed mostly uses hard coded values. We already use vertical and horizontal align to center it, however that is still too low, so we use line-height to correct it further.

JiuqingSong (approved):

PR #1659, opened 2023-03-23
Fix for pasting an Excel Online table inside RoosterJS table

Andres-CT98:
### Description:

When pasting an Excel Online table inside another table, it now merges them both using the place the cursor is located as a pivot. Previously, tables from Excel Online were nested.

### Fix

Tables copied from Excel desktop and online share the attribute `<meta name=ProgId content=Excel.Sheet>`, however only desktop tables have the attribute `xmlns:x="urn:schemas-microsoft-com:office:excel"`. Using this difference, it is possible to make necessary changes to only Excel online tables.

Changed the convertPastedContentFromExcel function to extract the `<table>` that comes inside the `<div>` that is created when copying a table in Excel Online.

### Cases tested:

Paste any table copied from Excel Online into a table created in Rooster.

**Before:** Pasting a table inside a table resulted in a nested table.

**After:** Pasting a table inside a table results in a new merged table.

### Warnings and implications

- Older versions of Excel were not tested to see if this change affects their pasting, only the Windows 11 and current Online ones were used.


packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
```diff
@@ -69,7 +69,9 @@ export default class Paste implements EditorPlugin {
                     convertPastedContentFromWord(event);
                     break;
                 case KnownSourceType.ExcelDesktop:
-                    // Handle HTML copied from Excel
+                // Handle HTML copied from Excel Desktop
```
JiuqingSong, line 72: You can merge this comment together with line 74 to be one single comment

packages/roosterjs-editor-plugins/lib/plugins/Paste/sourceValidations/isExcelDesktopDocument.ts
```diff
@@ -1,9 +1,8 @@
-import { PROG_ID_NAME } from './constants';
 import type { getSourceFunction, getSourceInputParams } from './getPasteSource';
 
-const EXCEL_ATTRIBUTE_NAME = 'xmlns:x';
+// Excel Online does not have this attribute
+export const EXCEL_DESKTOP_ATTRIBUTE_NAME = 'xmlns:x';
```
JiuqingSong, line 4: Since we already have a constant.ts, you can put this constant there.

Also please add comment "@internal" to avoid generating document for this export

packages/roosterjs-editor-plugins/lib/plugins/Paste/excelConverter/convertPastedContentFromExcel.ts
```diff
@@ -19,6 +19,15 @@ export default function convertPastedContentFromExcel(
     const { fragment, sanitizingOption, htmlBefore, clipboardData } = event;
     const html = excelHandler(clipboardData.html, htmlBefore);
 
+    // For Excel Online
+    if (fragment.childNodes.length > 0 && getTagOfNode(fragment.firstChild) == 'DIV') {
+        fragment.firstChild.childNodes.forEach(child => {
+            if (getTagOfNode(child) == 'TABLE') {
```
JiuqingSong, line 25: Is it possible there are multiple TABLE child under DIV?

packages/roosterjs-editor-plugins/lib/plugins/Paste/excelConverter/convertPastedContentFromExcel.ts
```diff
@@ -19,6 +19,15 @@ export default function convertPastedContentFromExcel(
...
+                event.fragment.replaceChildren(child);
+            }
+        });
+    }
+
     if (clipboardData.html != html) {
         const doc = new DOMParser().parseFromString(trustedHTMLHandler(html), 'text/html');
         moveChildNodes(fragment, doc?.body);
```
JiuqingSong, line 33: Actually this line will overwrite what you have done from line 22. Is that really ok?

packages/roosterjs-editor-plugins/lib/plugins/Paste/sourceValidations/isExcelOnlineDocument.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @internal
+ * Checks whether the Array provided contains strings that identify Excel Online documents
+ * @param props Properties related to the PasteEvent
+ * @returns
+ */
+const isExcelOnlineDocument: getSourceFunction = (props: getSourceInputParams) => {
+    const { htmlAttributes } = props;
+    return (
+        htmlAttributes[PROG_ID_NAME] == EXCEL_ONLINE_ATTRIBUTE_VALUE &&
+        htmlAttributes[EXCEL_DESKTOP_ATTRIBUTE_NAME] == undefined
```
JiuqingSong, line 18: I'm a little confused here.

So in the original code, we check:
```ts
htmlAttributes["xmlns:x"] == 'urn:schemas-microsoft-com:office:excel' ||
htmlAttributes["ProgId"] == 'Excel.Sheet'
```

After the change, it checks:
```ts
htmlAttributes['xmlns:x'] == 'urn:schemas-microsoft-com:office:excel'  // For Excel desktop
||
(htmlAttributes['ProgId'] == 'Excel.Sheet' && htmlAttributes['xmlns:x'] == undefined) // For Excel online
```

So the only difference is we also want htmlAttributes['xmlns:x'] must be undefined when it is for Excel Online. What is the reason to add this check?

packages/roosterjs-editor-plugins/lib/plugins/Paste/excelConverter/convertPastedContentFromExcel.ts
JiuqingSong: Does it mean fragment.firstChild should always only have one child?

packages/roosterjs-editor-plugins/lib/plugins/Paste/sourceValidations/isExcelOnlineDocument.ts
Andres-CT98: Since both share Excel.Sheet something needs to differentiate them. If I omit the explicit check for the presence or absence of 'xmlns:x' I would need to make clear that the check for Desktop should always come after Online. Since no other checks are bound by order I decided to not break that.

Does element have htmlAttributes['ProgId'] == 'Excel.Sheet'? If yes then it is confirmed Excel but not which type. If I check for Desktop first and Online second, without explicitly asking each time about Excel.Sheet, getPasteSource() will always return Online, so the order would matter.

packages/roosterjs-editor-plugins/lib/plugins/Paste/excelConverter/convertPastedContentFromExcel.ts
Andres-CT98: In my copies, the fragment always was:

    <div>
        <head>
            <meta/>
            <meta/>
            <meta/>
        <head/>
        <body>
            <table>
                 [Table]
            </table>
        <body/>
    </div>

packages/roosterjs-editor-plugins/lib/plugins/Paste/sourceValidations/isExcelOnlineDocument.ts
JiuqingSong: I see. Please add comment in code to explain this.

packages/roosterjs-editor-plugins/lib/plugins/Paste/excelConverter/convertPastedContentFromExcel.ts
JiuqingSong: I tried with your code ,and I see there are some META and TABLE under DIV, but no HEAD and BODY.

Anyway, I think we'd better check if the only child nodes are 0..n META and exactly 1 TABLE, only in that case we can replace the content.  In all other case just skip processing them, so that will be safer. Something like:

```ts
let hasTable = false;
firstChild.childNodes.every(x => {
  if (x.tagName == 'META') {
    return true;
  } else if (x.tagName == 'TABLE' && !hasTable) {
    hasTable = true;
    return true;
  } else {
    return false;
 });
```

packages/roosterjs-editor-plugins/lib/plugins/Paste/excelConverter/convertPastedContentFromExcel.ts
```diff
@@ -24,6 +24,15 @@ export default function convertPastedContentFromExcel(
...
     }
 
+    // For Excel Online
```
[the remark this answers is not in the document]
Andres-CT98: I did that at first, but since both will be sharing the same 'safeguard' functions of checking for html and chainSanitizerCallback I combined kept them as one.

JiuqingSong (approved):

PR #1673, opened 2023-03-27
Dispose data of picker plugin and add Strict Mode

juliaroldi:
When dispose plugin picker, also dispose the data provider and picker option to avoid memory leaks.


packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -132,6 +132,8 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
         this.currentInputLength = null;
         this.newInputLength = null;
         this.dataProvider.onDispose();
+        this.dataProvider = undefined;
```
JiuqingSong, line 135: nit, this may cause problem when we go to strict mode.

You can declare another local var to allow undefined and assign dataProvider to it. Then do null check when we use it.

JiuqingSong (approved):

JiuqingSong:
I see there is a big change in the latest iteration. So give me some time to read it again.

juliaroldi:
> I see there is a big change in the latest iteration. So give me some time to read it again.

I added strict mode, I'm testing in OWA to check that will cause a regression.

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -53,18 +53,65 @@ const UNIDENTIFIED_CODE = [0, 229];
...
+
+    public willHandleEventExclusively(event: PluginEvent) {
+        return this.core ? this.core.willHandleEventExclusively(event) : false;
+    }
+
+    public initialize(editor: IEditor) {
+        this.editor = editor;
+        if (this.editor && this.dataProvider && this.pickerOptions) {
+            this.core = new PickerPluginCore(this.editor, this.dataProvider, this.pickerOptions);
+            this.core.initialize(editor);
```
JiuqingSong, line 74: For the core class, when it is created, it already has everything (editor, provider, options) it needs, so you can let its ctor do initialize directly, no need to have a separate initialize function

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -53,18 +53,65 @@ const UNIDENTIFIED_CODE = [0, 229];
  */
 export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvider>
     implements EditorPlugin {
-    private editor: IEditor;
-    private eventHandledOnKeyDown: boolean;
-    private blockSuggestions: boolean;
-    private isSuggesting: boolean;
-    private lastKnownRange: Range;
+    private editor: IEditor | undefined = undefined;
```
JiuqingSong, line 56: I think editor is not required for this class at all. It is only needed in the core class and we have already pass it in when it is constructed.

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -53,18 +53,65 @@ const UNIDENTIFIED_CODE = [0, 229];
...
+    public initialize(editor: IEditor) {
+        this.editor = editor;
+        if (this.editor && this.dataProvider && this.pickerOptions) {
+            this.core = new PickerPluginCore(this.editor, this.dataProvider, this.pickerOptions);
+            this.core.initialize(editor);
+        }
+    }
+
+    public onPluginEvent(event: PluginEvent) {
+        if (this.editor && this.dataProvider && this.core) {
```
JiuqingSong, line 79: We only need to do null check for core.

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -53,18 +53,65 @@ const UNIDENTIFIED_CODE = [0, 229];
...
+        this.dataProvider = undefined;
+        this.pickerOptions = undefined;
+        if (this.core) {
+            this.core.dispose();
+            this.core = undefined;
+        }
+    }
+
+    public getName() {
+        return this.core ? this.core.getName() : '';
```
JiuqingSong, line 95: We can directly return the name string here, no need to get it from core

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -53,18 +53,65 @@ const UNIDENTIFIED_CODE = [0, 229];
...
+    public onPluginEvent(event: PluginEvent) {
+        if (this.editor && this.dataProvider && this.core) {
+            this.core.onPluginEvent(event);
+        }
+    }
+
+    public dispose() {
+        this.editor = undefined;
+        this.dataProvider = undefined;
+        this.pickerOptions = undefined;
```
JiuqingSong, line 87: Setting them to undefined should be helpful for mem leak, a side effect is that this plugin instance will become unusable because there is no chance to reset them.

This should be ok for now, since we don't expect an instance of plugin to be reused. But not sure if 3rd party people will do this, after all we didn't disable this way of use.

Let's open an issue to track this, and for this fix it is ok here I think.

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
juliaroldi: This editor is for the plugin picker class, we pass it in the PickerPluginCore constructor.

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -114,24 +160,25 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
-        this.isPendingInputEventHandling = null;
-        this.currentInputLength = null;
-        this.newInputLength = null;
+        this.isPendingInputEventHandling = false;
+        this.currentInputLength = undefined;
+        this.newInputLength = undefined;
         this.dataProvider.onDispose();
+        this.editor = null!;
+        this.dataProvider = null!;
+        this.pickerOptions = null!;
```
JiuqingSong, line 181: Please add comment to explain why we need to and it is safe to set to "null!" here

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
JiuqingSong: The fact is, we only get editor instance from initialize() function, then we pass it into core, then we will never use it again. That means the picker class doesn't need to hold it at all.

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -53,18 +53,65 @@ const UNIDENTIFIED_CODE = [0, 229];
...
+            this.core = null;
+        }
+    }
+
+    public getName() {
+        return 'Picker';
+    }
+}
+
+class PickerPluginCore<T extends PickerDataProvider = PickerDataProvider> implements EditorPlugin {
```
JiuqingSong, line 99: Actually, this class doesn't need to implement EditorPlugin interface, and that is why I suggest merge constructor and initialize function together.

juliaroldi: I see, I got it now.

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -53,32 +53,64 @@ const UNIDENTIFIED_CODE = [0, 229];
  */
 export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvider>
     implements EditorPlugin {
-    private editor: IEditor;
-    private eventHandledOnKeyDown: boolean;
-    private blockSuggestions: boolean;
-    private isSuggesting: boolean;
-    private lastKnownRange: Range;
+    private editor: IEditor | null = null;
```
JiuqingSong, line 56: this is still not removed

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -53,32 +53,64 @@ const UNIDENTIFIED_CODE = [0, 229];
...
+    constructor(dataProvider: T | undefined, pickerOptions: PickerPluginOptions) {
+        this.dataProvider = dataProvider;
+        this.pickerOptions = pickerOptions;
+    }
 
-    // For detecting backspace in Android
-    private isPendingInputEventHandling: boolean = false;
-    private currentInputLength: number;
-    private newInputLength: number;
+    public willHandleEventExclusively(event: PluginEvent) {
```
JiuqingSong, line 66: Originally this class doesn't have willHandleEventExclusively, so no need to add it

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -53,32 +53,64 @@ const UNIDENTIFIED_CODE = [0, 229];
...
+        this.editor = null;
+        this.dataProvider = undefined;
+        this.pickerOptions = undefined;
+        if (this.core) {
+            this.core.dispose();
+            this.core = null;
+        }
+    }
+
+    public getName() {
```
JiuqingSong, line 93: comment

JiuqingSong (approved):

PR #1678, opened 2023-03-28
Enable Strict Mode in packages\roosterjs-editor-api\lib\format

BryanValverdeU:
Enable Strict mode in `packages\roosterjs-editor-api\lib\format`


JiuqingSong (approved):

JiuqingSong:
#886

PR #1685, opened 2023-03-31
Strict mode in paste plugin

juliaroldi:
Add strict mode in Paste Plugin


JiuqingSong (approved):

JiuqingSong:
Do you need to change the parent tsconfig to include your new tsconfig file?

juliaroldi:
> Do you need to change the parent tsconfig to include your new tsconfig file?
Yes!! I forget, that is why that it does not caused any error in the build. I created a revert PR.

PR #1687, opened 2023-03-31
Add strict mode to paste

juliaroldi:
Add strict mode to Paste plugin and functions related to it.


packages/roosterjs-editor-plugins/lib/plugins/Paste/wordConverter/converterUtils.ts
```diff
@@ -28,120 +28,126 @@ const LINE_BREAKS = /[\n|\r]/gi;
...
-            } else if (!listMetadata.ignore && listMetadata.numberOfItems == 1) {
-                // This is the second item we've seen for this list.. we'll compare the 2 fake bullet
-                // items we have an decide if we create ordered or unordered lists based on this.
-                // This is the best way we can do this since we cannot read the metadata that Word
-                // puts in the head of the HTML...
-                let secondFakeBullet = getFakeBulletText(node, LOOKUP_DEPTH);
-                listMetadata.tagName =
-                    listMetadata.firstFakeBullet == secondFakeBullet ? 'UL' : 'OL';
-            }
+    if (args) {
```
BryanValverdeU, line 31: nit: We could add a guard check for args and return false when args is falsy. So we do not need to wrap all the code in the function

JiuqingSong (approved):

PR #1688, opened 2023-03-31
Content Model Shadow Edit Step 1: Support clone model

JiuqingSong:
In order to support shadow edit in Content Model, we will cache a model before edit, then we need to clone a model every time we change a format. This is the change to add API `cloneModel`.

We will do deep clone for a content model for format, blocks, segments, decorators. For cached element and entity wrapper, we don't need to clone them, but just copy a reference of the element.

Related test case added.


JiuqingSong:
#1691

PR #1689, opened 2023-03-31
Content Model Shadow Edit Step 2: Split Editor class

JiuqingSong:
In order to allow `ContentModelEditor` append more info into editor core and override core api, in this change I split the `Editor` class into two parts:
- EditorBase: A generic class that support a generic `EditorCore` object, and accept a editorCoreCreator function to create editor core
- Editor: Child class of `EditorBase`, provide creator of EditorCore

Also split some code out of Editor to be separate function:
- select: Move it out to be a Core API
- isFeatureEnabled: Move it out to be a separate function so that it can be called outside editor class.

In later step I'll make `ContentModelEditor` another child class of `EditorBase` and provide its own EditorCore type and editorCoreCreator so that we can override some behavior of Editor.


JiuqingSong:
#1691

Issue #1691, opened 2023-03-31
Content Model Shadow Edit


JiuqingSong:
Need to support shadow edit with Content Model

Currently there are some issues:
1. When using shadow edit with Content Model API, the cached the model can be changed and not change back
2. Table selection is not well handled

PR #1693, opened 2023-04-03
Content Model Shadow Edit Step 1: Split editor

JiuqingSong:
In order to allow `ContentModelEditor` append more info into editor core and override core api, in this change I split the `Editor` class into two parts:
- EditorBase: A generic class that support a generic `EditorCore` object, and accept a editorCoreCreator function to create editor core
- Editor: Child class of `EditorBase`, provide creator of EditorCore

This change is just code moving, no logic change


JiuqingSong:
#1691

BryanValverdeU (approved):

PR #1694, opened 2023-04-03
Content Model Shadow Edit Step 2: Refactor Editor class

JiuqingSong:
To allow do more customization to Editor class, in this change I did:
1. Make EditorBase a generic class, allow change the type of EditorCore and EditorOptions
2. Move the implementation if select function out of editor to be a core api
3. Move isFeatureEnabled function out of Editor class to be a normal function
4. Move the code of creating EditorCore out to be a replaceable function, so that child class can pass in different implementation and type


JiuqingSong:
#1691

BryanValverdeU (approved):

PR #1696, opened 2023-04-03
Content Model Shadow Edit Step 4: Finally support shadow edit with Content Model

JiuqingSong:
In order to support shadow edit with Content Model, we need to override the original core API `switchShadowEdit` and allow call other functions like create content model. So in this change I split some functions out of ContentModelEditor to be new core API:
- createContentModel
- createEditorContext
- setContentModel
- switchShadowEdit: This function will make sure there is a cached content model before entering shadow edit state, and use it to restore content after shadow edit is done. When format with content model during shadow edit, we will now clone a model from the cached one and do format change on it. If format using original API that is not based on Content Model, it will still use the cached fragment so that everything keeps no change.

Then we create a new function `createContentModelEditorCore` that will invoke original `createEditorCore` function and build editor core for `ContentModelEditor`, including appending these new core APIs, override shadow edit API when reuse model is allowed.

We also disable changing cached model during shadow edit to make sure the cached model is always the same one for the content before shadow edit.

Related test cases added.


JiuqingSong:
#1691

Issue #904, opened 2022-04-11
Default Format changed after switching to dark mode and then switching back


haven2world:
**Describe the bug**
If we set the value of the `textColor(s)`/`backgroundColor(s)` of `defaultFormat` to `undefined` when we initialize the RoosterEditor, and everything works well in light mode. But if we switch to dark mode and `reCalculateDefaultFormat` in `LifecyclePlugin` will be triggered and the default format will be changed. And even if we switch back to light mode, the default format won't be restored.

**To Reproduce**
1. Open roosterJS Demo site
2. Clear all content and then enter some new lines
3. Take a look at the inline style of the added lines
4. The inline style is empty
5. Turn on Dark mode
6. Clear all content and then enter some new lines
7. Take a look at the inline style of the added lines
8. The inline style has a dark background and white font color
9. Turn off dark mode
10. Clear all content and then enter some new lines
11. Take a look at the inline style of the added lines
12. There are still some inline style

**Expected behavior**

**Screenshots**
Step 4:
<img width="465" alt="image" src="https://user-images.githubusercontent.com/17299418/162714138-a90de422-ffe1-4afa-8a81-b1b692ca4d55.png">
Step 8:
<img width="1265" alt="image" src="https://user-images.githubusercontent.com/17299418/162714445-872cf0bd-5dc1-4ba2-8ae9-e1cb53fdcd32.png">
Step 12:
<img width="1089" alt="image" src="https://user-images.githubusercontent.com/17299418/162714880-c77fe92e-1004-4e64-ad91-7ec529ca5d9f.png">

**Device Information**
 - OS: Mac OS
 - Browser chrome
 - Version 100.0.4896.75 (Official Build) (x86_64)

**Additional context**

JiuqingSong:
Fixed in #1683

PR #1700, opened 2023-04-04
Fix #1683: Do not apply text and background color when not necessary

### LINKED ISSUE #1683: Do not apply text and background color when not necessary
When text color and background color is same with default one, no need to apply color in style.

Or at least, no need to add background color.

The issue causes text in table cell has background color so when apply shade color, text background color can cover shade color

JiuqingSong:
In the existing code, we will always get text color and background color no matter if we set them in default format in LifecyclePluign. This is not necessary for now, so we can remove them.

And we also do not need recalculate color for color mode change since they will be recalculated when apply.

After this change, we will only apply format for those have default values, and leave empty for others.

Also change the CDN address for react and react-dom in demo site since the original one is very unstable.


JiuqingSong:
#904 
#1683

juliaroldi (approved):

PR #1707, opened 2023-04-05
Remove max-width when editing image

juliaroldi:
When editing image mode we restore the original image and use the wrapper to overlay the part the was cut, then we cannot use max-width in the editing image, because it will fit the whole image inside the wrapper instead of overlay it. 
So if the selected image has max-width, when create the editing image, remove max-width. 

Before:
BeforeFixImageCrop

After: 
AfterFixImageCrop


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -404,6 +404,7 @@ export default class ImageEdit implements EditorPlugin {
             //Clone the image and insert the clone in a entity
             this.clonedImage = this.image.cloneNode(true) as HTMLImageElement;
             this.clonedImage.removeAttribute('id');
+            this.clonedImage.style.removeProperty('max-width');
```
BryanValverdeU, line 407: nit: add a unit test where we remove the max-width

BryanValverdeU (approved):

Issue #1322, opened 2022-10-10
Copying tables turns them into single line


Andres-CT98:
**Describe the bug**

Copying and pasting in a certain way causes the table to turn into a single line.

**To Reproduce**
1. Copy this table into the roosterjs demo site: 
<table><tr><td>1</td><td>2</td><td>3</td></tr><tr><td>4</td><td>5</td><td>6</td></tr><tr><td>7</td><td>8</td><td>9</td></tr></table>

2. Paste the table on the editor

3. Copy the table on the editor while using table selection, the table needs to look all greyed out

4. Paste it again, it will turn into a line

**Expected behavior**
Table pastes like normal. This behaviour can be seen if the table is copied using normal selection, where the contents of the table are highlighted blue.

**Device Information**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 8.33.1

jvillalobos:
I'm not able to reproduce this in Firefox. When you paste the table on the demo site, it's pasted with less formatting, but copying it and pasting it again works correctly. It seems there's something in the styling that is preserved on Edge and Chrome that is breaking the following copy/paste actions.

Issue #1065, opened 2022-06-28
Pasting a list item on top of another list item causes odd behavior


jvillalobos:
Steps to reproduce:

1. Start a list (bulleted or numbered)
2. Paste a **full** list item from another app. I think case I used OneNote and triple-clicked on the item, so the entire item, including the bullet, was selected.

The result is that when you paste it, you get nested bullets:

![image](https://user-images.githubusercontent.com/62568/176309955-f9d361cc-aef7-4dc7-9be1-03838df826a4.png)

If you click on the paste button and select Text Only, the result is also strange:

![image](https://user-images.githubusercontent.com/62568/176310767-8d4ae684-c40a-4edf-a9d7-2439ee5bf9bb.png)

I think the expected behavior in this case is that the lists should be merged, meaning that the pasted items should be at the same level as where the cursor is. They should probably use the bullet or number style from the origin, but I would consider that more of a nice to have.

jvillalobos:
Here are the clipboard contents: https://1drv.ms/u/s!AsQmD_D-z0z_gdk7rJ_RjOm-HbWEsw?e=CAsSJf 

And the before and after of the paste operation:

```
<div style="font-family: &quot;Segoe UI&quot;, &quot;Segoe UI Web (West European)&quot;, &quot;Segoe UI&quot;, -apple-system, BlinkMacSystemFont, Roboto, &quot;Helvetica Neue&quot;, sans-serif; font-size: 10.5pt; color: rgb(255, 255, 255); background-color: rgb(51, 51, 51);" data-ogsc="rgb(0,0,0)" data-ogsb="rgb(255,255,255)"><ul><li><br></li></ul></div><!--{"type":0,"isDarkMode":true,"start":[0,0,0,0],"end":[0,0,0,0]}-->
```

```
<div style="font-family: &quot;Segoe UI&quot;, &quot;Segoe UI Web (West European)&quot;, &quot;Segoe UI&quot;, -apple-system, BlinkMacSystemFont, Roboto, &quot;Helvetica Neue&quot;, sans-serif; font-size: 10.5pt; color: rgb(255, 255, 255); background-color: rgb(51, 51, 51);" data-ogsc="rgb(0,0,0)" data-ogsb="rgb(255,255,255)"><ul><li><ul style="unicode-bidi:embed;margin-top:0in;margin-bottom:0in" type="disc">
[...]

Issue #920, opened 2022-04-21
Pasting bullets doesn't preserve level styling


JiuqingSong:
1) Copy a bulleted list with multiple levels (I used OneNote)
![image](https://user-images.githubusercontent.com/23065085/164557956-6365717a-6e78-4203-9c3b-ad21f7b4b5a2.png)
2) Paste it in OWA
![image](https://user-images.githubusercontent.com/23065085/164557976-4a52f775-f02a-451a-a842-de7249ea63a6.png)
As you can see, the third level bullet doesn't use the correct bullet type. It's odd because the default third level bullet in OWA is the white square, so there's no reason to change it.
3) Making changes to the list, like creating a new level, fixes the third bullet point but also changes indentation

BryanValverdeU:
The pasting part is fixed, however we may need to also add support on the Content Edit Feature so we do not remove the margins of the list item when writing back

JiuqingSong:
Tested on lates build, everything is working well now.

Issue #456, opened 2020-07-22
Paste (insertNode) under B/I/U tag causes the inserted content has B/I/U style


JiuqingSong:

huahuahu:
Hi, any updates?

JiuqingSong:
This needs a large redesign about how we handle the B/I/U tags. It is in our plan but may not happen very soon.
A workaround will be having a plugin to handle BeforePasteEvent, and split existing B/I/U tags if any, then put cursor out of those tags, so that pasting will work correctly in most cases. But in case B/I/U is put outside of a block node (e.g. DIV), there will still be issue.

BryanValverdeU:
This should be fixed with

Experimental Feature: Paste with content model

PR #1716, opened 2023-04-13
Leverage Content Model to insert the Pasted content from clipboard

BryanValverdeU:
In order to start using content model to insert clipboard content to the editor we need to do some refactor on some utilities:

1. Move the createPasteFragment logic from roosterjs-editor-core package to roosterjs-editor-dom, so we can also use the logic in roosterjs-content-model.
2. Add a new property to BeforePasteEvent to let other plugins modify the way we process elements in the paste fragment, this will be used later in another PR.
3. Create a new core api `createPasteModel` that will use the PasteFragment to create and return a content model document.
4. In the ContentModelEditor Class, override the paste function so, instead of using insertContent, use mergeModel function

Bugs that may be fixes by leveraging Content Model with this change:
#456
#1065
#1322 partially, only the paste part is fixed, copy needs to fixed  in another PR.
#920


packages/roosterjs-content-model/lib/editor/coreApi/createPasteFragment.ts
```diff
@@ -0,0 +1,248 @@
...
+                },
+            }
+        );
+        contentModelToDom(
+            fragment.ownerDocument,
+            fragment,
+            model,
+            cmCore.api.createEditorContext(cmCore)
+        );
+    }
```
BryanValverdeU, line 185: This is a copy of the original createCopyFragment the only change is this:

packages/roosterjs-content-model/lib/publicTypes/event/PasteElementProcessor.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * A function type to process HTML element when do DOM to Content Model conversion in Paste Event.
+ * If the function returns true, means that the element processing was handled by the function.
+ * @param group Parent content model group
+ * @param element The element to process
+ * @param context The context object to provide related information
+ */
+
+export type PasteElementProcessor<T extends Node> = (
```
JiuqingSong, line 12: What is the difference between this interface and ElementProcessor?

packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
```diff
@@ -57,4 +60,51 @@ export default class ContentModelEditor
...
+     * false to keep original format.  Default value is false. When pasteAsText is true, this parameter is ignored
+     */
+    public paste(
+        clipboardData: ClipboardData,
+        pasteAsText: boolean = false,
+        applyCurrentFormat: boolean = false,
+        pasteAsImage: boolean = false
+    ) {
+        const core = this.getCore();
+        if (!clipboardData) {
```
JiuqingSong, line 78: We have already enabled strict mode, so we don't need this check here

packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
```diff
@@ -57,4 +60,51 @@ export default class ContentModelEditor
...
+            applyCurrentFormat,
+            pasteAsImage
+        );
+
+        if (model) {
+            const currentModel = this.createContentModel();
+
+            mergeModel(currentModel, model);
+
+            this.setContentModel(currentModel);
```
JiuqingSong, line 107: Use formatWithContentModel(), and pass in correct change source. Otherwise it won't add undo snapshot.

packages/roosterjs-content-model/lib/editor/coreApi/createPasteModel.ts
```diff
@@ -0,0 +1,70 @@
...
+        core,
+        clipboardData,
+        position,
+        pasteAsText,
+        applyCurrentStyle,
+        pasteAsImage,
+        event
+    );
+
+    return domToContentModel(wrap(fragment, 'span'), core.api.createEditorContext(core), {
```
JiuqingSong, line 36: is it necessary to call `wrap` here?

packages/roosterjs-content-model/lib/editor/createContentModelEditorCore.ts
```diff
@@ -54,10 +55,10 @@ export function promoteToContentModelEditorCore(
         // Only use Content Model shadow edit when reuse model is enabled because it relies on cached model for the original model
         cmCore.api.switchShadowEdit = switchShadowEdit;
     }
-
     cmCore.originalApi.createEditorContext = createEditorContext;
     cmCore.originalApi.createContentModel = createContentModel;
     cmCore.originalApi.setContentModel = setContentModel;
+    cmCore.api.createPasteModel = createPasteModel;
```
JiuqingSong, line 61: Let's move this up to be together with other api assignment, and also assign to originalApi

packages/roosterjs-content-model/lib/editor/coreApi/createPasteModel.ts
```diff
@@ -0,0 +1,70 @@
...
+        processorOverride: {
+            element: (group, element, context) => {
+                const wasHandled =
+                    event.elementProcessors.length > 0 &&
```
JiuqingSong, line 40: So if one plugin pass in a process that can process this element, other processors will be ignored?

packages/roosterjs-content-model/lib/publicTypes/event/PasteElementProcessor.ts
BryanValverdeU: Only that the return type needs to be a boolean.

packages/roosterjs-content-model/lib/editor/coreApi/createPasteModel.ts
BryanValverdeU: Yes, domToContentModel's first parameter only accept HTMLElement, Document fragment seem to be not compatible. Wrapping it a span element should work.

packages/roosterjs-content-model/lib/editor/coreApi/createPasteModel.ts
BryanValverdeU: Yes, I was thinking that. Otherwise if multiple plugins can process the same element, we would have duplicated elements in the model. Another option is instead only handle the last element processor that was provided in the event. Similar to what chainSanitizerCallback does.

packages/roosterjs-content-model/lib/editor/coreApi/createPasteModel.ts
JiuqingSong: Can you try to make domToContentModel accept Node instead of HTMLElement, so no need to wrap here

packages/roosterjs-content-model/lib/publicTypes/event/PasteElementProcessor.ts
JiuqingSong: In that case, I prefer change the common processor type to force it return a boolean and change all exiting processors to return true. Although this will change more files, but it makes things consitent.

BryanValverdeU: I think instead of that I will just pass in a domToModelOption to all the event so we can change the behavior as needed.

packages/roosterjs-content-model/lib/modelApi/common/insertContent.ts
```diff
@@ -12,11 +12,7 @@ export function insertContent(
     htmlContent: DocumentFragment | HTMLElement | ContentModelDocument,
     isFromDarkMode?: boolean
 ) {
-    if (safeInstanceOf(htmlContent, 'DocumentFragment')) {
-        htmlContent = wrap(htmlContent, 'span');
-    }
-
-    if (safeInstanceOf(htmlContent, 'HTMLElement')) {
+    if (safeInstanceOf(htmlContent, 'Node')) {
```
JiuqingSong, line 15: This is nice, thanks!

JiuqingSong (approved):

PR #1717, opened 2023-04-13
Handle Word Desktop Paste content

BryanValverdeU:
This PR adds: 

- [X] Add a new plugin for content model paste that will handle the BeforePasteEvent
- [X] Add parsers for Not supported Colors and Links that should be run always
- [X] Support to handle paste content when the source is Word Desktop by adding a new processor for Word Lists and Word Comments.

To test enable the Content model paste plugin in the Side pane -> Editor Options -> Plugins and disable the default paste plugin

Known issues:

1. Pasting content with Background color under the fragment is not pasted correctly.
2. Pasting Tables from Word Desktop may leave white background color


packages/roosterjs-content-model/lib/editor/plugins/PastePlugin/utils/linkParser.ts
```diff
@@ -0,0 +1,39 @@
...
+    } catch {
+        url = undefined;
+    }
+
+    if (
+        !url ||
+        !(
+            url.protocol === HTTP ||
+            url.protocol === HTTPS ||
+            url.protocol === NOTES
```
JiuqingSong, line 31: There are requests to allow more protocols, such as onenote.

We can open bug to discuss what should we do here. For now we can keep going with this.

packages/roosterjs-content-model/lib/editor/plugins/PastePlugin/utils/deprecatedColorParser.ts
```diff
@@ -0,0 +1,46 @@
...
+    'threedface',
+    'threedhighlight',
+    'threedlightshadow',
+    'threedfhadow',
+    'window',
+    'windowframe',
+    'windowtext',
+];
+
+const deprecatedColorParser = (
```
JiuqingSong, line 30: Add type declaration here

packages/roosterjs-content-model/lib/editor/plugins/PastePlugin/ContentModelPastePlugin.ts
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        const ev = event as ContentModelBeforePasteEvent;
+        if (!ev.domToModelOption) {
+            return;
+        }
+        const pasteSource = getPasteSource(event, false);
+        switch (pasteSource) {
+            case KnownPasteSourceType.WordDesktop:
+                handleWordDesktop(ev);
```
JiuqingSong, line 75: So this content will be processed by both old and new paste plugin, right? Is there any issue with that?

packages/roosterjs-content-model/lib/editor/plugins/PastePlugin/WordDesktop/handleWordDesktopPaste.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+function listLevelParser(
+    format: ContentModelListItemLevelFormat,
+    element: HTMLElement,
+    context: DomToModelContext,
+    defaultStyle: Readonly<Partial<CSSStyleDeclaration>>
+): void {
+    if (element.style.marginLeft != '') {
```
JiuqingSong, line 80: do we want to handle RTL?

JiuqingSong (approved):

packages/roosterjs-content-model/lib/editor/plugins/PastePlugin/ContentModelPastePlugin.ts
BryanValverdeU: My idea is to use this instead of having the functionality in the old plugin. Same for other Paste sources

JiuqingSong: So we can't enable this plugin until it can cover all functionalities of old plugin?

If that is the plan, I think it is fine.

packages/roosterjs-content-model/lib/editor/plugins/PastePlugin/WordDesktop/handleWordDesktopPaste.ts
BryanValverdeU: After checking, the default direction parser already handles that, but it seems that it is not displaying the list item correctly:
image

I think we may be missing text align for the list item

packages/roosterjs-content-model/lib/editor/plugins/PastePlugin/utils/linkParser.ts
BryanValverdeU: Added an issue for this requirement

PR #1721, opened 2023-04-14
Strict-mode on picker plugin 

juliaroldi:


JiuqingSong:
Given Andres already has a pr for PickerPlugin (https://github.com/microsoft/roosterjs/pull/1830), I think we can close this PR. @juliaroldi

PR #1735, opened 2023-04-18
[1/3] Enable Strict mode in TableCellSelection Utils + Plugin

BryanValverdeU:
Enable Strict mode in Table Cell Selection Utils and the Plugin class


packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/utils/restoreSelection.ts
```diff
@@ -8,8 +8,12 @@ import { updateSelection } from './updateSelection';
  * @internal
  */
 export function restoreSelection(state: TableCellSelectionState, editor: IEditor) {
+    if (!state.lastTarget || !state.firstTarget) {
+        return;
+    }
+
     if (state.firstTable) {
-        editor.select(state.firstTable, null);
+        editor.select(state.firstTable, undefined);
```
JiuqingSong, line 16: is the second parameter necessary here?

JiuqingSong (approved):

PR #1740, opened 2023-04-19
Content Model: Fix #1738 delete space issue

### LINKED ISSUE #1738: ContentModeEditPlugin causes space to be deleted
1. Enable ContentModelEditPlugin
2. Type "a b"
3. Press Backspace

Expect: show "a "
Actual: show "a"

This is because the last space is actual a \u0020, which is not shown for HTML. We need to convert it into `&nbsp;`

JiuqingSong:
This change addresses 3 issues:
1. For content such as "a b", and backspace after "b", we should not delete the space. This is done in deleteSelections.ts by changing the space after "a" to be `&nbsp;` when delete. Similar for DELETE key case
2. For content such as "a[SPACE][SPACE]b", when put cursor right before "b", browser will leave the second space after cursor so Content Model need to recognize this space and ignore it since it isn't rendering anything. This is done by creating a new model API `normalizeParagraph` and call it from `normalizeContentModel`. There are bunch of cases of spaces need to be handled there, see comments in code for more info
3. For `&nbsp;` at then end of a segment (not the last text segment), we need to replace it to be a normal space if the second last char is not space so that if there is another long segment after it (e.g. a hyperlink), browser can wrap before the link.

Related test cases added. And in this change I also enabled `ContentModelEditPlugin` in demo site by default.

### DIFF
--- a/demo/scripts/controls/sidePane/editorOptions/EditorOptionsPlugin.ts
+++ b/demo/scripts/controls/sidePane/editorOptions/EditorOptionsPlugin.ts
@@ -22,7 +22,7 @@ const initialState: BuildInPluginState = {
         contextMenu: true,
         autoFormat: true,
         contentModelFormat: true,
-        contentModelEdit: false,
+        contentModelEdit: true,
     },
     contentEditFeatures: getDefaultContentEditFeatureSettings(),
     defaultFormat: {},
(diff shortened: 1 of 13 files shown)


BryanValverdeU (approved):

JiuqingSong:
> Issue seem to repro with this snapshot. But only because the first chars are wrapped in a element. Not sure if it is in scope of this PR.
> 
> ```
> <div><span style="font-family: &quot;Segoe UI&quot;, &quot;Segoe UI Web (West European)&quot;, &quot;Segoe UI&quot;, -apple-system, BlinkMacSystemFont, Roboto, &quot;Helvetica Neue&quot;, sans-serif; font-size: 10.5pt;"><b>a&nbsp; </b>b</span></div>
> ```

Good catch. I have made a fix

PR #1742, opened 2023-04-20
Don't return focus to editor when blur on image

ianeli1:
Having the image be reselected on editor blur is causing focus to immediately return to editor in some scenarios, causing disturbance to surfaces which require focus to be shown. Fix by not selecting the image on `blur`.


ianeli1:
Is there any scenario that actually requires this behavior? Should we check if `blur`'s `relatedTarget` is contained in editor first?

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -394,7 +394,7 @@ export default class ImageEdit implements EditorPlugin {
      * quit editing mode when editor lose focus
      */
     private onBlur = () => {
-        this.setEditingImage(null, true);
+        this.setEditingImage(null, false);
```
JiuqingSong, line 397: Please add inline comment to "false"

JiuqingSong (approved):

PR #1748, opened 2023-04-25
Fix Excel Desktop pasting

Andres-CT98:
### Description:

When pasting an Excel Desktop, it now pastes properly as a table. Previously, tables from Excel Desktop would paste as plain text if they contained a link element.

### Fix

Use the full html from clip board data to do the Excel paste, instead of the first element of the fragment.

### Cases tested:

Paste any table copied from Excel Desktop that has a link in Rooster.

**Before:** Pasting a table would turn it into plain text.

**After:** Pasting a table results in its proper table format being kept.

### Warnings and implications

- Reverts the behaviour of pasting excel tables as images and plain text.


JiuqingSong:
Should we close this PR?

PR #1749, opened 2023-04-25
Revert "Paste as image feature" - convertPastedContentFromExcel

Andres-CT98:
Reverts microsoft/roosterjs#1710 - convertPastedContentFromExcel.ts

Changes made for Pasting as image were causing issues in pasting as table. Tables from Excel desktop were pasting as plain text.
Only convertPastedContentFromExcel.ts was reverted as reverting the rest will cause interface issues.

### Warnings and implications

- Reverts the behaviour of pasting excel tables as images and plain text.


JiuqingSong (approved):

romanisa:
@Andres-CT98 can you please add to the PR description, why we are reverting this PR by noting the bug that it caused.

romanisa (changes requested):
Reverting might not be a good idea, wait on that!

JiuqingSong (approved):

Issue #1682, opened 2023-03-31
Content Model: Table fidelity


JiuqingSong:
Complex table may lose format when recreate with content model.

An alternative way is to detect if a table is created using roosterjs, then for other tables, cache its TABLE/TR/TD elements and use General elements for all its children.

PR #1762, opened 2023-04-28
Content Model Fidelity improvement 7: Table

JiuqingSong:
Table fidelity is the hardest part. Instead of parse so many styles of TABLE/TR/TD, we want to reuse exist TABLE/TD/TH elements in most cases, unless they are changed.

To achieve this goal, we need to split the supported table related style into several parts:
1. Format that always need to rewrite, such as margin, because they can be impacted by outer container's align attribute
2. Format that only need to be reapply when something is changed. In that case, we delete the cached TD/TABLE element to indicate this element is dirty so we know we should recreate it. In the future we keep improving this by adding something like "isDirty" and no need to remove the cached element.
3. Other attributes that need special handling. This includes cell width/height. We need to reapply width/height to a cell when  there is no cached element (cell element is dirty) or this cell is always ready to apply size (use borderbox, and has metadata --- that means this is a table created from roosterjs), so it can help  handle table edit/resize case.
4. BlockFormat on table cell (direction, align, whiteSpace, lineHeight) will be handled in normal way, and its value will impact context.blockFormat. So that if there is textAlign from context and TD reset htmlAlign, textAlign will also be removed.


JiuqingSong:
#1682

BryanValverdeU (approved):

Issue #1698, opened 2023-04-03
Remove code for old dark mode


JiuqingSong:
We will always have DarkModeHandler for editor, so we can remove the code for the case DarkModeHandler not exist.

We can also make DarkModeHandler standalone from editor, expose some callback to let editor handle it to add/remove color var on content div.

PR #1767, opened 2023-05-01
Graduate VariableBasedDarkColor

JiuqingSong:
Graduate experimenatal feature VariableBasedDarkColor. Now `darkColorHandler` is required from `EditorCore` and always be created. When use it we don't need to do null check any more. Old dark color code that related to data-ogsc and data-ogsb are removed.

In Content Model `EditorContext` we still keep it as optional for now since Content Model code itself should be independent from editor code, so putting editor content DIV into Content Model EditorContext is not suitable. But Content Model code should also be self-contained, so we will need to do another refactoring to `DarkColorHandlerImpl` to move Content DIV out.

For other color API such as setColor and getColor, we also keep darkColorHandler as optional for backward compatibility only. But the code to handle color without this handler is removed.


JiuqingSong:
#1698

BryanValverdeU (approved):

PR #1771, opened 2023-05-02
Ensure list type increases correctly 

juliaroldi:
When commit a list chain, verify if the next element index is the same number of the start number of the next list. If verification returns true, increases the start number of the next list. Also use the start attribute to set numbering list style properly. 

testListIncresed

testSetNumbering


JiuqingSong:
I'm curious when was this regressed?

juliaroldi:
> I'm curious when was this regressed?

This code is being there for at least 6 month according to git history. Looking the unit tests, it was made to cause this behavior.

juliaroldi:
I merged both codes, now both scenarios are fixed. 
listIncrease

package.json
```diff
@@ -77,5 +77,8 @@
         "hooks": {
             "pre-commit": "pretty-quick --staged"
         }
+    },
+    "dependencies": {
+        "ori": "^0.1.9"
```
JiuqingSong, line 82: is this change intentional?

packages/roosterjs-editor-dom/lib/list/VListChain.ts
```diff
@@ -110,7 +110,13 @@ export default class VListChain {
 
         for (let i = 0; i < lists.length; i++) {
             const list = lists[i];
-            list.start = list.start > 1 ? list.start : lastNumber + 1;
+
+            //If there is a list chain sequence, ensure the list chain keep increasing correctly
+            if (list.start > 1) {
+                list.start = list.start === lastNumber ? lastNumber + 1 : list.start;
+            } else {
+                list.start = lastNumber + 1;
```
JiuqingSong, line 118: I tried with the latest code but the problem still exists

package.json
juliaroldi: No it is not. I'll remove it

JiuqingSong (approved):

PR #1772, opened 2023-05-03
Make deleteSelections function return the deleted content

BryanValverdeU:
In order to support copy, cut and paste with content model, I am updating the `deleteSelections` function to return the deleted blocks or segments.
Using that data we can update the tempDiv using contentModelToDom function. To perform the copy or cut operation. Which will be done in a future PR

### DIFF
--- a/packages/roosterjs-content-model/lib/modelApi/selection/deleteSelections.ts
+++ b/packages/roosterjs-content-model/lib/modelApi/selection/deleteSelections.ts
@@ -8,7 +8,7 @@ import { ContentModelEntity } from '../../publicTypes/entity/ContentModelEntity'
 import { ContentModelParagraph } from '../../publicTypes/block/ContentModelParagraph';
 import { ContentModelSegment } from '../../publicTypes/segment/ContentModelSegment';
 import { ContentModelSelectionMarker } from '../../publicTypes/segment/ContentModelSelectionMarker';
-import { ContentModelTable } from '../../publicTypes';
+import { ContentModelTable } from '../../publicTypes/block/ContentModelTable';
 import { createBr } from '../creators/createBr';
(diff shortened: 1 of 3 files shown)


packages/roosterjs-content-model/lib/modelApi/selection/deleteSelections.ts
```diff
@@ -111,6 +118,8 @@ const deleteSelectionStep1: DeleteSelectionStep = (context, options, model) => {
             let markerFormat = model.format;
             let insertMarkerIndex = 0;
 
+            saveDeletedElements(block, context, tableContext, deleteResult);
```
JiuqingSong, line 121: It is a good idea to return deleted content when delete. But I think this process can be integrated into the deleting process, but not let it be standalone here. You can move the deleted content into the target model when delete it, like L151 and L160

packages/roosterjs-content-model/lib/modelApi/selection/deleteSelections.ts
```diff
@@ -256,6 +266,33 @@ const DeleteSelectionSteps: DeleteSelectionStep[] = [
...
 ];
 
+function saveDeletedElements(
+    block: ContentModelBlock | undefined,
+    context: DeleteSelectionContext,
+    tableContext: TableSelectionContext | undefined,
+    deleteResult: ContentModelDocument
+) {
+    if (block) {
+        const newBlock = cloneBlock(block);
```
JiuqingSong, line 276: I feel using cloneBlock is too much overhead. It will clone the whole block including all segments even they are not selected.

We can just move the deleted content into target model. For blocks, just move and push into target model as a block. For segment, may need to create a new Paragraph, with the same format of source paragraph.

BryanValverdeU:
I will close this for now, will re open later

PR #1777, opened 2023-05-04
Add PasteType enum and check the paste type in PastePlugin

juliaroldi:
When create a paste fragment from clipboard data, add the paste type to the before paste event object, so the plugins can know if the fragment is being paste as plain text, as image, merging the format or just normal paste. 
pasteTableAsImage


packages/roosterjs-editor-dom/lib/clipboard/createFragmentFromClipboardData.ts
```diff
@@ -106,6 +109,10 @@ export default function createFragmentFromClipboardData(
         img.style.maxWidth = '100%';
         img.src = imageDataUri;
         fragment.appendChild(img);
+        // If the fragment is pasted as image, the excel attributes are not needed anymore, so remove it to avoid treat the fragment as a excel table.
+        if (pasteAsImage) {
```
JiuqingSong, line 113: This is so hacky. `createFragmentFromClipboardData` is a common function so it should not know the detail of paste source. We need another way to do this.

JiuqingSong: For example, in the code convert Excel table, we check if there is only an image in the fragment.

JiuqingSong: Or we can add a property in BeforePasteEvent to specify the paste option, so that plugins can read it and only change paste content when it is a "normal paste"

juliaroldi: is it safe to only change content when is "normal paste"? For now we are changing for all existing types of paste, maybe change the content only for normal paste can cause a regression.

JiuqingSong: Maybe not only "normal paste", but also "paste and merge format".

First we need to define what paste actions are supported. For now we support these 4:
- Normal paste  (default option)
- Paste as plain text: Then we will only get plain text and do not fix up any HTML
- Paste and merge format: Similar with normal paste, but change the format
- Paste image: Only image is pasted, all other HTML is removed, so plugin does not need to fix anything


In general, plugin knows what kind of fix it need to do, for exmaple, if we paste from Word, plugin that handles Word paste can know what scenario it should fix. If it also want to fix image paste, yes it can. So from a common code we just provide the basic information and let plugin decide what to do.

Normal paste/paste as plain text/paste image/ ... these are the different options selected by user, so plugins should also know this info and do proper handling.

juliaroldi: I made a new change that I added a enum with the paste types then createPasteFragment function provide the paste type data to the event data. Then I only added a check if the paste type is asImage, do not convert from excel. Do you think that I should add a check that only do any conversion when is Normal and MergeFormat paste types?

JiuqingSong: I think we just need to check Normal and MergeFormat

packages/roosterjs-editor-types/lib/enum/PasteType.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Enum for paste options
+ */
+export const enum PasteType {
+    /**
+     * Simple paste with not content change
```
JiuqingSong, line 6: Actually this is not accurate. Content might be changed. We just need to say this is t he default paste behavior and auto select the pasting format.

packages/roosterjs-editor-types/lib/enum/PasteType.ts
```diff
@@ -0,0 +1,24 @@
...
+     */
+    AsPlainText,
+
+    /**
+     * Apply the current style to pasted content
+     */
+    MergeFormat,
+
+    /**
+     * Paste the content as image element
```
JiuqingSong, line 21: What if there is not image? We need to explain this in this comment

packages/roosterjs-editor-types/lib/event/BeforePasteEvent.ts
```diff
@@ -37,6 +38,11 @@ export interface BeforePasteEventData {
      * Attributes of the root "HTML" tag
      */
     htmlAttributes: Record<string, string>;
+
+    /**
+     * Paste type option (as plain text, merge format, normal, as image)
+     */
+    pasteType?: PasteType;
```
JiuqingSong, line 45: Also need to allow `CompatiblePasteType`:

pasteType: PasteType | CompatiblePasteType;

Make sure to "import type" only for the Compatible type
And make this as required so that we know we can always read this value

JiuqingSong: Also make it as readonly, so plugin cannot change it. Because this value comes from user's decision, plugins should never change it.

packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
```diff
@@ -72,7 +78,9 @@ export default class Paste implements EditorPlugin {
                 case KnownPasteSourceType.ExcelDesktop:
                 case KnownPasteSourceType.ExcelOnline:
                     // Handle HTML copied from Excel
-                    convertPastedContentFromExcel(event, trustedHTMLHandler);
+                    if (event.pasteType !== PasteType.AsImage) {
```
JiuqingSong, line 81: As I said, check Normal and MergeFormat here

packages/roosterjs-editor-dom/lib/clipboard/createFragmentFromClipboardData.ts
```diff
@@ -106,13 +108,16 @@ export default function createFragmentFromClipboardData(
         img.style.maxWidth = '100%';
         img.src = imageDataUri;
         fragment.appendChild(img);
+        event.pasteType = pasteAsImage ? PasteType.AsImage : PasteType.Normal;
```
JiuqingSong, line 111: This value should be set even earlier, when create this event object.

packages/roosterjs-editor-core/test/coreApi/createPasteFragmentTest.ts
```diff
@@ -397,6 +397,7 @@ describe('createPasteFragment', () => {
                     metaName1: 'metaContent1',
                     metaName2: 'metaContent2',
                 },
+                pasteType: 0,
```
JiuqingSong, line 400: Let's use the enum value instead.

packages/roosterjs-editor-types/lib/event/BeforePasteEvent.ts
```diff
@@ -37,6 +39,11 @@ export interface BeforePasteEventData {
      * Attributes of the root "HTML" tag
      */
     htmlAttributes: Record<string, string>;
+
+    /**
+     * Paste type option (as plain text, merge format, normal, as image)
+     */
+    readonly pasteType?: PasteType | CompatiblePasteType;
```
JiuqingSong, line 46: Please remove "?" to make it not optional

packages/roosterjs-editor-plugins/lib/plugins/Paste/Paste.ts
```diff
@@ -60,7 +66,11 @@ export default class Paste implements EditorPlugin {
      * @param event PluginEvent object
      */
     onPluginEvent(event: PluginEvent) {
-        if (this.editor && event.eventType == PluginEventType.BeforePaste) {
+        if (
+            this.editor &&
+            event.eventType == PluginEventType.BeforePaste &&
+            (event.pasteType === PasteType.Normal || event.pasteType === PasteType.MergeFormat)
```
JiuqingSong, line 72: Suggest we only do this check for Excel handler. For others, we can change them if we see any problem

packages/roosterjs-editor-core/lib/coreApi/createPasteFragment.ts
```diff
@@ -60,5 +67,18 @@ function createBeforePasteEvent(core: EditorCore, clipboardData: ClipboardData):
         htmlBefore: '',
         htmlAfter: '',
         htmlAttributes: {},
+        pasteType: pasteType,
     };
 }
+
+function getPasteType(pasteAsText: boolean, applyCurrentStyle: boolean, pasteAsImage: boolean) {
+    if (!pasteAsImage && !applyCurrentStyle && !pasteAsText) {
```
JiuqingSong, line 75: Let's reorder this check, how about

if (pasteAsText) { return AsPlainText } 
else if (applyCurrentStyle) { ... }
else if (pasteAsImage) { ... }
else { return Normal}

JiuqingSong (approved):

packages/roosterjs-editor-dom/lib/clipboard/createFragmentFromClipboardData.ts
```diff
@@ -6,6 +6,7 @@ import HtmlSanitizer from '../htmlSanitizer/HtmlSanitizer';
 import moveChildNodes from '../utils/moveChildNodes';
 import toArray from '../jsUtils/toArray';
 import wrap from '../utils/wrap';
+
```
BryanValverdeU, line 9: nit: remove empty space

PR #1782, opened 2023-05-06
Add support for Wac Components Paste

BryanValverdeU:
This PR adds support for Wac components paste using Content Model to process the paste fragment.

To test:

Disable default Paste Plugin
Enable the Content Model Paste Plugin
Enable Experimental Feature Content Model paste
Paste content from Word Online


packages/roosterjs-content-model/lib/editor/coreApi/createContentModel.ts
```diff
@@ -32,7 +32,7 @@ function internalCreateContentModel(
         ...(option || {}),
     };
 
-    if (core.reuseModel) {
+    if (core.reuseModel && option?.allowCacheElement !== false) {
```
JiuqingSong, line 35: So that means default value is true?

In that case, suggest to add option as "disableCacheElement" which is more intuitive.

BryanValverdeU: default value is undefined. But in this scenario I set the value of the option to false. That way if the value is undefined or true we keep caching the elements.

JiuqingSong: Understand. I mean, by default its value is undefined, but undefined has the same meaning with "true" here, which is not intuitive. In most cases people will think undefined == false. That is why I suggest to change the option name to be opposite one

packages/roosterjs-content-model/lib/domToModel/context/createDomToModelContext.ts
```diff
@@ -59,8 +59,8 @@ export function createDomToModelContext(
         defaultFormatParsers: defaultFormatParsers,
     };
 
-    if (options?.allowCacheElement) {
-        context.allowCacheElement = true;
+    if (typeof options?.disableCacheElement == 'boolean') {
```
JiuqingSong, line 62: nit: 
```ts
if (options?.disableCacheElement) {
  context.disableCacheElement = true;
}
```

packages/roosterjs-content-model/lib/domToModel/context/createDomToModelContext.ts
```diff
@@ -59,8 +59,8 @@ export function createDomToModelContext(
...
+        context.disableCacheElement = options?.disableCacheElement;
```
JiuqingSong, line 63: Actually, I don't think we need to change context. Just change the options is ok.

BryanValverdeU: Removed the changes to the Context interface.

packages/roosterjs-content-model/lib/editor/coreApi/createContentModel.ts
```diff
@@ -32,8 +32,8 @@ function internalCreateContentModel(
         ...(option || {}),
     };
 
-    if (core.reuseModel) {
-        context.allowCacheElement = true;
+    if (core.reuseModel && !option?.disableCacheElement) {
```
JiuqingSong, line 35: We already copied this value in line 32, so here we just need to set disableCacheElement to true if core.reuseModel is false.

```ts
if (!core.reuseModel) {
  context.disableCacheElement = true;
}
```

In general, cache should only be used when both of the following to be true:
1. core.reuseModel
2. !option.disableCacheElement

Sof if any of them is not true, we should disable cache.

JiuqingSong (approved):

PR #1791, opened 2023-05-11
Undo with entity

JiuqingSong:
## Requirement

Entity is a concept introduced by roosterjs from earlier days. An entity is a special DOM node that created and owned by a plugin, roosterjs will not touch it. Instead, when an entity is removed by user or by other code, roosterjs will trigger event to let plugin know the entity is gone.

However, this was broken by Undo/Redo scenario. Today when user perform Undo/Redo, we just flush everything inside editor using a HTML string. So the existing entity DOM element will be removed and a new one will be created from the HTML. In most cases, when undo/redo, the entity should not be changed and we should persist its DOM element so that if a plugin caches the entity DOM node, we can make the cached DOM node still working. And if there is some additional work to initialize an entity, Undo/Redo should not trigger this initialization again.

Another requirement is, entity should has its own state for Undo/Redo. That means, when user did some change inside an entity, the related plugin should be able to trigger a new undo snapshot so that when user want to undo, we can undo the change from entity and keep other parts not changed.

In this change I'll solve the two requirements above.

## Persist entity DOM node

To persist the DOM node for an entity, you can handle `EntityOperationEvent` from your plugin, check the entity type and operation == `EntityOperation`.`NewEntity`, then set event.`shouldPersist` to be true. So that roosterjs knows that this entity will be persisted during undo/redo:

```ts
class MyEntityPlugin implements EditorPlugin {
       if (event.eventType == PluginEventType.EntityOperation && event.entity.type == 'MyEntityType' && event.operation == EntityOperation.NewEntity) {
           // ... Some code to hydrate your entity
          entity.shouldPersist = true;  // Make this entity persistable, it will not be replaced during undo/redo
      }
  }
}
```

## Add undo snapshot for entity

The API `addUndoSnapshot` now supports a new attribute in its `additionalData` parameter: `getEntityState`. It is a callback to allow you return id, type and a state string for your entity. This entity state will be saved together with undo snapshot. When user undo/redo to this snapshot, your plugin will receive a `EntityOperationEvent` with operation == `EntityOperation`.`UpdateEntityState`, and the state string will be passed through event.state:

```ts
class MyPluginEvent implements EditorPlugin {
   private onEntityChange() {
     // Add undo snapshot for entity
     editor.addUndoSnapshot(
       () => {
           // ... do you change to entity, or you can also pass null to this callback
       },
       'ChangeEntity', // or any string to specify the change source
       false, // Whether this change can be undo using Backspace key
      {
         getEntityState: () => { return 'MyChangeState1' }  // Return a string that you can recognize how to restore this state. You can also do JSON serialize to store a JSON object here
      }
   };
 }

 public onPluginEvent(event: PluginEvent) {
   // When restore the state
   if (event.eventType == PluginEventType.EntityOperation && event.type == 'MyEntityType' && event.operation == EntityOperation.UpdateEntityState) {
     this.updateMyEntity(event.state); // Do whatever change you need to update your entity according to the state string
   }
 }
}
```

## Demo

This change also includes a demo plugin SampleEntityPlugin to show how these new features work. To try it, pull the branch and start demo site using `npm start`, then press Ctrl+M inside the editor to insert a new entity. The entity will allow you to press a button to increase a counter. This count number can be undone/redone. And during undo/redo the entity DOM node will be persisted.

image


juliaroldi (approved):

Rain-Zheng:
Hi @JiuqingSong, the changes related `setContent` in this commit made a regression in Outlook mobile side. If the initial content contains some pre-defined styles in `<head>` part, these styles will be lost, because the `setContent` api only inserts the `<body>` part into the editor. 

image

PR #1800, opened 2023-05-15
Content Model: insertEntity API

JiuqingSong:
Implement Content Model version of `insertEntity` API.

Compare to the original API, Content Model insertEntity API has the following difference:
1. Add an `options` parameter for all those optional things
2. Move `contentNode` parameter into `options`, and make it optional. When not passed, we will just insert an empty entity, then plugin can change its content later
3. For `position` parameter, we merged `End` and `DomEnd` parameter into one parameter `end` since in Content Model there is not difference between End and DomEnd
4. Remove `isReadonly` parameter, we now only support readonly entity
5. Move `focusAfterEntity` into `options`
6. Add `wrapperDisplay` parameter in `options`, caller can directly set display style for wrapper via this parameter
7. Add `skipUndoSnapshot` parameter in `options`. By default `insertEntity` will add undo snapshot now and trigger `ContentChangedEvent`. Pass true to this parameter to skip this behavior
8. For block entity, we now check if we need to add an empty paragraph after it, and only add empty paragraph if necessary.

Related test cases added.


hipanwar:
Can we please add support to avoid adding a new line after the entity? Without it, clients have to resort to hacking up code to remove the new line.

JiuqingSong:
> Can we please add support to avoid adding a new line after the entity? Without it, clients have to resort to hacking up code to remove the new line.

New line will only be added when necessary now. That means, if there is no place to put focus after the entity, we will add an empty line.

PR #1808, opened 2023-05-19
Fix split cell shading for Content Model

Andres-CT98:
### Description:

When splitting a table cell that has a background colour different that the default, the new cell will now inherit the colour too. Previously, when using content model split the new cell reverted to the default colour.

### Fix

Added to the dataset of any new cells the property `editingInfo: '{"bgColorOverride":true}'` to make it maintain the inherited shading.

### Cases tested:

* Create a table in Rooster
* Shade all table as a new colour
* Select one or all cells and use Split from the content model ribbon

**Before:** After the split, all new cells would not inherit the shading from its parent cell.

**After:** After the split, all new cells will inherit the shading from its parent cell.

### Warnings and implications

- After the split, the new cell will still not inherit other format properties from its parent, such as font size.

### DIFF
--- a/packages/roosterjs-content-model/lib/modelApi/table/splitTableCellHorizontally.ts
+++ b/packages/roosterjs-content-model/lib/modelApi/table/splitTableCellHorizontally.ts
@@ -39,8 +39,7 @@ export function splitTableCellHorizontally(table: ContentModelTable) {
                             cell.format
                         );
 
-                        //Keep the background color of the original cell
-                        newCell.dataset = { editingInfo: '{"bgColorOverride":true}' };
+                        newCell.dataset = { ...cell.dataset };
 
                         if (rowIndex < sel.firstRow || rowIndex > sel.lastRow) {
                             newCell.spanLeft = true;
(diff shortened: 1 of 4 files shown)


packages/roosterjs-content-model/lib/modelApi/table/splitTableCellHorizontally.ts
```diff
@@ -39,6 +39,9 @@ export function splitTableCellHorizontally(table: ContentModelTable) {
                             cell.format
                         );
 
+                        //Keep the background color of the original cell
+                        newCell.dataset = { editingInfo: '{"bgColorOverride":true}' };
```
JiuqingSong, line 43: Good finding!

But instead of just copy "editingInfo" and "bgColorOverride", you can actually copy all the dataset so that if we add more metadata in the future, they will also be cared.

```ts
newCell.dataset = {...cell.dataset};
```

JiuqingSong:
Please also add test cases for these two files to cover this scenario. Thanks.

JiuqingSong:
Please also add test cases for these two files to cover this scenario. Thanks.

JiuqingSong (approved):

PR #1814, opened 2023-05-22
Content Model: Maintain selection on empty line

JiuqingSong:
Problem:
In roosterjs demo site, open side pane, switch to Content Model page. For an empty document, check the content model, the selection marker contains default format:
image

Then press the "Create DOM Tree" button from side pane, then check the format of selection marker, it becomes:
image

The format is gone.

This is because after create DOM tree, the selection is changed from "inside SPAN" to "inside DIV", however the format is under SPAN, so the focus is located out of a styled SPAN.

To solve this, we need to make sure after create DOM tree (rewrite DOM), focus is still under SPAN. With the fix, I create a dummy text node for the selection marker, and after optimize and normalize, the selection will be put under the SPAN, so it will still have the format.

Before rewrite:
```html
<div><span style="...">I<br></span></div>
```

Before fix:
```html
<div>I<span style="..."><br></span></div>
```

After fix:
```html
<div><span style="...">I<br></span></div>
```


JiuqingSong:
I have seen a problem that when append a node after an empty text node, it will be inserted before the text node.

```ts
const span = document.createElement('span');
const br = document.createElement('br');

span.appendChild(document.createTextNode('');
span.appendChild(br);

console.log(span.firstChild); // Expect to be txt node, but it is actually br
console.log(span.lastchild); // Expect to be Br node, but it is actually txt node
```

BryanValverdeU (approved):

BryanValverdeU (approved):

PR #1825, opened 2023-05-24
Clone entity when performing a copy

BryanValverdeU:
With the last change for copy and paste using content model, if we copy content that contains an entity, the entity is removed from the editor. This is because when we process the entity in the editor we store the Wrapper element inside of the Content Model. And we reuse the wrapper element when doing contentModelToDom into the tempDiv for copy.

To prevent this, clone the entity element. This is not always needed so added a new DomToContentModelOption to decide whether to clone the Entity element or not.

Before
beforeEntityCopy

After
afterEntityCopy


packages/roosterjs-content-model/lib/editor/corePlugins/ContentModelCopyPastePlugin.ts
```diff
@@ -83,72 +83,71 @@ export default class ContentModelCopyPastePlugin implements PluginWithState<Copy
...
-                const model = this.editor.createContentModel({
-                    disableCacheElement: true,
+        if (!this.editor) {
+            return;
+        }
+        const selection = this.editor.getSelectionRangeEx();
+        if (selection && !selection.areAllCollapsed) {
+            const model = this.editor.createContentModel({
+                disableCacheElement: true,
+                cloneEntityElement: true,
```
JiuqingSong, line 93: Question: Is it ok to always clone entity element if disableCacheEelemnt (line 92) is true? If that is ok, we don't need to introduce this option

BryanValverdeU: Since we only enable disableCacheElement & this new option for copy/paste, I think it make sense to do that.
Removing this new option

JiuqingSong (approved):

PR #1830, opened 2023-05-25
Strict mode for Picker Plugin

Andres-CT98:
Enable strict mode, and fix its issues, for the Picker Plugin


packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -288,7 +289,9 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
 
             return hasMatched;
         });
-        return createRange(startPos, endPos) || this.editor.getDocument().createRange();
+        return startPos && endPos
+            ? createRange(startPos, endPos)
```
JiuqingSong, line 293: This changes the logic.

createRange() can still return a valid range if endPos is null. But with this change when endPos is null, it will go to the other way.

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -306,29 +309,31 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
-                    trimmedWordBeforeCursor.split(' ').length <= 4)
-            ) {
-                this.dataProvider.queryStringUpdated(
-                    trimmedWordBeforeCursor,
-                    wordBeforeCursorWithoutTriggerChar == trimmedWordBeforeCursor
-                );
-                this.setLastKnownRange(this.editor.getSelectionRange());
-            } else {
-                this.setIsSuggesting(false);
+            if (wordBeforeCursor) {
```
JiuqingSong, line 312: This will change the logic if wordBeforeCursor is empty string

packages/roosterjs-editor-plugins/lib/plugins/Picker/PickerPlugin.ts
```diff
@@ -306,29 +309,31 @@ export default class PickerPlugin<T extends PickerDataProvider = PickerDataProvi
...
+                    wordBeforeCursor == this.pickerOptions.triggerCharacter ||
+                    (trimmedWordBeforeCursor &&
+                        trimmedWordBeforeCursor.length > 0 &&
+                        trimmedWordBeforeCursor.split(' ').length <= 4)
+                ) {
+                    this.dataProvider.queryStringUpdated(
+                        trimmedWordBeforeCursor,
+                        wordBeforeCursorWithoutTriggerChar == trimmedWordBeforeCursor
+                    );
+                    this.setLastKnownRange(this.editor?.getSelectionRange() ?? null);
```
JiuqingSong, line 333: Suggest do a null check for this.editor at the beginning of this function so that no need to check editor everywhere.

JiuqingSong (approved):

PR #1831, opened 2023-05-25
Support context menu in ImageSelection Plugin

juliaroldi:
Instead of select the image when right click, use Context Menu event to select the image when right click a image. 
Add copy/cut options to the image context menu to test image selection. 

**Before:**
contextMenuSelectionBlue

**After:**
contextMenuSelection

### DIFF
--- a/packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
@@ -1,4 +1,4 @@
-import { createRange, safeInstanceOf } from 'roosterjs-editor-dom';
+import { safeInstanceOf } from 'roosterjs-editor-dom';
 
 import {
     EditorPlugin,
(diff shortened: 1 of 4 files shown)


packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
```diff
@@ -92,6 +89,17 @@ export default class ImageSelection implements EditorPlugin {
                         }
                     }
                     break;
+                case PluginEventType.ContextMenu:
+                    const contextMenuTarget = event.rawEvent.target;
+                    const actualSelection = this.editor.getSelectionRangeEx();
+                    if (
+                        safeInstanceOf(contextMenuTarget, 'HTMLImageElement') &&
+                        (actualSelection.type !== SelectionRangeTypes.ImageSelection ||
+                            (actualSelection.type === SelectionRangeTypes.ImageSelection &&
```
JiuqingSong, line 98: You can remove the check to `actualSelection.type === SelectionRangeTypes.ImageSelection` because when it goes here, this must be true

PR #1838, opened 2023-05-26
Add Width Format Handler

BryanValverdeU:
When pasting external tables I found out that width is not preserved. Causing the table to look bad after pasting with the `Paste with content model` Experimental Feature.
Adding a new format handler to keep the width of the tables.

### DIFF
--- a/packages/roosterjs-content-model/lib/formatHandlers/common/widthFormatHandler.ts
+++ b/packages/roosterjs-content-model/lib/formatHandlers/common/widthFormatHandler.ts
@@ -1,0 +1,20 @@
+import { FormatHandler } from '../FormatHandler';
+import { WidthFormat } from '../../publicTypes/format/formatParts/WidthFormat';
+
+/**
+ * @internal
+ */
+export const widthFormatHandler: FormatHandler<WidthFormat> = {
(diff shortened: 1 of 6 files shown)


JiuqingSong:
We already have a sizeFormatHandler.

And for external table, ideally it should be treated as entity, can you double check why that does not work?https://github.com/microsoft/roosterjs/blob/769fef6f04c59969f76dd54c951f1ec059b2bb7b/packages/roosterjs-content-model/lib/domToModel/processors/tablePreProcessor.ts#L12

BryanValverdeU:
> We already have a sizeFormatHandler.
> 
> And for external table, ideally it should be treated as entity, can you double check why that does not work?
> 
> https://github.com/microsoft/roosterjs/blob/769fef6f04c59969f76dd54c951f1ec059b2bb7b/packages/roosterjs-content-model/lib/domToModel/processors/tablePreProcessor.ts#L12

Oh I see, using the size handler is also fixing the issue.

I think it is because the tablePreProcessor is only when getting the content from the editor itself. Since we parse the content from the paste fragment we still parse the content.

I guess we can also add the table preprocessor when creating the paste fragment

JiuqingSong (approved):

PR #1843, opened 2023-05-29
Fix #1322: Copying some table structure transform the table to single line text

### LINKED ISSUE #1322: Copying tables turns them into single line

BryanValverdeU:
The problem: #1322

Copying some tables are not adding the table tag, this seems to be a Browser issue. 

image

To fix this add an additional element to the temp div to prevent the table tags to not be added to the clipboard


packages/roosterjs-content-model/lib/editor/corePlugins/ContentModelCopyPastePlugin.ts
```diff
@@ -228,7 +228,16 @@ function selectionExToRange(
     let newRange: Range | null = null;
     if (selection.type === SelectionRangeTypes.TableSelection && selection.coordinates) {
         const table = tempDiv.querySelector(`#${selection.table.id}`) as HTMLTableElement;
-        newRange = createRange(table);
+        const elementAfter = table.parentElement?.insertBefore(
```
JiuqingSong, line 231: I think this should be done when we write content model back to DOM, no matter if it is table selection. For example, when handleTable (may need a wrapper handler for table for cut/copy), we directly write a SPAN around TABLE element

JiuqingSong: Then we convert the table selection directly to a regular selection, then here it can select the whole SPAN. (or maybe a DIV will be better)

BryanValverdeU: Sounds good.
Updated the PR to change the handleTable behavior to add a div around the table.

packages/roosterjs-content-model/lib/modelToDom/handlers/handleTable.ts
```diff
@@ -29,10 +29,12 @@ export const handleTable: ContentModelBlockHandler<ContentModelTable> = (
 
         moveChildNodes(tableNode);
     } else {
+        const tableContainer = doc.createElement('div');
```
JiuqingSong, line 32: Add comment to explain why we need this DIV.

Question: Do we want to apply this DIV to all tables, or only when cut/copy?

BryanValverdeU: We can keep it only for cut and paste.

packages/roosterjs-content-model/lib/editor/corePlugins/ContentModelCopyPastePlugin.ts
```diff
@@ -119,6 +123,11 @@ export default class ContentModelCopyPastePlugin implements PluginWithState<Copy
                 {
                     isDarkMode: false /* To force light mode on paste */,
                     darkColorHandler: this.editor.getDarkColorHandler(),
+                },
+                {
+                    modelHandlerOverride: {
+                        table: copyCutTableHandler,
```
JiuqingSong, line 129: You can leverage the onNodeCreated parameter from ModelToDomOption. It will be called when a node is created, then you can check if it is a table then wrap it.

Using a customized handler here is fine, but using cachedElement is not a good idea here.

JiuqingSong: There is an exmaple about how to use this parameter in packages\roosterjs-content-model\lib\publicApi\link\insertLink.ts

JiuqingSong (approved):

PR #1844, opened 2023-05-30
Make mergeModel, formatWithContentModel and onDeleteEntityCallback Public

BryanValverdeU:
Make the following APIs public mergeModel, formatWithContentModel and onDeleteEntityCallback.

I did the following to make them public:
- mergeModel: move it from ModelAP folder to public api folder and make it a default export.
- formatWithContentModel: Make it a default export.
- onDeleteEntityCallback: Make it a default export

Most of the file changes are due to updating the path of the files and importing the default export.


JiuqingSong:
As discussed offline, first I'd like to see if it is possible to just move paste function to be a core API and then no need to public these functions. If we still need to make them public, we can go ahead.

PR #1852, opened 2023-06-01
Make paste with content model a public api instead of a Editor Class member

BryanValverdeU:
This PR moves the functionality to paste with content model from Editor Class member to public api so it is easier to use.
Additionally remove createPasteModel coreApi as it will not be needed anymore.

Did a minor refactor on CreateFragmentFromClipboardData since some code of this function will also be used in the new paste public api.


packages/roosterjs-content-model/lib/publicTypes/ContentModelEditorCore.ts
```diff
@@ -78,6 +78,7 @@ export interface ContentModelCoreApiMap extends CoreApiMap {
     setContentModel: SetContentModel;
 
     /**
+     * @deprecated
```
JiuqingSong, line 81: Content Model interface is not stable now. So it is ok to just remove this function.

packages/roosterjs-content-model/lib/publicTypes/ContentModelEditorCore.ts
```diff
@@ -76,18 +76,6 @@ export interface ContentModelCoreApiMap extends CoreApiMap {
...
-     * Create a content model to use when pasting.
-     * @param core The ContentModelEditorCore object
-     * @param clipboardData Clipboard data retrieved from clipboard
-     * @param position The position to paste to
-     * @param pasteAsText True to force use plain text as the content to paste, false to choose HTML or Image if any
-     * @param applyCurrentStyle True if apply format of current selection to the pasted content,
-     * @param pasteAsImage Whether to force paste as image
-     * false to keep original format
-     */
-    createPasteModel: CreatePasteModel;
```
JiuqingSong, line 90: Remove the type declaration of CreatePasteModel from line 45 as well if it is not used any more

JiuqingSong (approved):

PR #1857, opened 2023-06-02
Fix bug when pasting from Outlook Win 32 to Rooster

BryanValverdeU:
When pasting to Rooster content of list items copied from Outlook Win32 the content is incorrect:

Pasted content:
image

Before 
image
**Note that the third item uses 1. instead of i., and the fourth item has incorrect spacing. If I press the Enter key after the fourth item, the third item corrects itself.

After
image


packages/roosterjs-content-model/lib/editor/plugins/PastePlugin/WordDesktop/handleWordDesktopPaste.ts
```diff
@@ -22,6 +22,11 @@ export function handleWordDesktop(ev: ContentModelBeforePasteEvent) {
     setProcessor(ev.domToModelOption, 'element', wordDesktopElementProcessor);
     addParser(ev.domToModelOption, 'block', removeNonValidLineHeight);
     addParser(ev.domToModelOption, 'listLevel', listLevelParser);
+    addParser(ev.domToModelOption, 'listItemElement', (format, element) => {
+        if (element.style.marginLeft) {
```
JiuqingSong, line 26: what about marginRight if it is RTL?

packages/roosterjs-content-model/lib/editor/plugins/PastePlugin/WordDesktop/handleWordDesktopPaste.ts
```diff
@@ -22,6 +22,11 @@ export function handleWordDesktop(ev: ContentModelBeforePasteEvent) {
     setProcessor(ev.domToModelOption, 'element', wordDesktopElementProcessor);
     addParser(ev.domToModelOption, 'block', removeNonValidLineHeight);
     addParser(ev.domToModelOption, 'listLevel', listLevelParser);
+    addParser(ev.domToModelOption, 'listItemElement', (format, element) => {
```
JiuqingSong, line 25: The callback does not use any other thing than the parameter, so it does not need to be a inline callback. Let's make it a regular function

packages/roosterjs-content-model/lib/editor/plugins/PastePlugin/WordDesktop/processWordLists.ts
```diff
@@ -12,6 +13,11 @@ const MSO_LIST = 'mso-list';
 const MSO_LIST_IGNORE = 'ignore';
 const LOOKUP_DEPTH = 5;
 
+interface WordDesktopListFormat extends DomToModelListFormat {
+    wordLevel: number | '';
+    wordKnownLevels: Map<string, ContentModelListItemLevelFormat[]>;
```
JiuqingSong, line 18: Let's make these 2 properties optional, so that in other code compiler can make sure we do null check

JiuqingSong (approved):

PR #1860, opened 2023-06-05
[Feat] Support excluding elements from being transformed

haven2world:
We have some features which we want to set the style under dark mode by ourselves. So we need to add this method to make it possiblie.

Before we achieved it by the API `onExternalContentTransform` which has been deprecated and not used in code.


haven2world:
Hi @JiuqingSong do you know how to revert the change of EOL?

JiuqingSong:
Would you please help me understand what kind of customization you want to do when do color transform? I'd like to put all dark color related logic into DarkColorHandler. So your requirement can help me decide how to design it.

haven2world:
Sure. For example we have a feature to beautify the link, and we will keep the same style of the link when the dark mode changes. And for another feature, we just prefer to set the style by ourselves when dark mode changes, rather than calculating it automatically.

haven2world:
> Hi @JiuqingSong do you know how to revert the change of EOL?

Fixed it by remaking the change on windows

haven2world:
@JiuqingSong We cannot bump roosterJS version until this change is released. Any concern to merge this?

JiuqingSong:
Instead of adding a new option, we'd better still reuse the `onExternalContentTransform` option. We should avoid add more and more options when there is already another one can do the same job.

haven2world:
Sure, if possible. But currently `onExternalContentTransform` is deprecated and in the code, it's not called anymore. And on the other hand, we also want to leverage the internal implementation of dark mode transform, rather than we implement it again in `onExternalContentTransform` but the internal implementation doesn't provide a way to exclude some elements which I think is helpful and necessary.

haven2world:
That's great! It's good to know we're going to have a new dark mode manager.
Sure. I'll close this PR. Thanks for your help!

PR #1861, opened 2023-06-05
Add null check to avoid exceptions

juliaroldi:
When write back a list into the editor, the placeholder of the parentNode of the root list is used to hold the list position. 
A null check was added to the parent node to avoid exceptions.


BryanValverdeU (approved):
Interesting, do you have a callstack when this is happening?

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -192,7 +192,7 @@ export default class VList {
         let lastList: Node;
 
         // Use a placeholder to hold the position since the root list may be moved into document fragment later
-        this.rootList.parentNode!.replaceChild(placeholder, this.rootList);
+        this.rootList.parentNode?.replaceChild(placeholder, this.rootList);
```
JiuqingSong, line 195: This means if parentNode is null, replaceChild won't happen.

Do you know in what case this can happen? We'd better understand why it can happen but not just do null check, because there might be underlying issue causes the issue so here the null issue is just a symptom of the real issue. If that is the case, checking null here will hide the real issue.

PR #1866, opened 2023-06-06
Merge Format when pasting.

BryanValverdeU:
When pasting and also using the Paste Option to merge the styles we need to apply the current format to the pasted content.
Following Word's behavior, if the pasted content contains text that is Bold/italic/Underline that format is going to be preserved.

image

Source: https://support.microsoft.com/en-us/office/control-the-formatting-when-you-paste-text-20156a41-520e-48a6-8680-fb9ce15bf3d6?ui=en-us&rs=en-us&ad=us

With this change we add a new option to mergeModel so we can set the format to the pasted model.

Also, modified the current ContentModelEditor so we can use the 'Merge Format' Option from RoosterUI

MergeFormatOnPaste1

### DIFF
--- a/packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
+++ b/packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
@@ -1,5 +1,3 @@
-import paste from '../publicApi/utils/paste';
-import { ClipboardData, ExperimentalFeatures } from 'roosterjs-editor-types';
 import { ContentModelDocument } from '../publicTypes/group/ContentModelDocument';
 import { ContentModelEditorCore } from '../publicTypes/ContentModelEditorCore';
 import { ContentModelSegmentFormat } from '../publicTypes/format/ContentModelSegmentFormat';
--- a/packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
+++ b/packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
@@ -1,10 +1,10 @@
-import { ClipboardData, ExperimentalFeatures } from 'roosterjs-editor-types/lib';
+import paste from '../publicApi/utils/paste';
+import { ClipboardData, ExperimentalFeatures } from 'roosterjs-editor-types';
 import { ContentModelDocument } from '../publicTypes/group/ContentModelDocument';
 import { ContentModelEditorCore } from '../publicTypes/ContentModelEditorCore';
 import { ContentModelSegmentFormat } from '../publicTypes/format/ContentModelSegmentFormat';
 import { createContentModelEditorCore } from './createContentModelEditorCore';
 import { EditorBase } from 'roosterjs-editor-core';
-import { paste } from '../publicApi';
 import {
     ContentModelEditorOptions,
     DomToModelOption,
@@ -72,6 +72,13 @@ export default class ContentModelEditor
         return core.defaultFormat;
     }
 
+    /**
+     * Paste into editor using a clipboardData object
+     * @param clipboardData Clipboard data retrieved from clipboard
+     * @param pasteAsText Force pasting as plain text. Default value is false
+     * @param applyCurrentStyle True if apply format of current selection to the pasted content,
+     * false to keep original format.  Default value is false. When pasteAsText is true, this parameter is ignored
+     */
     public paste(
         clipboardData: ClipboardData,
         pasteAsText: boolean = false,
--- a/packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
+++ b/packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
@@ -1,8 +1,10 @@
+import { ClipboardData, ExperimentalFeatures } from 'roosterjs-editor-types/lib';
 import { ContentModelDocument } from '../publicTypes/group/ContentModelDocument';
 import { ContentModelEditorCore } from '../publicTypes/ContentModelEditorCore';
 import { ContentModelSegmentFormat } from '../publicTypes/format/ContentModelSegmentFormat';
 import { createContentModelEditorCore } from './createContentModelEditorCore';
 import { EditorBase } from 'roosterjs-editor-core';
+import { paste } from '../publicApi';
 import {
     ContentModelEditorOptions,
     DomToModelOption,
@@ -68,5 +70,18 @@ export default class ContentModelEditor
         const core = this.getCore();
 
         return core.defaultFormat;
+    }
+
+    public paste(
+        clipboardData: ClipboardData,
+        pasteAsText: boolean = false,
+        applyCurrentFormat: boolean = false,
+        pasteAsImage: boolean = false
+    ) {
+        if (this.isFeatureEnabled(ExperimentalFeatures.ContentModelPaste)) {
+            paste(this, clipboardData, pasteAsText, applyCurrentFormat, pasteAsImage);
+        } else {
+            super.paste(clipboardData, pasteAsText, applyCurrentFormat, pasteAsImage);
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
```diff
@@ -69,4 +71,24 @@ export default class ContentModelEditor
...
     }
+
+    /**
+     * Paste into editor using a clipboardData object
+     * @param clipboardData Clipboard data retrieved from clipboard
+     * @param pasteAsText Force pasting as plain text. Default value is false
+     * @param applyCurrentStyle True if apply format of current selection to the pasted content,
+     * false to keep original format.  Default value is false. When pasteAsText is true, this parameter is ignored
+     */
+    public paste(
```
JiuqingSong, line 82: Is this necessary?

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
```diff
@@ -42,6 +43,13 @@ export interface MergeModelOption {
      * @default false
      */
     mergeCurrentFormat?: boolean;
+
+    /**
+     * When set to true, segment format of the insert position will be set into the content that is merged into current model.
+     * If the source model already has fontWeight, Italic or underline, it will not be overwritten.
+     * @default false
+     */
+    applyCurrentFormat?: boolean;
```
JiuqingSong, line 52: This optionis conflict with `mergeCurrentFormat`, and the two names together make me confused.

Can we merge them into one option? You can make its value to be a string literal type instead of boolean.

packages/roosterjs-content-model/lib/editor/ContentModelEditor.ts
BryanValverdeU: No, we can remove it. It is only for testing in demo site. Since PasteOption Plugin from RoosterReact calls this.editor.paste. Will Remove it.

JiuqingSong: For long term, we should have a separate demo site for content model.

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
```diff
@@ -37,11 +38,14 @@ export interface MergeModelOption {
...
+     * Use this to decide whether to change the source model format when doing the merge.
+     * 'MergeCurrentFormat': segment format of the insert position will be merged into the content that is merged into current model.
      * If the source model already has some format, it will not be overwritten.
-     * @default false
+     * 'ApplyCurrentFormat': format of the insert position will be set into the content that is merged into current model.
+     * If the source model already has fontWeight, Italic or underline, it will not be overwritten.
+     * @default undefined
      */
-    mergeCurrentFormat?: boolean;
+    mergeFormatOption?: 'MergeCurrentFormat' | 'ApplyCurrentFormat';
```
JiuqingSong, line 48: Better add a 'none' option to not merge. Although we can leave it undefined for "none", but in case the scenario that we already have an option object then we want to override it by assigning another option object to it and we want it not merge, then we can merge with "none" value.

JiuqingSong: Also, for other cases of string literal values, we use strings start with lower cases. Although there is not a rule but I think we'd better make them consistent.

JiuqingSong: also the names make me a little confusing.

In my understand, "merge" means both source and target format will be used, when there are conflict, we may have some rule to decide which one to use, here we may want to keep bold, italic, ...

And "apply" means only apply one side format to the other, if there are conflicts, always overwrite. 

So that is in a totally opposite way with your comment. Do you think we can switch them, or use some better words? I'm open for discussion.

JiuqingSong: Also the property name `mergeFormatOption`, it is already inside an `MergeModelOption` interface, everything here is an option, so no need to add the word "option" in its name again.

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
```diff
@@ -57,13 +61,13 @@ export function mergeModel(
         options?.insertPosition ?? deleteSelection(target, onDeleteEntity).insertPoint;
 
     if (insertPosition) {
-        if (options?.mergeCurrentFormat) {
+        if (options?.mergeFormatOption) {
```
JiuqingSong, line 64: If we added "none" as an option, need to check it is not none here

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
```diff
@@ -259,30 +263,56 @@ function insertBlock(markerPosition: InsertPoint, block: ContentModelBlock) {
...
                 });
                 break;
         }
     });
 }
+
+function getSemanticFormat(segment: ContentModelSegment): ContentModelSegmentFormat {
+    const result: ContentModelSegmentFormat = {};
+
+    if (segment.format.fontWeight) {
```
JiuqingSong, line 307: What if fontWeight is "normal", what is the expected result in that case? Especially the current format has fontWeight to be bold?

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
```diff
@@ -259,30 +263,56 @@ function insertBlock(markerPosition: InsertPoint, block: ContentModelBlock) {
...
+        result.fontWeight = segment.format.fontWeight;
+    }
+    if (segment.format.italic) {
```
JiuqingSong, line 310: Similar here, what if italic is explicitly set to false?

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
BryanValverdeU: I agree with you on all. 

What about something like this:

```
    /**
     * Use this to decide whether to change the source model format when doing the merge.
     * 'targetAndSource': segment format of the insert position will be merged into the content that is merged into current model.
     * If the source model already has some format, it will not be overwritten.
     * 'currentFormatAndSourceEmphasis': format of the insert position will be set into the content that is merged into current model.
     * If the source model already has emphasis format, such as, fontWeight, Italic or underline different than the default style, it will not be overwritten.
     * 'none' No format merging will be applied to the source model.
     * @default undefined
     */
    mergeFormat: 'targetAndSource' | 'targetFormatAndSourceEmphasis' | 'none';
```

This way we would understand easier what type of merging we would do to the content.

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
BryanValverdeU: Good point, I checked Word and if the fontWeight is normal or initial, we should not use the font weight from the source.
Will add this check here.

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
BryanValverdeU: Checking what Word does, if the source is not italic. We should not use Italic. 

Also if the target is italic and the source is not italic, upon merging we would set the pasted content to italic.

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
JiuqingSong: How about 'mergeAll' | 'keepSourceEmphasisFormat' | 'none'?

BryanValverdeU: Sounds good

JiuqingSong (approved):

PR #1868, opened 2023-06-07
Content Model: Ignore empty paragraph when turn on list

JiuqingSong:
When turn on list, ignore those empty paragraphs.

Before format:
image

After format, before change:
image

After format, after change:
image


JiuqingSong:
2 should be fixed by https://github.com/microsoft/roosterjs/pull/1872

BryanValverdeU (approved):

PR #1869, opened 2023-06-07
Fix table inserter for merged vertical cells

Andres-CT98:
### Description:

If some cells are merged vertically, the row inserter will not be accessible for any intersecting row except the first one. The issue is not present for horizontally merged cells and the column inserter.

Example table:
<div><table ><tbody><tr><td><br/></td><td><br/></td><td rowspan="3"><br/></td></tr><tr><td><br/></td><td><br/></td></tr><tr><td><br/></td><td><br/></td></tr></tbody></table>

### Fix

A new condition was added to check if the cursor's position is more than the left margin of the cell, plus other minor optimisations. This was done because the previous two conditions (less than right and less than bottom) were being met by any vertically merged cell in the wrong row, so the inserter would not appear for the intersecting rows of that cell, except the first one.

### Cases tested:

Using the above table as example, Insert a new row after the second one using the table inserter.

**Before:** The inserter is not showing, so the only way to insert the row is though a menu.

**After:** The row inserter is accessible

### Warnings and implications

- It could be considered to separate the behaviour of the Table Inserter and Table Resizer.
- The Table Resizer is not usable for left edges of a vertically merged cell, except the first one.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
@@ -13,10 +13,10 @@ import {
 } from 'roosterjs-editor-types';
 
 const INSERTER_HOVER_OFFSET = 6;
-const TOP_OR_SIDE = {
-    top: 0,
-    side: 1,
-};
+const enum TOP_OR_SIDE {
+    top = 0,
+    side = 1,
+}
 /**
  * @internal
  *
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
```diff
@@ -12,8 +12,11 @@ import {
     CreateElementData,
 } from 'roosterjs-editor-types';
 
-const INSERTER_HOVER_OFFSET = 5;
-
+const INSERTER_HOVER_OFFSET = 6;
+const TOP_OR_SIDE = {
```
JiuqingSong, line 16: let's use const enum to save bytes

JiuqingSong (approved):

PR #1882, opened 2023-06-13
Remove block background color on Merge Format after Paste

BryanValverdeU:
When pasting and merging format, we are going to remove the background color of block elements with background color to follow Word's behavior, this will be done by adding an additional block parser when Content Model from the clipboard fragment.

Additionally a small fix to also merge the format of list item format holder and paragraphs. When using mergeFormat. And also add background color parser to list level to keep the color of the list when pasting.
demoMerge


packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
```diff
@@ -289,23 +295,34 @@ function applyDefaultFormat(
...
+                    );
                 });
                 break;
         }
     });
 }
 
-function getSemanticFormat(segment: ContentModelSegment): ContentModelSegmentFormat {
+function mergeSegmentFormat(
+    applyDefaultFormatOption: 'mergeAll' | 'keepSourceEmphasisFormat',
```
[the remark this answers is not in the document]
BryanValverdeU: Hmm good point, I think for now we can keep it like this, just for consistency with other APIs options. For example, in [iterateSelections](https://github.com/microsoft/roosterjs/blob/2c44dc5406799bd6647ccd69a443bc614766d55e/packages/roosterjs-content-model/lib/modelApi/selection/iterateSelections.ts#LL20C1-L20C91)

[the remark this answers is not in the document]
JiuqingSong: Normally, if a type is used in multiple places, we should add a type. If it is only used by one public function, it is ok not to add a type.

Adding a string literal type is ok, but do not add enum. I'm trying to avoid using enum in Content Model because
- Normal enum will cause extra bundle increase
- Const enum has some problem when used by isolated typescript project

So string literal type is the best here.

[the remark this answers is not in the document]
juliaroldi: That is good to know!!

juliaroldi (approved):

packages/roosterjs-content-model/lib/publicApi/utils/paste.ts
```diff
@@ -153,3 +163,17 @@ function createFragmentFromClipboardData(
...
+
+/**
+ * For block elements that have background color style, remove the background color when user selects the merge current format
+ * paste option
+ */
+const blockElementParser: FormatParser<ContentModelBlockFormat> = (
+    format: ContentModelBlockFormat,
+    element: HTMLElement
+) => {
+    if (element.style.backgroundColor) {
```
JiuqingSong, line 175: This can fix the bug when paste text with black background. But what if the text has white color? In that case if we delete the dark block background, the text will become invisible. We may need to either remove text color under it, or do a dark to light conversion

BryanValverdeU: Hmm yes, I tested Word Desktop and it seems that they still show the text in white making it look invisible. For example:

Copy source
image

Word Content: 
image

BryanValverdeU: Let me double check with Jorge

BryanValverdeU: Checked with Jorge and this should be fine as long as we used the Merge Format option in the PasteOption Plugin. Without merge format we should still should the background color of the Div and the list. 
So I had to do additional changes to add background color to lists and fix an issue when merging format of paragraphs in mergeModel function

BryanValverdeU: Sorry, I meant that I added that change in this PR.

JiuqingSong (approved):

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
```diff
@@ -1,11 +1,11 @@
 import { addSegment } from './addSegment';
 import { applyTableFormat } from '../table/applyTableFormat';
 import { ContentModelBlock } from '../../publicTypes/block/ContentModelBlock';
+import { ContentModelBlockFormat } from 'roosterjs-content-model/lib/publicTypes';
```
JiuqingSong, line 4: please fix this import

PR #1892, opened 2023-06-16
Add TableLayout Format Handler

BryanValverdeU:
When pasting a table with table layout we are not parsing the style causing the table to look different.
Adding table layout to parse and apply this style.


JiuqingSong:
Please also add the new type to ContentModelTableFormat https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs-content-model/lib/publicTypes/format/ContentModelTableFormat.ts

JiuqingSong (approved):

PR #1896, opened 2023-06-19
Add size format handler to table cells

BryanValverdeU:
Last week we added table layout parser, but we also need to add size parser to table cells otherwise the table will not be rendered correctly. Eg:

Before:
image

After:
image


JiuqingSong:
Does this have impact to table resize?

e.g. paste a table that has width/height in TABLE element, then resize it, can it work well?

BryanValverdeU:
> Does this have impact to table resize?
> 
> e.g. paste a table that has width/height in TABLE element, then resize it, can it work well?

Checked again and it is working fine with Table Resize:
Resize2

JiuqingSong (approved):

PR #1899, opened 2023-06-19
Remove width and height after quit editing

juliaroldi:
Only set image width and height to the image in editing mode. After editing the image do set width and height style properties. This properties were removed, because it was causing an issue in Outlook after the image has been sent. 

**Bug:** 
Screenshot 2023-06-16 at 14 37 18

**Expected:** 
Screenshot 2023-06-16 at 14 37 00

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
@@ -405,6 +405,8 @@ export default class ImageEdit implements EditorPlugin {
             this.clonedImage = this.image.cloneNode(true) as HTMLImageElement;
             this.clonedImage.removeAttribute('id');
             this.clonedImage.style.removeProperty('max-width');
+            this.clonedImage.style.width = this.editInfo.widthPx + 'px';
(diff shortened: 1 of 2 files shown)


JiuqingSong (approved):

haven2world:
Hi @julia, this change made a regression on outlook mobile side. When we edit an image with a preset "width" style, which has a higher priority than element's "width" property, if we only apply the change on element's "width" property, then the image's size won't change.

haven2world:
I read the PR description. I just wonder whether an image like `` can still be resized by percentage after this change. Because after upgrading roosterJS, one test case like this failed on our end.
The cause is after this change, the new width only apply on the `width` attribute, for example, the user want to change its width to `200px`, then the element is changed to ``. But the width style still has a higher priority than attribute, so the image's width will still be `300px`

juliaroldi:
I can reproduce the issue you are reporting now. We should not have width/height style attributes in the final image. If the image has this style attributes, we should remove it and only let width/height attributes.

juliaroldi:
Hey @haven2world  we have a fix for the issue you reported https://github.com/microsoft/roosterjs/pull/1961. Do you need a patch fix for it or is it ok to wait for the next release?

PR #1906, opened 2023-06-21
Improve merge when source is Paragraph with Decorator and source is not. 

BryanValverdeU:
Merge decorator format too when mergin paragraphs with decorator. To make the functionality similar to Word Paste:

Copy Source: H1 and a paragraph:
image

Word:
WordExample

Before this change
BeforeMH

After this change:
AfterMH


packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
```diff
@@ -124,6 +124,10 @@ function mergeParagraph(
     if (segmentIndex >= 0) {
         newParagraph.segments.splice(segmentIndex, 0, ...newPara.segments);
     }
+
+    if (newPara.decorator) {
```
JiuqingSong, line 128: What if we paste a paragraph with decorator into an existing paragarph without decorator?

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
```diff
@@ -293,17 +297,20 @@ function applyDefaultFormat(
...
                     }
 
-                    segment.format = mergeSegmentFormat(
-                        applyDefaultFormatOption,
-                        format,
-                        segment.format
-                    );
+                    segment.format = mergeSegmentFormat(applyDefaultFormatOption, format, {
+                        ...segment.format,
+                        ...paragraphFormat,
```
JiuqingSong, line 308: should we put paragraph format before segment.format?

packages/roosterjs-content-model/lib/modelApi/common/mergeModel.ts
BryanValverdeU: Checking what Word does:

If we use 'paste as is' and the target contains a decorator it is going to be overwritten by the source decorator. Eg: Current Content is H1, if we paste content with H2, the result decorator is going to be H2.

But if we do 'merge format' we should only keep the emphasis format. So in the above example the result decorator will continue being H1.

JiuqingSong (approved):

PR #1907, opened 2023-06-21
Fix onExternalContentTransform, allow customize dark color handling

JiuqingSong:
Per Outlook Mobile team's requirement, we need to add back the support to onExternalContentTransform, and inside this function  we allow calling darkColorHandler to do color transformation using roosterjs's build in algorithm.

To do this, I moved the core logic of color transformation from core API transformColor() to DarkColorHandler, and let transformColor call this new function from DarkColorHandler. Then I added some new parameters to onExternalContentTransform so that external code can also call to DarkColorHandler to do color transformation.


packages/roosterjs-editor-types/lib/corePluginState/LifecyclePluginState.ts
```diff
@@ -31,7 +32,14 @@ export default interface LifecyclePluginState {
     /**
      * External content transform function to help do color transform for existing content
      */
-    onExternalContentTransform: ((htmlIn: HTMLElement) => void) | null;
+    onExternalContentTransform:
+        | ((
+              element: HTMLElement,
+              fromDarkMode: boolean,
```
BryanValverdeU, line 38: Should we make the new parameters optional to avoid breaking change? Or it is okay?

BryanValverdeU (approved):

packages/roosterjs-editor-types/lib/corePluginState/LifecyclePluginState.ts
JiuqingSong: Normally, adding new parameter should be optional. But here we don't expect this function to be called outside editor core API, but it is a callback that people can pass in. In that case even the passed-in callback doesn't have those new parameters it can still pass the build. So from this part it is not a breaking change.

PR #1911, opened 2023-06-22
Stop image animation while editing

juliaroldi:
When editing a gif image, convert the original image into a png file, then while editing the image the animation will be stopped. If the image were not cropped or rotated, the image animation must start again after quitting editing mode. 

stopGifWhileEditing


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/covertGifToPng.ts
```diff
@@ -0,0 +1,30 @@
+import generateDataURL from './generateDataURL';
+import ImageEditInfo from '../types/ImageEditInfo';
+
+/**
+ * If the image is a gif, use canvas to convert it to a png.
```
JiuqingSong, line 5: Add @internal

BryanValverdeU: There are other types of animated image extensions such as [WebP](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#webp_image), [AVIF](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#avif_image), [APNG](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#apng_animated_portable_network_graphics) in case we want to also support them.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/covertGifToPng.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @internal
+ * If the image is a gif, use canvas to convert it to a png.
+ * @param image to be converted
+ * @returns
+ */
+export function convertGifToPng(editInfo: ImageEditInfo) {
+    const { src, widthPx, heightPx, naturalHeight, naturalWidth } = editInfo;
+    if (src.indexOf('.gif') > -1 || src.indexOf('image/gif') > -1) {
```
BryanValverdeU, line 12: There are other types of animated image extensions such as [WebP](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#webp_image), [AVIF](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#avif_image), [APNG](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#apng_animated_portable_network_graphics) in case we want to also support them.

juliaroldi: I don't know if canvas can convert this file of types. According to this source, it is not supported https://community.canvaslms.com/t5/Instructor-Guide/What-types-of-media-files-can-I-upload-in-Canvas-as-an/ta-p/1142

BryanValverdeU: Oh I see. sounds good.

PR #1912, opened 2023-06-23
Uncaught TypeError: Cannot read properties of undefined (reading 'cells')

BryanValverdeU:
Fix Uncaught TypeError: Cannot read properties of undefined (reading 'cells')

Callstack
webpack://.../roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts cells 117
webpack://.../roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts onMouseMove 125
webpack://.../roosterjs-editor-core/lib/coreApi/attachDomEvent.ts beforeDispatch 27

Do not create the TableEditor if the table do not have rows. to prevent the issue. Also add a null check on the line that cause the regression

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
@@ -130,7 +130,7 @@ export default class TableResize implements EditorPlugin {
             this.disposeTableEditor();
         }
 
-        if (!this.tableEditor && table && this.editor) {
+        if (!this.tableEditor && table && this.editor && table.rows.length > 0) {
             this.tableEditor = new TableEditor(
                 this.editor,
                 table,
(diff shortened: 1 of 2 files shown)


juliaroldi:
Do you know which scenario this files can be undefined?

BryanValverdeU:
> Do you know which scenario this files can be undefined?

A table with no rows it looks like. I am able to repro adding <table></table> and mouse over it. 
I dont know why would someone add a table like that though

juliaroldi (approved):

juliaroldi:
> > Do you know which scenario this files can be undefined?
> 
> A table with no rows it looks like. I am able to repro adding `<table></table>` and mouse over it. I dont know why would someone add a table like that though

Do you if some issue is just deleting the cells and not the table element?

BryanValverdeU:
> > > Do you know which scenario this files can be undefined?
> > 
> > 
> > A table with no rows it looks like. I am able to repro adding `<table></table>` and mouse over it. I dont know why would someone add a table like that though
> 
> Do you if some issue is just deleting the cells and not the table element?

I think when that happens we delete the table

JiuqingSong (approved):

PR #1915, opened 2023-06-23
Add an experimental feature to allow disabling list chain functionality

JiuqingSong:
Per Yammer team's requirement, we want to allow disabling list chain functionality. So add an experimental feature to support this.

When list chain is allowed (default):
allowlistchain

When list chain is disabled:
disablelistchain


juliaroldi (approved):

juliaroldi:
Is this experimental feature permanent? Should we try to pass a parameter to always disable list chain?

JiuqingSong:
> Is this experimental feature permanent? Should we try to pass a parameter to always disable list chain?

Because I don't want to add too much single feature check to IEditor, so reuse ExperimentalFeatures. This might e permanent, or in the future version we may introduce a better approach to customize editor.

PR #1918, opened 2023-06-27
Add Unit test for Paste source from M365 Apps 

BryanValverdeU:
Add unit test using clipboard data from M365 Apps.


packages/roosterjs-editor-plugins/test/paste/e2e/pasteFromWordTest.ts
```diff
@@ -0,0 +1,46 @@
...
+                '<p style="margin:0in 0in 8pt;font-size:11pt;font-family:Calibri, sans-serif">Test</p><!--{"start":[0,0,0],"end":[0,0,4]}-->',
+            htmlFirstLevelChildTags: ['P', 'P'],
+            html:
+                '<p style="margin:0in 0in 8pt;font-size:11pt;font-family:Calibri, sans-serif">Test</p><p style="margin:0in 0in 8pt;font-size:11pt;font-family:Calibri, sans-serif">asdsad</p>',
+        });
+
+        editor.paste(clipboardData);
+
+        expect(editor.getContent()).toEqual(
+            Browser.isFirefox
```
JiuqingSong, line 40: Why Firefox can generate different HTML than others?

BryanValverdeU: Not sure about it. I think that is how Firefox handles the HTML. It seems they add family name style to paragraph element. That is the only difference I see.

JiuqingSong: Can we also verify the content model inside? Then we may understand why they are different

BryanValverdeU: Added the content model tests to the new package.

BryanValverdeU: I am just going to check whether the function that process the content was called. 
We already have other unit tests to make sure the content is sanitized correctly. These unit tests will just make sure we do not break the logic to call the function.

JiuqingSong (approved):

PR #1919, opened 2023-06-27
Fix table selector and resizer

Andres-CT98:
### Description:

The Table selector and resizer are currently inserted on the document, thus putting them on a different container to the table they belong to. This creates issues in some scenarios such as when a dialog is created above the document with a high z-index. They also have a fixed position which causes issues in some scrolling scenarios, where they stayed fixed in place instead of moving or disappearing.

### Fix

Changed the insertion point of where the squares are located. The Selector is inserted before the table and the Resizer after it. Also changed its position from fixed to absolute and made changes to the CSS.

### Change

Fixed scenarios:

- Squares no longer stick on screen when scrolling.
- Squares are in the same context as the table, so they are never hidden under the editor or overlap outside of it.

**Before:**
image

**After:** 
image

### Warnings and implications

- The same scroll scenario where the squares remained on screen still has issues after this change. Both operations are not functional, but it is a rare scenario to encounter.
- The same issue exists form the table Inserters and cell resizers.


packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableSelector.ts
```diff
@@ -41,12 +47,13 @@ export default function createTableSelector(
     div.id = TABLE_SELECTOR_ID;
     div.style.width = `${TABLE_SELECTOR_LENGTH}px`;
     div.style.height = `${TABLE_SELECTOR_LENGTH}px`;
-    document.body.appendChild(div);
+    table.insertAdjacentElement('beforebegin', div);
```
JiuqingSong, line 50: So this is to insert the handle DIV into editor content, is that right?

I can think about 2 issues:
1. If an auto save happens while the resize handle is showing, it will be saved as part of the content
2. Please try when compose is zoomed (in OWA), if the handle position is still right.

JiuqingSong: Maybe it is ok to insert here. But we need to remove it when getContent() is called, by handling `ExtractContentWithDomEvent` event

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -85,6 +85,8 @@ export default class TableResize implements EditorPlugin {
             case PluginEventType.ContentChanged:
             case PluginEventType.Scroll:
             case PluginEventType.ZoomChanged:
+            case PluginEventType.ExtractContentWithDom:
```
JiuqingSong, line 88: This is not what I mean. You need to handle extractContentWithDom event separately, it will pass you a DocumentFragment, and you just need to remove the selector from that fragment, but no need to remove it from current editor. (The passed-in fragment is a cloned DOM tree of editor, so operating on that tree won't imapct editor)

That means, we just don't want the table selector to be saved, but from current editor, it should still show. User's resizing action should not be interrupted by auto save.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableResizer.ts
```diff
@@ -35,7 +35,7 @@ export default function createTableResizer(
 
     div.style.width = `${TABLE_RESIZER_LENGTH}px`;
     div.style.height = `${TABLE_RESIZER_LENGTH}px`;
-    document.body.appendChild(div);
+    table.insertAdjacentElement('afterend', div);
```
JiuqingSong, line 38: do we also need to do the same change for other table resize helper elements? Actually there are 6:
1. Whole table selector
2. Whole table resizer (these 2 are visible as a gray square)
3. horizontal resizer
4. vertical resizer
5. new column inserter (a circle with "+" sign when hover above a table)
6. new row inserter (a circle with "+" sign when hover on left side (or right side when RTL) of a table)

3,4 are invisible, but it will change mouse cursor when hover
5,6 will be visible when hover.

I can see your change fixed 1 and 2, so we also need same fix for 3-6. And maybe we should have a unified function for all of them to reduce duplicated code.

Andres-CT98: I noticed that but wanted to do that on a separate PR. I think a better approach would be to do something like Image Edit does, where the helpers are in in a span.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
Andres-CT98: How can I test auto save?

JiuqingSong: To simulate an auto save, you can add code to call window.setTimeout() with a delayed time, say 10sec. And in the callback you can call editor.getContent(), which will trigger this event.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableResizer.ts
JiuqingSong: It's ok to do it separately, and please open an issue to track it.

I think it should be fine to use a SPAN for all the resizer UI, although this seems like a larger change.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -85,9 +85,15 @@ export default class TableResize implements EditorPlugin {
             case PluginEventType.ContentChanged:
             case PluginEventType.Scroll:
             case PluginEventType.ZoomChanged:
+            case PluginEventType.SelectionChanged:
                 this.setTableEditor(null);
                 this.invalidateTableRects();
                 break;
+            case PluginEventType.ExtractContentWithDom:
+                e.clonedRoot.querySelectorAll('div.tableEditFeature').forEach(div => {
```
JiuqingSong, line 93: Make this class name a constant, and only declare once, then let all other places import it. So we only have one single definition.

And better use a more specified name, such as "_roosterjsTableEditHelper"

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableResizer.ts
```diff
@@ -12,7 +12,7 @@ const MIN_CELL_HEIGHT = 20;
  */
 export default function createTableResizer(
     table: HTMLTableElement,
-    zoomScale: number,
+    editor: IEditor,
```
JiuqingSong, line 15: What is the purpose to change this?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
Andres-CT98: Which 'all other places'?

JiuqingSong: For example, Line 27 of TableResizer.ts:

image

And Line 39 of TableSelector.ts:

image

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
```diff
@@ -24,6 +24,10 @@ const enum TOP_OR_SIDE {
     top = 0,
     side = 1,
 }
+
+// Class name for Table edit feature divs
+export const TEF_CLASS_NAME = '_roosterjsTableEditFeature';
```
JiuqingSong, line 29: Please add comment like below, to avoid creating js document for this internal export:
/**
 * @internal
 */

JiuqingSong (approved):

PR #1920, opened 2023-06-27
Make Select All Table Cells, change the color of table.

BryanValverdeU:
1. When selecting all the table, instead of adding the styles to all the cells, we can just add the styles to the whole table.
2. Reactivate the select table tests to see if with the changes they are no longer unstable
3. Modify how we handle the TableSelectionRange when the whole table is selected. In this case we should not add ranges for each row as we normally do, instead just select the whole table.
4. Add to the TableSelectionRange `isWholeTableSelected` property, so consumers do not need to create the vTable and invoke isWholeTableSelected everytime they need to check if whole table is selected.


packages/roosterjs-editor-core/lib/coreApi/selectTable.ts
```diff
@@ -119,26 +127,27 @@ function buildCss(
...
-                        tag,
-                        tdCount
-                    );
-                    const elementsSelector = selector + ' *';
-
-                    selectors.push(selector);
-                    selectors.push(elementsSelector);
-                    firstSelected = firstSelected || table.querySelector(selector);
-                    lastSelected = table.querySelector(selector);
+                    if (!isAllTableSelected) {
```
JiuqingSong, line 134: I think we can move this check out of for loop

JiuqingSong (approved):

Issue #1937, opened 2023-07-10
iOS/macOS app's TableFeatureSettings.upDownInTable is false by default


miku1958:
**Describe the bug**
TableFeatureSettings.upDownInTable only enable for Browser.isChrome or Browser.isSafari
https://github.com/microsoft/roosterjs/blob/55b2135615844d8348e75f3fed95e395ad9d45b5/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/tableFeatures.ts#L175

but the up down arrow behavior in iOS/macOS app is also wrong, should it be disable by !Browser.isWebKit?

```[tasklist]
### Tasks
- [x] Add a draft title or issue reference here
```

**Device Information**
 - OS: iOS/macOS
 - Browser: WKWebView

miku1958:
Hi @JiuqingSong, do you know why defaultDisabled is using (!Browser.isChrome && !Browser.isSafari) here?

JiuqingSong:
@miku1958 I think the only reason is Firefox can work well even without this feature. But it should be fine to also enable it to firefox.

miku1958:
@JiuqingSong Ok, let me submit a change to remove this line

miku1958:
Fixes #1941

PR #1941, opened 2023-07-12
Remove disable controls to other webkit-based iOS/macOS clients

miku1958:
`defaultDisabled: !Browser.isChrome && !Browser.isSafari` is added to exclude FireFox, but it also disable this feature for other webkit-based iOS/macOS clients


miku1958:
Fixes #1937

JiuqingSong (approved):

PR #1944, opened 2023-07-12
Make cell vertical alignment 'top' as default.

Andres-CT98:
### Description:

The vertical alignment of cells should start as top not middle.

### Fix

Add to both (non-content model and content model) the default value of vertical align as top to inserted table cells and change the tests to reflect it. This is done on InsertTable by adding it to the initial format. On the functions that modify the vertical alignment, a new metadata is added to override the default, as well as metadata for the whole table. Changes were made in other places to accommodate this new metadata value.

### Cases tested:

1. Insert a new table in both (non-content model and content model)
2. Resize any row
3. Observe the vertical alignment of the cells

**Before:** Alignment defaults to middle.

**After:** Alignment defaults to top.

### Warnings and implications

- The new vAlignOverride value can also be used for text alignment.
- Clear format leaves the Table metadata value for verticalAlignment as null, defaulting to browser behaviour.


juliaroldi:
Could we set the vertical-align to top when insert a table and do not save this info in the metadata. FormatTable is related to table format of the whole table, such as predefined styles. I think the change you is more related to cell style. If need to save this info in a metadata, I think you should save it in TableCellMetadata.

JiuqingSong:
> Could we set the vertical-align to top when insert a table and do not save this info in the metadata? FormatTable is related to table format of the whole table, such as predefined styles. I think the change you made is more related to cell style. If need to save this info in a metadata, I think you should save it in TableCellMetadata.

I feel it is needed to put this setting into metadata. Consider this two scenarios:
1. Insert a new table from roosterjs, we want the text to align to top by default. Then we add a new row, it should keep text to align to top in the new row
2. Now we paste a table from other place that has text align to middle (default value), then add a new row. Where should we align to text in the new row? I think we should keep it align to middle to be consistent with other rows.

Then how do we know which behavior we should obey? Metadata is the way we are using for other formats such as background color, and should be same for this alignment.

The only question is, if user explicitly changed alignment in one cell, the reformat the table. How can we know where it should align to? I think we need an extra metadata for that cell to let the format table code know it should keep its alignment.

JiuqingSong:
> > > Could we set the vertical-align to top when insert a table and do not save this info in the metadata? FormatTable is related to table format of the whole table, such as predefined styles. I think the change you made is more related to cell style. If need to save this info in a metadata, I think you should save it in TableCellMetadata.
> > 
> > 
> > I feel it is needed to put this setting into metadata. Consider this two scenarios:
> > 
> > 1. Insert a new table from roosterjs, we want the text to align to top by default. Then we add a new row, it should keep text to align to top in the new row
> > 2. Now we paste a table from other place that has text align to middle (default value), then add a new row. Where should we align to text in the new row? I think we should keep it align to middle to be consistent with other rows.
> > 
> > Then how do we know which behavior we should obey? Metadata is the way we are using for other formats such as background color, and should be same for this alignment.
> > The only question is, if user explicitly changed alignment in one cell, the reformat the table. How can we know where it should align to? I think we need an extra metadata for that cell to let the format table code know it should keep its alignment.
> 
> I think I got it, the vertical-align define all the table cells to top, so it make sense in TableFormatMetadata. I think if the user change cell alignment, we should add the info in TableCellMetadata, since it information related to a particular cell.

Yes, that is what I mean.

Andres-CT98:
> I feel it is needed to put this setting into metadata. Consider this two scenarios:

The two scenarios are covered by this change, and both behave as suggested. When new rows are inserted they inherit the alignment of their parent row/column, as it happens with shading too.

> Metadata is the way we are using for other formats such as background color, and should be same for this alignment.

I can only see metadata being used Only for background color, while things like borders are handled like alignment. It would be useful if we defined all parameters that need to be metadata.

packages/roosterjs-editor-api/lib/table/applyCellShading.ts
```diff
@@ -23,7 +28,11 @@ export default function applyCellShading(editor: IEditor, color: string | ModeIn
                         true /** shouldAdaptFontColor */,
                         editor.getDarkColorHandler()
                     );
-                    saveTableCellMetadata(region.rootNode, { bgColorOverride: true });
+                    const meta = getTableCellMetadata(region.rootNode);
+                    saveTableCellMetadata(region.rootNode, {
+                        ...meta,
+                        ...{ bgColorOverride: true },
```
JiuqingSong, line 34: Here you just need to do:
```ts
{
  ...meta,
  bgColorOverride: true
}
```

packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
```diff
@@ -43,15 +48,16 @@ export function applyTableFormat(
         };
 
         const bgColorOverrides = updateBgColorOverrides(rows, !keepCellShade);
+        const vAlignOverrides = updateVAlignOverrides(rows);
```
JiuqingSong, line 51: These two functions can be merged into one, and no need to parse metadata for each cell twice

JiuqingSong: Then let it directly return metaOverrides

packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
```diff
@@ -202,47 +226,54 @@ function formatBackgroundColors(
 
                 setTableCellBackgroundColor(cell, color);
             }
+
+            // Format Vertical Align
+            if (!metaOverrides.vAlignOverrides[rowIndex][colIndex]) {
+                cell.format.verticalAlign = 'top';
```
JiuqingSong, line 232: Instead of hard code "top" here, can we put it as part of the table metadata, then read this value from table metadata?

Otherwise, if we reformat an existing table which has valign to middle, all cells will suddenly align to top since they don't have cell override.

The align to top behavior should only affect new table, but for existing tables, the should keep no change.

Andres-CT98: How is Shading handled in that case? Because Shading would have the same problem in that case since it is not part of the table Metadata.

JiuqingSong: Shading is defined as part of the format metadata, and we read it from line 218:

image

JiuqingSong: I feel that if we put this setting inside the format metadata, we should only set its value when insert new table. For other cases that format an existing table, do not touch this value and keep whatever value it has. That means:
1. For new table, we apply a format that has all other existing values (borders, ...) as well as this vertical alignment
2. When apply format to an existing table, no matter if it already has format or vertical align, do not change its vertical align value, keep whatever it has
3. When apply the format, read this value from table, and read if it has override from cell. If it has value from table but not overridden from cell, apply this value.

Andres-CT98: I think the way shading is done is misnamed. If it is supposed to be the default shading it should say it somewhere, but the name implies it will only be used on even rows. Should I do a name change, or just add comments?

As for new tables, applyTableFormat has no way to differentiate if the table is new or existing. So it will override existing format as it currently does for shading in content model tables. For example, if a tables' 'bgColorEven' metadata is different from null, when applyTableFormat runs it will change it to null, unless it had an override.

JiuqingSong: The name is a little confusing. But for compatibility, let's just keep it. Feel free to add comment.

applyTableFormat doesn't need to know if a table is new. We just need to let insertTable API build a correct table format and pass into applyTableFormat.

For example, for a new table, insertTable() can add something like "verticalAlign: top" into the table format, then call applyTableFormat(). For existing table, just read the table format from the table, then combine with the new format. If the table format already has verticalAlign, it will be preserved. If not, we don't add it.

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -30,6 +31,7 @@ const DEFAULT_FORMAT: Required<TableFormat> = {
     headerRowColor: '#ABABAB',
     tableBorderFormat: TableBorderFormat.DEFAULT,
     keepCellShade: false,
+    verticalAlign: 'top',
```
JiuqingSong, line 34: We should not add this here. We should only apply this change when insert a new table. Here it is common code and will be used by all table scenarios.

packages/roosterjs-editor-dom/lib/metadata/metadata.ts
```diff
@@ -57,9 +57,22 @@ export function setMetadata<T>(element: HTMLElement, metadata: T, def?: Definiti
 }
 
 /**
- * Remove metadata from the given element if any
+ * Remove specific metadata from the given element if any
  * @param element The element to remove metadata from
+ * @param metadata The metadata tag to remove
  */
-export function removeMetadata(element: HTMLElement) {
+export function removeMetadata<T>(element: HTMLElement, metadata: string) {
```
JiuqingSong, line 64: This is a breaking change. For new added parameter, we need to make it optional, and when it is not passed, the function should work in the same way with the code before change.

Otherwise, other people who is using this function will be broken

packages/roosterjs-editor-dom/lib/metadata/metadata.ts
```diff
@@ -57,9 +57,22 @@ export function setMetadata<T>(element: HTMLElement, metadata: T, def?: Definiti
 }
 
 /**
- * Remove metadata from the given element if any
+ * Remove specific metadata from the given element if any
  * @param element The element to remove metadata from
+ * @param metadata The metadata tag to remove
```
JiuqingSong, line 62: I understand you'd like to allow delete a single item of metadata, so we'd better name it `metadataName` or `metadataKey`.

packages/roosterjs-editor-dom/lib/metadata/metadata.ts
```diff
@@ -57,9 +57,22 @@ export function setMetadata<T>(element: HTMLElement, metadata: T, def?: Definiti
...
+        delete currentMetadata[metadata];
+        element.dataset[MetadataDataSetName] = JSON.stringify(currentMetadata);
+    }
+}
+
+/**
+ * Remove all metadata from the given element if any
+ * @param element The element to remove metadata from
+ */
+export function removeAllMetadata(element: HTMLElement) {
```
JiuqingSong, line 76: Same with previous one, the functionality of this one can be merged into the previous function

packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
```diff
@@ -66,29 +70,32 @@ function clearCache(rows: ContentModelTableRow[]) {
...
             updateTableCellMetadata(cell, metadata => {
                 if (metadata && forceClear) {
-                    currentRow.push(false);
+                    bgColorOverrides.push(false);
                     delete metadata.bgColorOverride;
                 } else {
-                    currentRow.push(!!metadata?.bgColorOverride);
+                    bgColorOverrides.push(!!metadata?.bgColorOverride);
                 }
+                vAlignOverrides.push(!!metadata?.vAlignOverride);
```
JiuqingSong, line 91: Question: do we also need to delete metadata.verticalAlign when forceClear is true?

packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
```diff
@@ -178,71 +195,67 @@ function formatBorders(rows: ContentModelTableRow[], format: TableMetadataFormat
...
 
 function setFirstColumnFormat(
     rows: ContentModelTableRow[],
     format: Partial<TableMetadataFormat>,
-    bgColorOverrides: boolean[][]
+    metaOverrides: MetaOverrides
 ) {
-    rows.forEach((row, rowIndex) => {
-        row.cells.forEach((cell, cellIndex) => {
-            if (format.hasFirstColumn && cellIndex === 0) {
```
JiuqingSong, line 216: why remove this check?

Andres-CT98: It is not removed, but format.hasFirstColumn is constant so I moved it outside.

packages/roosterjs-editor-dom/lib/metadata/metadata.ts
Andres-CT98: Does that mean I can never change function names if they are exported?

packages/roosterjs-editor-dom/lib/table/VTable.ts
Andres-CT98: I mentioned the same about shading, but it also is in the default format.

What is the purpose of the default format constant if not to apply by default to all scenarios that do not explicitly have a format of their own? If default format should just be used for new inserted tables, I suggest we move it there in both models, however that change should be done separately.

packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
JiuqingSong: Ok, in that case we should rename this parameter to reflect the fact that it will only impact sharde

packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
JiuqingSong: But this changes the logic.

In original code, if format.hasFirstColumn is false, it will still execute line 228 (line 240 after change) since it is in "else" block, but after this change nothing will be executed.

packages/roosterjs-editor-dom/lib/metadata/metadata.ts
JiuqingSong: Changing a function name means we need to bump major version (v8 => v9). We don't want this happen just because we want to rename a function.

JiuqingSong: Sometimes we do want to rename a function, e.g. the original function name has a typo. In that case, will still keep the old one and mark it as deprecated. For example:

```ts
export function newFunc() { ... }

/**
 * @deprecated Use newFunc instead
 */
export const oldFunc = newFunc;
```

In that way, people who don't want to make change to their code can still upgrade version without breaking the build.

packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
Andres-CT98: You are right, I did not notice.

packages/roosterjs-editor-dom/lib/table/VTable.ts
JiuqingSong: I think it makes sense to move this format to somewhere we start to apply table format, so VTable will be fully decoupled from the formats.

But given the code here is just working, we'd better not change it, to avoid regression. If we want to refactor it, that can be a separate PR.

packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
```diff
@@ -23,6 +23,12 @@ const DEFAULT_FORMAT: Required<TableMetadataFormat> = {
     bgColorOdd: '#ABABAB20',
     headerRowColor: '#ABABAB',
     tableBorderFormat: TableBorderFormat.DEFAULT,
+    verticalAlign: 'top',
```
JiuqingSong, line 26: Remove this, specify it in insertTable()

packages/roosterjs-editor-dom/lib/table/VTable.ts
```diff
@@ -17,7 +18,7 @@ import {
 } from 'roosterjs-editor-types';
 import type { CompatibleTableOperation } from 'roosterjs-editor-types/lib/compatibleTypes';
 
-const DEFAULT_FORMAT: Required<TableFormat> = {
+const DEFAULT_FORMAT: Partial<TableFormat> = {
```
JiuqingSong, line 21: Sorry I misunderstood.

So here we should keep it as Required, and you may need to let the verticalAlign allow an empty value. e.g.
```ts
interface TableFormat {
 ...
 verticalAlign?: 'top' | 'bottom' | 'middle' | '';
}
```

Then put the empty string value here, which means it is not specified. Then when apply the value, check if the cell as vertical align override and if the format value is not empty, e.g.
```ts
if (!override[x][y] && format.verticalAlign != '' ) {
  ...
}
```

JiuqingSong: or use null as the empty value, should also ok

Andres-CT98: The implementation of alignment in non-content model is done differently, by adding it to the row on creation on insertTable.

JiuqingSong (approved):

JiuqingSong:
There is still one thing missing: When set table cell vertical alignment, need to set the override metadata to be true. We can do this in separate change.

PR #1952, opened 2023-07-14
If not merging Paragraphs, keep the source format

BryanValverdeU:
If we merge some content with Table and a paragraph. The block format of the table (All the borders,etc) is leaking to the paragraphs that are after the table.

So, if we are not merging To Current Paragraph, keep using the format of the source block.

See gifs attached here for the before and after:
https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/215985/


packages-content-model/roosterjs-content-model-editor/lib/modelApi/common/mergeModel.ts
```diff
@@ -132,6 +132,10 @@ function mergeParagraph(
     if (newPara.decorator) {
         newParagraph.decorator = { ...newPara.decorator };
     }
+
+    if (!mergeToCurrentParagraph) {
+        newParagraph.format = newPara.format;
```
JiuqingSong, line 137: Isn't this format already applied in line 228?

JiuqingSong: Oh I got it, so this is to remove existing format from source block

JiuqingSong (approved):

PR #1959, opened 2023-07-17
Remove SegmentFormat when merging Heading

BryanValverdeU:
When merging a Heading to a Paragraph with DefaultFormat (segmentFormat), the format of the decorator is overriden by the segment decorator on writeback.

To follow the same behavior as the HeadingProcessor, delete the segmentFormat if the source Paragraph is a heading.

https://github.com/microsoft/roosterjs/blob/b4604ae59f5ff56d560eac13b9b973c64107cb86/packages-content-model/roosterjs-content-model-dom/lib/domToModel/processors/headingProcessor.ts#L22-28


packages-content-model/roosterjs-content-model-editor/lib/modelApi/common/mergeModel.ts
```diff
@@ -131,6 +133,9 @@ function mergeParagraph(
 
     if (newPara.decorator) {
         newParagraph.decorator = { ...newPara.decorator };
+        if (HeadingTags.indexOf(newParagraph.decorator.tagName) > -1) {
+            delete newParagraph.segmentFormat;
```
JiuqingSong, line 137: do we really need to delet ethe whole segmentFormat? Or we just want to delete the conflict ones?

BryanValverdeU: Let me verify this. I think in theory it should be okay to remove all, at least with the testing I did. But if we can keep some properties would be better.

BryanValverdeU: Just tested and I think it should be okay to just remove the conflict formats. Added that to the latest commits

packages-content-model/roosterjs-content-model-editor/lib/modelApi/common/mergeModel.ts
```diff
@@ -131,6 +134,16 @@ function mergeParagraph(
 
     if (newPara.decorator) {
         newParagraph.decorator = { ...newPara.decorator };
+        if (HeadingTags.indexOf(newParagraph.decorator.tagName) > -1) {
+            const keys: (keyof ContentModelSegmentFormat)[] = getObjectKeys(
+                newParagraph.decorator.format
+            );
+            keys.forEach(key => {
+                if (newParagraph.segmentFormat?.[key]) {
```
JiuqingSong, line 142: when the value is false, or 0, this will also return false. Is that expected?

BryanValverdeU: Good catch. Fixed in latest commit

JiuqingSong (approved):

PR #1961, opened 2023-07-18
Remove style width/height

juliaroldi:
When an image is resized, the width/height width attributes are changed to set the new size. If the image already has width/height style properties, those properties must be removed, because they have high priority and it will prevent the image from resize.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
@@ -75,5 +75,8 @@ export default function applyChange(
     if (wasResizedOrCropped || state == ImageEditInfoState.FullyChanged) {
         image.width = targetWidth;
         image.height = targetHeight;
+        // Remove width/height style so that it won't affect the image size, since style width/height has higher priority
+        image.style.removeProperty('width');
+        image.style.removeProperty('height');
     }
 }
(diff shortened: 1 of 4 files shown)


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
```diff
@@ -75,5 +75,8 @@ export default function applyChange(
     if (wasResizedOrCropped || state == ImageEditInfoState.FullyChanged) {
         image.width = targetWidth;
         image.height = targetHeight;
+        // Remove width/height style so that it won't affect the image size, since style width/height has higher priority
+        image.style.removeProperty('width');
```
JiuqingSong, line 79: Hmm, should we keep the CSS style and HTML attribute have the same value

juliaroldi: We remove this properties due this issue https://github.com/microsoft/roosterjs/pull/1899

JiuqingSong: I see.

Do we also need to change Content Model for image? Because we will write back image size using CSS

packages-content-model\roosterjs-content-model-dom\lib\formatHandlers\common\sizeFormatHandler.ts

juliaroldi: It seems content model will write back the image after when quit editing, so the style width/height will be already removed.

JiuqingSong (approved):

PR #1965, opened 2023-07-20
Couple of fixes for Word Online Paste

BryanValverdeU:
For some reason, table container have negative margin in Word Online. Which causes the table to look truncated when pasted into rooster.
Remove that margin on paste when the source is Word Online

Before
image

After
image

Second bugfix is to keep the EOP Elements when pasting from Word Online since these elements can sometimes be present in new lines, and if we remove them can cause that the paste result will not have some new lines. (https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/218022/)


packages-content-model/roosterjs-content-model-editor/lib/editor/plugins/PastePlugin/WacComponents/processPastedContentWacComponents.ts
```diff
@@ -259,3 +263,13 @@ const wacListProcessor: ElementProcessor<HTMLOListElement | HTMLUListElement> =
         context.defaultElementProcessors.ul?.(group, element as HTMLUListElement, context);
     }
 };
+
+const wacBlockParser: FormatParser<ContentModelBlockFormat> = (
+    format: ContentModelBlockFormat,
+    element: HTMLElement
+) => {
+    if (element.classList.contains(TABLE_CONTAINER) && element.style.marginLeft.startsWith('-')) {
+        element.style.marginLeft = '0px';
```
JiuqingSong, line 272: I'm not sure if we should change this margin value here. Normally a parser should not modify source element. 

Does it still work if we don't change the margin?

BryanValverdeU: You are right, I understand why this was not working. I should have been using the container formatParser instead of the block format parser. 

Just removed this line

JiuqingSong (approved):

PR #1971, opened 2023-07-21
Dont run Test in FF

BryanValverdeU:
This test is throwing errors whe run in FF. due to FF handling the styles different. Skipping this test when browser is firefox


packages-content-model/roosterjs-content-model-editor/test/editor/plugins/paste/processPastedContentFromWacTest.ts
```diff
@@ -1593,6 +1593,10 @@ describe('wordOnlineHandler', () => {
     });
 
     it('Table with merged cells.', () => {
```
JiuqingSong, line 1595: You can use "itChromeOnly" instead

JiuqingSong:
This change is already included in https://github.com/microsoft/roosterjs/pull/1973
So close. @BryanValverdeU

PR #1972, opened 2023-07-21
Handle key down events in content model only if was not handled by EditFeature

juliaroldi:
To avoid conflicts between EditPlugin and ContentModelEditPlugin, the `handleByEditFeature` parameter was creates in PluginKeyDownEvent data. Therefore, if the event was already handle by a content edit feature, we set `handleByEditFeature` to `true`,  then `ContentModelEditPlugin` know that it the event should not be handled again.


packages/roosterjs-editor-types/lib/event/PluginDomEvent.ts
```diff
@@ -79,7 +89,8 @@ export type PluginMouseEvent = PluginMouseDownEvent | PluginMouseUpEvent | Plugi
  * This interface represents a PluginEvent wrapping native KeyDown event
  */
 export interface PluginKeyDownEvent
```
JiuqingSong, line 91: Please change CompatiblePluginKeyDownEvent as well

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -599,6 +600,9 @@ const MergeListOnBackspaceAfterList: BuildInEditFeature<PluginKeyboardEvent> = {
         return false;
     },
     handleEvent: (event, editor) => {
+        if (event.eventType == PluginEventType.KeyDown) {
+            event.handledByEditFeature = true;
```
JiuqingSong, line 604: Instead of setting this flag here, you can set it in EditPlugin, when an edit feature is found and used, set this flag so the flag is common for all features.

packages-content-model/roosterjs-content-model-editor/lib/editor/plugins/ContentModelEditPlugin.ts
```diff
@@ -91,7 +91,9 @@ export default class ContentModelEditPlugin implements EditorPlugin {
                     break;
 
                 case PluginEventType.KeyDown:
-                    this.handleKeyDownEvent(this.editor, event.rawEvent);
+                    if (!event.handledByEditFeature) {
```
JiuqingSong, line 94: Move this down to line 120 to merge with other checks

packages-content-model/roosterjs-content-model-editor/lib/editor/plugins/ContentModelEditPlugin.ts
```diff
@@ -91,7 +91,11 @@ export default class ContentModelEditPlugin implements EditorPlugin {
                     break;
 
                 case PluginEventType.KeyDown:
-                    this.handleKeyDownEvent(this.editor, event.rawEvent);
+                    this.handleKeyDownEvent(
+                        this.editor,
+                        event.rawEvent,
```
JiuqingSong, line 96: Let's pass the whole event object in, so no need to pass in two parameter separately.

packages-content-model/roosterjs-content-model-editor/lib/editor/plugins/ContentModelEditPlugin.ts
```diff
@@ -112,13 +116,17 @@ export default class ContentModelEditPlugin implements EditorPlugin {
...
 
-    private handleKeyDownEvent(editor: IContentModelEditor, rawEvent: KeyboardEvent) {
+    private handleKeyDownEvent(
+        editor: IContentModelEditor,
+        rawEvent: KeyboardEvent,
+        handledByEditFeature: boolean
+    ) {
         const which = rawEvent.which;
 
         if (!this.editWithContentModel || rawEvent.defaultPrevented) {
```
JiuqingSong, line 126: You can add the check to handledByEditFeature. We can treat handledByEditFeature in the same way with rawEvent.defaultPrevented

JiuqingSong (approved):

PR #1984, opened 2023-07-27
Fix table selector and resizer insertion

Andres-CT98:
### Description:

Redo of #1919

The Table selector and resizer are currently inserted on the document, thus putting them on a different container to the table they belong to. This creates issues in some scenarios such as when a dialog is created above the document with a high z-index. They also have a fixed position which causes issues in some scrolling scenarios, where they stayed fixed in place instead of moving or disappearing.

### Fix

Changed the insertion point of where the squares are located. The Selector and Resizer are inserted to where the Plugin specifies, if no selector is given it will default to the previous behaviour. The intent is for the selector to contain the editor. Additionally added a check for the resizer to not render if out of editor bounds.

### Warnings and implications

- The same scroll scenario where the squares remained on screen still remained due to the use of fixed position. Both operations are not functional, but it is a rare scenario to encounter.
- The same issue exists in the table Inserters and cell resizers, which will be fixed the same way on a different PR.
 - The name of the Plugin 'TableResize' could be changed, as it is very similar to 'TableResizer' and is not descriptive of its full function..


packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -30,7 +30,8 @@ export default class TableResize implements EditorPlugin {
         private onShowHelperElement?: (
             elementData: CreateElementData,
             helperType: 'CellResizer' | 'TableInserter' | 'TableResizer' | 'TableSelector'
-        ) => void
+        ) => void,
+        private containerId?: string
```
JiuqingSong, line 34: From line 136, this value is actually a selector, but not just an id, right? Then we should name it containerSelector instead.

Also please update comment

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableSelector.ts
```diff
@@ -20,7 +25,7 @@ export default function createTableSelector(
         elementData: CreateElementData,
         helperType: 'CellResizer' | 'TableInserter' | 'TableResizer' | 'TableSelector'
     ) => void,
-    contentDiv?: EventTarget | null
+    contentDiv?: HTMLElement
```
JiuqingSong, line 28: In line 32, it checks if table is visible to editor using this contentDiv. However, now contentDiv is changed to be the "container", which can be different than before, so isTableTopVisible() may return different result.

I think we should not change this contentDiv, but add one more parameter for the container.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableResizer.ts
```diff
@@ -137,11 +144,40 @@ function onDragging(
     }
 }
 
-function setResizeDivPosition(context: DragAndDropContext, trigger: HTMLElement) {
+function setResizeDivPosition(
```
JiuqingSong, line 147: What if we don't change this part and keep using position:fix, can it still work?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableSelector.ts
Andres-CT98: Specific to this case, the top value of the container and contentDiv should be the same, and if they are not what happens is that it renders a TableSelector square that is not visible. It is the same that happens currently with TableResizer and the bottom.

JiuqingSong: So that added an additional requirement that the container needs to have the same top position with editor.

We can actually remove this limitation by adding another parameter, and try keep using position:fix to simplify t he calculation of position.

position:fixed means the position is related to the whole window, no matter who the parent is, which is a very convenient style to use, especially when there is complex scrolling, zooming, margin, ...

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableResizer.ts
Andres-CT98: Fixed has some issues with scrolling, and changing it would mean reworking the values for left and top again. As it is now, just changing it does not work.

JiuqingSong: Scrolling is a separate issue. We can solve it by reapply the position when scroll. And it can be fixed separately.

For now, there is no complain about the scrolling issue yet so it should be lower priority

JiuqingSong: In this change we just need to solve the zIndex issue.

Scroll issue has been there for long time, so no hurry to fix

demo/scripts/controls/BuildInPluginState.ts
```diff
@@ -34,6 +34,7 @@ export default interface BuildInPluginState {
     experimentalFeatures: ExperimentalFeatures[];
     forcePreserveRatio: boolean;
     isRtl: boolean;
+    container: string;
```
JiuqingSong, line 37: Need a better name, e.g. tableResizerContainerSelector

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableEditor.ts
```diff
@@ -211,23 +212,24 @@ export default class TableEditor {
...
                 this.table,
-                this.editor.getZoomScale(),
-                this.isRTL,
+                this.editor,
                 this.onStartTableResize,
                 this.onFinishEditing,
-                this.onShowHelperElement
+                this.onShowHelperElement,
+                this.contentDiv,
+                this.anchorContainer
```
JiuqingSong, line 232: what about tableInserter?

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableResizer.ts
```diff
@@ -44,18 +60,19 @@ export default function createTableResizer(
...
-        setResizeDivPosition,
+        setDivPosition,
         {
             onDragStart,
             onDragging,
             onDragEnd,
         },
-        zoomScale
+        zoomScale,
+        undefined
```
JiuqingSong, line 75: Please add an inline comment to explain what this "undefined" is.

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableResizer.ts
```diff
@@ -137,14 +154,34 @@ function onDragging(
     }
 }
 
-function setResizeDivPosition(context: DragAndDropContext, trigger: HTMLElement) {
+function setDivPosition(context: DragAndDropContext, trigger: HTMLElement) {
     const { table, isRTL } = context;
     const rect = normalizeRect(table.getBoundingClientRect());
 
     if (rect) {
+        trigger.style.position = 'fixed';
```
JiuqingSong, line 162: We already set position: fixed at line 42, so need to set again here

packages/roosterjs-editor-plugins/lib/plugins/TableResize/editors/TableSelector.ts
```diff
@@ -61,12 +69,13 @@ export default function createTableSelector(
...
-        setSelectorDivPosition,
+        setDivPosition,
         {
             onDragEnd,
         },
-        zoomScale,
-        getOnMouseOut
+        context.zoomScale,
+        getOnMouseOut,
+        undefined
```
JiuqingSong, line 78: same here

JiuqingSong (approved):

PR #1993, opened 2023-07-31
Customize merge model when paste

juliaroldi:
Add `customizedMerge` to the `ContentModelBeforePasteEventData` interface so can we use it as callback to change the way the copied fragment is merged to content model.
Also, split the `getSelectedSegments` functions from the other collect selections functions, so it can be exported.


packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
```diff
@@ -120,14 +125,19 @@ function createBeforePasteEvent(
     };
 }
 
-function createFragmentFromClipboardData(
+function createBeforePasteEventAndFragment(
```
JiuqingSong, line 128: Actually this fucntion does not only create event and fragment, but also trigger the event and let plugins handle it (same for the origianl code). So I think we need a better name.

packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
```diff
@@ -151,12 +161,16 @@ function createFragmentFromClipboardData(
     }
 
     // Step 4: Trigger BeforePasteEvent so that plugins can do proper change before paste
-    editor.triggerPluginEvent(PluginEventType.BeforePaste, event, true /* broadcast */);
+    const pluginEvent = editor.triggerPluginEvent(
+        PluginEventType.BeforePaste,
+        eventData,
+        true /* broadcast */
+    ) as ContentModelBeforePasteEvent;
```
JiuqingSong, line 168: Is this type case required? Try remove it and see if it can pass the build.

packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
```diff
@@ -45,32 +47,32 @@ export default function paste(
         clipboardData.snapshotBeforePaste = editor.getContent(GetContentMode.RawHTMLWithSelection);
     }
 
-    const event = createBeforePasteEvent(
+    const eventData = createBeforePasteEventData(
```
BryanValverdeU, line 50: Is this needed? The event object should be accesable fine before the change

juliaroldi: I changed this function to create a object that has only the event data and not the eventType, then we use it to trigger the plugin event and use the return value of triggerPluginEvent that had been the modified by plugin actions to format the model. Without this change, we were using the object created in the paste file and not the one edited by plugins.

packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
juliaroldi: I renamed it to triggerPluginEventAndCreatePasteFragment and added a comment

JiuqingSong: sounds good

packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
JiuqingSong: hmm, that is a real problem

packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
BryanValverdeU: Let me test some scenarios, since I need to fix the unit test for this function

BryanValverdeU (approved):

PR #1999, opened 2023-08-02
Content Model: Improve cache behavior

### LINKED ISSUE #1997: Content Model: "Copy" can delete entity
1. Open demo site on a Mac OS, and switch to Content Model page (http://localhost:3000/?cm=1#Snapshot)
2. Restore the undo snapshot below:
3. Press "B" button to toggle Bold
4. From top menu bar, click "Edit", "Copy"

Expect: Content is copied, and content is not changed in editor

Actual: Content is copied, entity is gone

Snapshot:
```html
<div>aaaa</div><div class="_Entity _EType_aaaa _EId_aaaa_2 _EReadonly_1" contenteditable="false"><span><b>bbb</b></span></div><div>cccc</div><!--{"type":0,"isDarkMode":false,"start":[0,0,2],"end":[2,0,2]}-->
```

JiuqingSong:
- Move `allowCacheElement` from `DomToModelFormatContext` to `EditorContext`, so we will only allow cache element in Content Model when create content model from editor
- Remove `disableCacheElement` from `DomToModelOption` since now we always allow cache element for Content Model from editor, and always disallow cache element when create Content Model from other sources
- In API `cloneModel`, add an options parameter, and now by default cachedElement will not be included in cloned model. For entity and general model, we will put a cloned element as the wrapper element. We will include cached element and directly put the same reference to entity only when options.includeCachedElement is true. So that when clone model for cut/copy, the new model can be fully disconnected from editor content
- Fix some selection bugs in ContentModelCopyPastePlugin
- Fix #1997, copy content using Content Model won't make entity disappear even cached model is used
- Use common API `formatWithContentModel` for cut action in `ContentModelCopyPastePlugin`
- Fix all related test cases


juliaroldi (approved):

BryanValverdeU:
It seems that copying a entity is changing the color and the background color.

```
<div>asdsadsad&nbsp;<span class="entityDelimiterBefore">​</span><span style="display: inline-block;" class="_Entity _EType_asdasd _EId_asdasd_35 _EReadonly_1" contenteditable="false"><span>| entity |</span></span><span class="entityDelimiterAfter">​</span>&nbsp;asdsadsadsad</div><!--{"type":0,"isDarkMode":true,"start":[0,0,0],"end":[0,4,13]}-->
```

With this change:
image

Before
image

packages-content-model/roosterjs-content-model-dom/lib/domToModel/processors/entityProcessor.ts
```diff
@@ -23,22 +23,7 @@ export const entityProcessor: ElementProcessor<HTMLElement> = (group, element, c
         context,
         { segment: isBlockEntity ? 'empty' : undefined, paragraph: 'empty' },
         () => {
-            const wrapperToUse = context.allowCacheElement
-                ? element
-                : (element.cloneNode(true /* deep */) as HTMLElement);
-
-            if (!context.allowCacheElement) {
-                wrapperToUse.style.backgroundColor = element.style.backgroundColor || 'inherit';
-                wrapperToUse.style.color = element.style.color || 'inherit';
```
BryanValverdeU, line 32: I think maybe this is what might be causing the above issue.

JiuqingSong: You are right, I also just debugged to here.

JiuqingSong:
> It seems that copying a entity is changing the color and the background color.
> 
> ```
> <div>asdsadsad&nbsp;<span class="entityDelimiterBefore">​</span><span style="display: inline-block;" class="_Entity _EType_asdasd _EId_asdasd_35 _EReadonly_1" contenteditable="false"><span>| entity |</span></span><span class="entityDelimiterAfter">​</span>&nbsp;asdsadsadsad</div>
> ```
> 
> With this change, copying and pasting the first line image
> 
> Before image

Thanks for caching this. I fixed it by adding it back into `handleEntity`.

Now entity model will always have the original wrapper element, and we only handle the separation when
1. clone model
2. write back

BryanValverdeU (approved):

PR #2000, opened 2023-08-02
Fix rotator and side resize handles visibility

juliaroldi:
Update visibility rotate handle and side resize handles according to image size. 
If image has width/height less than the sum of 3 resize handles diameters plus 2 resize margins, hide side resize handles and rotator handle. 

dynamic_resizer_delight


JiuqingSong (approved):

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -406,7 +406,7 @@ export default class ImageEdit implements EditorPlugin {
      * quit editing mode when editor lose focus
      */
     private onBlur = () => {
-        this.setEditingImage(null, false /* selectImage */);
+        // this.setEditingImage(null, false /* selectImage */);
```
JiuqingSong, line 409: change back

PR #2002, opened 2023-08-03
Content Model: improve formatWithContentModel 2

JiuqingSong:
The goal of this change is to make the API `formatWithContentModel` can handle more common work, and simplify its parameter.
1. Add a context parameter to the format callback, so callback function can set more contextual result into this object
2. Move `skipUndoSnapshot` from options to this context object. So that callback can set its value according to execution result
3. Allow pass in `rawEvent` object into options
4. Allow callback set deleted entities into context, and remove `getOnDeleteEntityCallback`, then let `formatWithContentModel` handle the deletion of entities. Now entities will always be deleted if user want to, and plugin can't cancel it.
5. Pass in the context into delete* functions, so `deleteSegment` and `deleteBlock` are decoupled from `getOnDeleteEntityCallback`, and just need to put deleted entities into this context and they will finally be handled by `formatWithContentModel`
6. Fix related test cases


packages-content-model/roosterjs-content-model-editor/lib/modelApi/edit/deleteSelection.ts
```diff
@@ -1,31 +1,31 @@
...
 } from './utils/DeleteSelectionStep';
 
 /**
  * @internal
  */
 export function deleteSelection(
     model: ContentModelDocument,
-    onDeleteEntity: OnDeleteEntity,
-    additionalSteps: (DeleteSelectionStep | null)[] = []
+    additionalSteps: (DeleteSelectionStep | null)[] = [],
```
BryanValverdeU, line 17: I saw that this param was added in some places like the same default value. (mergeModel, ContentModelCopyPastePlugin, insertTable)
What if we move format context before additionalSteps? I think that way we do not need to add the empty array if we do not have additional steps, but still need the format context.

BryanValverdeU (approved):

packages-content-model/roosterjs-content-model-editor/lib/modelApi/edit/deleteSelection.ts
JiuqingSong: I considered about this.

You may noticed that the code under modelApi folder are only operating on an existing Content Model, but have no dependency to editor. That means, I'd like to make all "Model API" are independent to editor, when they are called from editor, we pass in a context object to collect the info that editor needs to know. When they are not called from an editor (that is supported, e.g. a model is created for paste), we don't need to pass any editor related thing. Here this context is something that only editor need to know.

But I think it may be a good idea to reuse EditorContext here, rather than creating a new `FormatWithContentModelContext`. Let me try it.

JiuqingSong: Ok, since the context object also carry in rawEvent which is not directly comes from editor, I think I'd better just keep current change.

PR #2011, opened 2023-08-04
Skip Trigger plugin event on plain text paste

BryanValverdeU:
When pasting as plain text, there is no need to let plugins handle the event.
So just skip trigger event function in this scenario.

### DIFF
--- a/packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
+++ b/packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
@@ -139,7 +139,7 @@ function triggerPluginEventAndCreatePasteFragment(
     pasteAsText: boolean,
     pasteAsImage: boolean,
     eventData: ContentModelBeforePasteEventData
-): ContentModelBeforePasteEvent | ContentModelBeforePasteEventData {
+): ContentModelBeforePasteEventData {
     const event = {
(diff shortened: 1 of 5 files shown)


packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
```diff
@@ -135,7 +139,7 @@ function triggerPluginEventAndCreatePasteFragment(
     pasteAsText: boolean,
     pasteAsImage: boolean,
     eventData: ContentModelBeforePasteEventData
-): { pluginEvent: ContentModelBeforePasteEvent; fragment: DocumentFragment } {
+): ContentModelBeforePasteEvent | ContentModelBeforePasteEventData {
```
JiuqingSong, line 142: Since `ContentModelBeforePasteEvent` inherits from `ContentModelBeforePasteEventData` and we didn't use anything in `ContentModelBeforePasteEvent` but not in `ContentModelBeforePasteEventData`, we can just return  `ContentModelBeforePasteEventData`

JiuqingSong (approved):

packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
BryanValverdeU: Good catch. fixed in latest commit.

PR #2014, opened 2023-08-08
Change default table border colour to black

Andres-CT98:
Change default border colour for all new inserted tables from #ABABAB to #000000
image

### DIFF
--- a/packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
+++ b/packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
@@ -12,9 +12,9 @@ import {
 } from 'roosterjs-content-model-types';
 
 const DEFAULT_FORMAT: Required<TableMetadataFormat> = {
-    topBorderColor: '#ABABAB',
-    bottomBorderColor: '#ABABAB',
-    verticalBorderColor: '#ABABAB',
+    topBorderColor: '#000000',
+    bottomBorderColor: '#000000',
+    verticalBorderColor: '#000000',
     hasHeaderRow: false,
     hasFirstColumn: false,
     hasBandedRows: false,
--- a/packages-content-model/roosterjs-content-model-editor/test/modelApi/common/clearModelFormatTest.ts
+++ b/packages-content-model/roosterjs-content-model-editor/test/modelApi/common/clearModelFormatTest.ts
@@ -710,7 +710,6 @@ describe('clearModelFormat', () => {
 
         clearModelFormat(model, blocks, segments, tables);
 
-        console.log('FAILED', model);
         expect(model).toEqual({
             blockGroupType: 'Document',
             blocks: [
--- a/packages-content-model/roosterjs-content-model-editor/test/modelApi/common/clearModelFormatTest.ts
+++ b/packages-content-model/roosterjs-content-model-editor/test/modelApi/common/clearModelFormatTest.ts
@@ -710,6 +710,7 @@ describe('clearModelFormat', () => {
 
         clearModelFormat(model, blocks, segments, tables);
 
+        console.log('FAILED', model);
         expect(model).toEqual({
             blockGroupType: 'Document',
             blocks: [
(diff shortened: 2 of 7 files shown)


packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
```diff
@@ -12,9 +12,9 @@ import {
 } from 'roosterjs-content-model-types';
 
 const DEFAULT_FORMAT: Required<TableMetadataFormat> = {
-    topBorderColor: '#ABABAB',
-    bottomBorderColor: '#ABABAB',
-    verticalBorderColor: '#ABABAB',
+    topBorderColor: '#000000',
```
JiuqingSong, line 15: This changes existing behavior, and will impact other roosterjs customers. I think we should either hide this change under a ExperimentalFeature, or allow pass in the border color as parameter, then keep the default behavior not changed.

Andres-CT98: This change is to explicitly change the default colour, as it is already possible to pass any desired colour as format.

JiuqingSong: That is the problem. We'd better not change the default color, because all other clients of roosterjs will be impacted by this change.

Since we can already pass in the format, I think we can just change OWA side to pass in the black border so that others won't be impacted.

Andres-CT98: I already asked and was told to change it from Rooster. I do not see why default values can not be changed, the idea is to impact all clients too.

JiuqingSong: We should not impact all clients. Otherwise we will later get a lot of ask why this is changed and how to change it back. Other people didn't ask for a change because they are satisfied with current behavior. When we want to change the behavior, we should always keep default behavior not changed (unless default behavior is not acceptable), and only change to the clients who want the change.

jvillalobos: I suspect most clients will stick with whatever default there is because there won't be strong opinions either way. If we change the default we will probably hear from some because any change will do that, but that doesn't mean it's the wrong change to make.

I proposed making the change at the Rooster level because I thought it would be easier to have a single default instead of two, but I don't feel too strongly about it, so if we just want to make the change on the OWA side that's fine too. However, we should have a way to make these sorts of changes if we think they're necessary. Maybe that means waiting for a major version bump or something like that.

JiuqingSong: Thanks @jvillalobos 

My top concern here is, with this change we have changed the content that end user generates, but not the customer developer's UI. So there are 3 groups of people:
1. RoosterJs developer, us.
2. RoosterJs consumers, people who uses roosterjs to build their apps.
3. End user. People who uses an app that is build on top of roosterjs by 2.

If our code change only impacts 2, that is kind of acceptable since 2 decides to use code from 1 so they also need to accept the changes made by 1 when they do version bump.
But since this change impact 3, and 3 doesn't know their content is also controlled by 1. So 3 will blame 2, then 2 has no idea how this happen and finally they found that it is caused by 1.

If 2 decide to make this change happen, that is fine. Because when 3 found something changed, 3 will go to 2, and 2 can explain why this change happens. But here it is not that case. 2 can't explain why 1 made change can impact 3.

jvillalobos: Understood. And that's why I mentioned the major version as a way to do that. If devs know that major version bumps can have an impact on 3, they'll be careful about reading version notes and testing changes before deploying. I just want to make sure that we have a way make those changes safely if we need to.

This isn't a big one, so I'm okay taking a different approach and just doing it for OWA.

JiuqingSong: yeah, agree that when doing major version bump we can change this behavior

Issue #2015, opened 2023-08-09
Content Model: Paste a table causes extra border in next paragraph


JiuqingSong:
Restore the following snapshot:
```html
<table style="border-collapse:collapse;border:1px solid black" id="tableSelected0"><tbody><tr><td>aa</td></tr></tbody></table><!--{"type":1,"tableId":"tableSelected0","isDarkMode":false,"firstCell":{"x":0,"y":0},"lastCell":{"y":1,"x":1}}-->
```
Copy,
Paste

Expect: Paste a table only
Actual: 
<img width="453" alt="image" src="https://github.com/microsoft/roosterjs/assets/23065085/fdab2ff9-a608-4f8a-b7e3-2aea4305a1b6">

PR #2016, opened 2023-08-09
MergeModel, do not inherit the styles of table when splitting the param

BryanValverdeU:
#2015
We should not add the table format when splitting the paragraph, as this can lead to border styles to be incorrectly applied to the splitted pagraph

Before
image

After
image

### DIFF
--- a/packages-content-model/roosterjs-content-model-editor/lib/modelApi/common/mergeModel.ts
+++ b/packages-content-model/roosterjs-content-model-editor/lib/modelApi/common/mergeModel.ts
@@ -287,7 +287,7 @@ function splitParagraph(markerPosition: InsertPoint, newParaFormat: ContentModel
 
 function insertBlock(markerPosition: InsertPoint, block: ContentModelBlock) {
     const { path } = markerPosition;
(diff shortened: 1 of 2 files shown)


packages-content-model/roosterjs-content-model-editor/lib/modelApi/common/mergeModel.ts
```diff
@@ -287,7 +287,8 @@ function splitParagraph(markerPosition: InsertPoint, newParaFormat: ContentModel
 
 function insertBlock(markerPosition: InsertPoint, block: ContentModelBlock) {
     const { path } = markerPosition;
-    const newPara = splitParagraph(markerPosition, block.format);
+    const newParaFormat = block.blockType === 'Table' ? {} : block.format;
```
JiuqingSong, line 290: I feel we should get this format only when block is paragraph. For all other block types, such as ListItem, Table, Divider, Entity, we should not inherit its format. What do you think?

BryanValverdeU: I agree with you.  I was not too sure to handle those block types too in this PR. But now that you mention it I will add that change too.

PR #2022, opened 2023-08-11
Update logic to decide if we need to merge a table on paste.

BryanValverdeU:
Update merge table logic when doing a paste.
We will merge a table on paste if:
1. Paste Model contains a single table
2. Paste Model contains a table and a paragraph with a BR element, as this is usually added to the clipboard with current implementation of copy 

Also add and update unit tests accordingly to the change


packages-content-model/roosterjs-content-model-editor/lib/editor/corePlugins/ContentModelCopyPastePlugin.ts
```diff
@@ -230,9 +230,7 @@ function selectionExToRange(
     let newRange: Range | null = null;
     if (selection.type === SelectionRangeTypes.TableSelection && selection.coordinates) {
         const table = tempDiv.querySelector(`#${selection.table.id}`) as HTMLTableElement;
-        const elementToSelect =
-            table.parentElement?.childElementCount == 1 ? table.parentElement : table;
-        newRange = createRange(elementToSelect);
+        newRange = createRange(table);
```
JiuqingSong, line 233: I remember there used to be a bug that when copy table, `<TABLE>` tag is not copied sometimes, and that is why we add a DIV around the table on purpose. Will this make the bug back?

BryanValverdeU: I see, I vaguely remember now that you mention it. 
I may need to rethink this then.

packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
```diff
@@ -105,6 +99,50 @@ export default function paste(
...
+    if (
+        pasteModel.blocks.length == 2 &&
+        pasteModel.blocks[0].blockType === 'Table' &&
+        pasteModel.blocks[1].blockType === 'Paragraph' &&
+        pasteModel.blocks[1].segments.length === 1 &&
+        pasteModel.blocks[1].segments[0].segmentType === 'Br'
+    ) {
+        pasteModel.blocks.splice(1);
+        return true;
+    }
```
JiuqingSong, line 142: nit: How about:

```ts
if (pasteModel has table and paragraph) {
  pasteModel.blocks.splice(1);
}

return pastModel has only table;

JiuqingSong (approved):

Issue #2023, opened 2023-08-11
Writing on a selected cell does not replace all text as expected


Andres-CT98:
**Describe the bug**
Writing on a selected cell does not replace all text as expected.
 
**To Reproduce**
1. Create a table
2. Select one or more cells
3. Type something

**Expected behavior**
When selecting more than two cells, the top left most one is the one expected to be replaced.

![CellSelectNoReplace](https://github.com/microsoft/roosterjs/assets/107568016/eaa398cf-eef4-411a-8563-5bba6ddfaa76)

PR #2024, opened 2023-08-11
Fix Triple clicking a single cell selecting more than one

Andres-CT98:
### Description:

Triple clicking a single cell will select the next cell too, or if it is at the end of a row, it will select the full current and next row.

### Fix

Catch if more than 3 clicks have been made in a table cell, and only select the clicked cell.

### Cases tested:

Triple click the first cell in a table with more than two cells.

**Before:** The clicked cell and its neighbour will be selected.

**After:** Only the clicked cell will be selected.

### Warnings and implications

- Issue: #2023 , Will be more visible with this change.


JiuqingSong:
Curious, about #2023, what is the actual behavior before and after this change?

Andres-CT98:
> Curious, about #2023, what is the actual behavior before and after this change?

The same as selecting one cell by dragging, it is just that triple clicking makes it more evident. I also have a fix for it which will be done separately.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/mouseUtils/handleMouseDownEvent.ts
```diff
@@ -53,65 +53,99 @@ export function handleMouseDownEvent(
...
-        clearState(state, editor);
+    if (which == LEFT_CLICK) {
+        // Triple clicking a cell will select that cell
+        if (detail >= 3) {
+            editor.runAsync(editor => {
+                const sel = editor.getDocument().defaultView?.getSelection();
+                const first = getCellAtCursor(editor, sel?.anchorNode);
+                const firstTable = getTableAtCursor(editor, first);
+                const targetTable = getTableAtCursor(editor, first);
+                if (firstTable! == targetTable! && safeInstanceOf(first, 'HTMLTableCellElement')) {
```
JiuqingSong, line 68: In what case firstTable != targetTable? From line 66 and 67, right side of "=" are the same, so they should be always the same

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/mouseUtils/handleMouseDownEvent.ts
```diff
@@ -53,65 +53,99 @@ export function handleMouseDownEvent(
...
+                    state.firstTarget = first;
+                    state.lastTarget = first;
+                    selectTable(editor, state);
 
-        if (getTableAtCursor(editor, event.rawEvent.target)) {
-            const doc = editor.getDocument() || document;
+                    state.tableSelection = true;
+                    state.firstTable = firstTable as HTMLTableElement;
+                    state.targetTable = targetTable;
+                    updateSelection(editor, first, 0);
```
JiuqingSong, line 88: Curious, can this be replaced by calling editor.select({... (a table selection rangeEx) })?

I understand the original code is also like this, but just want to try improve it

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/mouseUtils/handleMouseDownEvent.ts
```diff
@@ -53,65 +53,99 @@ export function handleMouseDownEvent(
...
+                    };
+
+                    state.firstTarget = first;
+                    state.lastTarget = last;
+                    selectTable(editor, state);
+
+                    state.tableSelection = true;
+                    state.firstTable = firstTable as HTMLTableElement;
+                    state.targetTable = targetTable;
+                    updateSelection(editor, first, 0);
```
JiuqingSong, line 145: This part is almost the same with the previous part. Can we try to merge them?

JiuqingSong: For example, first we always need to get selected table and cell, then coordinates, this part should be the same. Then after that we can check how many time mouse is clicked, then do different action? Not sure if this is possible.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/mouseUtils/handleMouseDownEvent.ts
BryanValverdeU: For triple click I think editor.select should be enough. For the other scenarios I suggest testing if the selection is not broken. since setting selection with range does not mantain the initial and the last node selected.

Andres-CT98: I tried using editor.select but it does not seem to work as expected.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/mouseUtils/handleMouseDownEvent.ts
Andres-CT98: I implemented triple click inside of it now, by making first and last the same.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/mouseUtils/handleMouseDownEvent.ts
```diff
@@ -53,65 +54,67 @@ export function handleMouseDownEvent(
...
         }
-    }
 
-    if (which == LEFT_CLICK && shiftKey) {
-        editor.runAsync(editor => {
-            const sel = editor.getDocument().defaultView?.getSelection();
-            const first = getCellAtCursor(editor, sel?.anchorNode);
-            const last = getCellAtCursor(editor, sel?.focusNode);
-            const firstTable = getTableAtCursor(editor, first);
-            const targetTable = getTableAtCursor(editor, first);
```
JiuqingSong, line 87: @BryanValverdeU I saw L86 and L87 are actually the same. Please review if it is ok to remove this line. Thanks.

JiuqingSong:
> > Curious, about #2023, what is the actual behavior before and after this change?
> 
> The same as selecting one cell by dragging, it is just that triple clicking makes it more evident. I also have a fix for it which will be done separately.

I know what you mean now. Thanks for opening but to track.

JiuqingSong (approved):

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/mouseUtils/handleMouseDownEvent.ts
BryanValverdeU: I see, should be okay to remove it.

BryanValverdeU (approved):

PR #2030, opened 2023-08-15
Replace first cell content if input while on cell selection

Andres-CT98:
### Description:

Fix of #2023 

Writing on a selected cell, or group of cells, does not replace the content of the first cell as expected.

### Fix

On the TableCellSelection plugin, the handle key down event used to end the cell selection when typing. Now it makes only the first cell selected and creates a new selection of all its children.

### Cases tested:

1. Create a table
2. Select one or more cells
3. Type something

**Before:** The place where the cursor was placed is where the typed text will be inserted.

**After:** The first cell in the selection, top left one, will be cleared and the typed text will be inserted.

### Warnings and implications

- Keys like the function keys and Caps Lock will end cell selections. This behaviour exits before this change. However for the case of typing, after pressing said keys the full contents of the cell will still be selected, so typing and replacing the content should not be affected.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
@@ -93,6 +93,13 @@ export default class TableCellSelection implements EditorPlugin {
                         event.rawEvent.preventDefault();
                     }
                     break;
+                case PluginEventType.Input:
+                    if (!this.state.startedSelection) {
+                        this.handleInputEvent(this.editor);
+                    } else {
+                        event.rawEvent.preventDefault();
+                    }
+                    break;
                 case PluginEventType.Scroll:
                     if (this.state.startedSelection) {
                         handleScrollEvent(this.state, this.editor);
@@ -126,5 +133,19 @@ export default class TableCellSelection implements EditorPlugin {
             state.tableSelection = true;
             editor.select(selection.table, null);
         }
+    }
+
+    private handleInputEvent(editor: IEditor) {
+        // If typing on any cell selection, clear the first cell
+        const range = editor.getSelectionRangeEx();
+        if (range.type == SelectionRangeTypes.TableSelection) {
+            const row = range.ranges[0];
+            const firstCell = row.startContainer.childNodes[row.startOffset];
+            firstCell.childNodes.forEach(node => {
+                node.remove();
+            });
+        }
+        // Add undo snapshot after content deletion
+        editor.addUndoSnapshot();
     }
 }
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
@@ -93,13 +93,6 @@ export default class TableCellSelection implements EditorPlugin {
                         event.rawEvent.preventDefault();
                     }
                     break;
-                case PluginEventType.Input:
-                    if (!this.state.startedSelection) {
-                        this.handleInputEvent(this.editor);
-                    } else {
-                        event.rawEvent.preventDefault();
-                    }
-                    break;
                 case PluginEventType.Scroll:
                     if (this.state.startedSelection) {
                         handleScrollEvent(this.state, this.editor);
@@ -133,19 +126,5 @@ export default class TableCellSelection implements EditorPlugin {
             state.tableSelection = true;
             editor.select(selection.table, null);
         }
-    }
-
-    private handleInputEvent(editor: IEditor) {
-        // If typing on any cell selection, clear the first cell
-        const range = editor.getSelectionRangeEx();
-        if (range.type == SelectionRangeTypes.TableSelection) {
-            const row = range.ranges[0];
-            const firstCell = row.startContainer.childNodes[row.startOffset];
-            firstCell.childNodes.forEach(node => {
-                node.remove();
-            });
-        }
-        // Add undo snapshot after content deletion
-        editor.addUndoSnapshot();
     }
 }
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/keyUtils/handleKeyDownEvent.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/keyUtils/handleKeyDownEvent.ts
@@ -73,7 +73,13 @@ export function handleKeyDownEvent(
         editor.getSelectionRangeEx()?.type == SelectionRangeTypes.TableSelection &&
         (!isCtrlOrMetaPressed(event.rawEvent) || which == Keys.HOME || which == Keys.END)
     ) {
-        editor.select(null);
+        // Collapse selection to first selected cell
+        const range = editor.getSelectionRangeEx();
+        if (range.type == SelectionRangeTypes.TableSelection) {
+            const row = range.ranges[0];
+            const firstCell = row.startContainer.childNodes[row.startOffset];
+            editor.select(firstCell);
+        }
     }
 }
 
--- a/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/keyUtils/handleKeyDownEvent.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/keyUtils/handleKeyDownEvent.ts
@@ -39,7 +39,6 @@ export function handleKeyDownEvent(
     }
 
     const range = editor.getSelectionRangeEx();
-
     if (shiftKey) {
         if (!state.firstTarget) {
             const pos = editor.getFocusedPosition();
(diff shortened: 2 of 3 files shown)


packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -127,4 +134,18 @@ export default class TableCellSelection implements EditorPlugin {
...
+        if (range.type == SelectionRangeTypes.TableSelection) {
+            const row = range.ranges[0];
+            const firstCell = row.startContainer.childNodes[row.startOffset];
+            firstCell.childNodes.forEach(node => {
+                node.remove();
+            });
+        }
+        // Add undo snapshot after content deletion
+        editor.addUndoSnapshot();
+    }
```
JiuqingSong, line 150: This will also trigger if type out of table. 

The problem is, we should not always add undo snapshot in that case. You can try in an empty doc, type something, then undo. Now each undo only remove one character, which is wrong.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -93,6 +93,13 @@ export default class TableCellSelection implements EditorPlugin {
                         event.rawEvent.preventDefault();
                     }
                     break;
+                case PluginEventType.Input:
+                    if (!this.state.startedSelection) {
+                        this.handleInputEvent(this.editor);
+                    } else {
+                        event.rawEvent.preventDefault();
```
JiuqingSong, line 100: Is Input event preventable?

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/keyUtils/handleKeyDownEvent.ts
```diff
@@ -73,7 +73,13 @@ export function handleKeyDownEvent(
         editor.getSelectionRangeEx()?.type == SelectionRangeTypes.TableSelection &&
         (!isCtrlOrMetaPressed(event.rawEvent) || which == Keys.HOME || which == Keys.END)
     ) {
-        editor.select(null);
+        // Collapse selection to first selected cell
+        const range = editor.getSelectionRangeEx();
+        if (range.type == SelectionRangeTypes.TableSelection) {
```
JiuqingSong, line 78: We already checked selection type in line 73

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
JiuqingSong: It seems we just need to handle keydown event, it will be good enough. I tried to comment out the handling code of input event, and it still works well

Andres-CT98: There is an issue with undo that is mitigated by using the Input method. It only happens if you type one character and do undo, the inputted character and the last written thing will both be undone. This issue is present today and not introduced by this change.

The reason I clear the full cell is that I do not know if select is enough to cover all possible elements in the cell for deletion. And for clearing I want to make sure it is an input and not other keys like caps lock.

JiuqingSong: About the undo issue, can you provide a full end to end repro step? I want to understand more about it. Thanks

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
```diff
@@ -127,4 +134,18 @@ export default class TableCellSelection implements EditorPlugin {
             editor.select(selection.table, null);
         }
     }
+
+    private handleInputEvent(editor: IEditor) {
+        // If typing on any cell selection, clear the first cell
+        const range = editor.getSelectionRangeEx();
+        if (range.type == SelectionRangeTypes.TableSelection) {
```
JiuqingSong, line 141: Actually, when I debug your code, Input event happens after KeyDown event. Since you have already changed selection when KeyDown event, it will fail the check of table selection range type, so will go directly to line 149, so code from line 142 to line 146 is never executed.

image

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
Andres-CT98: 1. Create a table
2. Fill it with values, remembering which one was the last one
3. Make a cell selection
4. Type one character
5. Undo

Notice that the typed character and the last filled value are both undone. The only one undone should be the typed one.
tableUndoWrong

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
Andres-CT98: Interesting. I assumed it would retain the selection type.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
JiuqingSong: Ok, in that case we do need to call addUndoSnapshot() when type with table selection, but only need to do once.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
JiuqingSong: Because you have called editor.select(), that will change selection type.

JiuqingSong: And because of that, the executing result is correct now. But if line 43-46 are executed, the new typed content will be removed, just because when Input event is triggered, the new typed content is already added into the view.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
JiuqingSong: And I think it should be good enough to just select everything inside the table cell when keydown. Because:
1. User made a table selection, then type
2. User selected everything in a table cell, then type

Both of them should get the same result. So making a selection to the first select cell means we convert 2 to 1, so that is the right operation.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
Andres-CT98: I have noticed that, would it work if I did the deletion before the Input event? I would need to add it to PluginEventType.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
Andres-CT98: I have been trying both selection types, but each have an issue. editor.select with the Td as parameter has the above issue, while editor.select with the table and coordinates as parameter has an issue of replacing everything in the cell when typing. For example if the cell has multiple lines, only the first one is replaced.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
JiuqingSong: My feeling is you just need to keep you current code to select the content of first selected cell, (and call addUndoSnapshot() in a proper place), then let browser handle everything else, so no need to handle Input event at all.

Because Input event is triggered after type, so anything there is too late.

And maybe only need to handle character types (you can use `isCharacterValue` function to check) so for other keys like caps lock, leave it to browser

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/TableCellSelection.ts
Andres-CT98: I made changes. No I select all child nodes of the cell. I have not found how to solve the above issue with undo.

JiuqingSong:
I made the following change locally to make undo work.

1. In handleKeydownEvent.ts:
image

2. In UndoPlugin.ts:
image

The problem is although we added new undo snapshot, but the flag "hasNewContenet" is also cleared. So we need to set it back when there is really input.

packages/roosterjs-editor-plugins/lib/plugins/TableCellSelection/keyUtils/handleKeyUpEvent.ts
```diff
@@ -26,7 +26,7 @@ export function handleKeyUpEvent(
         !state.preventKeyUp &&
         IGNORE_KEY_UP_KEYS.indexOf(which) == -1
     ) {
-        clearState(state, editor);
+        editor.addUndoSnapshot(() => clearState(state, editor));
```
BryanValverdeU, line 29: Using Arrow Keys, Escape or some other keys that do not add content will add a unneeded undo snapshot. Is Undo snapshot really needed?

JiuqingSong: Makes sense. In that case, my change in handleKeydownEvent.ts to addUndoSnapshot() is not needed

Andres-CT98: I can use isCharacterValue here to only add the snapshot if it is true.

JiuqingSong (approved):

PR #2038, opened 2023-08-17
Add border collapse to table when pasting from Excel.

BryanValverdeU:
Add borderCollapse to table when pasting from Excel, to fix the table layout looking incorrect after paste.

Before, (notice the gaps between the borders)
image

After
image


JiuqingSong:
What does it look like before/after this change?

packages-content-model/roosterjs-content-model-editor/lib/editor/plugins/PastePlugin/Excel/processPastedContentFromExcel.ts
```diff
@@ -43,6 +43,11 @@ export function processPastedContentFromExcel(
         }
     }
 
+    addParser(event.domToModelOption, 'table', (format, element) => {
+        if (!element.style.borderCollapse) {
+            format.borderCollapse = true;
```
JiuqingSong, line 48: I don't understand here. Does it mean if the source element has no borderCollapse, we always treat it as collapsed? 
Then what does it look like when paste without content model?

BryanValverdeU:
> What does it look like before/after this change?

Added the before and after in the description

packages-content-model/roosterjs-content-model-editor/lib/editor/plugins/PastePlugin/Excel/processPastedContentFromExcel.ts
BryanValverdeU: Yes, the problem is that since excel does not add the table tag to the copied HTML, we add the table tag with the excel handler regex.  this causes the table tag to to have no stylings from the Excel content, and missing the border collapse style.

In short, the issue here is that Excel provides an incorrect clipboard data and we need to add this otherwise it would look bad. WIthout content model we also add this style on paste, but I was not able to find where exactly it is.

PR #2044, opened 2023-08-25
Support more border styles

BryanValverdeU:
Add support for more border styles. These styles are added in this PR:

    'borderTopColor',
    'borderRightColor',
    'borderBottomColor',
    'borderLeftColor',
    'borderTopStyle',
    'borderRightStyle',
    'borderBottomStyle',
    'borderLeftStyle',
    'borderTopWidth',
    'borderRightWidth',
    'borderBottomWidth',
    'borderLeftWidth',

Word online adds styles to table borders like:
image

Since border-X-color was not supported before this change, when pasting some tables from WordOnline to rooster it causes the table to not have border, since the default color for a border style like `border-top: 1pt solid` is equal to none.


packages-content-model/roosterjs-content-model-dom/lib/formatHandlers/common/borderFormatHandler.ts
```diff
@@ -10,6 +10,18 @@ export const BorderKeys: (keyof BorderFormat & keyof CSSStyleDeclaration)[] = [
...
+    'borderBottomColor',
+    'borderLeftColor',
+    'borderTopStyle',
+    'borderRightStyle',
+    'borderBottomStyle',
+    'borderLeftStyle',
+    'borderTopWidth',
+    'borderRightWidth',
+    'borderBottomWidth',
+    'borderLeftWidth',
```
JiuqingSong, line 24: Actually, borderTop/Right/Bottom/Left is already the combination of color, style and width. So we should not add redundant properties, but always calculate the values from those existing properties. There are existing API `extractBorderValues` and `combineBorderValue` to do this work

BryanValverdeU: I see, just an example to make sure I understand correctly. 
From this snapshot:

```
<table data-editing-info="{&quot;topBorderColor&quot;:&quot;#ABABAB&quot;,&quot;bottomBorderColor&quot;:&quot;#ABABAB&quot;,&quot;verticalBorderColor&quot;:&quot;#ABABAB&quot;,&quot;hasHeaderRow&quot;:false,&quot;hasFirstColumn&quot;:false,&quot;hasBandedRows&quot;:false,&quot;hasBandedColumns&quot;:false,&quot;bgColorEven&quot;:null,&quot;bgColorOdd&quot;:&quot;#ABABAB20&quot;,&quot;headerRowColor&quot;:&quot;#ABABAB&quot;,&quot;tableBorderFormat&quot;:0,&quot;verticalAlign&quot;:&quot;top&quot;}" style="box-sizing: border-box; border-collapse: collapse; border-spacing: 0px;"><tbody><tr><td style="border-top: 1pt solid;border-color: inherit;"><br></td></tr></tbody></table><div><br></div><!--{"type":0,"isDarkMode":true,"start":[0,0,0,0,0],"end":[0,0,0,0,0]}-->
```

I have a single table cell with these styles:
image

Calling element.style.borderTop, from the only table cell present in the content, is going to return an empty string. But if I call each property (color/width/style) they return the value correctly.
image

 If I understand correctly, you mean we need to add this new functionality to those function to handle this scenario?

JiuqingSong: The problem with this snapshot is the border color is "inherit" which can't be put into a combined style. And actually today's content model code can't handle it. Rewrite with content model will cause the border to be lost (need to clear cached element)

Roosterjs should never generate border with color "inherit", it is only possible to see such value when paste or load existing content. Maybe we should do a normalization of border color in those cases?

JiuqingSong: I think we can handle "inherit" separately, but focus on the major feature work for now, and open a bug to track.

BryanValverdeU: Got it. Since this is only for paste, I will handle this diferently and only for paste scenario.

PR #2045, opened 2023-08-25
Remove deprecated colors from borders, text and background.

BryanValverdeU:
Win32 desktop apps add borders to tables like: `1pt solid windowtext`, window text color is not supported. So remove those styles on paste

Added before and after here:
https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/220620/

Also, remove the deprecated colors using the textColorFormatHandler.parse and backgroundColorFormatHandler.parse


JiuqingSong:
Actually, I think we also need to filter out unsupported colors in regular content model parsing process, but not only paste.

You can apply this process to 
- textColorFormatHandler.parse
- backgroundColorFormatHandler.parse

And consider the perf impact, we don't want always extract the border info when parse, so you can keep doing this for border only when paste.

Then this change can also fix https://outlookweb.visualstudio.com/Outlook%20Web/_queries/edit/209841/?triage=true

BryanValverdeU:
> Actually, I think we also need to filter out unsupported colors in regular content model parsing process, but not only paste.
> 
> You can apply this process to
> 
> * textColorFormatHandler.parse
> * backgroundColorFormatHandler.parse
> 
> And consider the perf impact, we don't want always extract the border info when parse, so you can keep doing this for border only when paste.
> 
> Then this change can also fix https://outlookweb.visualstudio.com/Outlook%20Web/_queries/edit/209841/?triage=true

Sounds good, added support to remove these styles to those functions.

JiuqingSong (approved):

PR #2049, opened 2023-09-04
Keep image selection when editing

juliaroldi:
When select a image that is editing mode, it will change the editor selection to the shadow root. Therefore, when mouse up, check if the image is editing mode, then reselect it.


JiuqingSong:
Question, without this change, is there anything borken?

juliaroldi:
> Question, without this change, is there anything borken?

The getSelectionRangeEx are not returning the image selection when we click over a image that is in editing mode (image editing wrapper on), because it selects the span that has the shadow dom element and not the image.

JiuqingSong:
> > Question, without this change, is there anything borken?
> 
> The getSelectionRangeEx are not returning the image selection when we click over a image that is in editing mode (image editing wrapper on), because it selects the span that has the shadow dom element and not the image.

I see.

JiuqingSong (approved):

PR #2055, opened 2023-09-07
Content Model: Add solid paragraph in new table cell

JiuqingSong:
When insert new table, currently we are adding an implicit paragraph in each table cell, it will be represented as a BR tag. So if user keep typing and press ENTER, it will keep using BR as line splitter. This causes some problem such as applying bullet list causes only one list item is created.

To fix this, we add a solid (not implicit) paragraph in each table cell.


BryanValverdeU:
I was testing, and noticed that when using table selection, the public API to set the segment format is not applying to new formats
Unrelated to this PR, but just an FYI

BryanValverdeU (approved):

JiuqingSong:
> I was testing, and noticed that when using table selection, the public API to set the segment format is not applying to new formats Unrelated to this PR, but just an FYI not sure sure if it is a recent regression

Good finding! Let me check.

PR #2062, opened 2023-09-11
[FHL] Deprecate AutoFormatList experimental feature

juliaroldi:
Deprecated the `AutoFormatList` experimental feature and always enable `AutoFormatList` feature. Also always the disable `AutoBullet` content edit feature.

### DIFF
--- a/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
+++ b/packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
@@ -34,13 +34,11 @@ import {
     Keys,
     PluginKeyboardEvent,
     QueryScope,
-    RegionBase,
     ListType,
     ExperimentalFeatures,
     PositionType,
     NumberingListType,
     BulletListType,
-    IPositionContentSearcher,
 } from 'roosterjs-editor-types';
 
 const PREVIOUS_BLOCK_CACHE_KEY = 'previousBlock';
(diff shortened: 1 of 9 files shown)


packages/roosterjs-editor-types/lib/enum/ExperimentalFeatures.ts
```diff
@@ -144,6 +144,7 @@ export const enum ExperimentalFeatures {
     TabKeyTextFeatures = 'TabKeyTextFeatures',
 
     /**
+     * @deprecated This feature is always enabled
```
JiuqingSong, line 147: Please move it up before line 139 to put this feature flag together with other deprecated ones.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -305,19 +304,16 @@ const AutoBullet: BuildInEditFeature<PluginKeyboardEvent> = {
             true /*canUndoByBackspace*/
         );
     },
+    defaultDisabled: true,
```
JiuqingSong, line 307: Besides this, I think you can now clear the content of this feature, keep the feature itself but make it do nothing. So that 
1. Even it is added, it will not be conflict with the new feature
2. Save some bytes.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -248,76 +248,22 @@ function isAListPattern(textBeforeCursor: string) {
...
-                    prepareAutoBullet(editor, textRange);
-                    toggleNumbering(editor, num);
-                }
-                searcher.getRangeFromText(textBeforeCursor, true /*exactMatch*/)?.deleteContents();
-            },
-            undefined /*changeSource*/,
-            true /*canUndoByBackspace*/
-        );
-    },
+    shouldHandleEvent: (event, editor) => {},
```
JiuqingSong, line 255: Not sure if this can pass the build. If not, just let it return false.

JiuqingSong (approved):

PR #2073, opened 2023-09-18
Remove Segment format from the pasted content when the format merge option is equal to `none`

BryanValverdeU:
When copying to the editor without merging the formats, if the editor have default format, the pasted content also get the format, which should not happen as we did not select the option to merge formats.

This is because the parent paragraph still have the segment format.
To fx, when we paste without merging the formats and split the first paragraph, we need to remove the segmentFormat from the parent paragraph and apply it to the already existing segments, that way the segment format will not get propagated to the pasted elements.

Before
9182023B44

After
9182023B4


juliaroldi (approved):

packages-content-model/roosterjs-content-model-editor/lib/modelApi/common/mergeModel.ts
```diff
@@ -120,14 +120,21 @@ export function mergeModel(
...
 ) {
     const { paragraph, marker } = markerPosition;
     const newParagraph = mergeToCurrentParagraph
         ? paragraph
         : splitParagraph(markerPosition, newPara.format);
     const segmentIndex = newParagraph.segments.indexOf(marker);
 
+    if (option?.mergeFormat == 'none' && mergeToCurrentParagraph) {
+        newParagraph.segments.forEach(segment => {
+            segment.format = { ...(newParagraph.segmentFormat || {}), ...segment.format };
```
juliaroldi, line 134: Are we passing the paragraph format to segment and them removing the paragraph format, right? If some segment already has a format that is different from the paragraph format, will the segment lose the format and apply only the paragraph format?

JiuqingSong, line 134: According how format is working on segment today, existing segments should already have a copy of segment format from the paragraph. That means even we don't rewrite format on segments, it should still keep existing formats. (of cause for the new merged segments, we still need to copy the format).

BryanValverdeU: Right, it should not. since the segment.format object we are destructing to create the new format would override the newParagraph.segmentFormat. If the format keys are duplicated. But maybe need to update this due to Jason's comment.

BryanValverdeU: I see, so, we only need to remove the segment format from the paragraph then, right? 
Since we already set the segment (default) format from container paragraph to each child segment's format.

JiuqingSong: For those newly merged segments, we still need to merge the format since there may be some format specified in block.segmentFormat but not exist in new segment. For those existing segments, nothing need to be done.

But I think to simplify the code, it should be fine to just do merge for all segments like you already did.

JiuqingSong (approved):

PR #2077, opened 2023-09-18
Delete link and image

juliaroldi:
When an image contains a hyperlink, the image element is encapsulated within an anchor (a) element, like so:
`<a ...>

</a>
`
Therefore both image and selection marker segments has link properties. if the image is deleted and there is no other segment with link decorator, remove selection marker link decorator. 
Also, change the image font-size to image wrapper, so a additional span element do not be created. 

Before: 
DeleteImageBefore

After: 
DeleteImageAfter


packages-content-model/roosterjs-content-model-editor/lib/modelApi/edit/utils/deleteSegment.ts
```diff
@@ -26,11 +26,12 @@ export function deleteSegment(
 
     switch (segmentToDelete.segmentType) {
         case 'Br':
-        case 'Image':
         case 'SelectionMarker':
             segments.splice(index, 1);
             return true;
-
+        case 'Image':
+            segments.splice(index, segmentToDelete.link ? 2 : 1);
```
JiuqingSong, line 33: What are the 2 segments when delete?

juliaroldi: The selection marker and the image.

JiuqingSong: So that means selection marker will be removed from model? how does browser know where to focus after this?

JiuqingSong: I feel this issue will happen to all kinds of segments, but not only image.

Maybe we just need to remove the link decorator (as well as code decorator) in `normalizeContentModel` if a selection marker has link and its sibling segments have not.

packages-content-model/roosterjs-content-model-dom/lib/modelApi/common/normalizeParagraph.ts
```diff
@@ -47,3 +49,12 @@ function removeEmptySegments(block: ContentModelParagraph) {
         }
     }
 }
+
+function removeEmptyLinks(paragraph: ContentModelParagraph) {
+    const segments = paragraph.segments;
+    const noMarkerSegments = segments.filter(x => x.segmentType != 'SelectionMarker');
+    const marker = segments.find(x => x.segmentType == 'SelectionMarker');
+    if (marker && marker.link && noMarkerSegments.every(x => !x.link)) {
+        delete marker.link;
```
JiuqingSong, line 58: what if
```html
<a href="..">test1</a>
test2
[SELECTION MARKER HERE]
```
?

JiuqingSong: You can do something like:

paragraph.segments.forEach((segment, i) => {
  if (segment.link && areSameFormat(segment.link, paragraph.segments[i-1]?.link) && areSameFormat(segment.ink, paragraph.segments[i+1]?.link) {
   delete segment.link;
}
}

juliaroldi: I think we only need to delete segment.link when the segment is selection marker. I think we might lose the link in other segment types if we check all segments.

PR #2091, opened 2023-09-21
Fix text color in in cells with shade. 

juliaroldi:
When normalizeTable and applyTableFormat, apply the cell text color to the segment inside the cells, so the default color do not override it. 

cellShadeTest


packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/setTableCellBackgroundColor.ts
```diff
@@ -16,7 +16,8 @@ const Black = '#000000';
 export function setTableCellBackgroundColor(
     cell: ContentModelTableCell,
     color: string | null | undefined,
-    isColorOverride?: boolean
+    isColorOverride?: boolean,
+    overrideTextColor?: boolean
```
JiuqingSong, line 20: Suggested name: applyToSegments

JiuqingSong (approved):

PR #2100, opened 2023-09-25
Fix link underline on images

juliaroldi:
When add link decorator, do not set underline for non text segments.


packages-content-model/roosterjs-content-model-editor/lib/publicApi/link/insertLink.ts
```diff
@@ -43,14 +43,16 @@ export default function insertLink(
...
         const linkData = matchLink(url);
-        const link: ContentModelLink = {
-            dataset: {},
-            format: {
-                href: linkData ? linkData.normalizedUrl : applyLinkPrefix(url),
-                anchorTitle,
-                target,
-                underline: true,
-            },
+        const link = (underline: boolean = true): ContentModelLink => {
```
JiuqingSong, line 46: suggest move this out to be a normal function, and pass anything need as parameter.

JiuqingSong (approved):

PR #2104, opened 2023-09-26
Do not add link to trailing spaces

juliaroldi:
When insert link to not insert it in trailing spaces segments. Split the trailing spaces in the selected segments and only apply link to non trailing spaces segments
Before: 
Screenshot 2023-09-26 at 14 03 34

After: 
Screenshot 2023-09-26 at 14 02 53

### DIFF
--- a/packages-content-model/roosterjs-content-model-editor/lib/publicApi/link/insertLink.ts
+++ b/packages-content-model/roosterjs-content-model-editor/lib/publicApi/link/insertLink.ts
@@ -1,6 +1,6 @@
 import getSelectedSegments from '../selection/getSelectedSegments';
 import { ChangeSource } from 'roosterjs-editor-types';
-import { ContentModelLink } from 'roosterjs-content-model-types';
+import { ContentModelLink, ContentModelSegment } from 'roosterjs-content-model-types';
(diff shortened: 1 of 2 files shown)


packages-content-model/roosterjs-content-model-editor/lib/publicApi/link/insertLink.ts
```diff
@@ -153,3 +169,32 @@ function checkXss(link: string): string {
...
+            segments.push(trailingSpaceSegment);
+        } else {
+            segments.push(segment);
+        }
+    });
+    return segments;
+};
+
+const splitTrailingSpace = (segment: ContentModelSegment) => {
+    if (segment.segmentType == 'Text' && segment.text.trimRight().length !== segment.text.length) {
```
JiuqingSong, line 188: Use isTrailingSpace()

JiuqingSong:
Please test the following cases. (Each case is an array of segment, with text value from each item of the array)
Better add test cases for them
1. [' test ']
2. [' ', 'test', ' ']
3. [' test1', ' test2 ', 'test3 ']
4. ['test1', ' test2 ', 'test3']

PR #2109, opened 2023-09-27
Add announce Plugin

BryanValverdeU:
- Add announce plugin
- Add a AnnounceHandler, that will be in charge of announcing the messages by using a aria-live element, this handler will require a string map with the localized strings to announce messages from built-in RoosterJS features.
- Add an additional callback property to ContentEditEventData, getAnnounceData, used in the Announce Plugin
- Add first announcing logic when indenting/outdenting list

AnnounceListIndent


JiuqingSong:
In general, I'd like to discuss about the design.

Announcing is not a core function of editor, but should only be applied when A11y is cared. So can we make it as a plugin instead? And editor/other plugins just provides as few info as possible (I think your code already did this part).

For example, with ContentChangedEvent, we an add something about announcing, then the plugin see the event and see there is announcing info, then it will announce.

So if there is not such plugin added, we don't announce at all.

What do you thing? I'm open to disucss.

BryanValverdeU:
> In general, I'd like to discuss about the design.
> 
> Announcing is not a core function of editor, but should only be applied when A11y is cared. So can we make it as a plugin instead? And editor/other plugins just provides as few info as possible (I think your code already did this part).
> 
> For example, with ContentChangedEvent, we an add something about announcing, then the plugin see the event and see there is announcing info, then it will announce.
> 
> So if there is not such plugin added, we don't announce at all.
> 
> What do you thing? I'm open to disucss.

Thanks for checking Jason,

I agree with all you said. 
Adding the additional announce info to the content changed event may need some specific plugin array ordering, which I think may be confusing to be used. 
Because if there is a plugin that wanted us to announce something and it is after the new plugin, it may be missed.

Maybe we can just add a new PluginEvent for announcing and let the new plugin handle it exclussively. Do you think this could be a good way?

JiuqingSong:
> > In general, I'd like to discuss about the design.
> > Announcing is not a core function of editor, but should only be applied when A11y is cared. So can we make it as a plugin instead? And editor/other plugins just provides as few info as possible (I think your code already did this part).
> > For example, with ContentChangedEvent, we an add something about announcing, then the plugin see the event and see there is announcing info, then it will announce.
> > So if there is not such plugin added, we don't announce at all.
> > What do you thing? I'm open to disucss.
> 
> Thanks for checking Jason,
> 
> I agree with all you said. Adding the additional announce info to the content changed event may need some specific plugin array ordering, which I think may be confusing to be used. Because if there is a plugin that wanted us to announce something and it is after the new plugin, it may be missed.
> 
> Maybe we can just add a new PluginEvent for announcing and let the new plugin handle it exclussively. Do you think this could be a good way?

I'm ok to have a new event type. But I also feel ContentChangedEvent should be fine, think about this:
- AnnouncePlugin
- SomeOtherPlugin

When `SomeOtherPlugin` trigger a `ContentChangedEvent`, it is a new event so that `AnnouncePlugin` will get it, it doesn't matter where `SomeOtherPlugin` is.

packages/roosterjs-editor-types/lib/interface/AnnounceData.ts
```diff
@@ -0,0 +1,21 @@
+import type { DefaultAnnounceStrings } from '../enum/DefaultAnnounceStrings';
+
+/**
+ * Represents data, that can be used to announce text to screen reader.
+ */
+export default interface AnnounceData {
+    /**
+     * Default announce strings built in Rooster
+     */
+    defaultStrings?: DefaultAnnounceStrings;
```
JiuqingSong, line 10: Is this optional? If yes, please also add @optional in comment

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
```diff
@@ -0,0 +1,69 @@
...
+    return element;
+}
+
+function formatString(text: string | undefined, formatStrings: string[]) {
+    if (text == undefined) {
+        return text;
+    }
+
+    formatStrings.forEach((value, index) => {
+        text = text?.replace(`{${index}}`, value);
```
JiuqingSong, line 65: Question: do we want to allow `{{0}}` to render `{0}` in case we don't want it to be replaced?

Maybe no need to do now, but just keep in mind

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    element.style.clip = 'rect(0 0 0 0)';
+    element.style.clipPath = 'inset(100%)';
+    element.style.height = '1px';
+    element.style.overflow = 'hidden';
+    element.style.position = 'absolute';
+    element.style.whiteSpace = 'nowrap';
+    element.style.width = '1px';
+
+    element.ariaLive = 'assertive';
```
JiuqingSong, line 52: Consider use roosterjs API [createElement](https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs-editor-dom/lib/utils/createElement.ts)

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
```diff
@@ -0,0 +1,69 @@
+import type { AnnounceData, DefaultAnnounceStrings } from 'roosterjs-editor-types';
+
+/**
+ * @internal
+ */
+export default class AnnounceHandler {
```
JiuqingSong, line 6: Do we need to have this class? Can this be merged into the plugin?

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
```diff
@@ -0,0 +1,69 @@
...
+    private ariaLiveElement: HTMLDivElement | undefined;
+
+    constructor(
+        private readonly document: Document,
+        private stringsMap?: Map<DefaultAnnounceStrings, string>
```
JiuqingSong, line 11: Do we want to support switching language when plugin is still alive? Maybe not a hard requirement.

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
```diff
@@ -0,0 +1,69 @@
...
+    ) {}
+
+    public announce(announceData: AnnounceData) {
+        const { text, defaultStrings, formatStrings = [] } = announceData;
+        let textToAnnounce = formatString(
+            (defaultStrings && this.stringsMap?.get(defaultStrings)) || text,
```
JiuqingSong, line 17: What if a 3rd part feature also want to use announce, do they need to modify DefaultAnnounceStrings enum?

Maybe just use a string or string-enum instead, so that 3rd party code can also use its own string key, similar with `ChangeSource`

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
```diff
@@ -1,31 +1,33 @@
...
-    blockFormat,
-    commitListChains,
-    setIndentation,
-    toggleBullet,
-    toggleNumbering,
-    toggleListType,
-} from 'roosterjs-editor-api';
 import {
     Browser,
+    cacheGetEventData,
```
JiuqingSong, line 5: Is this order change generated by some extension? It looks nice, please consider add the extension to `extensions.json`

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
BryanValverdeU: In that case they can use the "text" string property from the AnnounceData instead. 
I will update the comment with this info.

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
BryanValverdeU: I wil propose to keep it simple for now. If requested we can do it but it is not needed right now.

Also since this this will be the last format place I feel that doing an additional check for {{0}} is not too relevant since after this function we will just announce the content.

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
BryanValverdeU: Done in latest commits

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
BryanValverdeU: Merged both classes as suggested

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
BryanValverdeU: I think it was done with miclo.sort-typescript-imports which is already in the file. Not quite sure why it wasnt ordered before.

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
BryanValverdeU: How do you think it would be best? 

I can think of: 
1. Adding a public function to the plugin to set the string map
2. Something similar to setZoomScale.

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnouncePlugin.ts
```diff
@@ -0,0 +1,115 @@
...
+     * @returns
+     */
+    public getAriaLiveElement() {
+        return this.ariaLiveElement;
+    }
+}
+
+function createAriaLiveElement(document: Document): HTMLDivElement {
+    const element = createElement(
+        {
```
JiuqingSong, line 89: Suggest move this part out above the class to be a const.

packages/roosterjs-editor-plugins/lib/plugins/ContentEdit/features/listFeatures.ts
JiuqingSong: I see. So let's just leave it there.

packages/roosterjs-editor-types/lib/enum/KnownAnnounceStrings.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Known announce strings
+ */
+export enum KnownAnnounceStrings {
```
JiuqingSong, line 4: Regular enum will cause additional code which can increase bundle size. We can use const enum, the our build code will general Compatible* type which is not const so that other people who don't want to use const enum can use it.

Then for all those places where this enum is used, also let them accept the Compatible* enum type and do "import type". 

You can search const enum for examples.

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
JiuqingSong: Ok, sounds good.

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnounceHandlerImpl.ts
JiuqingSong: Maybe no need to support it now. But if we want, we can do something like

```ts
constructor(..., stringMap: Map<...> | (key) => string)
```

So that we can provide a callback function to return the string.

BryanValverdeU: Sounds good, added the callback

JiuqingSong (approved):

PR #2111, opened 2023-09-27
Content Model: Move copy entity related code to copyPastePlugin

JiuqingSong:
In entity handler code, we have logic to always add text color and background color. This is only used for cut/copy. 

Now that we have clone callback in `ContentModelCopyPastePlugin`, we can also move this logic there to keep entity handler code clean.


JiuqingSong:
@BryanValverdeU I remember we add this clone and color code to fix some bug. I tested to copy entity in dark mode, it works fine to me. Do you remember if there is any other scenario we need to test?

Thanks.

BryanValverdeU:
> @BryanValverdeU I remember we add this clone and color code to fix some bug. I tested to copy entity in dark mode, it works fine to me. Do you remember if there is any other scenario we need to test?
> 
> Thanks.

Yes, the issue was when copying an entity in dark mode.
I am also not able to repro the original issue after this change.

BryanValverdeU (approved):

JiuqingSong:
> > @BryanValverdeU I remember we add this clone and color code to fix some bug. I tested to copy entity in dark mode, it works fine to me. Do you remember if there is any other scenario we need to test?
> > Thanks.
> 
> Yes, the issue was when copying an entity in dark mode. I am also not able to repro the original issue after this change.

Thanks!

PR #2119, opened 2023-10-02
Announce Plugin, add features to current plugin

BryanValverdeU:
1. Add new features to announce like on last Cell and when creating a new list item (These are already in OWA, just moving them to this plugin) plus a new parameter in the constructor to not handle them if needed or add new features if needed (mostly used for Unit test right now but could be usefull).
2. Moved the getAnnounceDataForList to dom package since it will be used in multiple places & removed the export of get Roman and alpha bullets from index.ts in rooster dom package
3. Renamed Couple of KnownAnnounceStrings to have a better name, these changes still havent been released so should not be a breaking change.
4. Add related unit tests


packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnouncePlugin.ts
```diff
@@ -65,23 +84,49 @@ export default class Announce implements EditorPlugin {
...
         }
+
+        if (ev.eventType == PluginEventType.KeyDown && this.editor) {
+            this.handleFeatures(ev, this.editor);
+        }
+    }
+
+    private handleFeatures(event: PluginKeyDownEvent, editor: IEditor) {
+        editor.runAsync(() => {
+            this.features
```
JiuqingSong, line 118: need to check disposed for editor, or use the editor passed in from the parameter of runAsync callback:

editor.runAsync((editor) => { .... editor.***;

JiuqingSong:
Is it possible to keep `getAnnounceDataForList` in plugins package since it is only used for announcing? We should be careful of the bundle size of dom package since it is widely used.

BryanValverdeU:
> Is it possible to keep `getAnnounceDataForList` in plugins package since it is only used for announcing? We should be careful of the bundle size of dom package since it is widely used.

Moved the util to plugins package as suggested.

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnouncePlugin.ts
```diff
@@ -65,23 +84,52 @@ export default class Announce implements EditorPlugin {
...
         }
+
+        if (ev.eventType == PluginEventType.KeyDown && this.editor) {
+            this.handleFeatures(ev, this.editor);
+        }
+    }
+
+    private handleFeatures(event: PluginKeyDownEvent, editorInput: IEditor) {
+        editorInput.runAsync(editor => {
+            if (!editor || editor.isDisposed()) {
```
JiuqingSong, line 118: minor: when use this editor parameter, it is guaranteed that editor must not be disposed. So no need to check.

JiuqingSong (approved):

packages/roosterjs-editor-plugins/lib/plugins/Announce/AnnouncePlugin.ts
BryanValverdeU: I see, removed this check since it is not needed. Thanks.

PR #2126, opened 2023-10-05
Paste/Insert Image according to the visible viewport

juliaroldi:
When insert or paste image, add max height/width to the image, so large images can be resized to the size of the editor.   
adaptImageSize

### DIFF
--- a/packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/formatWithContentModel.ts
+++ b/packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/formatWithContentModel.ts
@@ -42,7 +42,7 @@ export function formatWithContentModel(
         newEntities: [],
         deletedEntities: [],
         rawEvent,
-        images: [],
+        newImages: [],
     };
     let selection: DOMSelection | undefined;
 
@@ -134,14 +134,14 @@ function handleDeletedEntities(
 }
 
 function handleImages(editor: IContentModelEditor, context: FormatWithContentModelContext) {
-    if (context.images.length > 0) {
+    if (context.newImages.length > 0) {
         const viewport = editor.getVisibleViewport();
         if (viewport) {
             const { top, bottom, left, right } = viewport;
             const minMaxImageSize = 10;
             const maxWidth = Math.max(right - left, minMaxImageSize);
             const maxHeight = Math.max(bottom - top, minMaxImageSize);
-            context.images.forEach(image => {
+            context.newImages.forEach(image => {
                 image.format.maxHeight = `${maxHeight}px`;
                 image.format.maxWidth = `${maxWidth}px`;
             });
--- a/packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/formatWithContentModel.ts
+++ b/packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/formatWithContentModel.ts
@@ -42,6 +42,7 @@ export function formatWithContentModel(
         newEntities: [],
         deletedEntities: [],
         rawEvent,
+        images: [],
     };
     let selection: DOMSelection | undefined;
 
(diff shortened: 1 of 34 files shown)


packages-content-model/roosterjs-content-model-editor/lib/publicTypes/parameter/FormatWithContentModelContext.ts
```diff
@@ -35,6 +36,11 @@ export interface FormatWithContentModelContext {
      */
     readonly deletedEntities: DeletedEntity[];
 
+    /**
+     * Images inserted in the editor that needs to have their size adjusted
+     */
+    readonly images?: ContentModelImage[];
```
JiuqingSong, line 42: Can we make it required?

packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/formatWithContentModel.ts
```diff
@@ -130,3 +132,16 @@ function handleDeletedEntities(
         }
     );
 }
+
+function handleImages(editor: IContentModelEditor, context: FormatWithContentModelContext) {
+    const viewport = editor.getVisibleViewport();
```
JiuqingSong, line 137: Move this line inside "if" since we don't need it if there is no image

packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/formatWithContentModel.ts
```diff
@@ -130,3 +132,16 @@ function handleDeletedEntities(
...
+    if (viewport && context.images) {
```
JiuqingSong, line 138: Also check images.length > 0.

For empty array, no need to do anything

packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/formatWithContentModel.ts
```diff
@@ -130,3 +132,16 @@ function handleDeletedEntities(
...
+        const { top, bottom, left, right } = viewport;
+        const maxWidth = right - left;
```
JiuqingSong, line 140: Better have a minimum value such as 10, to avoid it to be 0 or negitve number

JiuqingSong: const minMaxImageSize = 10;

maxWidth = Math.max(right-left, minMaxImageSize)

packages-content-model/roosterjs-content-model-editor/lib/publicTypes/parameter/FormatWithContentModelContext.ts
```diff
@@ -35,6 +36,11 @@ export interface FormatWithContentModelContext {
      */
     readonly deletedEntities: DeletedEntity[];
 
+    /**
+     * Images inserted in the editor that needs to have their size adjusted
+     */
+    readonly images: ContentModelImage[];
```
JiuqingSong, line 42: Suggest name it "newImages". Because we don't put existing images into it.

packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/formatWithContentModel.ts
```diff
@@ -130,3 +132,19 @@ function handleDeletedEntities(
         }
     );
 }
+
+function handleImages(editor: IContentModelEditor, context: FormatWithContentModelContext) {
+    if (context.images.length > 0) {
+        const viewport = editor.getVisibleViewport();
```
BryanValverdeU, line 138: Re running getVisibleViewport could have high compute use if used for multiple images, since it runs getBoundingClientRect for the editor div for every image, 
I wonder if we could cache the value if it is used once when formatting, maybe in the context.
@juliaroldi @JiuqingSong do you know if this is a big concern or if caching the value should be a good fix?

JiuqingSong (approved):

PR #2127, opened 2023-10-05
Standalone editor: Remove more dependencies

JiuqingSong:
- moveChildNodes: Create a copy in `roosterjs-content-model-dom` package
- createRange: Use native `document.createRange` instead since it is not very commonly used
- Position and NodePosition: Stop using them, but only use its member node and offset directly
- arrayPush: Use '...' instead

### DIFF
--- a/packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
+++ b/packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
@@ -8,11 +8,12 @@ import type {
 } from 'roosterjs-content-model-types';
 import type { FormatWithContentModelContext } from '../../publicTypes/parameter/FormatWithContentModelContext';
 import type { IContentModelEditor } from '../../publicTypes/IContentModelEditor';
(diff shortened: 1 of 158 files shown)


packages-content-model/roosterjs-content-model-editor/lib/publicApi/utils/paste.ts
```diff
@@ -67,7 +67,6 @@ export default function paste(
             } = triggerPluginEventAndCreatePasteFragment(
                 editor,
                 clipboardData,
-                null /* position */,
```
JiuqingSong, line 70: @BryanValverdeU  want to check with you why this parameter is always passed as null, and it is ok I remove it?

BryanValverdeU: I think it can be removed. Just need to pass null to the inner util that is still used in the normal editor.

JiuqingSong: Sounds good, thanks.

BryanValverdeU (approved):

PR #2128, opened 2023-10-05
Recreate Content Model for table and image selection

JiuqingSong:
With advanced cache enabled in Content Model editor, when we start a table selection, table id can be modified. When that happens, we need to update cached model to reflect this change.

For now the easiest way is to clear cache and rebuild content model when new selection is table/image.

### DIFF
--- a/packages-content-model/roosterjs-content-model-editor/lib/editor/utils/contentModelDomIndexer.ts
+++ b/packages-content-model/roosterjs-content-model-editor/lib/editor/utils/contentModelDomIndexer.ts
@@ -41,12 +41,6 @@ function isIndexedSegment(node: Node): node is IndexedSegmentNode {
     );
 }
 
-function isIndexedTable(element: HTMLTableElement): element is IndexedTableElement {
-    const { tableRows } = (element as IndexedTableElement).__roosterjsContentModel ?? {};
-
-    return Array.isArray(tableRows) && tableRows.every(row => Array.isArray(row.cells));
-}
-
 function onSegment(
     segmentNode: Node,
     paragraph: ContentModelParagraph,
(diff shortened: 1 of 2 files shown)


juliaroldi (approved):

juliaroldi:
How table/image id can be changed during compose? Are we changing this this because of particular bug?

JiuqingSong:
> How table/image id can be changed during compose? Are we changing this this because of particular bug?

https://github.com/microsoft/roosterjs/blob/b6f69616f7b14266ef6347bd856c8e7a34da8970/packages/roosterjs-editor-core/lib/coreApi/selectTable.ts#L40

If the selected image or table does not have id (it is the common case), we add id to it when select.

juliaroldi:
> > How table/image id can be changed during compose? Are we changing this this because of particular bug?
> 
> https://github.com/microsoft/roosterjs/blob/b6f69616f7b14266ef6347bd856c8e7a34da8970/packages/roosterjs-editor-core/lib/coreApi/selectTable.ts#L40
> 
> If the selected image or table does not have id (it is the common case), we add id to it when select.

I understood by the PR description that if the table/image that has id might have it modified.

JiuqingSong:
> > > How table/image id can be changed during compose? Are we changing this this because of particular bug?
> > 
> > 
> > https://github.com/microsoft/roosterjs/blob/b6f69616f7b14266ef6347bd856c8e7a34da8970/packages/roosterjs-editor-core/lib/coreApi/selectTable.ts#L40
> > 
> > If the selected image or table does not have id (it is the common case), we add id to it when select.
> 
> I understood by the PR description that if the table/image that has id might have it modified.

If there is already id, and it is unique, then it won't be changed. In theory we can check whether it is changed, but it need to go through the whole model to find the table, so just clear the cached model should be easier. This is not ideal, but it should fix a blocker issue for enabling cache. Later I may find a better way.

PR #2133, opened 2023-10-09
Trigger apply changed when mouse up

juliaroldi:
Added a new parameters `applyChangesOnMouseUp`. If this parameter is true, when editing an image the last changes will be applied when mouse up event occurs. `applyChangesOnMouseUp` will be false by default.


packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
```diff
@@ -179,6 +180,11 @@ export default class ImageEdit implements EditorPlugin {
                     e.preventDefault();
                 }
             },
+            mouseup: e => {
```
JiuqingSong, line 183: we already have mouse up event handler at line 185

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/editInfoUtils/applyChange.ts
```diff
@@ -57,6 +57,10 @@ export default function applyChange(
             newSrc,
         });
         newSrc = event.newSrc;
+    } else {
```
JiuqingSong, line 60: check the new parameter as well.

packages/roosterjs-editor-plugins/lib/plugins/ImageEdit/ImageEdit.ts
juliaroldi: The handler in the switch case is not triggered when the mouse up event happens in the image wrapper.

JiuqingSong: I see. In that case, we should only attach this event handler when image wrapper is showing, and detach it when hide.

Also please try if we can attach the event to the wrapper element only, but not the whole window.

juliaroldi: Unless the user always have the cursor on the handles during the whole rotate and resize process, the event will not be triggered. Sometimes the cursor can move out from the handles. 

resizeAndRotate

JiuqingSong: Try pass true as the 3rd parameter when addEventListener, see if that can fix this problem:

```ts
node.addEventListener('mouseup', handler, true /*useCapture*/);
```

If that works, also need to pass true when remove event listener.

JiuqingSong (approved):

PR #2138, opened 2023-10-11
Add font weight to format state

juliaroldi:
Add font weight to format state. 
fontWeightProperty


packages/roosterjs-editor-types/lib/interface/FormatState.ts
```diff
@@ -174,6 +174,11 @@ export interface StyleBasedFormatState {
      * Direction of the element ('ltr' or 'rtl')
      */
     direction?: string;
+
+    /**
+     * Font weight
+     */
+    fontWeight?: string;
```
JiuqingSong, line 181: Can you help me understand why we need this? We already have isBold in FormatState.

juliaroldi: it is a request for a partner team.

juliaroldi: They need semibold font-weight.

JiuqingSong: I see. Since we modified FormatState and it is also shared with Content Model editor, we also need to let content model version of getFormatState returns this value. Otherwise other people see there is a fontWeight defined in FormatState but it never has value from Content model.

JiuqingSong (approved):

PR #2146, opened 2023-10-13
Add Table Borders basic functionality

Andres-CT98:
### Description:

Implement the ability to modify table borders.

### Fix

* Add Table border functions to modify and apply changes to table borders to selected cells.

The apply function takes the cell selection and applies the given operation accordingly. The border needs to be applied to the selected cells as well as adjacent ones (perimeter) which share the same border to ensure format visibility and persistence in situations such as copy.

* Modified Demo site to Include Table Borders functionality: Buttons to set color, width, and style and apply it to selected cells.

Storage of the table border format is made as a state in MainPaneBaseState.

* Add metadata to handle modified borders.

Any border modification will result in the whole cell being marked as modified.

### Warnings and implications

- Only the All Borders functionality is currently implemented


demo/scripts/controls/ribbonButtons/contentModel/tableBorderColorButton.ts
```diff
@@ -0,0 +1,29 @@
...
+export const tableBorderColorButton: RibbonButton<'buttonNameTableBorderColor'> = {
+    ...originalButton,
+    unlocalizedText: 'Table Border Color',
+    iconName: 'ColorSolid',
+    isDisabled: formatState => !formatState.isInTable,
+    onClick: (editor, key) => {
+        // This check will always be true, add it here just to satisfy compiler
+        if (key != 'buttonNameTableBorderColor' && isContentModelEditor(editor)) {
+            editor.setTableBorderColor(
+                editor.isDarkMode()
```
JiuqingSong, line 22: For content model we should be able to always specify light mode color only, and let content model handle dark mode color.

packages-content-model/roosterjs-content-model-editor/lib/publicApi/table/editTable.ts
```diff
@@ -30,9 +32,12 @@ import {
  * @param editor The editor instance
  * @param operation The table operation to apply
  */
-export default function editTable(editor: IContentModelEditor, operation: TableOperation) {
+export default function editTable(
+    editor: IContentModelEditor,
+    operation: TableOperation,
+    border?: Border
```
JiuqingSong, line 38: All other editing operations don't have parameter, but border needs parameter. So suggest split border into a different API, no need to merge into editTable.

JiuqingSong: Similar with set table cell shade color.

packages-content-model/roosterjs-content-model-editor/lib/publicTypes/ContentModelEditorCore.ts
```diff
@@ -1,5 +1,5 @@
 import type { ContentModelPluginState } from './pluginState/ContentModelPluginState';
-import type { CoreApiMap, EditorCore } from 'roosterjs-editor-types';
+import type { Border, CoreApiMap, EditorCore } from 'roosterjs-editor-types';
```
JiuqingSong, line 2: For the new types that is only used by Content Model, declare it in `roosterjs-content-model-types` rather than `roosterjs-editor-types`.

Content Model will be moved out and totally get rid of old editor code.

demo/scripts/controls/ribbonButtons/contentModel/tableBorderColorButton.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export const tableBorderColorButton: RibbonButton<'buttonNameTableBorderColor'> = {
+    ...originalButton,
+    unlocalizedText: 'Table Border Color',
+    iconName: 'ColorSolid',
+    isDisabled: formatState => !formatState.isInTable,
+    onClick: (editor, key) => {
+        // This check will always be true, add it here just to satisfy compiler
+        if (key != 'buttonNameTableBorderColor' && isContentModelEditor(editor)) {
+            editor.setTableBorderColor(
```
JiuqingSong, line 21: What is this function for? Is it to change border color of current selected table (if any) or just set different table border color?

packages/roosterjs-editor-types/lib/interface/EditorOptions.ts
```diff
@@ -145,6 +145,11 @@ export default interface EditorOptions {
      */
     imageSelectionBorderColor?: string;
 
+    /**
+     * Initial Table Border Format
+     */
+    initialTableBorder?: Border;
```
JiuqingSong, line 151: what is this value for?

packages-ui/roosterjs-react/lib/contextMenu/index.ts
```diff
@@ -15,4 +15,5 @@ export {
     TableEditShadeMenuItemStringKey,
     TableEditMenuItemStringKey,
     TableEditAlignTableMenuItemStringKey,
+    TableBorder_unfinished,
```
JiuqingSong, line 18: Currently roosterjs-react is not supporting Content Model yet. So if possible, declare any UI related thing in demo site code only. Later we will do dedicate work on port roosterjs-react to use Content Model.

packages-content-model/roosterjs-content-model-editor/lib/publicTypes/ContentModelEditorCore.ts
Andres-CT98: That would mean I need to import a content model type for MainPaneBase, which I currently do by mistake. If something does not depend on content model it should not import things from it.

packages/roosterjs-editor-types/lib/interface/EditorOptions.ts
Andres-CT98: The Initial Table Border Format passed to the editor.

demo/scripts/controls/ribbonButtons/contentModel/tableBorderColorButton.ts
Andres-CT98: Just set different border color. The function that applies changes is named applyBorderFormat.

JiuqingSong: Can you elebrate more about the funcationality of this function?

Assume I have two tables, and one is selected. Now I call this function to set to a different border color, then what is the expected result?

packages/roosterjs-editor-types/lib/interface/EditorOptions.ts
JiuqingSong: Why editor need to know the initial table border color?

Andres-CT98: It is an optional parameter, same as imageSelectionBorderColor, to allow customization of what will be the initial format if you were to apply table borders.

demo/scripts/controls/ribbonButtons/contentModel/tableBorderColorButton.ts
Andres-CT98: Internally the editor core will change the border color that will be applied by applyBorderFormat if called. No changes to any table in the editor.

JiuqingSong: Then if I call applyBorderFormat, will both table be formatted, or just one?

packages/roosterjs-editor-types/lib/interface/EditorOptions.ts
JiuqingSong: Oh, do you mean it is allowed that user format table border without specifying a format?

I think this should be part of the UI layer, but not the roosterjs editor layer.

For example, in roosterjs we should always require a valid set of parameters of the border when do border format. Then in UI layer (here it means OWA/Monarch) it can store the default settings somewhere, and when format table border, pass the stored default value to roosterjs. In that case, roosterjs doesn't need to know the concept of default border format at all.

Please refer to the text color and background color button in OWA, they have the similar behavior.

demo/scripts/controls/ribbonButtons/contentModel/tableBorderColorButton.ts
Andres-CT98: Only what is selected will be formatted.

packages/roosterjs-editor-types/lib/interface/EditorOptions.ts
Andres-CT98: Table borders should not be allowed to apply without a format. If the user does not specify one on editor creation, a default one will be set in createContentModelEditorCore.ts.

I do not think the current standard of undefined for Text color or background color are right, but even those are also at the editor lever through defaultFormat. Why would this one not be?

demo/scripts/controls/ribbonButtons/contentModel/tableBorderColorButton.ts
JiuqingSong: In that case, why not just format the table directly, and no need to store the format inside editor.

packages/roosterjs-editor-types/lib/interface/EditorOptions.ts
JiuqingSong: I understand what you mean. So there are two different features.

About the text default format, it is the format for text even before user do any format change. Editor still need to format it as the default format, so editor needs to remember this format.

For table border, I agree there needs to be a default format, and actually there is one inside insertTable.ts. User can change to use a different format when insertTable is called. But since it is only used when user insert a table or format a table, there is entrypoint to pass the format in, not like the default text format. So that editor does not need to remember table border format.

demo/scripts/controls/ribbonButtons/contentModel/tableBorderColorButton.ts
Andres-CT98: Because that is the requirement of the function. There needs to be an editor-wide format, independent of any table, that only gets applied though the Apply Borders functions, currently only All Borders is implemented.

packages-content-model/roosterjs-content-model-editor/lib/modelApi/table/applyTableFormat.ts
```diff
@@ -71,14 +72,20 @@ function clearCache(rows: ContentModelTableRow[]) {
...
+    };
 
     rows.forEach(row => {
         const bgColorOverrides: boolean[] = [];
         const vAlignOverrides: boolean[] = [];
+        const borderOverrides: boolean[] = [];
 
         overrides.bgColorOverrides.push(bgColorOverrides);
         overrides.vAlignOverrides.push(vAlignOverrides);
+        overrides.borderOverrides?.push(borderOverrides);
```
JiuqingSong, line 88: no need to add "?" here

packages-content-model/roosterjs-content-model-editor/lib/publicApi/table/applyTableBorderFormat.ts
```diff
@@ -0,0 +1,154 @@
...
+                if (Perimeter.Right && sel.lastCol + 1 < tableModel.rows[0].cells.length) {
+                    for (let rowIndex = sel.firstRow; rowIndex <= sel.lastRow; rowIndex++) {
+                        const cell = tableModel.rows[rowIndex].cells[sel.lastCol + 1];
+                        applySingleBorderFormat(cell, borderFormat, 'borderLeft');
+                    }
+                }
+            }
+
+            // Apply format changes
+            applyTableFormat(tableModel, undefined /*newFormat*/, true /*keepCellShade*/);
```
JiuqingSong, line 122: why do we need to call applyTableFormat here?

packages/roosterjs-editor-dom/lib/table/applyTableFormat.ts
```diff
@@ -55,6 +55,18 @@ function hasValign(cell: VCell) {
     return !!getTableCellMetadata(cell.td)?.vAlignOverride;
 }
 
+/**
+ * Check if the cell has any modified border
+ * @param cell
+ * @returns
+ */
+function hasBorder(cell: VCell) {
```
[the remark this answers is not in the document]
Andres-CT98: Should this also not apply to all others?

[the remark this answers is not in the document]
JiuqingSong: I'm ok to change all of them.

packages-content-model/roosterjs-content-model-editor/lib/publicTypes/enum/BorderOperations.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * All Border operations
+ */
+export const enum BorderOperations {
+    AllBorders,
```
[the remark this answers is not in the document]
Andres-CT98: It has a comment. Do you mean a comment for each operation?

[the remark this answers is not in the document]
JiuqingSong: yes, for those public types, let's add comment for each item, so our document generation tool can generate document for them.

packages-content-model/roosterjs-content-model-editor/lib/publicApi/table/applyTableBorderFormat.ts
```diff
@@ -0,0 +1,149 @@
...
+                Bottom: false,
+                Left: false,
+                Right: false,
+            };
+
+            // Create border format with table format as backup
+            let borderFormat = '';
+            const format = tableModel.format;
+            const { width, style, color } = border;
+            const extractedBorder = extractBorderValues(format.borderTop);
```
JiuqingSong, line 39: I just realized that this format is from table, and we only check its top border now. What if table cell has different border than table, and what if the borders in a cell are different, e.g. top vs bottom?

JiuqingSong: Can we just force the type `Border` must have all properties have value? Then we just need to read the value from the parameter, and no need to read from current table. Caller code can decide what value to pass in.

For example, if user has specified border color, but didn't specify style and width, caller code pass in the specified color together with default width and style? Then in this function we can always assume parameter already passes in enough information.

Andres-CT98: It is just a fallback value. Forcing changes in Border affects Image Borders too.

Andres-CT98: I could make the Border type Required<Border> for the function.

PR #2150, opened 2023-10-16
Remove margin block on list 

juliaroldi:
Set margin-block to 0px to make the style the same as Office Apps. 
margins1
gifMargin2


packages-content-model/roosterjs-content-model-types/lib/format/formatParts/MarginBlockFormat.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Format of margin-block
+ */
+export type MarginBlockFormat = {
```
JiuqingSong, line 4: What will happen if both margin-block-start and margin-top are specified with different values?

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -205,7 +206,11 @@ export default class VList {
             }
 
             item.writeBack(listStack, this.rootList, shouldReuseAllAncestorListElements);
-            const topList = listStack[1];
+            const topList = listStack[1] as HTMLElement;
+            if (topList) {
+                topList.style.marginBlockStart = '0px';
```
JiuqingSong, line 211: What about an existing list that has margins before and after? Will this cause the margins go away suddenly?

juliaroldi:
> Use the snapshot below, then press TAB key, the existing margin will disappear.
> 
> ```
> <ol><li>aa</li></ol>
> ```
 I added a check to only trigger in the first list and not in the list chain

packages/roosterjs-editor-dom/lib/list/VList.ts
```diff
@@ -182,7 +182,11 @@ export default class VList {
      * @param shouldReuseAllAncestorListElements Optional - defaults to false.
      * @param disableListChain Whether we want to disable list chain functionality. @default false
      */
-    writeBack(shouldReuseAllAncestorListElements?: boolean, disableListChain?: boolean) {
+    writeBack(
+        shouldReuseAllAncestorListElements?: boolean,
+        disableListChain?: boolean,
+        newList?: boolean
```
JiuqingSong, line 188: I'm always trying to avoid this kind of parameter. It is hard to say whether it is a new list from the caller side.

Can we just check if it is a new list within our code? For exmaple, if there is already a OL/UL and we reuse it, it should not be new list. Otherwise we can treat it as new?

JiuqingSong:
When I type "1. " to start an auto list, it still has margins

JiuqingSong:
> Use the snapshot below, then press TAB key, the existing margin will disappear.
> 
> ```
> <ol><li>aa</li></ol>
> ```

Now this issue is back.

JiuqingSong:
I think we can't let VList to decide if it should add margin or not. VList should just keep whatever the list has, but don't change it.

What we need to change is the places where we create a list, such as using toggle list API, change list type API, and auto list. They know there is a new list.

For example, here is one entrypoint we should change https://github.com/microsoft/roosterjs/blob/dd3d6910e6763dc0e22124185c336d1b57878fe3/packages/roosterjs-editor-api/lib/utils/toggleListType.ts#L53C1-L53C1

When create new list, we can let VList know this list should not have top/bottom margin. So that when write back, it just apply whatever it has. And when create a VList from existing list element, just read the style and store it in VList.

Also need to take care the split existing list scenario, make sure the splitted list has the same style with original one.

JiuqingSong:
Use the following snapshot, press Enter, the margin is gone.

```html
<div>aa</div><ol><li>bb</li></ol><div>cc</div><!--{"type":0,"isDarkMode":false,"start":[1,0,0,2],"end":[1,0,0,2]}-->
```

JiuqingSong:
Once you fixes the issue, please consider add test cases for all the cases above.

JiuqingSong:
Type "*" and space, it starts a new bullet list and there is still margin. Do you also want to fix this case?

JiuqingSong (approved):

PR #2166, opened 2023-10-24
Fix #236416

JiuqingSong:
https://outlookweb.visualstudio.com/Outlook%20Web/_queries/edit/236416/?triage=true

The root cause is delimiterProcessor didn't handle selection well if the start/endContainer is set to text under delimiter, then the offset won't match. And now emoji is wrapped with entity, so that if a selection range is ended right before an entity, it is not treated as end so everything after are still treated as selected.

For an entity delimiter, we actually don't need to care if selection offset matches, but just check if the selection container is contained by the delimiter element, then we can treat it as the right place, and do proper handling.

Fix by using Node.contains() to check if we should start/end the selection in a delimiter.


JiuqingSong:
> If we do not treat emoji as entity, do we need this change?

yes. Emoji as entity is just a trigger of the bug. The problem is the code to handle selection has the bug.

juliaroldi (approved):

PR #2168, opened 2023-10-26
Add image size checkmarks to context menu

ianeli1:
Adds checkmarks to the "Size" menu inside context menu

image

For small, medium, and original sizes, `isResizedTo` is used to check the sizing. This function has been modified to allow for resized images with odd sizes, such as 271px, as this would get divided in half and then rounded up, while HTML would have the number divided in half and rounded down, which would cause issues.

For best fit, we check if "width" and "height" attributes are set.


packages-ui/roosterjs-react/lib/contextMenu/menus/createImageEditMenuProvider.tsx
```diff
@@ -188,6 +203,14 @@ function shouldShowImageEditItems(editor: IEditor, node: Node) {
     return safeInstanceOf(node, 'HTMLImageElement') && node.isContentEditable;
 }
 
+function getImageFromTarget(target: Node) {
```
juliaroldi, line 206: which scenario triggers this code? Maybe we should consider add it to image selection code and relay only in 'getSelectionRangeEx'

JiuqingSong: For roosterjs-react, I think it is ok to only support image selection.

packages-ui/roosterjs-react/lib/contextMenu/types/ContextMenuItem.ts
```diff
@@ -42,6 +42,8 @@ export default interface ContextMenuItem<TString extends string, TContext = neve
      */
     shouldShow?: (editor: IEditor, targetNode: Node, context?: TContext) => boolean;
 
+    getSelectedId?: (editor: IEditor, targetNode: Node) => TString | null;
```
JiuqingSong, line 45: Please add comment

packages-ui/roosterjs-react/lib/contextMenu/menus/createImageEditMenuProvider.tsx
```diff
@@ -77,6 +78,20 @@ const ImageResizeMenuItem: ContextMenuItem<ImageEditMenuItemStringKey, ImageEdit
             }
         });
     },
+    getSelectedId: (editor, target) => {
```
juliaroldi, line 81: I see you are passing the node parameter for this function in the interface. Should we use the node her instead getting the selection?

packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
```diff
@@ -54,7 +59,8 @@ export default class ImageSelection implements EditorPlugin {
                     if (
                         mouseSelection &&
                         mouseSelection.type === SelectionRangeTypes.ImageSelection &&
-                        mouseSelection.image !== mouseTarget
+                        (!safeInstanceOf(mouseTarget, 'HTMLElement') ||
+                            !mouseTarget.contains(mouseSelection.image))
```
juliaroldi, line 63: What happens if the mouseTarget has more content than just image the image?

ianeli1: reverted

packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
```diff
@@ -43,7 +43,7 @@ export default class ImageSelection implements EditorPlugin {
                     if (
                         safeInstanceOf(target, 'HTMLImageElement') &&
                         target.isContentEditable &&
-                        event.rawEvent.button === mouseLeftButton
+                        event.rawEvent.button != mouseMiddleButton
```
juliaroldi, line 46: do we need this change?

packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
```diff
@@ -233,11 +233,15 @@ export default class DOMEventPlugin implements PluginWithState<DOMEventPluginSta
...
         const searcher = this.editor?.getContentSearcherOfCursor();
         const elementBeforeCursor = searcher?.getInlineElementBefore();
+        const selection = this.editor?.getSelectionRangeEx();
 
         let eventTargetNode = event.target as Node;
         if (event.button != 2 && elementBeforeCursor) {
             eventTargetNode = elementBeforeCursor.getContainerNode();
         }
+        if (event.button == 2 && selection?.type === SelectionRangeTypes.ImageSelection) {
+            eventTargetNode = selection.image;
```
juliaroldi, line 243: Do we need this changes?

packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
ianeli1: ImageSelection is broken on context menu without it

packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
ianeli1: Not really, but having image selection be the default provider of the node in this case makes sense. We rarely use the wrapper target on any action

packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
JiuqingSong, line 46: There is a copy of this file under content model folder: https://github.com/microsoft/roosterjs/blob/master/packages-content-model/roosterjs-content-model-adapter/lib/corePlugins/ImageSelection.ts

Would you please do the same change to the copied file? Thanks.

ianeli1: done

JiuqingSong (approved):

PR #2181, opened 2023-10-30
Fix Mouseout behavior to hide table editors

BryanValverdeU:
Right now, we attach a event listener for the mouse out event, so when mouse leaves the editor div, dismiss the editors. 
But this can cause that the table selector is hard to be used if the Editor Div has 0 margins. To fix, attach the listener to the scroll container instead.

Before
tableSelectorMouseMoveBefore

After
tableSelectorMouseMoveAfter


packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -129,7 +132,7 @@ export default class TableResize implements EditorPlugin {
         this.tableEditor?.onMouseMove(x, y);
     };
 
-    private setTableEditor(table: HTMLTableElement | null, e?: MouseEvent) {
+    public setTableEditor(table: HTMLTableElement | null, e?: MouseEvent) {
```
JiuqingSong, line 135: Please add comment for public function

JiuqingSong (approved):

packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
BryanValverdeU: Sounds good, done

Andres-CT98 (approved):

PR #2184, opened 2023-11-01
Fix apply Table Inside borders operation and Demo site

Andres-CT98:
### Description:

Using the apply Inside borders table operation results in the wrong results for single row/column selections. Additionally, all Table border operations are unusable in the current demo site.

### Fix

The case for single cell or single row/column needs to be accounted for separately. All other cases are unaffected and already accounted for. Demo operations were fixed as well as one Table style.

### Cases tested:

1. Create a Table
2. Change the default colour and width for Table border operations
3. Select a single row or column
4. Apply Inside borders

**Before:**

wrongInsideTableBorders

**After:**

rightInsideTableBorders


demo/scripts/controls/ribbonButtons/contentModel/tableBorderApplyButton.ts
```diff
@@ -1,16 +1,20 @@
...
-const TABLE_OPERATIONS = {
-    menuNameTableAllBorder: 'AllBorders',
-    menuNameTableNoBorder: 'NoBorders',
-    menuNameTableLeftBorder: 'LeftBorders',
-    menuNameTableRightBorder: 'RightBorders',
-    menuNameTableTopBorder: 'TopBorders',
-    menuNameTableBottomBorder: 'BottomBorders',
-    menuNameTableInsideBorder: 'InsideBorders',
-    menuNameTableOutsideBorder: 'OutsideBorders',
+const TABLE_OPERATIONS: Record<string, BorderOperations> = {
```
Andres-CT98, line 9: @JiuqingSong Following up on [this comment](https://github.com/microsoft/roosterjs/pull/2146#discussion_r1364228960), Changing from enum to string literals makes is so that if the values are modified it also needs to be changed everywhere it is used.

I made TABLE_OPERATIONS a Record to prevent the issue of all operations not working, but it is still the case that changing the value of a string literal makes it harder to modify everywhere than an enum.

Just an observation.

JiuqingSong (approved):

PR #2194, opened 2023-11-07
Ensure image parent node exist, before select it. 

juliaroldi:
When select a position in Image selection plugin, check if the image parent node exist, so can it be selected.


packages/roosterjs-editor-core/lib/corePlugins/ImageSelection.ts
```diff
@@ -70,19 +70,15 @@ export default class ImageSelection implements EditorPlugin {
...
                         } else if (key === Delete) {
                             this.editor.deleteNode(keyDownSelection.image);
                             event.rawEvent.preventDefault();
-                        } else {
-                            const position = new Position(
-                                keyDownSelection.image,
-                                PositionType.Before
-                            );
-
+                        } else if (position) {
```
JiuqingSong, line 81: actually position can never be null. The problem could be there is no parent element so that when we create Position, we stored null as node:

image

Then in createRange, when we get nodeType, it throws:

image

To avoid this, we need to check if there is parent node of image. If not, it is impossible to select the image so we just do nothing.

JiuqingSong (approved):

Issue #2190, opened 2023-11-06
Add module entry to package.json for better bundling support.


dake3601:
**Describe the bug**
We are using esbuild to bundle an app that uses roosterjs packages. However, we encountered some problems with the bundling process because the roosterjs packages are missing a module entry in their package.json files. Having a module entry helps bundlers such as esbuild know the package is an ES module and find the right entry to the package, while avoid breaking changes for other consumers.

**Actual behavior**
Navigate to each roosterjs' package.json and only a main entry pointing to the lib/index.js (CommonJS) is provided. This causes esbuild to fail to bundle the packages correctly. For example: [roosterjs-react](https://github.com/microsoft/roosterjs/blob/master/packages-ui/roosterjs-react/package.json), [roosterjs](https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs/package.json), [roosterjs-editor-dom](https://github.com/microsoft/roosterjs/blob/master/packages/roosterjs-editor-dom/package.json) and others.

**Expected behavior**
Each roosterjs package that has exported `.mjs` code should have a `"module"` entry in their package.json file, pointing to the corresponding `.mjs` file. For example:
```json
{
    "name": "package-name",
    "description": "Package Description",
    "dependencies": { },
    "version": "0.0.0",
    "main": "./lib/index.js"
    "module": "./lib-mjs/index.js"
}
```

**Additional context**
We are using roosterjs' packages and using esbuild to bundle an app for a fast inner dev loop.

@JiuqingSong let me know what you think.

PR #2197, opened 2023-11-08
Adding module entry to package.json

dake3601:
Adding a module entry to roosterjs' package.json files. Having a module entry helps bundlers such as esbuild know the package is an ES module and find the right entry to the package, while avoid breaking changes for other consumers.

Fixes: #2190


JiuqingSong:
Actually we have a script to regenerate package.json files when build: https://github.com/microsoft/roosterjs/blob/528caf4984d86bc919b815eb90f4c7a86a71d58b/tools/buildTools/normalize.js#L47

Can you try change this file so that no need to change all other files.

JiuqingSong (approved):

PR #2210, opened 2023-11-15
Standalone editor: Merge DOMEventPlugin and MouseUpPlugin

JiuqingSong:
The existing core plugin `DOMEventPlugin` translate raw DOM event to roosterjs event, and `MouseUpPlugin` triggers roosterjs `MouseUp` event when mouse is up even mouse is ouside of editor as long as it was inside editor when mouseDowne event happens.

This change merges these two plugins together and move into `roosterjs-content-model-core` package to be a ported core plugin. Also add test case for the new plugin.


juliaroldi (approved):

packages-content-model/roosterjs-content-model-editor/lib/editor/createEditorCore.ts
```diff
@@ -64,9 +67,12 @@ export function createEditorCore(
         ...createStandaloneEditorDefaultSettings(options),
 
         environment: {
-            // It is ok to use global window here since the environment should always be the same for all windows in one session
             isMac: window.navigator.appVersion.indexOf('Mac') != -1,
```
BryanValverdeU, line 70: I think window can be undefined in Server side rendering like in NextJS. We had a similar report some time ago in the Browser utility in roosterjs editor dom, so we added a undefined check there. Probably not a big deal for this PR. but just an FYI

BryanValverdeU (approved):

packages-content-model/roosterjs-content-model-editor/lib/editor/createEditorCore.ts
JiuqingSong: Good point.

Browser object was initialized when the code is loaded, so it is possible to be null if do server side render.
Now in Content Model editor I'm removing the usage of Browser object, and use this environment object instead. It is initalized only when we create an editor, it is triggered by js code and should not be rendered from server side directly.

PR #2214, opened 2023-11-16
Auto format dashes

juliaroldi:
Change auto format dashes to have the same behavior of other office apps.
autoFormatPlugin


packages/roosterjs-editor-plugins/lib/plugins/AutoFormat/AutoFormat.ts
```diff
@@ -56,47 +63,84 @@ export default class AutoFormat implements EditorPlugin {
                 this.lastKeyTyped = '';
             }
 
+            const searcher = this.editor.getContentSearcherOfCursor(event);
```
JiuqingSong, line 66: It seems this will run for all key press, no matter what key is. I'm a little concerned about the perf.

PR #2219, opened 2023-11-18
Standalone Editor: Port LifecyclePlugin

JiuqingSong:
Port core plugin `LifecyclePlugin` into standalone editor with the following changes:

1. Remove a bunch of shadow edit related properties that are not used any more:
   - shadowEditEntities
   - shadowEditSelectionPath
   - shadowEditTableSelectionPath
   - shadowEditImageSelectionPath
2. Move `customData` from `LifecyclePluginState` into `ContentModelEditorCore` since we don't want to keep supporting this in standalone editor
3. When init editor, we now accept a Content Model instead of HTML string to set initial content. If no init content model is passed in, we create an empty one with a paragraph and default format. So that we can remove the call to `ensureTypeInContainer` from editor ctor.
4. StandaloneEditor now accept default format in `ContentModelSegmentFormat` instead of `DefaultFormat`. The wrapper ContentModelEditor class is now doing all those option translate work.


packages-content-model/roosterjs-content-model-core/lib/corePlugin/LifecyclePlugin.ts
```diff
@@ -0,0 +1,205 @@
...
+     */
+    dispose() {
+        this.editor?.triggerPluginEvent(PluginEventType.BeforeDispose, {}, true /*broadcast*/);
+
+        if (this.disposer) {
+            this.disposer();
+            this.disposer = null;
+            this.initializer = null;
+        }
+
```
BryanValverdeU, line 115: Just curious, do we need to dispose the initialModel and the state?

BryanValverdeU (approved):

packages-content-model/roosterjs-content-model-core/lib/corePlugin/LifecyclePlugin.ts
JiuqingSong: state only has some simple boolean values, so there is nothing to dispose.
For initialModel, although it is possible large, but it is still just a JSON object without any external links, so it should be fine not dispose it. But I think I can overwrite it using an empty model to make sure cutting off all the links.

PR #2227, opened 2023-11-29
Auto format - Auto format hyphen

juliaroldi:
Change auto format hyphen to transform dashes into hyphen when they are between spaces. 
autoFormat


JiuqingSong (approved):

JiuqingSong:
I just realized that we have the code to check text before in existing code. Since this change does not introduce additional perf overhead, I'm ok with it for now. Later when we port it to use Content Model, we may need to rethink about the whole flow.

PR #2228, opened 2023-11-29
Standalone Editor: Selection API step 1: Add a SelectionPlugin

JiuqingSong:
The general goal is to remove the old selection related core API (`select`, `selectRange`, `selectImage`, `selectTable`, `getSelectionRange`, `getSelectionRangeEx`), and use the selection API (`setDOMSelection`, `getDOMSelection`) instead, and use a new core Plugin `SelectionPlugin` to handle all selection related events.

This is the first step of porting selection API. It creates a new `SelectionPlugin`, and move selection related things into `SelectionPluginState`.


juliaroldi (approved):

JiuqingSong:
> Will SelectionPlugin replace TableCellSelection and ImageSelectionPlugin in the future?

yes

PR #2233, opened 2023-11-30
Allow each package has its own version when publish

JiuqingSong:
This change allows us add a version override in versions.json to each package so when we do version bump, no need to bump up those packages that is not changed.

For example
```js
{
    "packages": "8.59.0",
    "packages-ui": "8.54.0",
    "packages-content-model": "0.20.0",
    "overrides": {
        "roosterjs-editor-plugins": "8.59.1"
    }
}
```

With this config, all other roosterjs packages are still in 8.59.0, only `roosterjs-editor-plugins` will be bumped to 8.59.1.

### DIFF
--- a/tools/buildTools/normalize.js
+++ b/tools/buildTools/normalize.js
@@ -19,7 +19,7 @@ function normalize() {
 
     allPackages.forEach(packageName => {
         const versionKey = findPackageRoot(packageName);
-        const version = versions[versionKey];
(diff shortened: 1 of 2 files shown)


juliaroldi (approved):

haven2world:
I'm just concerned that if that's the case, it will be harder to find versions that all packages match with each other. How can we restrict the override to only the patched version?

PR #2248, opened 2023-12-07
Remove whole list when delete

juliaroldi:
When cut the whole list from the editor, remove all elements. 
Before:
ListBug
After:
listFixed

### DIFF
--- a/packages-content-model/roosterjs-content-model-core/lib/publicApi/selection/deleteSelection.ts
+++ b/packages-content-model/roosterjs-content-model-core/lib/publicApi/selection/deleteSelection.ts
@@ -1,5 +1,9 @@
+import hasSelectionInBlock from './hasSelectionInBlock';
+import hasSelectionInBlockGroup from './hasSelectionInBlockGroup';
 import { deleteExpandedSelection } from '../../modelApi/edit/deleteExpandedSelection';
+import { getClosestAncestorBlockGroupIndex } from '../model/getClosestAncestorBlockGroupIndex';
 import type {
+    ContentModelBlock,
     ContentModelDocument,
     DeleteSelectionContext,
     DeleteSelectionResult,
@@ -13,12 +17,14 @@ import type {
  * @param model The model to delete selected content from
  * @param additionalSteps @optional Addition delete steps
  * @param formatContext @optional A context object provided by formatContentModel API
+ * @param isCut @optional True if this is a cut operation, false if this is a delete operation
  * @returns A DeleteSelectionResult object to specify the deletion result
  */
 export function deleteSelection(
     model: ContentModelDocument,
     additionalSteps: (DeleteSelectionStep | null)[] = [],
-    formatContext?: FormatWithContentModelContext
+    formatContext?: FormatWithContentModelContext,
+    isCut?: boolean
 ): DeleteSelectionResult {
     const context = deleteExpandedSelection(model, formatContext);
 
@@ -33,6 +39,7 @@ export function deleteSelection(
     });
 
     mergeParagraphAfterDelete(context);
+    deleteEmptyList(context, isCut);
 
     return context;
 }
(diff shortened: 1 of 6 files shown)


packages-content-model/roosterjs-content-model-core/lib/publicApi/selection/deleteSelection.ts
```diff
@@ -13,12 +17,14 @@ import type {
...
  * @param formatContext @optional A context object provided by formatContentModel API
+ * @param isCut @optional True if this is a cut operation, false if this is a delete operation
  * @returns A DeleteSelectionResult object to specify the deletion result
  */
 export function deleteSelection(
     model: ContentModelDocument,
     additionalSteps: (DeleteSelectionStep | null)[] = [],
-    formatContext?: FormatWithContentModelContext
+    formatContext?: FormatWithContentModelContext,
+    isCut?: boolean
```
JiuqingSong, line 27: deleteSelection is a very common API and is used in a lot of places, so it should not know anything about copy/cut.

Take one step back, in what case we will delete selection but want to keep an empty list number?

juliaroldi: When we merge models, when we paste content into a list. With this code, the list will be removed. Maybe we can should add this code after into normalize the model?

JiuqingSong: In that case, you can wrap the delete list step as a DeleteSelectionStep function, and pass to the `additionalSteps` parameter (the 2nd parameter of deleteSelection), no need to introduce a new parameter. Then from the cut code, pass that step in.

JiuqingSong: I mean only when cut, pass in this additional step. For all others, keep no change.

packages-content-model/roosterjs-content-model-core/lib/publicApi/selection/deleteSelection.ts
```diff
@@ -59,3 +66,46 @@ function mergeParagraphAfterDelete(context: DeleteSelectionContext) {
...
+                    : undefined;
+            const nextBlock =
+                listItemIndex > -1
+                    ? insertPoint.path[index + 1].blocks[listItemIndex + 1]
+                    : undefined;
+            if (
+                hasSelectionInBlockGroup(item) &&
+                (!previousBlock || hasSelectionInBlock(previousBlock)) &&
+                nextBlock &&
+                isEmptyBlock(nextBlock)
```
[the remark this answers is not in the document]
juliaroldi: We check if the next block is empty to check if it was also deleted in the selection.

packages-content-model/roosterjs-content-model-core/lib/corePlugin/utils/deleteEmptyList.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    return !!block;
+}
+
+/**
+ * @internal
+ * If the first item o the list is selected in a expanded selection, we need to remove the list item levels
+ * @param context A context object provided by formatContentModel API
+ */
+export function deleteEmptyList(context: DeleteSelectionContext) {
```
JiuqingSong, line 25: Declare this function like this:

```ts
export const deleteEmptyList: DeleteSelectionStep = (context) => {
...
}
```

packages-content-model/roosterjs-content-model-core/lib/publicApi/selection/deleteSelection.ts
```diff
@@ -13,27 +14,28 @@ import type {
...
  * @param formatContext @optional A context object provided by formatContentModel API
+ * @param additionalStepsResult  @optional The delete result to trigger the additional steps @default 'notDeleted'
  * @returns A DeleteSelectionResult object to specify the deletion result
  */
 export function deleteSelection(
     model: ContentModelDocument,
     additionalSteps: (DeleteSelectionStep | null)[] = [],
-    formatContext?: FormatWithContentModelContext
+    formatContext?: FormatWithContentModelContext,
+    additionalStepsResult: DeleteResult = 'notDeleted'
```
JiuqingSong, line 24: I can understand you need tot execute the step when delete result is range. But this parameter is not easy to understand.

I think we can remove the check at line 32, and move it into all existing delete steps to do this check there. So that you can check the delete result to be 'range' in your function directly.

packages-content-model/roosterjs-content-model-core/lib/corePlugin/utils/deleteEmptyList.ts
```diff
@@ -0,0 +1,54 @@
...
+                    : undefined;
+            const nextBlock =
+                listItemIndex > -1
+                    ? insertPoint.path[index + 1].blocks[listItemIndex + 1]
+                    : undefined;
+            if (
+                hasSelectionInBlockGroup(item) &&
+                (!previousBlock || hasSelectionInBlock(previousBlock)) &&
+                nextBlock &&
+                isEmptyBlock(nextBlock)
```
JiuqingSong, line 48: Still, why nextBlock needs to be empty?

JiuqingSong (approved):

PR #2255, opened 2023-12-11
Create list in a indented paragraph

juliaroldi:
Pass the style alignment and margin from the paragraph to the list item. When toggle list, pass the style from the list item to the paragraph. 

listgif1
listgif2
listgif3


packages-content-model/roosterjs-content-model-api/lib/modelApi/list/setListType.ts
```diff
@@ -37,19 +37,16 @@ export function setListType(model: ContentModelDocument, listType: 'OL' | 'UL')
...
-
-                    if (block.format.marginRight) {
-                        x.format.marginRight = block.format.marginRight;
-                    }
-
-                    if (block.format.textAlign) {
-                        x.format.textAlign = block.format.textAlign;
-                    }
-                });
+                const listBlock = block.blocks[0];
```
JiuqingSong, line 40: Move this line one line up. So that the original line 39 can also use this `listBlock`.

And add comment to explain why we need to copy these styles.

JiuqingSong (approved):

PR #2264, opened 2023-12-12
Copy cell dataset when inserting new row/column

Andres-CT98:
### Description:

When inserting a new column or row using the Content model Table Insert functions, the dataset of the cell is not copied. This results in the loss of shading, vertical alignment, and borders if they were changed from default. Using the Inserter handles does a full copy of the cell, so the problem is not present there.

### Fix

The dataset is now copied when creating the new cells to insert.

### Cases tested:

1. Insert a table
2. Set new borders for the first row and column
3. Insert new row and column the first cell by using the Content model Insert function, not the Inserter handle or the context menu

**Before:** 
image

**After:**
image


juliaroldi (approved):

packages-content-model/roosterjs-content-model-dom/lib/modelApi/creators/createTableCell.ts
```diff
@@ -27,6 +29,6 @@ export function createTableCell(
         spanLeft,
         spanAbove,
         isHeader: !!isHeader,
-        dataset: {},
+        dataset: dataset ?? {},
```
JiuqingSong, line 32: clone this object rather than just directly assign it:

dataset: {...dataset}

Otherwise two cells will share the same dataset object, modifying one of them will cause the other one also got changed.

JiuqingSong:
Please add test cases for the new logic.

You can add test for createTableCell in creatorsTest.ts

packages-content-model/roosterjs-content-model-dom/lib/modelApi/creators/createTableCell.ts
```diff
@@ -27,6 +29,6 @@ export function createTableCell(
         spanLeft,
         spanAbove,
         isHeader: !!isHeader,
-        dataset: {},
+        dataset: { ...dataset } ?? {},
```
JiuqingSong, line 32: Actually, just do "{...dataset}" is ok, no need to add " ?? {}" since it must not be null.

JiuqingSong (approved):

PR #2271, opened 2023-12-13
Delete Image format

juliaroldi:
**Problem**: Selection marker was coping the image format after deletion, causing other inserted image to have the same styles of the removed image. 
Before delete: 
Screenshot 2023-12-13 at 14 12 08
After delete: 
Screenshot 2023-12-13 at 14 12 28

**Solution**: Before create a new marker when delete, verify if the segment is not a image. If the segment is image, do not copy the image format to the selection marker.


packages-content-model/roosterjs-content-model-core/lib/modelApi/edit/deleteExpandedSelection.ts
```diff
@@ -128,6 +130,10 @@ export function deleteExpandedSelection(
     return context;
 }
 
+function getMarkerFormat(segment: ContentModelSegment): ContentModelSegmentFormat | undefined {
+    return segment.segmentType !== 'Image' ? segment.format : undefined;
```
JiuqingSong, line 134: Instead of remove all format from imageFormat, we should keep those format for text.

I have a function added to do this but the code is not ready for review yet. You can refactor it and add in your PR now and make it public.

https://github.com/microsoft/roosterjs/blob/ea972ffb8449e828fd071658c52b787043b92a67/packages-content-model/roosterjs-content-model-core/lib/publicApi/model/paste.ts#L112

packages-content-model/roosterjs-content-model-core/lib/publicApi/domUtils/getCurrentSegmentTextFormat.ts
```diff
@@ -0,0 +1,29 @@
+import type { ContentModelSegment, ContentModelSegmentFormat } from 'roosterjs-content-model-types';
+
+/**
+ * Get the text format of the current selected segment
+ * @param currentSegment The segment to get format from
+ * @returns
+ */
+export function getCurrentSegmentTextFormat(
```
JiuqingSong, line 8: We are actually getting format from a given segment, so you can remove the word "Current" from function name and parameter name.

packages-content-model/roosterjs-content-model-core/lib/publicApi/domUtils/getCurrentSegmentTextFormat.ts
```diff
@@ -0,0 +1,29 @@
+import type { ContentModelSegment, ContentModelSegmentFormat } from 'roosterjs-content-model-types';
+
+/**
+ * Get the text format of the current selected segment
```
JiuqingSong, line 4: Please update the comment to explain what is the difference between the segment format and the result of this function

packages-content-model/roosterjs-content-model-core/lib/publicApi/domUtils/getCurrentSegmentTextFormat.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    const textFormat: Record<string, string | undefined> = {
+        fontFamily,
+        fontSize,
+        textColor,
+        backgroundColor,
+        letterSpacing,
+        lineHeight,
+    };
+    const format: Record<string, string | undefined> = {};
```
JiuqingSong, line 22: You can use type ContentModelSegmentFormat here

JiuqingSong (approved):

PR #2275, opened 2023-12-18
Move Paste from publicApi to coreApi to leverage the same domToModelOptions

BryanValverdeU:
The problem:
Paste Code is not respecting the domToModelOptions that are passed in when instantiating the editor. 

Change:
Move the paste code from publicApi to coreApi to be able to access the defaultDomToModelOptions that are stored in the StandaloneCoreEditor Object.


juliaroldi (approved):

packages-content-model/roosterjs-content-model-core/lib/coreApi/paste.ts
```diff
@@ -176,12 +186,22 @@ function createBeforePasteEventData(
...
+        fragment: core.contentDiv.ownerDocument.createDocumentFragment(),
         sanitizingOption: options,
         htmlBefore: '',
         htmlAfter: '',
         htmlAttributes: {},
-        domToModelOption: {},
+        domToModelOption: Object.assign(
+            {},
+            ...[
+                core.defaultDomToModelOptions,
```
JiuqingSong, line 197: I have a change that contains a similar fix with this one. https://github.com/microsoft/roosterjs/tree/u/jisong/pasteapi3
I feel it is ok to move paste back to be core API. Just two problems
1. It is better to make the public paste API accept `PasteType` as parameter rather than "isImage", "isText", ... . We can solve this when we build the real standalone editor since for now it will be conflict with IEditor interface.
2. Here we override the options to use editor's default options, which includes an override for Table, then we override it back to default one. I think we should just keep the options passed in by user, and do not add the table processor override at all here. To do that, we can save those override options separately in StandaloneEditorCore. E.g.:
```ts
interface StandaloneEditorCore {
  ...
  editingDomToModelOptions: DomToModelOptions; // For the table override
  customizedDomToModelOptions: DomToModelOptions; // For options passed in from editorOptions
}
```

Another option is to remove the tablePreProcessor at all.

BryanValverdeU: If you have a change already doing this, I think we can just wait for your PR then. Closing this for now.

JiuqingSong (approved):

PR #2276, opened 2023-12-18
Remove gif conversion

juliaroldi:
Remove the code that turns gif into a png after editing a gif, since this code is not necessary anymore.


JiuqingSong (approved):

juliaroldi:
> Question: how is GIF working today?

if we rotate or cropped, the gif is turned into png. I'm removing this code, because it does not have any impact in OWA/Monarch, since we cannot access the file extension to do the conversion. Also, removing this code, we mitigate a issue of the gif becoming transparent in popout.

PR #2286, opened 2023-12-19
Change borderLeft/Right to borderInlineStart/End

Andres-CT98:
### Description:

The `border-left` and `border-right` css properties have issues when changing to RTL mode. This is due to the `direction` property mirroring the table, and some borders looking flipped to the user. For example:

Before: 
image

After: 
image

Here since the columns got mirrored, the borders look out of place, but the css is the same. There are two identified issues:

Issue 1 - Applying borders to a RTL table behaves incorrectly
Issue 2 - Changing the direction of a table using 'setDirection' causes the borders to look incorrect

### Fix

For Issue 1, the Table border controls used to apply changes now work as expected. The directions listed on the function are applied independent of the `direction`. Changes were made in `applyTableBorderFormat` to accommodate the `direction` of the table.

For Issue 2, changes were made in `setModelDirection` to flip all side borders in the case of a whole table direction switch. Only cells with modified borders are flipped for optimisation.

Test cases were added for both fixes.

### Cases tested:

Issue 1

1. Set the editor to RTL mode
2. Create a 3x3 table
3. Chose a wider border width
4. Appy 'All borders' to the central column

Issue 2

1. Set the editor to LTR mode
2. Create a 3x3 table
3. Chose a wider border width
4. Appy 'All borders' to the central column
5. Select the whole table and change its direction to RTL

**Before:**
image

**After:**
image

### Tried Approaches:

Changed the css attribute `border-left` to `border-inline-start` and `border-right` to `border-inline-end` in tables. This is due to those new properties adapting to the set `direction`. This approach is effective at solving Issue 2 arts of Issue 1, but the 'borderInlineStart/End" properties are not supported to the extent needed. The approach could be revisited at a much later date.

Using the 'borderFormatHandler' was also not viable, since Issue 1 was unrelated to it and for Issue 2 the handler is not aware of the previous `direction` set, so a contextual flip would be impossible.

### Warnings and implications

- There are still issues with Table templates


packages-content-model/roosterjs-content-model-dom/lib/formatHandlers/common/borderFormatHandler.ts
```diff
@@ -9,6 +9,7 @@ export const BorderKeys: (keyof BorderFormat & keyof CSSStyleDeclaration)[] = [
     'borderRight',
     'borderBottom',
     'borderLeft',
+    'borderRadius',
```
JiuqingSong, line 12: Merging mistake?

packages-content-model/roosterjs-content-model-types/lib/format/formatParts/BorderFormat.ts
```diff
@@ -26,4 +26,16 @@ export type BorderFormat = {
      * Radius to be applied in all borders corners
      */
     borderRadius?: string;
+
+    /**
+     * Inline Start border in format 'width style color'
+     * Left border in Left to right languages, right border in right to left languages
+     */
+    borderInlineStart?: string;
```
JiuqingSong, line 34: consider remove original properties `borderLeft` and `borderRight` to keep "single source of truth"

JiuqingSong:
Please double check if the new styles can be supported by Win 32 Outlook. For example:
1. create new message,
2. insert table, and set border in RTL
3. send
4. Open the message in Win32 Outlook
Check if the border is correctly rendered.

packages/roosterjs-editor-dom/lib/table/applyTableFormat.ts
```diff
@@ -148,8 +148,8 @@ function setCellFormat(
  */
 function setBorderColors(td: HTMLTableCellElement, format: Partial<TableFormat>) {
     td.style.borderTop = getBorderStyle(format.topBorderColor);
-    td.style.borderLeft = getBorderStyle(format.verticalBorderColor);
-    td.style.borderRight = getBorderStyle(format.verticalBorderColor);
+    td.style.borderInlineStart = getBorderStyle(format.verticalBorderColor);
```
JiuqingSong, line 151: Suggest we fix content model editor only since bumping version of old package will cause additional cost but we don't see any benefit from it.

The old editor code will be deprecated.

packages-content-model/roosterjs-content-model-types/lib/format/formatParts/BorderFormat.ts
juliaroldi: We are using `borderLeft` and `borderRight`, if you remove this properties, please test the impact in the image border feature.

JiuqingSong: I just did some test, and I found that borderLeft/Right and inlineBorderStart/End can have different values.

When they are conflict, inlineBorderStart/End will overwrite borderLeft/Right when render
When they are not conflict, both take effect.

This cause a problem that if we use inlineBorderStand/End, we also need to take care of borderLeft/Right, which means if there is already borderLeft/Right, when we set inlineBorderStart/End, it may not fully overwrite the original one. For example:

```html
<div style="border-left: solid 1px red; border-inline-start: solid 10px red; direction: rtl">test</div>
```

Because this DIV is in RTL, border-inline-start and border-left mean different thing, so we will see two borders on left and right. To solve this, we need to know if the DIV is in LTR or RTL when get/set border. That makes the benefit of using inline-border-start/end are gone.

In fact, in some format APIs, we are reading direction and do different handling for direction, for example: https://github.com/microsoft/roosterjs/blob/9aafcb65241bf936cc0cf12dbbb1f4ab5c827547/packages-content-model/roosterjs-content-model-api/lib/modelApi/block/setModelIndentation.ts#L53

And for some other cases, we are doing a translate of the property value in format handler to let the property not direction-sensitive, for example: https://github.com/microsoft/roosterjs/blob/9aafcb65241bf936cc0cf12dbbb1f4ab5c827547/packages-content-model/roosterjs-content-model-dom/lib/formatHandlers/utils/dir.ts#L30

The second one looks more elegant, the down side is it makes the related format need to be combined with DirectionFormat: https://github.com/microsoft/roosterjs/blob/9aafcb65241bf936cc0cf12dbbb1f4ab5c827547/packages-content-model/roosterjs-content-model-dom/lib/formatHandlers/block/textAlignFormatHandler.ts#L9

image

@Andres-CT98 I think either way above can work here. What do you think?

Andres-CT98: How does the example div happen in the editor? Using our Insert table functions all tables should have the correct side border properties, and we could handle paste cases. If the example div is willingly inserted on the editor as HTML then the behaviour is as expected.

packages/roosterjs-editor-dom/lib/table/applyTableFormat.ts
Andres-CT98: The issue is also present in the old version, but I agree that since it will be deprecated it is better to do it only in content model.``

packages-content-model/roosterjs-content-model-types/lib/format/formatParts/BorderFormat.ts
JiuqingSong: Possibly load existing HTML from draft.

Andres-CT98: Ok, I think I can make the change in borderFormatHandler, however I do not know how it works or when is it called. Ideally all instances of borderLeft/Right would be changed accordingly.

There is also a recent change with border widths that I have to change as well, but I could not find out why that change was made.

JiuqingSong: `*FormatHandler.apply` is called when we write Content Model back into editor DOM.

For example, when there is a DIV with border-left to be red, and direction is ltr, we read them and store the value as
```ts
// borderFormatHandler.parse
format.borderStart = 'red'
```

Then in a format API we change its direction to rtl, and border-start to green:

```ts
format.direction = 'rtl';
format.borderStart = 'gren';
```

Then we write it back, the borderFormatHandler.apply will be called and we will write border-right instead of border-left since we know we are in rtl now:
```ts
// borderFormatHandler.apply
if (format.direction == 'rtl') {
   element.style.borderRight = format.borderStart;
   element.style.borderLeft = format.borderEnd;
}  else {
   element.style.borderLeft = format.borderStart;
   element.style.borderRight = format.borderEnd;
}
```

That makes the direction is totally separated from format API. In format API we only need to care borderStart/End, no need to know if it is ltr or rtl. They will always be handled by formatHandler.

Andres-CT98: After trying the handler approach, I realised that the issue was both in the application of the borders, and with the RTL direction change itself.

packages-content-model/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
```diff
@@ -54,12 +65,32 @@ function internalSetDirection(format: ContentModelBlockFormat, direction: 'ltr'
         setProperty(format, 'marginRight', marginLeft);
         setProperty(format, 'paddingLeft', format.paddingRight);
         setProperty(format, 'paddingRight', paddingLeft);
+
+        // If whole Table direction changed, flip cell side borders
+        if (block && block.blockType == 'Table') {
+            block.rows.forEach(row => {
+                row.cells.forEach(cell => {
+                    // Optimise by skipping cells with unchaged borders
```
JiuqingSong, line 73: typo: "unchanged".

packages-content-model/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
```diff
@@ -54,12 +65,32 @@ function internalSetDirection(format: ContentModelBlockFormat, direction: 'ltr'
...
+                    updateTableCellMetadata(cell, metadaata => {
+                        if (metadaata?.borderOverride) {
```
JiuqingSong, line 75: typo: "metadata"

packages-content-model/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
```diff
@@ -54,12 +65,32 @@ function internalSetDirection(format: ContentModelBlockFormat, direction: 'ltr'
...
+                            setProperty(cell.format, 'borderLeft', cell.format.borderRight);
+                            setProperty(cell.format, 'borderRight', storeBorderLeft);
+                        }
+                        return metadaata;
+                    });
+                });
+            });
+
+            // Apply changed borders
+            applyTableFormat(block, undefined, true);
```
JiuqingSong, line 86: Please add inline format for the const parameters, for example
applyTableFormat(block, undefined /*....*/, true /*...*/)

packages-content-model/roosterjs-content-model-api/test/modelApi/block/setModelDirectionTest.ts
```diff
@@ -2,15 +2,25 @@ import { ContentModelDocument } from 'roosterjs-content-model-types';
...
+        expectedResult: boolean,
+        tableTest?: boolean
     ) {
         const result = setModelDirection(model, direction);
 
         expect(result).toBe(expectedResult);
+
+        if (tableTest) {
+            model.blocks[0]['dataset'] = {};
+        }
```
JiuqingSong, line 23: please fix build. If it is expected to be "any" type, you can add a type declaration like:
model.block[0]['dataset'] = {} as any;

packages-content-model/roosterjs-content-model-api/test/modelApi/block/setModelDirectionTest.ts
```diff
@@ -60,7 +70,7 @@ describe('setModelDirection', () => {
         );
     });
 
-    it('set direction for divider', () => {
+    xit('set direction for divider', () => {
```
JiuqingSong, line 73: any reason to skip this test case?

JiuqingSong (approved):

JiuqingSong (approved):

PR #2287, opened 2023-12-19
Adjust Selection on Cut/Copy first table cell 

BryanValverdeU:
The problem: 
When selecting the first cell of the table and copy/paste, the whole table cell is copied. 

Fix
Adjust the selection for the model on copy/cut so if the selection marker is before the table, move it to the first paragraph position, so we do not copy the cell and only copy the selected text

Before
CopyCellBefore

After 
CopyCellAfter

[Bug 243714](https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/243714): [Table] The cells are also pasted after pasting the text in cells


packages-content-model/roosterjs-content-model-core/lib/corePlugin/ContentModelCopyPastePlugin.ts
```diff
@@ -250,6 +256,45 @@ class ContentModelCopyPastePlugin implements PluginWithState<CopyPastePluginStat
...
+            }
+            return true;
+        }
+
+        if (
+            !selectionMarker &&
+            segments?.length == 1 &&
+            segments[0].segmentType == 'SelectionMarker'
+        ) {
+            if (tableContext) {
```
JiuqingSong, line 287: do we need this check?

What if there is a table nested in another table, the select content from the first cell of inner table?

packages-content-model/roosterjs-content-model-core/lib/corePlugin/ContentModelCopyPastePlugin.ts
```diff
@@ -250,6 +256,45 @@ class ContentModelCopyPastePlugin implements PluginWithState<CopyPastePluginStat
...
+    iterateSelections(pasteModel, (_, tableContext, block, segments) => {
+        if (selectionMarker) {
+            if (
+                tableContext &&
+                block &&
+                firstBlock?.blockType == 'Paragraph' &&
+                firstBlock.segments.includes(selectionMarker) &&
+                block.blockType == 'Paragraph'
+            ) {
+                block.segments?.unshift(selectionMarker);
```
JiuqingSong, line 276: I think we don't need to insert the selection marker into the table.

packages-content-model/roosterjs-content-model-core/lib/corePlugin/ContentModelCopyPastePlugin.ts
BryanValverdeU: Good point, removing this to handle this case.

packages-content-model/roosterjs-content-model-core/lib/corePlugin/ContentModelCopyPastePlugin.ts
BryanValverdeU: You are right, removed this line.

packages-content-model/roosterjs-content-model-core/lib/corePlugin/ContentModelCopyPastePlugin.ts
```diff
@@ -250,6 +256,41 @@ class ContentModelCopyPastePlugin implements PluginWithState<CopyPastePluginStat
...
+ */
+export function adjustSelectionForCopyCut(pasteModel: ContentModelDocument) {
+    let selectionMarker: ContentModelSelectionMarker | undefined;
+    let firstBlock: ContentModelBlock | undefined;
+
+    iterateSelections(pasteModel, (_, tableContext, block, segments) => {
+        if (selectionMarker) {
+            if (
+                tableContext &&
+                block &&
```
JiuqingSong, line 271: Maybe we don't event need to check block here. Even block doesn't exist, we can still do the same handling.

packages-content-model/roosterjs-content-model-core/lib/corePlugin/ContentModelCopyPastePlugin.ts
```diff
@@ -250,6 +256,41 @@ class ContentModelCopyPastePlugin implements PluginWithState<CopyPastePluginStat
     };
 }
 
+/**
+ * @internal
+ * Exported only for unit testing
+ */
+export function adjustSelectionForCopyCut(pasteModel: ContentModelDocument) {
+    let selectionMarker: ContentModelSelectionMarker | undefined;
+    let firstBlock: ContentModelBlock | undefined;
```
JiuqingSong, line 265: You can declare firstBlock as ContentModelParagraph and only set is value when the block is Paragraph. Then no need to check at line 272

packages-content-model/roosterjs-content-model-core/lib/corePlugin/ContentModelCopyPastePlugin.ts
```diff
@@ -250,6 +256,41 @@ class ContentModelCopyPastePlugin implements PluginWithState<CopyPastePluginStat
...
+
+    iterateSelections(pasteModel, (_, tableContext, block, segments) => {
+        if (selectionMarker) {
+            if (
+                tableContext &&
```
JiuqingSong, line 270: Not sure if it is a valid scenario but I feel we should also need to save the first tableContext, then just need to check this tableContext is different than the first one. And tableContext can be null here. That will cover the case like:

```html
<table><tr><td>
text [SelectionMarker]
</td></tr></table>
[Selected: text.... ]
```

JiuqingSong: Consider add test case for this case

BryanValverdeU: Ok, I think we can do that to.

packages-content-model/roosterjs-content-model-core/lib/corePlugin/ContentModelCopyPastePlugin.ts
BryanValverdeU: Agreed, removed that

packages-content-model/roosterjs-content-model-core/lib/corePlugin/ContentModelCopyPastePlugin.ts
BryanValverdeU: Sounds good, but I think I will need to move the same check, because the callback block is ContentModelBlock, so to type check we still need to verify that the blockType is equal to Paragraph.

JiuqingSong (approved):

PR #2288, opened 2023-12-20
Remove format of  trailing space segment

juliaroldi:
**Problem 1**: When adding link to a text with trailing spaces, the trailing space is also receiving link styles. 
linkWithSpacesBug

**Problem 2**: When double clicking a text segment with trailing space and applying underline, the trailing space also receives the underline style. 
underline bug

**Fix**: Adjust selection to split the text from the trailing and remove the style from trailing space. 
linkWithTrailingSpace
underlineFix

**Model before insert link:** 

Screenshot 2023-12-20 at 16 39 53

**Model after insert link:** 
Screenshot 2023-12-20 at 16 40 54


packages-content-model/roosterjs-content-model-api/lib/modelApi/selection/adjustTrailingSpaceSelection.ts
```diff
@@ -0,0 +1,65 @@
...
+                    segment.isSelected &&
+                    segment.segmentType == 'Text' &&
+                    hasTrailingSpace(segment.text) &&
+                    !isTrailingSpace(segment.text)
+                ) {
+                    splitTextSegment(block.segments, segment, index);
+                }
+            });
+        }
+        return true;
```
BryanValverdeU, line 27: In a multiple paragraph selection, if we return true here, we would only fix the first paragraph. Should we instead return false to continue iterating the next paragraphs?

juliaroldi: good point!!

packages-content-model/roosterjs-content-model-api/lib/modelApi/selection/adjustTrailingSpaceSelection.ts
```diff
@@ -0,0 +1,63 @@
...
+    ContentModelSegment,
+    ContentModelText,
+} from 'roosterjs-content-model-types';
+
+/**
+ * @internal
+ */
+export function adjustTrailingSpaceSelection(model: ContentModelDocument) {
+    iterateSelections(model, (_, __, block) => {
+        if (block?.blockType == 'Paragraph' && block.segments.length == 1) {
```
JiuqingSong, line 14: If the selected paragraph has multiple segments, we don't handle it?

packages-content-model/roosterjs-content-model-api/lib/modelApi/selection/adjustTrailingSpaceSelection.ts
```diff
@@ -0,0 +1,63 @@
...
+            const segment = block.segments[0];
+            if (
+                segment.isSelected &&
```
JiuqingSong, line 17: You can get selected segments from the fourth parameter of this callback

packages-content-model/roosterjs-content-model-api/lib/modelApi/selection/adjustTrailingSpaceSelection.ts
```diff
@@ -0,0 +1,63 @@
...
+                splitTextSegment(block.segments, segment);
+            }
+        }
+
+        return false;
+    });
+}
+
+function hasTrailingSpace(text: string) {
+    return text.length > 0 && text.trimRight().length < text.length;
```
JiuqingSong, line 31: how about "text.trimRight() != text"

packages-content-model/roosterjs-content-model-api/lib/modelApi/selection/adjustTrailingSpaceSelection.ts
```diff
@@ -0,0 +1,63 @@
...
+    const text = textSegment.text.trimRight();
+    const trailingSpace = textSegment.text.substring(text.length);
+    const newText = createText(text, textSegment.format, textSegment.link, textSegment.code);
+    newText.isSelected = true;
+    const trailingSpaceLink = textSegment.link
+        ? {
+              ...textSegment.link,
+              format: {
+                  ...textSegment.link?.format,
+                  underline: false,
```
JiuqingSong, line 51: not sure if this is ok. What if we are turning off underline? Maybe that can never happen?

At least we need some comment here.

packages-content-model/roosterjs-content-model-api/lib/modelApi/selection/adjustTrailingSpaceSelection.ts
```diff
@@ -0,0 +1,63 @@
...
+    segments: ContentModelSegment[],
+    textSegment: Readonly<ContentModelText>
+) {
+    const text = textSegment.text.trimRight();
+    const trailingSpace = textSegment.text.substring(text.length);
+    const newText = createText(text, textSegment.format, textSegment.link, textSegment.code);
+    newText.isSelected = true;
+    const trailingSpaceLink = textSegment.link
+        ? {
+              ...textSegment.link,
```
JiuqingSong, line 48: So that means we still let this space segment have link, but don't show underline? Is that expected?

packages-content-model/roosterjs-content-model-api/lib/modelApi/selection/adjustTrailingSpaceSelection.ts
```diff
@@ -0,0 +1,63 @@
...
+import type {
+    ContentModelDocument,
+    ContentModelSegment,
+    ContentModelText,
+} from 'roosterjs-content-model-types';
+
+/**
+ * @internal
+ */
+export function adjustTrailingSpaceSelection(model: ContentModelDocument) {
```
JiuqingSong, line 12: Given that this function is just for adjusting space selection after insert link or toggle underline, it is not a very common function. We need to add comment to explain it. So that in case other people want to use this function in the future, they need to understand what this function is doing.

packages-content-model/roosterjs-content-model-api/lib/modelApi/selection/adjustTrailingSpaceSelection.ts
juliaroldi: The bug usually happens when the user does double click in a segment with trailing spaces. I made a change to cover more scenarios

packages-content-model/roosterjs-content-model-api/lib/modelApi/selection/adjustTrailingSpaceSelection.ts
juliaroldi: that is what word does. Do you think we should remove everything?

packages-content-model/roosterjs-content-model-api/lib/modelApi/selection/adjustTrailingSpaceSelection.ts
juliaroldi: If we turn off underline, it will change the segment format and not the link

packages-content-model/roosterjs-content-model-api/lib/modelApi/selection/adjustTrailingSpaceSelection.ts
JiuqingSong: Ok, we can follow Word's behavior.

JiuqingSong (approved):

PR #2301, opened 2023-12-28
[Mac] Do not handle enter input on mac

juliaroldi:
**Problem:** Asians characters relay on enter key to the transformation on mac, so it selects the text range before to the conversion, it will make content model handle the enter input wrongly, removing the asians character. 
demosite-japanese-bug

**Fix:** Do not handle not collapsed ranges in mac. 
japanse-fix

### DIFF
--- a/packages-content-model/roosterjs-content-model-plugins/lib/edit/keyboardInput.ts
+++ b/packages-content-model/roosterjs-content-model-plugins/lib/edit/keyboardInput.ts
@@ -8,9 +8,8 @@ import type { DOMSelection } from 'roosterjs-content-model-types';
  */
 export function keyboardInput(editor: IContentModelEditor, rawEvent: KeyboardEvent) {
     const selection = editor.getDOMSelection();
-    const isMac = editor.getEnvironment().isMac;
 
-    if (shouldInputWithContentModel(selection, rawEvent, isMac)) {
+    if (shouldInputWithContentModel(selection, rawEvent, editor.isInIME())) {
         editor.takeSnapshot();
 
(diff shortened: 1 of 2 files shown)


packages-content-model/roosterjs-content-model-plugins/lib/edit/keyboardInput.ts
```diff
@@ -44,14 +45,18 @@ export function keyboardInput(editor: IContentModelEditor, rawEvent: KeyboardEve
...
+    isMac?: boolean
+) {
     if (!selection) {
         return false; // Nothing to delete
     } else if (
         !isModifierKey(rawEvent) &&
         (rawEvent.key == 'Enter' || rawEvent.key == 'Space' || rawEvent.key.length == 1)
     ) {
-        return selection.type != 'range' || !selection.range.collapsed; // TODO: Also handle Enter key even selection is collapsed
+        return selection.type != 'range' || (!selection.range.collapsed && !isMac); // TODO: Also handle Enter key even selection is collapsed
```
JiuqingSong, line 59: Thanks for handling this quickly.

Actually, can you try checking editor.isInIME() or rawEvent.isComposing instead? We still need to handle the ENTER key for expanded range when not using IME.

You can refer to here https://github.com/microsoft/roosterjs/blob/236d0afdfb5f94f80151abaa02040a82d49197de/packages-content-model/roosterjs-content-model-core/lib/corePlugin/ContentModelFormatPlugin.ts#L93

JiuqingSong (approved):

PR #2304, opened 2024-01-04
Content Model: Improve paste and sanitization behavior

JiuqingSong:
1. When paste, delete width from DIV (keep consistent with original behavior)
2. Add parameters to allow customize CSS style and HTML attribute behavior when sanitize element for paste
3. To better handle "white-space: pre" when paste, create a `pasteTextProcessor` to replace continuous spaces to `&nbsp;`
4. Fix and add test cases


packages-content-model/roosterjs-content-model-core/lib/override/pasteGeneralProcessor.ts
```diff
@@ -1,12 +1,22 @@
...
+import type {
+    DomToModelOptionForPaste,
+    ElementProcessor,
+    ValueSanitizer,
+} from 'roosterjs-content-model-types';
+
+/**
+ * @internal Export for test only
+ */
+export const removeDisplayFlex: ValueSanitizer = value => {
```
juliaroldi, line 12: Why do we remove display:flex? Does remove it can break the formatting of the copied fragment?

JiuqingSong: It was from this change: https://github.com/microsoft/roosterjs/pull/2031

When paste with display:flex, it may cause some unexpected editing behavior. For example, with horizontal flex display, moving cursor Up/Down will cause cursor move to the next container (on right side) rather than the one below it. So we decide to remove this style.

Although there is still some other debate about this, for now I'll still keep this behavior, until we have a better solution.

juliaroldi (approved):

packages-content-model/roosterjs-content-model-core/lib/utils/paste/mergePasteContent.ts
```diff
@@ -45,12 +47,16 @@ export function mergePasteContent(
...
             processorOverride: {
+                '#text': pasteTextProcessor,
                 entity: createPasteEntityProcessor(domToModelOption),
                 '*': createPasteGeneralProcessor(domToModelOption),
             },
             formatParserOverride: {
                 display: pasteDisplayFormatParser,
             },
+            additionalFormatParsers: {
+                container: [containerWidthFormatParser],
```
BryanValverdeU, line 58: Question: in the Paste plugin we add some parsers that are always added to the domToModel options.

Should we move those parsers here? I can do that if needed in another PR.

BryanValverdeU (approved):

packages-content-model/roosterjs-content-model-core/lib/utils/paste/mergePasteContent.ts
JiuqingSong: Paste Plugin is an optional plugin. We use this plugin when we want to handle content copied from some special known sources such as Word. 

Here it is some common behavior that even without PastePlugin, they will still take effect.

Also, PastePlugin has the equal rights to other 3rd plugins, so the behavior of parsers added in PastePlugin may overwrite parsers added by other plugins, and is also able to be overwritten by other plugins, depends on the plugin order. Here it is the default behavior so it will not overwrite any other plugin's behavior, it can always be overwritten.

PR #2306, opened 2024-01-04
[Content Model] Add text align justify option 

juliaroldi:
Include a feature that allows users to align text as justified, as well as ensuring that when a list item is aligned, the text within the list item is also aligned.
JustifyAlignText
multipleListItemAlign

### DIFF
--- a/packages-content-model/roosterjs-content-model-api/lib/modelApi/block/setModelAlignment.ts
+++ b/packages-content-model/roosterjs-content-model-api/lib/modelApi/block/setModelAlignment.ts
@@ -7,8 +7,8 @@ import type {
 } from 'roosterjs-content-model-types';
 
 const ResultMap: Record<
-    'left' | 'center' | 'right',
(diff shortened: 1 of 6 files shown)


packages-content-model/roosterjs-content-model-api/lib/modelApi/block/setModelAlignment.ts
```diff
@@ -56,15 +56,24 @@ export function setModelAlignment(
     );
 
     paragraphOrListItemOrTable.forEach(({ block }) => {
-        const newAligment = ResultMap[alignment][block.format.direction == 'rtl' ? 'rtl' : 'ltr'];
-        if (block.blockType === 'Table') {
+        const newAlignment =
+            alignment === 'justify'
```
JiuqingSong, line 60: Please update the value map at line 27, so that no need to have a separate check for 'justify' here.

We should make the code follow the same flow as much as possible.

JiuqingSong (approved):

Issue #2305, opened 2024-01-04
[Content Model] The style on SelectionMarker is not up to date after setFontSize more than one time


haven2world:
**Describe the bug**
If user set font size twice, the `fontSize` format on `SelectionMarker` won't be updated to the latest state. But if `applyDefaultFormat` in `ContentModelFormatPlugin` is called, it will use the style on the current selectoin marker to generate the new pending format and lose the latest font size style. 

**To Reproduce**
1. Open demo site under content model
2. set font size to 8 pt
3. set font size to 12 pt
4. check the format on selection marker
5. trigger `applyDefaultFormat` (in my case, the plugin received a `KeyDown` event with `ProcessKey`)

**Expected behavior**
The format on selection marker should be updated to the latest so we can always generate correct pending format

**Screenshots**
<img width="1106" alt="image" src="https://github.com/microsoft/roosterjs/assets/17299418/a5faa2f0-7da6-481b-a84e-ea5b312a5c83">

**Device Information**
 - OS: iOS
 - Browser WKWebView

haven2world:
<img width="1112" alt="image" src="https://github.com/microsoft/roosterjs/assets/17299418/5dbb1c54-a4fe-4f6a-b1a6-535de9f00297">
Do you think this fix makes sense?

JiuqingSong:
Can you mark that part that you have changed? Feel free just submit a PR so I can take a deeper look.

Thanks.

haven2world:
I created the PR #2309

JiuqingSong:
@haven2world  Please verify if https://github.com/microsoft/roosterjs/pull/2331 can fix this problem. Thanks.

PR is not fully done yet, I still need to fix test case and add more test cases

haven2world:
@JiuqingSong  it works like a charm!

PR #2309, opened 2024-01-05
Content Model #2305 Update format on selection marker in formatSegmentWithContentModel

haven2world:
As mentioned in #2305,If user set font size twice, the `fontSize` format on `SelectionMarker` won't be updated to the latest state. But if `applyDefaultFormat` in `ContentModelFormatPlugin` is called, it will use the style on the current selectoin marker to generate the new pending format and lose the latest font size style. 

To resolve this issue, when there is a pending format, and the selection is collapsed, we need to also update the format on selection marker.

### DIFF
--- a/packages-content-model/roosterjs-content-model-api/lib/publicApi/utils/formatSegmentWithContentModel.ts
+++ b/packages-content-model/roosterjs-content-model-api/lib/publicApi/utils/formatSegmentWithContentModel.ts
@@ -59,6 +59,15 @@ export function formatSegmentWithContentModel(
                 ? [[pendingFormat, null, null]]
                 : segmentAndParagraphs.map(item => [item[0].format, item[0], item[1]]);
 
+            //  If there is a pending format and the selection is collapsed, we need to apply the format to the selection marker either
+           if (pendingFormat && isCollapsedSelection) {
+                formatsAndSegments.push([
+                    segmentAndParagraphs[0][0].format,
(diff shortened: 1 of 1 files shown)


packages-content-model/roosterjs-content-model-api/lib/publicApi/utils/formatSegmentWithContentModel.ts
```diff
@@ -59,6 +59,15 @@ export function formatSegmentWithContentModel(
                 ? [[pendingFormat, null, null]]
                 : segmentAndParagraphs.map(item => [item[0].format, item[0], item[1]]);
 
+            //  If there is a pending format and the selection is collapsed, we need to apply the format to the selection marker either
+           if (pendingFormat && isCollapsedSelection) {
```
JiuqingSong, line 63: Now I understand what you mean. I think a better fix should be when creating Content Model, directly set the format of SelectionMarker to be the pending one if it matches current pending format selection. Then all further cases where selection marker format is used will apply this result.

I can make a fix later. Or if you don't want to wait, please go ahead to make the change.

haven2world: I'm not in a hurry, since we cannot upgrade roosterjs for now. If you think this change makes sense, I can wait for your fix, and overrides the `formatSegmentWithContentModel` in our side as a workaround.

PR #2314, opened 2024-01-08
Override TextColorParser on paste to not handle variable based styles

BryanValverdeU:
Override TextColorParser on paste to not handle variable based styles, since these styles are usually not provided in the clipboard. And instead just use the container text color

### DIFF
--- a/packages-content-model/roosterjs-content-model-core/lib/override/pasteTextColorFormatParser.ts
+++ b/packages-content-model/roosterjs-content-model-core/lib/override/pasteTextColorFormatParser.ts
@@ -28,7 +28,8 @@ function isVariableBasedStyle(color: string) {
             .split(COMMA)
             .every(
                 variable =>
-                    variable.trim().startsWith(TWO_DASHES) || variable.trim().startsWith(VAR_PREFIX)
+                    variable.trim().startsWith(TWO_DASHES) ||
+                    variable.trim().startsWith(VAR_PREFIX + TWO_DASHES)
             )
     );
 }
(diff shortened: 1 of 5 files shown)


juliaroldi (approved):

packages-content-model/roosterjs-content-model-core/lib/override/pasteTextColorFormatParser.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @internal
+ */
+export const pasteTextColorFormatParser: FormatParser<ContentModelSegmentFormat> = (
+    format,
+    element,
+    context,
+    defaultStyle
+): void => {
+    if (!element.style.color.startsWith(VAR_PREFIX)) {
+        context.defaultFormatParsers.textColor?.(format, element, context, defaultStyle);
```
JiuqingSong, line 15: Is there any issue when we just use the default parser?

packages-content-model/roosterjs-content-model-core/lib/override/pasteTextColorFormatParser.ts
```diff
@@ -0,0 +1,32 @@
...
+};
+
+function isVariableBasedStyle(element: HTMLElement) {
+    const color = element.style.color;
+    return (
+        color.startsWith(VAR_PREFIX) &&
+        color
+            .substring(color.indexOf(START_BRACKET) + 1, color.lastIndexOf(END_BRACKET))
+            .split(COMMA)
+            .every(variable => variable.trim().startsWith(TWO_DASHES))
```
JiuqingSong, line 30: the second half does not need to start with "--"

BryanValverdeU: I added an additional check for var prefixes too, to handle this kind of variables:

`var(--variable, var(--variable2, white))`

PR #2315, opened 2024-01-08
Allow TableResize's onShowHelperElement callback to get affected element

florian-msft:
The TableResize plugin allows customization of helper elements using the `onShowHelperElement` callback. This callback gets triggered with some information on the type of helper, but it does not provide information about the element (table or table cell) affected. The changes in this PR pass in the element to the callback to allow customizing the helper element based on which table is being acted on.


packages/roosterjs-editor-plugins/lib/plugins/TableResize/TableResize.ts
```diff
@@ -32,7 +32,8 @@ export default class TableResize implements EditorPlugin {
     constructor(
         private onShowHelperElement?: (
             elementData: CreateElementData,
-            helperType: 'CellResizer' | 'TableInserter' | 'TableResizer' | 'TableSelector'
+            helperType: 'CellResizer' | 'TableInserter' | 'TableResizer' | 'TableSelector',
+            tableOrTd: HTMLTableElement | HTMLTableCellElement
```
JiuqingSong, line 36: Please make  this parameter optional to be compatible with older builds

florian-msft: I might be wrong, but I think that TS will allow passing in functions that only take two arguments and the type check won't fail. But I am open to making the change also if I am missing something!

JiuqingSong: hmm. ok. That makes sense to me.

JiuqingSong (approved):

PR #2316, opened 2024-01-08
U/juliaroldi/delete list

juliaroldi:


packages-content-model/roosterjs-content-model-plugins/lib/edit/deleteSteps/deleteList.ts
```diff
@@ -0,0 +1,24 @@
...
+import type { DeleteSelectionStep } from 'roosterjs-content-model-types';
+
+/**
+ * @internal
+ */
+export const deleteList: DeleteSelectionStep = context => {
+    const { paragraph, marker, path } = context.insertPoint;
+
+    if (context.deleteResult == 'nothingToDelete' || context.deleteResult == 'notDeleted') {
+        const index = getClosestAncestorBlockGroupIndex(path, ['ListItem']);
```
JiuqingSong, line 11: Please add the 3rd parameter "['TableCell']" so if there is table under list, do not delete list. And add test case for this scenario

packages-content-model/roosterjs-content-model-plugins/lib/edit/deleteSteps/deleteList.ts
```diff
@@ -0,0 +1,24 @@
...
+        const item = path[index];
+        if (
+            index >= 0 &&
+            paragraph.segments[0] == marker &&
+            item.blockGroupType == 'ListItem' &&
+            (paragraph.segments.length == 1 ||
+                (paragraph.segments.length == 2 && paragraph.segments[1].segmentType == 'Br'))
+        ) {
+            item.levels = [];
+            context.deleteResult = 'singleChar';
```
JiuqingSong, line 21: I feel here it would be better to use 'range' so that undo snapshot will be added.

packages-content-model/roosterjs-content-model-plugins/lib/edit/deleteSteps/deleteList.ts
```diff
@@ -0,0 +1,24 @@
...
+        const index = getClosestAncestorBlockGroupIndex(path, ['ListItem', 'TableCell']);
+        const item = path[index];
+        if (
+            index >= 0 &&
+            paragraph.segments[0] == marker &&
+            item.blockGroupType == 'ListItem' &&
+            (paragraph.segments.length == 1 ||
+                (paragraph.segments.length == 2 && paragraph.segments[1].segmentType == 'Br'))
+        ) {
+            item.levels = [];
```
BryanValverdeU, line 20: Just curious, if the list is a nested list, should we remove all the list like right now, or should we outdent to the previous level? Probably with `levels.pop()`

JiuqingSong: I personally am ok with either. Maybe @jvillalobos can comment here?

JiuqingSong (approved):

packages-content-model/roosterjs-content-model-plugins/lib/edit/deleteSteps/deleteList.ts
juliaroldi: When we delete a nested list, the delete result returns 'Range', then it not impacted by this code. I tested in word desktop and it removes the list.

PR #2325, opened 2024-01-10
Table fidelity Improvement: Add each border radius when the shorthand css is not provided and `borderCollapse: separate` support

### LINKED ISSUE #2267: Copying a table does not carry over inline styles and class as expected
**Describe the bug**
When copy-pasting a table using the new content model, some inline styles and class names (and maybe also other properties) get removed when pasting.

**To Reproduce**
1. Open this repro example: https://codepen.io/florian-msft/pen/RwvmwGv
2. Click on the button labeled "click me" (this will insert a table with the bottom right cell in red with rounded corners and a class set)
3. Select and copy the table
4. Insert some line breaks and paste the table below

**Expected behavior**
The table will be inserted with the rounded corners and the same class

**Actual behavior**
The table does not have the rounded corners and the class name got removed

**Screenshots**
image

image

**Device Information**
 - OS: Windows 11
 - Browser: Edge
 - Version: 119.0.2151.97 (Official build) (64-bit)

**Additional context**
Our code relied on the class being copied over to run additional code when we detected "special" tables.

BryanValverdeU:
Add support to each of the border corners if the borderRadius shorthand style is not provided and also support separate border collapse.

#2267


packages-content-model/roosterjs-content-model-dom/lib/formatHandlers/common/borderFormatHandler.ts
```diff
@@ -42,15 +49,27 @@ export const borderFormatHandler: FormatHandler<BorderFormat> = {
...
+
+                if (value) {
+                    format[key] = value;
+                }
+            });
         }
     },
     apply: (format, element) => {
-        BorderKeys.forEach(key => {
+        BorderKeys.concat(BorderRadiusKeys).forEach(key => {
```
JiuqingSong, line 63: Let make the result of "BorderKeys.concat(BorderRadiusKeys)" to be a global constant so no need to calculate them every time.

BryanValverdeU: Sure, done.

JiuqingSong (approved):

PR #2326, opened 2024-01-10
[List Feature] Auto bullet/Auto Numbering 

juliaroldi:
Port auto format list feature from Content Edit to a new Content Model Auto Format Plugin. 
listTriggers


demo/scripts/controls/ContentModelEditorMainPane.tsx
```diff
@@ -24,6 +24,7 @@ import { getDarkColor } from 'roosterjs-color-utils';
 import { PartialTheme } from '@fluentui/react/lib/Theme';
 import { trustedHTMLHandler } from '../utils/trustedHTMLHandler';
 import {
+    ContentModelAutoFormatPlugin,
```
JiuqingSong, line 27: can you also add this new plugin to `StandaloneEditorMainPane.tsx`?

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/ContentModelAutoFormatPlugin.ts
```diff
@@ -0,0 +1,87 @@
...
+    IStandaloneEditor,
+    KeyDownEvent,
+    PluginEvent,
+} from 'roosterjs-content-model-types';
+import type { IContentModelEditor } from 'roosterjs-content-model-editor';
+
+/**
+ * Auto Format plugin handles auto formatting, such as transforming * characters into a bullet list.
+ * It can be customized with options to enable or disable auto list features.
+ * @param options An optional parameter that takes in an object of type AutoFormatOptions, which includes the following properties:
```
JiuqingSong, line 13: Move this comment down before `constructor` (line 24)

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/ContentModelAutoFormatPlugin.ts
```diff
@@ -0,0 +1,87 @@
...
+ *  - autoBullet: A boolean that enables or disables automatic bullet list formatting. Defaults to true.
+ *  - autoNumbering: A boolean that enables or disables automatic numbering formatting. Defaults to true.
+ */
+export class ContentModelAutoFormatPlugin implements EditorPlugin {
+    private editor: IContentModelEditor | null = null;
```
JiuqingSong, line 18: Use `IStandaloneEditor`

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/ContentModelAutoFormatPlugin.ts
```diff
@@ -0,0 +1,87 @@
...
+    private options = {
+        autoBullet: true,
+        autoNumbering: true,
+    };
+
+    constructor(options?: { autoBullet: boolean; autoNumbering: boolean }) {
```
JiuqingSong, line 24: Suggest move the parameter out to be a type so that it is easier to add comment, and add more features later.

Also I think we need a default value for options

JiuqingSong: For example:

```ts
const DefaultOptions: Required<YourTypeName> = {
  autoBullet: true;
  autoNumbering: true;
};

class ContentModelAutoFormatPlugin {
  constructor(private options: YouTypeName = DefaultOptions) { ... }
}
```

When you use this.options, no need to do null check.

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/ContentModelAutoFormatPlugin.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * The first method that editor will call to a plugin when editor is initializing.
+     * It will pass in the editor instance, plugin should take this chance to save the
+     * editor reference so that it can call to any editor method or format API later.
+     * @param editor The editor object
+     */
+    initialize(editor: IStandaloneEditor) {
+        // TODO: Later we may need a different interface for Content Model editor plugin
+        this.editor = editor as IContentModelEditor;
```
JiuqingSong, line 45: no need to do type cast.

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/ContentModelAutoFormatPlugin.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /**
+     * The first method that editor will call to a plugin when editor is initializing.
+     * It will pass in the editor instance, plugin should take this chance to save the
+     * editor reference so that it can call to any editor method or format API later.
+     * @param editor The editor object
+     */
+    initialize(editor: IStandaloneEditor) {
+        // TODO: Later we may need a different interface for Content Model editor plugin
```
JiuqingSong, line 44: remove this comment

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/keyboardListTrigger.ts
```diff
@@ -0,0 +1,59 @@
...
+    updateListMetadata,
+} from 'roosterjs-content-model-core';
+import type { IStandaloneEditor } from 'roosterjs-content-model-types';
+
+/**
+ * @internal
+ */
+export default function keyboardListTrigger(
+    editor: IStandaloneEditor,
+    shouldSearchForBullet: boolean = true,
```
JiuqingSong, line 15: Can we reuse the options object here?

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/keyboardListTrigger.ts
```diff
@@ -0,0 +1,59 @@
...
+            shouldSearchForBullet,
+            shouldSearchForNumbering
+        );
+        if (listStyleType) {
+            const { listType, styleType, index } = listStyleType;
+
+            const paragraphOrListItems = getOperationalBlocks(model, ['ListItem'], []);
+            paragraphOrListItems.forEach(({ block, parent }, itemIndex) => {
+                if (!isBlockGroupOfType(block, 'General')) {
+                    const blockIndex = parent.blocks.indexOf(block);
```
JiuqingSong, line 30: can we just use existing format API? e.g. `setListType`

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/utils/convertAlphaToDecimals.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * @internal
+ * Convert english alphabet numbers into decimal numbers
+ * @param letter The letter that needs to be converted
+ * @returns
+ */
+export default function convertAlphaToDecimals(letter: string): number | undefined {
```
JiuqingSong, line 7: I think we already have these methods in `updateListMetadata.ts`, can we just reuse?

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/keyboardListTrigger.ts
juliaroldi: We can reuse, but options might have properties not related to auto list. Also we are using this properties before calling this function, because if both autoBullet and autoNumbering are false, we do not need to call keyboardListTrigger.

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/keyboardListTrigger.ts
juliaroldi: I tried to use it, but if we type * and space, it will create the list: 

- *

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/utils/convertAlphaToDecimals.ts
juliaroldi: We have `convertDecimalsToAlpha` that convert the number to the character, here we need to transform the character into number.

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/keyboardListTrigger.ts
JiuqingSong: ok

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/keyboardListTrigger.ts
JiuqingSong: I see. Maybe we can do some kind of refactoring. I wish we can only have one place to create the list properties.

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/keyboardListTrigger.ts
JiuqingSong: So as an internal function, we don't expect it to be called from other place, so no need to have default values of the parameters.

JiuqingSong:
Please add description of this PR to track what we are doing here.

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/ContentModelAutoFormatPlugin.ts
```diff
@@ -91,6 +92,7 @@ export class ContentModelAutoFormatPlugin implements EditorPlugin {
                 case ' ':
                     const { autoBullet, autoNumbering } = this.options;
                     if (autoBullet || autoNumbering) {
+                        event.rawEvent.preventDefault();
```
JiuqingSong, line 95: This causes we cannot input space at all

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/keyboardListTrigger.ts
```diff
@@ -1,59 +1,55 @@
...
 ) {
-    editor.formatContentModel((model, _) => {
+    editor.formatContentModel((model, _context) => {
         const listStyleType = getListTypeStyle(
             model,
             shouldSearchForBullet,
             shouldSearchForNumbering
         );
         if (listStyleType) {
+            const paragraph = getSelectedSegmentsAndParagraphs(model, false)[0][1];
```
JiuqingSong, line 21: null check for getSelectedSegmentsAndParagraphs(model, false)[0]

packages-content-model/roosterjs-content-model-plugins/lib/autoFormat/keyboardListTrigger.ts
```diff
@@ -1,59 +1,55 @@
...
+const triggerList = (
+    editor: IStandaloneEditor,
+    model: ContentModelDocument,
+    listType: 'OL' | 'UL',
+    styleType: number,
+    index?: number
+) => {
+    setListType(model, listType);
+    const isOrderedList = listType == 'OL';
+    if (index && isOrderedList) {
```
JiuqingSong, line 42: what if index is 0?

JiuqingSong: is that allowed?

juliaroldi: index 0 is not allowed

JiuqingSong: Ok, in that case please add a comment to tell we don't expect index to be 0 here.

BryanValverdeU (approved):

PR #2331, opened 2024-01-12
Try fix #2305 (in progress)

### LINKED ISSUE #2305: [Content Model] The style on SelectionMarker is not up to date after setFontSize more than one time

JiuqingSong:
Try fix #2305 

Apply pending format to SelectionMarker when create content model. So `formatSegmentWithContentModel` won't need handle it.


packages-content-model/roosterjs-content-model-api/lib/publicApi/utils/formatSegmentWithContentModel.ts
```diff
@@ -71,7 +68,7 @@ export function formatSegmentWithContentModel(
 
             afterFormatCallback?.(model);
 
-            if (!pendingFormat && isCollapsedSelection) {
+            if (isCollapsedSelection) {
```
haven2world, line 71: Could we merge this `if` statement into the next one? since they have the same condition.

haven2world (approved):

PR #2332, opened 2024-01-12
Fix safari selection issue

JiuqingSong:
When editor loses focus, we saves a last selection by handling `onblur` event. However, in Safari when we try to get latest selection in `onblur` event handler, it is already too late, so that we can't know selection in Safari if editor doesn't have focus. This happens when user click on format buttons.

Originally we solved this issue by handling more event such as keydown, mousedown, ... Now we realize that if mobile Safari we still miss a scenario when user close soft keyboard, we don't receive any event but the editor already loses focus.

To fix this, now we always save a selection in Safari even editor still has focus. When we call `editor.getSelectionRange()`, `editor.getSelectionRangeEx()`, `standaloneEditor.getDOMSelection()`, we first check if editor has focus. Then we use this cached selection only when:
1. Editor does not have focus. Or
2. Editor has focus, but current cached selection is table selection or image selection

With this fix, editor will always has correct selection in Safari no matter if it has focus.


JiuqingSong:
Found a regression in Safari in old editor:
1. type something
2. click B, I, U buttons one by one
3. keep typing

The second time typed text appears at the beginning of doc. This only happens with this change in old editor in Safari. I need to check why.

JiuqingSong:
> Found a regression in Safari in old editor:
> 
> 1. type something
> 2. click B, I, U buttons one by one
> 3. keep typing
> 
> The second time typed text appears at the beginning of doc. This only happens with this change in old editor in Safari. I need to check why.

fixed.

juliaroldi (approved):

Issue #1942, opened 2023-07-12
Ctrl + K shortcut support for inserting the link


p-rk:
Currently, the ctrl + k shortcut seems to be not implemented for inserting the link into the editor. If we select some text and use the shortcut ctrl + k it should open an insert link popup. 

**To Reproduce**
1. Add some text to the editor
2. Select the word where you want to embed the link
3. Press shortcut ctrl + k
4. It is by default going to the browser's search bar.

**Expected behavior**
It should open the create link popup with the shortcut ctrl + k.

**Device Information**
 - Edge, Chrome

**Additional context**
This is supported in Outlook, office applications, and ADO also. It would be great if we can add this shortcut support.

JiuqingSong:
The insert link dialog is part of roosterjs-react. So The hotkey needs to be implemented there as well.

@p-rk do you want to submit a fix?

p-rk:
@JiuqingSong - Sure can give it a try, but I need some reference in implementing this. I saw earlier PRs related to shortcuts for reference and could see shortcuts were implemented as part of the ContentEdit plugin. So, need some reference if it is implemented in any of the other components in roosterjs-react library.

JiuqingSong:
The easiest fix should be just handle KeyDown event in RibbonPlugin, and check if the key is Ctrl+K, then invoke the same command of insert link button

sathishsridhar:
@p-rk / @JiuqingSong Hey anyone working on this ? if not, I might give it a try.

JiuqingSong:
@sathishsridhar  please feel free to start on it. Thanks.

Apoorvakothari:
@JiuqingSong Is this issue still open? If so, I would like to make my first open source contribution to this repository through this issue.

romanisa:
@Apoorvakothari Sure feel free to submit a PR and add Jiuqing as reviewer

Ghanem10:
Hey, if no one is working on this I have a PR ready for review @JiuqingSong

Ghanem10:
Hey, @JiuqingSong could you please take a look at the PR when you can, thank you!

JiuqingSong:
Solved by #2333

PR #2333, opened 2024-01-12
Add shortcut support Ctrl+K for insertLink 

gm-al:
Implement Ctrl+K shortcut for inserting a link on the KeyDown event in roosterjs-react, reference issue #1942.


packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
```diff
@@ -49,6 +50,15 @@ class RibbonPluginImpl implements RibbonPlugin {
      * @param event PluginEvent object
      */
     onPluginEvent(event: PluginEvent) {
+        if (
+            event.eventType == PluginEventType.KeyDown &&
+            event.rawEvent.key == 'k' &&
+            event.rawEvent.ctrlKey
```
JiuqingSong, line 56: do you also want to support Meta key on MacOS?

JiuqingSong: And better also check Alt key is not pressed

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
```diff
@@ -49,6 +50,15 @@ class RibbonPluginImpl implements RibbonPlugin {
...
+        ) {
+            this.onButtonClick(insertLink, 'insertLinkTitle', undefined);
```
JiuqingSong, line 58: Suggest we check if insertLink button is added to ribbon. It is possible that user doesn't want this feature.

Or we can also add a parameter to allow disable this feature.

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
Ghanem10: Yes! I can do that, I'll add a check for the Alt key and the Meta key on MacOS.

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
Ghanem10: Allowing users to disable this feature would be a good to add. Do you suggest a specific way to implement this?

Ghanem10: So, check if insertLink button is added to ribbon could you elaborate on this a bit. Thank you :)

Ghanem10: I guess if there is a setting option/button with a dropdown menu to allow users to adjust their preferences about shortcuts, instead of adding a dropdown menu on inertLinkButton directly. This would make much sense, and by default, we can disable this feature and enable it on a switch button from false to true if the user wants to use this feature, do you think?

Ghanem10:
@microsoft-github-policy-service agree

Ghanem10:
@JiuqingSong I should also point out that I didn't add any checks for the RibbonContentModel. Should we add a check there as well?

packages-ui/roosterjs-react/lib/ribbon/type/RibbonButton.ts
```diff
@@ -22,6 +22,12 @@ export default interface RibbonButton<T extends string> {
      */
     flipWhenRtl?: boolean;
 
+    /**
+     * Set to false if we need to enable insert link button on KeyDown event.
+     * @default True insert link button on KeyDown event is disabled by default.
+     */
+    disableInsertLink?: boolean;
```
JiuqingSong, line 29: RibbonButton is a common interface. We should not add a property that is only used by one single button. When people create new button they will be confused why I need to care something related to "insertlink"

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
```diff
@@ -49,6 +50,21 @@ class RibbonPluginImpl implements RibbonPlugin {
      * @param event PluginEvent object
      */
     onPluginEvent(event: PluginEvent) {
+        if (
+            event.eventType == PluginEventType.KeyDown &&
+            event.rawEvent.key == 'k' &&
+            (event.rawEvent.ctrlKey || event.rawEvent.metaKey) &&
```
JiuqingSong, line 56: you can use function `isCtrlOrMetaPressed()` (packages\roosterjs-editor-dom\lib\event\isCtrlOrMetaPressed.ts) which will also help check if current OS is Mac. Because on Window metakey can be true when Win key is pressed.

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
```diff
@@ -49,6 +50,21 @@ class RibbonPluginImpl implements RibbonPlugin {
...
+            !event.rawEvent.altKey
+        ) {
+            if (insertLink.disableInsertLink) {
+                event.rawEvent.preventDefault();
+                return;
+            }
+
+            this.onButtonClick(insertLink, 'insertLinkTitle', undefined);
```
JiuqingSong, line 64: Instead of directly invoking the button, you can extract the onButtonClick function of insertLink button to be a common function, and let both your code and the insertLinkButton to call that function. So RibbonPlugin doesn't need to have dependency to the button.

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
```diff
@@ -49,6 +50,21 @@ class RibbonPluginImpl implements RibbonPlugin {
      * @param event PluginEvent object
      */
     onPluginEvent(event: PluginEvent) {
+        if (
+            event.eventType == PluginEventType.KeyDown &&
+            event.rawEvent.key == 'k' &&
+            (event.rawEvent.ctrlKey || event.rawEvent.metaKey) &&
+            !event.rawEvent.altKey
+        ) {
+            if (insertLink.disableInsertLink) {
```
JiuqingSong, line 59: By saying "adding a check to the feature", I mean you can add code to RibbonPlugin class, for example:
```ts
interface RibbonPluginOptions {
  allowInsertHotKey? boolean;
}

class RibbonPlugin {
  constructor(options?: RibbonPluginOptions) { ... }
}
```

Then you can pass an options parameter to the plugin to enable this feature.

gm-al: I see what you meant there, thank you for clarifying!

gm-al:
@microsoft-github-policy-service agree

packages-ui/roosterjs-react/lib/common/type/RibbonPluginOptions.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Interface to allow insert link on hot key press in ribbon plugin.
+ */
+export default interface RibbonPluginOptions {
+    /**
+     * Set the allowInsertHotKey property to false when the user doesn't want to use this feature
+     * @default true
+     */
+    allowInsertHotKey?: Boolean;
```
JiuqingSong, line 9: Let's name it `allowInsertLinkHotKey`

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
```diff
@@ -15,12 +17,18 @@ class RibbonPluginImpl implements RibbonPlugin {
     private timer = 0;
     private formatState: FormatState | null = null;
     private uiUtilities: UIUtilities | null = null;
+    private allowInsertHotKey: Boolean = true;
```
JiuqingSong, line 20: Instead of saving each feature setting here, we can just save the whole options object. So that if we add more features in options, no need to add more members here.

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
```diff
@@ -49,6 +57,17 @@ class RibbonPluginImpl implements RibbonPlugin {
      * @param event PluginEvent object
      */
     onPluginEvent(event: PluginEvent) {
+        if (
```
JiuqingSong, line 60: let's move this block down into the switch block.

for example:

```ts
switch(event.eventType) {
  ...
  case PluginEventType.KeyDown:
   if (event.rawEvent.key == 'k' && isCtrlOrMetaPressed(event.rawEvent) && !event.rawEvent.altKey && this.options?.allowInsertLinkHotKey
        ) {
            this.handleButtonClick(insertLink, 'insertLinkTitle', undefined);
   }
   this.delayUpdate();
   break;

  case PluginEventType.MouseUp:
    this.delayUpdate();
    break;
}
```

gm-al:
@JiuqingSong Sorry, for the delay. I did remove the check on whether the insertLinkButton was added or not from the Ribbon component since it's not necessary and I did add an option for a user to enable or disable this feature.

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
gm-al: Updated.

packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
Faeour: I completely agree. I think this is a cleaner way of doing it if we need to add more features.

gm-al: I completely agree, I think this is a cleaner way of doing it if we need to add more features.

gm-al:
Should we write some unit tests for roosterJs-react, currently the test dir doesn't have any tests

JiuqingSong (approved):

JiuqingSong:
Please fix the build error:
/home/runner/work/roosterjs/roosterjs/packages-ui/roosterjs-react/lib/ribbon/plugin/createRibbonPlugin.ts
Error:    2:31  error  '/home/runner/work/roosterjs/roosterjs/node_modules/roosterjs-editor-dom/lib/index.ts' imported multiple times  import/no-duplicates
Error:    5:37  error  '/home/runner/work/roosterjs/roosterjs/node_modules/roosterjs-editor-dom/lib/index.ts' imported multiple times  import/no-duplicates
Error:   75:[13](https://github.com/microsoft/roosterjs/actions/runs/7587600875/job/20668322946?pr=2333#step:5:14)  error  Expected a 'break' statement before 'case'                                                                      no-fallthrough

JiuqingSong (approved):

PR #2334, opened 2024-01-12
Content Model: Support format parser and applier for text node

JiuqingSong:
This is not a common case, but we still want to allow customize the format parser/applier for text node.

To use them:
```ts
const parser: TextFormatParser = ...
const applier: TextFormatAppler = ...

const editor = new StandaloneEditor(div, {
   defaultDomToModelOptions: {
      additionalFormatParsers: {
          text: [parser]
      },
   },
   defaultModelToDomOptions: {
      additionalFormatAppliers: {
          text: [applier]
      }
  }
});
```


juliaroldi:
Could you please provide a example where we will this change will be need?

BryanValverdeU (approved):

JiuqingSong:
> Could you please provide a example where we will this change will be need?

It will be used in OWA by the auto correction feature. We will add some expanded property on text node to tell if a word should be auto-corrected. Then content model need to keep this property when rewrite.

PR #2340, opened 2024-01-17
Update start of lists when pasting from Word Online

BryanValverdeU:
When pasting multiple lists from Word Online to Rooster, if there are multiple numbering lists, the list start is not updated. 
This PR adds functionality to update the start override of the level when needed. By checking the start attribute of the list.

This is because Word Online List content is most of the time provided like:

```
<ol start=1><li></li><ol>
<ol start=2><li></li><ol>
```

Since each of the list items are wrapped inside of ol/ul the list format is lost, because the default list processors remove the levels after processing each of the list element, so we need a custom behavior, 

To do so, add: 
1. Add a new property to keep the list level cached, and use it when creating the new list levels.
2. Add a list thread property to catch the previous list number. This will help to check if the start property should be set in the list being processed

Source: 
image

| Before | After |
|---|---| 
|image|image|


packages-content-model/roosterjs-content-model-plugins/lib/paste/WacComponents/processPastedContentWacComponents.ts
```diff
@@ -21,8 +23,16 @@ import type {
 const LIST_ELEMENT_TAGS = ['UL', 'OL', 'LI'];
 const LIST_ELEMENT_SELECTOR = LIST_ELEMENT_TAGS.join(',');
 
-const COMMENT_BG_COLOR_REST = 'rgba(209, 209, 209, 0.5)';
-const COMMENTS_TEXT_HIGHLIGHT_CLICKED = 'rgba(197, 139, 204, 0.5)';
+interface WacContext extends DomToModelContext {
```
JiuqingSong, line 26: Instead of extend DomToModelContext, can we extend its member listFormat (type DomToModelListFormat)?

BryanValverdeU: Sure, updated to extend DomToModelListFormat instead in the latest iteration.

JiuqingSong (approved):

PR #2350, opened 2024-01-22
Add support to cursor around Block entities.

BryanValverdeU:
Currently if user adds a block entity, they do not have the option to set the cursor around it.

This PR adds the initial functionality to be able to set the cursor around a block entity by updating how a block entity is represented in RoosterJS

Before this change, a block entity is defined as a div element. With this change we are setting a block entity as a span element with `width: 100%` and `display: inline-block`.

With this mix of styles, user is going to be able to insert the cursor around the entity using the delimiter logic we already have for inline entities.


packages-content-model/roosterjs-content-model-types/lib/format/formatParts/EntityInfoFormat.ts
```diff
@@ -16,4 +16,9 @@ export type EntityInfoFormat = {
      * Type of this entity
      */
     entityType?: string;
+
+    /**
+     * Whether this entity is a block entity
+     */
+    isBlock?: boolean;
```
JiuqingSong, line 23: Question: do we really need to know if an entity is block now? What if we just always remember its style like width and display, then just write them back as it is, so we always treat the entity as inline?

BryanValverdeU: Discussed with Jason, I added this because on paste browser removed the percentage and set px width. 

We will remove this and move it to only copy and paste handlers

packages-content-model/roosterjs-content-model-core/lib/override/pasteCopyBlockEntityParser.ts
```diff
@@ -0,0 +1,37 @@
+import { isElementOfType, isNodeOfType } from 'roosterjs-content-model-dom';
```
BryanValverdeU, line 1: @JiuqingSong added this file to handle the issue when copying and Browser removing the percentage width. Do you think this is a good approach?

packages-content-model/roosterjs-content-model-dom/lib/modelApi/creators/createEntity.ts
```diff
@@ -7,6 +7,7 @@ import type { ContentModelEntity, ContentModelSegmentFormat } from 'roosterjs-co
  * @param segmentFormat @optional Segment format of this entity
  * @param type @optional Type of this entity
  * @param id @optional Id of this entity
+ * @param isBlock @optional Whether the entity will be a block
```
JiuqingSong, line 10: remove this comment

packages-content-model/roosterjs-content-model-api/lib/publicApi/entity/insertEntity.ts
```diff
@@ -58,16 +57,23 @@ export default function insertEntity(
...
         wrapper.appendChild(contentNode);
     }
 
-    const entityModel = createEntity(wrapper, true /*isReadonly*/, undefined /*format*/, type);
+    const entityModel = createEntity(
+        wrapper,
+        true /*isReadonly*/,
+        undefined /*format*/,
+        type,
+        undefined /*id*/
```
JiuqingSong, line 75: no need to pass id here

packages-content-model/roosterjs-content-model-core/lib/override/pasteCopyBlockEntityParser.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const BLOCK_ENTITY_CLASS = '_EBlock';
+const ONE_HUNDRED_PERCENT = '100%';
+
+/**
+ * @internal
+ */
+export const onCreateCopyEntityNode: OnNodeCreated = (model, node) => {
+    if (
+        (model as ContentModelEntity)?.wrapper &&
```
JiuqingSong, line 17: better also check segmentType in case other object type can have wrapper in the future

packages-content-model/roosterjs-content-model-dom/lib/modelApi/creators/createEntity.ts
BryanValverdeU: Fixed

packages-content-model/roosterjs-content-model-core/lib/override/pasteCopyBlockEntityParser.ts
BryanValverdeU: added the segment check here too. Thanks

JiuqingSong (approved):

PR #2358, opened 2024-01-24
Port Table edit plugin

Andres-CT98:
### Plugin needs
|       Function      |                Need from DOM               |        Cursor position      |     Is Drag and Drop?    |
|:-------------------:|:------------------------------------------:|:---------------------------:|:------------------------:|
|     Mover (Selector)        |     Table position on Document             |     Always outside table    |             No           |
|     Resizer         |     Table position on Document             |     Always outside table    |            Yes           |
|     Inserter        |     Table and cell position on Document    |     Always outside table    |             No           |
|     Cell resizer    |     Table and cell position on Document    |     Always inside table     |            Yes           |

## Progress

- [x] Copy plugin to Content model
     - The pluginUtils Disposable, DragAndDropHandler, and DragAndDropHelper were copied too #2388
- [x] Change IEditor to IStandaloneEditor in plugin and make necessary adaptations
     - TableResize renamed to TableEditPlugin
     - TableSelector renamed to TableMover
- [x] Make plugin usable on Demo site
     - Old plugin is removed from demo
- [x] Change functions to use already existing Content model ones where applicable
     - Table Mover and Table Inserter now use CM fully on their functionality
     - The handles of the all plugins still only use js and css
     - Table resizer and Cell resizer use CM on its functionality of resizing and use js for writeback
- [x] Bring to Standalone Editor needed functions
     - createElement and Rect related plugin utilities were added #2388
     - isMobileOrTablet added to core, getScrollContainer and getVisibleViewport added to editor #2400 
     - Modified insertTableRow/Column to select any new insertion. #2412 
     - Addition of getDomStyle to Dom Helper #2419 
- [x] Remove all uses of non Content model dependencies
     - No plugin has dependencies outside content model packages
     - onShowHelperElement method removed
     - Functionality to resize all columns equally while pressing shift removed, as it was non functioning and unused.
     - Normalization of cells occurs only in normalizeTable
     - A check for Min Width was used in CellResizer when resizing columns due to the possibility of normalisation causing a wrong writeback
- [x] Add test cases
     - Port and rework of unit tests for Table edit plugin. Scenarios tested:
          * Table Edit Plugin:
               * Dismiss table editor on mouse out
               * Do not dismiss table editor on mouse out, related target is contained in scroll container
               * Do not dismiss table editor on mouse out, table editor not
               * Do not dismiss table editor on mouse out, related target null
               * Do not dismiss table editor on mouse out, currentTarget null
          * Table Mover
               * Display component on mouse move inside table
               * Display component, Top of table is visible in the scroll container scrolled down
               * Do not display component, top of table is no visible in the container
               * Do not display component, Top of table is no visible in the scroll container
               * On Click
          * Cell Resizer
               * Increase the height of the first row of a standard table
               * Decrease the height of the first row of a standard table
               * Increase the height of the last row of a standard table
               * Decrease the height of the last row of a standard table
               * Increase the width of the first column of a standard table
               * Decrease the width of the first column of a standard table
               * Increase the width of the last column of a standard table
               * Decrease the width of the last column of a standard table
          * Table Resizer
               * Removes table resizer on input
               * Removes table resizer on content change
               * Removes table resizer on scrolling
               * Increases the width of a standard table
               * Increases the height of a standard table
               * Increases the width and height of a standard table
               * Decreases the width of a standard table
               * Decreases the height of a standard table
               * Decreases the width and height of a standard table
          * Table Inserter
               * Adds a new column if the vertical inserter is detected and clicked
               * Adds a new row if the horizontal inserter is detected and clicked
               * Does not add inserter if top left corner hovered
               * Does not add new column if top middle clicked on merged top row
               * Does not add new row if left middle clicked on merged first column 

### Warnings and implications

-  An issue was found when changing zoom: The CSS of the table not reflecting the values of that table in Content Model Table widths.


demo/scripts/controls/ContentModelEditorMainPane.tsx
```diff
@@ -110,6 +111,7 @@ class ContentModelEditorMainPane extends MainPaneBase<ContentModelMainPaneState>
     private formatPainterPlugin: ContentModelFormatPainterPlugin;
     private pastePlugin: ContentModelPastePlugin;
     private sampleEntityPlugin: SampleEntityPlugin;
+    private tableEditPlugin: TableEditPlugin;
```
JiuqingSong, line 114: Can you also do the change in demo\scripts\controls\StandaloneEditorMainPane.tsx?

You can test it with parameter "?cm=2"

Later we will use this new page to replace the old one.

demo/scripts/controls/ContentModelEditorMainPane.tsx
```diff
@@ -198,6 +201,12 @@ class ContentModelEditorMainPane extends MainPaneBase<ContentModelMainPaneState>
             this.sampleEntityPlugin,
         ];
 
+        for (let i = 0; i < plugins.length; i++) {
```
JiuqingSong, line 204: you can remove it from demo\scripts\controls\getToggleablePlugins.ts then no need to remove from here

demo/scripts/controls/ContentModelEditorMainPane.tsx
```diff
@@ -253,6 +262,7 @@ class ContentModelEditorMainPane extends MainPaneBase<ContentModelMainPaneState>
                                 this.contentModelRibbonPlugin,
                                 this.formatPainterPlugin,
                                 this.pastePlugin,
+                                this.tableEditPlugin,
```
JiuqingSong, line 265: nice

packages-content-model/roosterjs-content-model-api/lib/publicApi/table/editTable.ts
```diff
@@ -34,14 +34,20 @@ import {
...
+) {
     editor.focus();
 
     editor.formatContentModel(
-        model => {
+        (model, context) => {
             const [tableModel, path] = getFirstSelectedTable(model);
 
             if (tableModel) {
+                context.skipUndoSnapshot = skipSnapshot;
```
JiuqingSong, line 49: any reason we need to skip undo snapshot here?

packages-content-model/roosterjs-content-model-core/lib/editor/StandaloneEditor.ts
```diff
@@ -392,6 +393,20 @@ export class StandaloneEditor implements IStandaloneEditor {
...
+     * Get the scroll container of the editor
+     */
+    getScrollContainer(): HTMLElement {
+        return this.getCore().domEvent.scrollContainer;
+    }
+
+    /**
+     * Retrieves the rect of the visible viewport of the editor.
+     */
+    getVisibleViewport(): Rect | null {
```
JiuqingSong, line 406: I think it is ok to add these 2 functions. You can remove the functions with same names from `ContentModelEditor` now.

JiuqingSong:
Generally, I'm so glad to see this PR. It is definitely on the right direction.

To answer your questions:

- getScrollContainer
- getVisibleViewport

Let's copy them into `StandaloneEditor` as you have done in your code, and we can remove them from `ContentModelEditor` since it is the sub class of `StandaloneEditor`

- normalizeRect
- getIntersectedRect
- normalizeTableCellSize from VTable

We can copy these functions into the `roosterjs-content-model-plugins` package for now and do "@internal" export. If later we find out it is also used by other place, we can move it.

- safeInstanceOf

You can use `isNodeOfType` and `isElementOfType` instead.

- createElement
- CreateElementData

You can copy to your code. And now we don't need to keep the `KnownCreateElementDataIndex` in the new code since everything is not exported out of package now.

- getComputedStyle

Depends on how you use it, you can use `window.getComputedStyles` directly. And if possible, use Content Model of editor instead of directly getting style from DOM element (this is a "nice to have", if that is too complex to do, we can leave it for now)

- contains

You can use `IStandlaoneEditor.isNodeInEditor` instead

- Browser

You can use `IStandaloneEditor.getEnvironment` instead. If the property you need is not in the return value, you can add it. We are getting rid of global object now.

- function to obtain Editor background color, part of getDefaultFormat

Maybe we can put default format into the result of getEnvironment().

- onShowHelperElement is an optional function implemented in the plugin, should it be kept?

I remember it was added by some 3rd party to support customizing the resizing element. I think we can remove it for now. If we get requirement, we can add it back later.

- TableSelector will also have the function of moving the table, maybe it should be renamed TableMover as it is the more complex behaviour?

Sounds good to me

* * *
I didn't see you list `queryElements` in the question list. Is it still in use?

packages-content-model/roosterjs-content-model-api/lib/publicApi/table/editTable.ts
Andres-CT98: The workflow is:
1. Click the table inserter
2. Internally the inserter selects a cell in the table and calls editTable after
3. editTable gets the table with a selected cell and does the operation

The problem is that editTable will create a snapshot with the selected cell, so between 1 and 2, a pre-selection snapshot is created and editTable is made to not create another one.

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/editors/features/CellResizer.ts
```diff
@@ -0,0 +1,232 @@
...
+        type: 'table',
+        firstColumn: 0,
+        firstRow: 0,
+        lastColumn: 0,
+        lastRow: 0,
+        table: table,
+    });
+
+    // Get the table content model
+    const cmTable = getFirstSelectedTable(editor.createContentModel())[0];
```
JiuqingSong, line 110: I have removed the API `createContentModel`, and replaced with a new one `getContentModelCopy`

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/TableEditPlugin.ts
```diff
@@ -0,0 +1,177 @@
...
+            mousemove: { beforeDispatch: this.onMouseMove },
+        });
+        const scrollContainer = this.editor.getScrollContainer();
+        scrollContainer.addEventListener('mouseout', this.onMouseOut);
+    }
+
+    private onMouseOut = ({ relatedTarget, currentTarget }: MouseEvent) => {
+        if (
+            relatedTarget instanceof HTMLElement &&
+            currentTarget instanceof HTMLElement &&
```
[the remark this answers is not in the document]
Andres-CT98: isNodeOfType does not accept events as Nodes.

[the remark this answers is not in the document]
JiuqingSong: You can do isNodeOfType(target as Node)

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableInserter.ts
```diff
@@ -0,0 +1,191 @@
...
+            lastColumn: columnIndex,
+            lastRow: rowIndex,
+            table: this.table,
+        });
+
+        // Insert row or column, skip snapshot as it was already taken
+        editTable(
+            this.editor,
+            this.isHorizontal ? 'insertBelow' : 'insertRight',
+            true /* skipSnapshot */
```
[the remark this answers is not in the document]
Andres-CT98: If the snapshot is made inside editTable it will be a wrong.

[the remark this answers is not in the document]
JiuqingSong: Can you add more detail here? What will be wrong?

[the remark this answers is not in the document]
Andres-CT98: I mentioned in a previous comment:

The workflow is:
1. Click the table inserter
2. Internally the inserter selects a cell in the table and calls editTable after
3. editTable gets the table with a selected cell and does the operation

The problem is that editTable will create a snapshot with the selected cell, so between 1 and 2, a pre-selection snapshot is created and editTable is made to not create another one.

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableInserter.ts
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        this.onMouseOutEvent = null;
+    }
+
+    private insertTd = () => {
+        // Get cell coordinates
+        const columnIndex = this.td.cellIndex;
+        const row =
+            this.td.parentElement instanceof HTMLTableRowElement
```
JiuqingSong, line 111: same here, use isElementOfType(). "instanceof" cannot work correctly in projection popout.

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableInserter.ts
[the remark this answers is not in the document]
JiuqingSong: After some thinking, I think we can do the format and selection like this:

```ts
// Replace from line 120
this.editor.formatContentModel(model => {
  const [tableModel, path] = getFirstSelectedTable(model);

  if (isHorizontal) {
     insertTableRow(tableModel, 'insertBelow');
 } else {
     insertTableColumn(tableModel, 'insertBelow');
} 

  // Table normalization and ensure empty paragraph here, you can extract from line 107 to line 123 of editTable.ts to be a new public function and call it here

  return true;
}, {
  selectionOverride: { // This will override the original selection to create a new Content Model, and no additional snapshot added
    type: 'table',
    table: table,
    firstRow: rowIndex,
    lastRow: rowIndex,
    firstColumn: columnIndex,
    lastColumn: columnIndex  
  },
  changeSource: ChangeSource.Format
});
```

So we can still change selection in model, no addition undo snapshot added, and no need to add new parameter to editTable.ts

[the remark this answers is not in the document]
JiuqingSong: To make it easier, I created a draft PR here https://github.com/microsoft/roosterjs/pull/2399 to extract a common API `formatTableWithContentModel`, and let existing `editTable` API use it. You can also call it from your code here and pass in the overridden table selection.

packages-content-model/roosterjs-content-model-api/lib/publicApi/table/editTable.ts
```diff
@@ -21,7 +21,11 @@ import {
  * @param editor The editor instance
  * @param operation The table operation to apply
  */
-export default function editTable(editor: IStandaloneEditor, operation: TableOperation) {
+export default function editTable(
+    editor: IStandaloneEditor,
+    operation: TableOperation,
+    skipSnapshot?: boolean
```
JiuqingSong, line 27: now we can remove this parameter.

packages-content-model/roosterjs-content-model-core/lib/editor/createStandaloneEditorCore.ts
```diff
@@ -61,6 +61,21 @@ function createEditorEnvironment(contentDiv: HTMLElement): EditorEnvironment {
     const navigator = contentDiv.ownerDocument.defaultView?.navigator;
     const userAgent = navigator?.userAgent ?? '';
     const appVersion = navigator?.appVersion ?? '';
+    let isMobileOrTablet = false;
```
JiuqingSong, line 64: bad merge.

JiuqingSong:
Found two issues:
1. Resize by dragging the lines (not the square at the corner) won't work
2. Change the zoom scale of editor (there is a zoom button in toolbar), then do whole table resize, the table size is wrong.

Andres-CT98:
> Found two issues:
> 
> 1. Resize by dragging the lines (not the square at the corner) won't work
> 2. Change the zoom scale of editor (there is a zoom button in toolbar), then do whole table resize, the table size is wrong.

Issue 1 is mentioned in the description as not working. As for Issue 2, changes where made to how the zoom is obtained. Is the new way to obtain the zoom tested?

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+function isTableTopVisible(
+    editor: IStandaloneEditor,
+    rect: Rect | null,
+    contentDiv?: EventTarget | null
+): boolean {
+    const visibleViewport = editor.getVisibleViewport();
+    if (contentDiv instanceof HTMLElement && visibleViewport && rect) {
```
BryanValverdeU, line 132: Can change this to use isNodeOfType util instead.

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/editors/TableEditor.ts
```diff
@@ -0,0 +1,384 @@
...
+        }
+    };
+
+    private getOnMouseOut = (feature: HTMLElement) => {
+        return (ev: MouseEvent) => {
+            if (
+                feature &&
+                ev.relatedTarget != feature &&
+                this.contentDiv instanceof HTMLElement &&
+                ev.relatedTarget instanceof HTMLElement &&
```
BryanValverdeU, line 377: Use IsNodeOfType util too.

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableResizer.ts
```diff
@@ -0,0 +1,246 @@
...
+    trigger.style.visibility = 'hidden';
+}
+
+function isTableBottomVisible(
+    editor: IStandaloneEditor,
+    rect: Rect | null,
+    contentDiv?: EventTarget | null
+): boolean {
+    const visibleViewport = editor.getVisibleViewport();
+    if (contentDiv instanceof HTMLElement && visibleViewport && rect) {
```
BryanValverdeU, line 235: Here we can also use isNodeOfType util instead

juliaroldi (commented):
Should we have the test cases in this PR to ensure this plugin behaves as expected?

Andres-CT98:
> Should we have the test cases in this PR to ensure this plugin behaves as expected?

I prefer it be done in a separate one to keep the focus of this PR on the code of the implementation.

juliaroldi:
> > Should we have the test cases in this PR to ensure this plugin behaves as expected?
> 
> I prefer it be done in a separate one to keep the focus of this PR on the code of the implementation.
> I'm worried that if someone makes a change that could potentially break this feature, without submitting unit tests, it might go unnoticed until the feature breaks, because we do not have unit test. But if you already have a PR for unit test ready, I think it is ok the way it is.

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/TableEditPlugin.ts
```diff
@@ -0,0 +1,175 @@
...
+    private tableRectMap: { table: HTMLTableElement; rect: Rect }[] | null = null;
+    private tableEditor: TableEditor | null = null;
+
+    /**
+     * Construct a new instance of TableResize plugin
+     * @param anchorContainerSelector An optional selector string to specify the container to host the plugin.
+     * The container must not be affected by transform: scale(), otherwise the position calculation will be wrong.
+     * If not specified, the plugin will be inserted in document.body
+     */
+    constructor(private anchorContainerSelector?: string) {}
```
BryanValverdeU, line 23: Just curious, are we removing the onShowHelperElement helper to allow users customize the TableEditors?

packages-content-model/roosterjs-content-model-core/lib/publicApi/table/normalizeTable.ts
```diff
@@ -6,6 +6,10 @@ import type {
     ContentModelTableCell,
 } from 'roosterjs-content-model-types';
 
+/**
+ * Minimum width for a table cell
+ */
+export const MIN_WIDTH: number = 30;
```
BryanValverdeU, line 12: Instead of exporting this constant, could we just create a constant in the file that is needed with the same value? (I believe Cell Resizer).
SplitTableCellHorizontally is also doing the same.

Andres-CT98: I think it is better if the value is to reman constant to export it rather than have multiple files with the same value. If the value changes in one place it would have to be manually changed in all places. I think it should also be exported to SplitTableCellHorizontally or maybe removed since the table will be normalised after the split..

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/TableEditPlugin.ts
Andres-CT98: Yes. If it is requested it will be re- added later.

BryanValverdeU: Any technical implication why we are doing this? Sounds like regressing functionality from the previous version.

packages-content-model/roosterjs-content-model-core/lib/publicApi/table/normalizeTable.ts
BryanValverdeU: Also, I have not seen that we are exporting const strings in this packages. I would probably ask Jason when he is back what would be the best practice/guideline for the content model packages

JiuqingSong: I think it is ok to export some const, but agree with Bryan, an exported const should have a more specified name.

For this particular case, instead of export the const directly, can we export a function instead? e.g.

```ts
export function getEffectiveTableCellWidth(originalWidth: number): number {
   return Math.max(originalWidth, MIN_WIDTH);
}
```

So that we don't need to export this width and other places do not need to do the math separately.

packages-content-model/roosterjs-content-model-core/lib/publicApi/table/normalizeTable.ts
```diff
@@ -75,6 +79,8 @@ export function normalizeTable(
     for (let i = 0; i < columns; i++) {
         if (table.widths[i] === undefined) {
             table.widths[i] = getTableCellWidth(columns);
+        } else if (table.widths[i] < MIN_WIDTH) {
```
JiuqingSong, line 82: Then here you can just do
else {
 table.width[i] = getEffectiveTableCellWidth(table.width[i]);
}

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/TableEditPlugin.ts
JiuqingSong: At the first step of porting a plugin, I prefer to just port the necessary functionalities. For now we don't guarantee 100% functional compatible to earlier builds. Later when need, we can add the lost functionalities back.

packages-content-model/roosterjs-content-model-plugins/lib/tableEdit/editors/features/CellResizer.ts
```diff
@@ -0,0 +1,233 @@
...
+    allWidths: number[];
+}
+
+function onDragStart(context: DragAndDropContext, event: MouseEvent): DragAndDropInitValue {
+    const { td, onStart } = context;
+    const rect = normalizeRect(td.getBoundingClientRect());
+
+    // Get cell coordinates
+    const columnIndex = td.cellIndex;
+    const row = td.parentElement instanceof HTMLTableRowElement ? td.parentElement : undefined;
```
JiuqingSong, line 81: please do not use instanceof

JiuqingSong:
The dev code change loos good to me now.

JiuqingSong (approved):

PR #2363, opened 2024-01-25
Standalone Editor: Provide a DOMHelper to allow access DOM tree

JiuqingSong:
When port existing plugins to use `StandaloneEditor`, we found that we still need access to DOM tree in some cases. To avoid providing too much API to access DOM tree directly in `IStandaloneEditor`, I'm creating a `DOMHelper` type and put all DOM access related functions here.

In this change we provide the first DOM access API: `queryElements`. For example:

```ts
const editor: IStandaloneEditor = ... 
const tables = editor.getDOMHelper().queryElements('table');
// Now we have all tables inside editor
```

Note that DOM helper access DOM directly so that it will not skip elements under entity, so we need to be careful when use this object.

### DIFF
--- a/packages-content-model/roosterjs-content-model-types/lib/parameter/DOMHelper.ts
+++ b/packages-content-model/roosterjs-content-model-types/lib/parameter/DOMHelper.ts
@@ -1,0 +1,22 @@
+/**
+ * A helper class to provide DOM access APIs
+ */
+export interface DOMHelper {
+    /**
+     * Query HTML elements in editor by tag name.
+     * Be careful of this function since it will also return element under entity.
+     * @param tag Tag name of the element to query
+     * @returns HTML Element array of the query result
+     */
+    queryElements<TTag extends keyof HTMLElementTagNameMap>(
+        tag: TTag
+    ): HTMLElementTagNameMap[TTag][];
+
+    /**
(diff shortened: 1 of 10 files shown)


Andres-CT98 (approved):

packages-content-model/roosterjs-content-model-types/lib/parameter/DOMHelper.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * A helper class to provide DOM access APIs
+ */
+export interface DOMHelper {
```
BryanValverdeU, line 4: So all Editor API that access DOM should be under this helper?
Like getElementAtCursor in current editor?

JiuqingSong: For every necessary DOM access API, I'd like to put them here, so the API provided by IStandaloneEditor directy will only has access to Content Model. This is to encourage people to use Content Model more, and reduce using DOM access.

For `getElementAtCursor`, can we use `getDOMSelection` instead?

BryanValverdeU: Let me check that.

BryanValverdeU (approved):

PR #2368, opened 2024-01-29
Add `isReverted` to DOM Selection

BryanValverdeU:
1. This PR adds a new Util to get the focused position using a DOMSelection as parameter.
1.1. To accomplish this, also adds a new property to DOMSelection `isReverted` to determine if the selection in the DOM was in Document order (From right to left/Top to bottom) or in reverse Document order (Left to right/Bottom to top). By Checking the Focus node of the selection and the start container of the range of the selection.  
2. Add the support to maintain the selection order when formating and when restoring an undo snapshot if the selection was in reverse of the document.


packages-content-model/roosterjs-content-model-core/lib/coreApi/getDOMSelection.ts
```diff
@@ -26,7 +27,8 @@ function getNewSelection(core: StandaloneEditorCore): DOMSelection | null {
     return range && core.contentDiv.contains(range.commonAncestorContainer)
         ? {
               type: 'range',
-              range: range,
+              range,
+              isReverted: isSelectionReverted(selection),
```
JiuqingSong, line 31: It seems the function isSelectionReverted is only used here, so you can move it into this file and not export it

JiuqingSong: I think it just need to be a one line code between line 25 and 27, such as:
```ts
const range = ...
const isReverted = ....
```

packages-content-model/roosterjs-content-model-dom/lib/domUtils/getFocusedPosition.ts
```diff
@@ -0,0 +1,21 @@
+import type { DOMSelection, Position } from 'roosterjs-content-model-types';
+
+/**
+ * Gets the focused position using the DOM Selection provided.
+ * @param selection DOM Selection to use when extracting the focused position
+ * @returns container and offset of the focused position
+ */
+export function getFocusedPosition(selection: DOMSelection): Position | undefined {
```
JiuqingSong, line 8: Where will this function be used? If it is not used very widely, maybe we just need to directly read isRevert when we use, so no need to have this public function so no need to add `Position` interface.

packages-content-model/roosterjs-content-model-core/lib/coreApi/getDOMSelection.ts
BryanValverdeU: I think it is also used in contentModelToDom. But I think we could just add the code to both files, It is a one liner check if it is reversed

packages-content-model/roosterjs-content-model-dom/lib/domUtils/getFocusedPosition.ts
BryanValverdeU: Agree, I think I will just remove this file.

packages-content-model/roosterjs-content-model-types/lib/parameter/Position.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Represent a position in DOM tree by the container and its offset index
+ */
+export interface Position {
```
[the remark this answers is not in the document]
BryanValverdeU: Sounds good, removed this interface as it will not be needed and we can just retrieve the position using the range and whether the selecion is reverted or not.

packages-content-model/roosterjs-content-model-dom/lib/modelToDom/contentModelToDom.ts
```diff
@@ -27,6 +27,7 @@ export function contentModelToDom(
     onNodeCreated?: OnNodeCreated
 ): DOMSelection | null {
     context.onNodeCreated = onNodeCreated;
+    context.regularSelection.isReverted = isSelectionReverted(doc.getSelection());
```
JiuqingSong, line 30: selection in content model may not be the same with current selection, so here we can't use doc.getSelection().

Actually when create Content Model, I think the isReverted info is lost, because we only have "isSelected" in Content Model but we don't store if it is reverted.

Do we need this info when write back from Content Model? If not, we can always set to false and leave a TODO here.

BryanValverdeU: Ok, lets remove it. 
The only problem here is that if the selection was reverse of the document, after model to dom, the reverse order could be lost. But it is probably a small issue.

JiuqingSong (approved):

PR #2399, opened 2024-02-06
Add API formatTableWithContentModel

JiuqingSong:
Split the existing `editTable` API, extract a new common function `formatTableWithContentModel`, so we can easily write new table format API with the common code.

This is a preparation step of TableEdit plugin porting.

### DIFF
--- a/packages-content-model/roosterjs-content-model-api/lib/publicApi/utils/formatTableWithContentModel.ts
+++ b/packages-content-model/roosterjs-content-model-api/lib/publicApi/utils/formatTableWithContentModel.ts
@@ -18,7 +18,11 @@ import type {
 } from 'roosterjs-content-model-types';
 
 /**
- * @internal
+ * Invoke a callback to format the selected table using Content Model
+ * @param editor The editor object
+ * @param apiName Name of API this calling this function. This is mostly for logging.
+ * @param callback The callback to format the table. It will be called with current selected table. If no table is selected, it will not be called.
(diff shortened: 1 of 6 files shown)


packages-content-model/roosterjs-content-model-api/lib/publicApi/utils/formatTableWithContentModel.ts
```diff
@@ -0,0 +1,68 @@
...
+                }
+
+                return true;
+            } else {
+                return false;
+            }
+        },
+        {
+            apiName,
+            selectionOverride,
```
Andres-CT98, line 65: On name of the parameter alone, I am not clear on what part of the process is the selection overriden. Before the format change or after?

Based on the comment I assume before but it also creates the issue of the selection appearing on the snapshot when it should not.

The behaviour on other editors seems to be to always select any new inserted row or column, so maybe that could be made a core part of the InsertRow/Column function.

JiuqingSong: These are two separate parts.
1. Yes, the override happens before format. It is just used for creating content model, not really select the document. So it won't impact the undo snapshot.
2. For insert row/column, that needs to happen after format. And yes, it needs to be handled in the insertTableRow/Column API, so you can leave it there for now, it should not block your work.

Andres-CT98 (approved):

PR #2402, opened 2024-02-07
Content model: fix keyboard delete issue on Android

Rain-Zheng:
Some Android IMEs (such as Gboard and Samsung keyboard) don't support `KeyboardEvent` well. The `key` property of keyboard events will always be `Unidentified` when using these IMEs. To workaround this issue, we need to handle `InputEvent` instead. 

The PR changes the DOM event handling logic in `EditPlugin`:
1. When receiving `Unidentified` keyboard event on Android, check the next `beforeinput` event. 
2. Create a keyboard event with correct key according to the type of input event: `deleteContentBackward` corresponds to `Backspace` key, and `deleteContentForward` corresponds to `Delete` key.
3. Use the created keyboard event to trigger `keyboardDelete`.


packages-content-model/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
```diff
@@ -31,6 +36,11 @@ export class EditPlugin implements EditorPlugin {
      */
     initialize(editor: IStandaloneEditor) {
         this.editor = editor;
+        this.disposer = this.editor.attachDomEvent({
```
JiuqingSong, line 39: I think only Android needs this, so we can add a check for android (you can call editor.getEnvironment().isAndroid)

JiuqingSong (approved):

PR #2403, opened 2024-02-07
Fix REM Unit behavior

BryanValverdeU:
When normalizing the font size, We were checking the parent element when the unit was rem, however this is wrong. For rem, we should be checking the root element font size, the `html` element fontsize.

This PR addresses this by checking the html element font size to set the normalized font size.

[rem](https://developer.mozilla.org/en-US/docs/Web/CSS/length#rem)
Represents the [font-size](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size) of the root element (typically `<html>` [link](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/html)). When used within the root element [font-size](https://developer.mozilla.org/en-US/docs/Web/CSS/font-size), it represents its initial value. A common browser default is 16px, but user-defined preferences may modify this.


juliaroldi (approved):

packages-content-model/roosterjs-content-model-dom/lib/formatHandlers/utils/parseValueWithUnit.ts
```diff
@@ -41,6 +43,23 @@ export function parseValueWithUnit(
             case 'in':
                 result = num * PixelPerInch;
                 break;
+            case 'rem':
+                if (currentSizePxOrElement && typeof currentSizePxOrElement != 'number') {
+                    const doc = currentSizePxOrElement.ownerDocument;
+                    const htmlRoot = doc.querySelector('html');
```
JiuqingSong, line 49: can we put the result of root font size into editorContext, since during the calculation it should not be changed, then we don't need to calculate it multiple times.

packages-content-model/roosterjs-content-model-dom/lib/formatHandlers/segment/fontSizeFormatHandler.ts
```diff
@@ -49,12 +59,18 @@ function normalizeFontSize(fontSize: string, contextFont: string | undefined): s
...
         fontSize.endsWith('em') ||
-        fontSize.endsWith('%')
+        fontSize.endsWith('%') ||
+        fontSize.endsWith('rem')
     ) {
         if (!contextFont) {
             return undefined;
         } else {
-            const existingFontSize = parseValueWithUnit(contextFont, undefined /*element*/, 'px');
+            const existingFontSize = parseValueWithUnit(
```
JiuqingSong, line 68: Instead of parsing the value here, for rem, can we just store a pixel value in editorContext? Then here we can directly use it as a number

packages-content-model/roosterjs-content-model-core/lib/coreApi/createEditorContext.ts
```diff
@@ -16,6 +16,7 @@ export const createEditorContext: CreateEditorContext = (core, saveIndex) => {
         allowCacheElement: true,
         domIndexer: saveIndex ? cache.domIndexer : undefined,
         zoomScale: domHelper.calculateZoomScale(),
+        rootDocumentFormat: {},
```
JiuqingSong, line 19: Here, we can call getComputedStyle() to get the pixel value of root font size, put it into editor context. Then we don't need to calculate it anymore.

And we don't need to add a new format type here, just use a simple number or string value here is ok.

JiuqingSong: For example

interface EditorContext {
  ...
  rootFontSize: getComputedStyle(....).fontSize
}

BryanValverdeU: Ok, we can do that. With the current code we would only get the computed style if we detect a rem style. But I guess with this it would be easier to read.

packages-content-model/roosterjs-content-model-dom/lib/formatHandlers/segment/fontSizeFormatHandler.ts
BryanValverdeU: Made the change, let me know if it looks good.

packages-content-model/roosterjs-content-model-core/lib/coreApi/createEditorContext.ts
BryanValverdeU: I am going to set the rootFontSize as number so we can pass it in to the parseValueWithUnit.

packages-content-model/roosterjs-content-model-core/lib/coreApi/createEditorContext.ts
```diff
@@ -30,3 +39,15 @@ function checkRootRtl(element: HTMLElement, context: EditorContext) {
         context.isRootRtl = true;
     }
 }
+
+function getRootComputedStyle(core: StandaloneEditorCore) {
+    const document = core.contentDiv.ownerDocument;
+    let rootElement: HTMLElement | undefined;
+    document.childNodes.forEach(node => {
```
JiuqingSong, line 46: can we just use document.body here?

BryanValverdeU: I think the root most of the time will be the <html> tag. Maybe document.body.parentElement?

JiuqingSong (approved):

Issue #2234, opened 2023-12-01
Content Model: need to have a more robust way to check default format to apply


JiuqingSong:
Sometimes we should apply default format but we didn't, it is possible because one of the container element already has `style` tag which causes we bypassed the default format applying.

We need to make this process smarter.

Related code: https://github.com/microsoft/roosterjs/blob/e03f3db89274b5e7da59d56960ac5c4006aab967/packages-content-model/roosterjs-content-model-core/lib/corePlugin/utils/applyDefaultFormat.ts#L26

PR #2406, opened 2024-02-09
Improve default format behavior

JiuqingSong:
1. Move the code for checking if should apply default format from function `applyDefaultFormat` to `FormatPlugin`, so that `applyDefaultFormat` is now only applying default format
2. Improve the code to check if should apply default format. 
   - If current focus is directly under editor DIV, we should apply default format.
   - If current focus is under some styled container but it doesn't include all the styles in default format, still need to apply default format. For now this check only contains font family, font size, text color and back color. B/I/U is not included.
3. To improve performance, cache the last checked container element so that if user is keep typing under the same container, no need to check again


JiuqingSong:
Fix #2234

juliaroldi (approved):

PR #2408, opened 2024-02-09
[Context Menu Plugin] Pass PointerEvent for contextmenu plugin

vinguyen12:
Issue: Customized context menu needs to prevent default behavior of both 'contextmenu' touchevent and 'contextmenu' mouseevent. Currently, we only pass mouseevent back to editor but not touchevent.

### DIFF
--- a/packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
+++ b/packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
@@ -217,7 +217,7 @@ export default class DOMEventPlugin implements PluginWithState<DOMEventPluginSta
         event.stopPropagation();
     };
 
-    private onContextMenuEvent = (event: MouseEvent) => {
+    private onContextMenuEvent = (event: PointerEvent) => {
         const allItems: any[] = [];
         const searcher = this.editor?.getContentSearcherOfCursor();
         const elementBeforeCursor = searcher?.getInlineElementBefore();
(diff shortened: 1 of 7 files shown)


packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
```diff
@@ -236,7 +236,7 @@ export default class DOMEventPlugin implements PluginWithState<DOMEventPluginSta
             }
         });
         this.editor?.triggerPluginEvent(PluginEventType.ContextMenu, {
-            rawEvent: event,
+            rawEvent: event as PointerEvent,
```
JiuqingSong, line 239: can we only change content model code, but leave the old editor code not changed?

JiuqingSong:
PointerEvent is a sub interface of MouseEvent. So if we didn't really use anything from PointerEvent, we can leave it as MouseEvent. We only need to cast it to PointerEvent when we really need it. Another concern is, it is really possible that the event is really just a mouse event but no pointer event. So if we always pass out as PointerEvent at core level, it may cause other potential issue that cannot be caught at build time.

packages/roosterjs-editor-core/lib/corePlugins/DOMEventPlugin.ts
vinguyen12: I reverted the change in old editor code.

vinguyen12:
> PointerEvent is a sub interface of MouseEvent. So if we didn't really use anything from PointerEvent, we can leave it as MouseEvent. We only need to cast it to PointerEvent when we really need it. Another concern is, it is really possible that the event is really just a mouse event but no pointer event. So if we always pass out as PointerEvent at core level, it may cause other potential issue that cannot be caught at build time.

I reverted the change and add a new separated property for the pointer event so that we can use it to handle in editor.

PR #2410, opened 2024-02-12
Outdent parent format container if necessary

JiuqingSong:
When there is margin in parent level of FormatContainer, `outdent` command should be able to outdent the parent level margin if current level doesn't have margin.

To repro:
Snapshot:
```html
<div style="margin-left: 40px; border-width: 1px; border-style: solid; border-color: black;"><div>test1</div><div style="margin-left: 0px;">test2</div></div><!--{"type":"range","start":[0,1,0,2],"end":[0,1,0,2],"isReverted":false,"isDarkMode":false}-->
```

Then press outdent button,

Before fix (nothing happens):
image

After fix: (whole format container is outdented):
image


JiuqingSong:
> Found an issue with this snapshot, the button is outdenting the whole container.
> 
> ```
> <div style="margin-left: 40px; border-width: 1px; border-style: solid; border-color: black;"><div>test1</div><div style="margin-left: 0px;">test2</div></div>
> ```

Yes, for now I'll say it is by design since we don't have a better way to only outdent one single paragraph under a format container.

User can still manually change indent to other paragraphs. It is better than "cannot outdent"

BryanValverdeU (approved):

PR #2412, opened 2024-02-13
Select newly inserted row/column

Andres-CT98:
### Description:

After using insertTableRow/Column, the newly inserted row/column should be selected.

### Fix

After getting selected cells to insert from, clear the table selection, including removing any selection marker present.
When creating the new cells to insert, set isSelected to true before inserting on row.

### Cases tested:

1. Create a 3x3 cell table
2. Triple click the central cell to select it
3. Use insertTableRow/Column

**Before:** The newly inserted row/column is not selected

**After:** The newly inserted row/column is selected

### Warnings and implications

- None


JiuqingSong:
Please fix test

packages-content-model/roosterjs-content-model-api/lib/modelApi/table/insertTableColumn.ts
```diff
@@ -1,12 +1,12 @@
...
 import { createTableCell } from 'roosterjs-content-model-dom';
-import { getSelectedCells } from 'roosterjs-content-model-core';
 import type {
     ContentModelTable,
     TableHorizontalInsertOperation,
 } from 'roosterjs-content-model-types';
 
 /**
- * @internal
+ * Insert a column to the table
```
JiuqingSong, line 9: Please add comment for parameters

packages-content-model/roosterjs-content-model-api/lib/modelApi/table/insertTableRow.ts
```diff
@@ -1,32 +1,35 @@
...
 import { createTableCell } from 'roosterjs-content-model-dom';
-import { getSelectedCells } from 'roosterjs-content-model-core';
 import type {
     ContentModelTable,
     TableVerticalInsertOperation,
 } from 'roosterjs-content-model-types';
 
 /**
- * @internal
+ * Insert a row to the table
```
JiuqingSong, line 9: same

packages-content-model/roosterjs-content-model-core/lib/publicApi/table/clearSelectedCells.ts
```diff
@@ -0,0 +1,28 @@
+import hasSelectionInBlock from '../selection/hasSelectionInBlock';
+import type { ContentModelTable, TableSelectionCoordinates } from 'roosterjs-content-model-types';
+
+/**
+ * Clear selection of a table.
+ * @param table The table model where the selection is to be cleared
+ * @param sel The selection coordinates to be cleared
+ */
+export function clearSelectedCells(table: ContentModelTable, sel: TableSelectionCoordinates) {
```
JiuqingSong, line 9: Curious: can this be done by using `setSelection` and do not pass selectable parameter?

JiuqingSong: Actually, I mean, it is possible to just use setSelection() function instead of this new function.

For example, consider you have a content model object:
```ts
const model: ContentModelDocument = { ... };
```

Now you want to clear any selection from this model, can you just do:
```ts
setSelection(model); // Passing in just the model parameter means we want to clear existing selections if any
```

This should clear all selections, including selection on the table. So we don't need to know the original selection at all.

JiuqingSong (approved):

PR #2419, opened 2024-02-15
Add getDomStyle to DOMHelper

Andres-CT98:
Add the getDomStyle method to DOM helper to get any CSS attribute from the contentDiv of the editor.


packages-content-model/roosterjs-content-model-types/lib/parameter/DOMHelper.ts
```diff
@@ -48,4 +48,10 @@ export interface DOMHelper {
      * @param name Name of the attribute
      */
     getDomAttribute(name: string): string | null;
+
+    /**
+     * Get DOM style of editor content DIV
+     * @param style Name of the style
+     */
+    getDomStyle(style: keyof CSSStyleDeclaration): string | number | object | undefined;
```
JiuqingSong, line 56: Instead of returning string|number|..., can you try this:
```ts
getDomStyle<T extends keyof CSSStyleDeclaration>(style: T): CSSStyleDeclaration[T];
```

So we have a strong-typed return value

packages-content-model/roosterjs-content-model-core/lib/editor/DOMHelperImpl.ts
```diff
@@ -36,6 +36,10 @@ class DOMHelperImpl implements DOMHelper {
     getDomAttribute(name: string): string | null {
         return this.contentDiv.getAttribute(name);
     }
+
+    getDomStyle(style: keyof CSSStyleDeclaration): string | number | object | undefined {
+        return this.contentDiv.style[style]?.valueOf();
```
JiuqingSong, line 41: Then you can remove "valueOf()" here

packages-content-model/roosterjs-content-model-core/lib/editor/DOMHelperImpl.ts
```diff
@@ -36,6 +36,10 @@ class DOMHelperImpl implements DOMHelper {
...
+    }
 }
```
JiuqingSong, line 43: please add a test case.

packages-content-model/roosterjs-content-model-types/lib/parameter/DOMHelper.ts
Andres-CT98: I made the change but it looks like now all return values are of type string.

JiuqingSong (approved):

packages-content-model/roosterjs-content-model-types/lib/parameter/DOMHelper.ts
JiuqingSong: hmm you are right. But I think it is by design. When you go to the declaration of CssStyleDeclaration, the type of its members are really string:

image

JiuqingSong: And that also makes sense. Is there any example of CSS style value that is not string?

Andres-CT98: The one I can think of related to the editor is z-index, which is a number.

JiuqingSong: According to its definition, it is also string:

image

JiuqingSong: And according to https://www.w3schools.com/jsref/prop_style_zindex.asp, the javascript property of zIndex is also assigned as string:

image

So, kind of per standard.

Andres-CT98: Seeing as it is easy to convert string to number, I think it can stay this way.

PR #2420, opened 2024-02-15
Handle enter on expanded range 

juliaroldi:
When enter on a list with expanded selection, a new list item must not be created. The list just need to be adjusted to have correct style and order. 

Expected behavior:
listExpanded selection
adjustCursorList
expandedLIstEnter


JiuqingSong:
After press Enter when select two list items, we should put focus at the beginning of the second line, but not the end of the previous line

JiuqingSong (approved):

PR #2440, opened 2024-02-26
Take snapshot before paste

juliaroldi:
Instead of check if should store undo snapshot in `formatContentModel` function, add a more rigorous checking in `addUndoSnapshot` function. 
The snapshot will be stores if: 
1. It is the first snapshot
2. It has HTML different from previous one
3. It has a different entityState from the previous one
4. it has a entity state and the previous one does not has a entity state 

Before: 
bugFocus

After:
fixedFocus

### DIFF
--- a/packages-content-model/roosterjs-content-model-core/lib/coreApi/paste.ts
+++ b/packages-content-model/roosterjs-content-model-core/lib/coreApi/paste.ts
@@ -1,3 +1,4 @@
+import { addUndoSnapshot } from './addUndoSnapshot';
 import { cloneModel } from '../publicApi/model/cloneModel';
 import { convertInlineCss } from '../utils/convertInlineCss';
 import { createPasteFragment } from '../utils/paste/createPasteFragment';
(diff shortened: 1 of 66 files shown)


packages-content-model/roosterjs-content-model-core/lib/coreApi/paste.ts
```diff
@@ -30,6 +31,7 @@ export const paste: Paste = (
     pasteType: PasteType = 'normal'
 ) => {
     core.api.focus(core);
+    addUndoSnapshot(core, false);
```
JiuqingSong, line 34: we should have already taken undo snapshot when call `formatContentModel` from `mergePasteContent`, so I don't think we need to add undo snapshot here.

packages-content-model/roosterjs-content-model-core/lib/editor/SnapshotsManagerImpl.ts
```diff
@@ -129,3 +125,13 @@ class SnapshotsManagerImpl implements SnapshotsManager {
 export function createSnapshotsManager(snapshots?: Snapshots): SnapshotsManager {
     return new SnapshotsManagerImpl(snapshots);
 }
+
+function shouldAddSnapshot(currentSnapshot: Snapshot, snapshot: Snapshot) {
```
JiuqingSong, line 129: let's add some test case to cover this part

JiuqingSong (approved):

PR #2455, opened 2024-02-28
Unselect image when CTRL + A

juliaroldi:
When a image is selected, if ctrl + a is pressed, unselect the image and select the whole content.

### DIFF
--- a/packages-content-model/roosterjs-content-model-core/lib/corePlugin/SelectionPlugin.ts
+++ b/packages-content-model/roosterjs-content-model-core/lib/corePlugin/SelectionPlugin.ts
@@ -125,26 +125,21 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
                 selection = this.editor.getDOMSelection();
 
                 if (
-                    !isModifierKey(rawEvent) &&
                     !rawEvent.shiftKey &&
                     selection?.type == 'image' &&
                     selection.image.parentNode
                 ) {
-                    if (key === 'Escape') {
+                    if (key === 'Escape' && !isModifierKey(rawEvent)) {
                         this.selectBeforeImage(this.editor, selection.image);
                         event.rawEvent.stopPropagation();
-                    } else if (key !== 'Delete' && key !== 'Backspace') {
+                    } else if (
+                        (key !== 'Delete' && key !== 'Backspace' && !isModifierKey(rawEvent)) ||
+                        (key !== 'Escape' &&
+                            isModifierKey(rawEvent) &&
+                            !this.getClickingImage(rawEvent))
+                    ) {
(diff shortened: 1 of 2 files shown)


JiuqingSong:
With this change I can't copy image using Ctrl+C

JiuqingSong:
Suggest we unselect image according to selection change (browser selectionchange event) rather than the keyboard event

packages-content-model/roosterjs-content-model-core/lib/corePlugin/SelectionPlugin.ts
```diff
@@ -221,6 +223,16 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
             if (newSelection?.type == 'range') {
                 this.state.selection = newSelection;
             }
```
JiuqingSong, line 225: Let's only do the original part for Safari, to reduce risk

packages-content-model/roosterjs-content-model-core/lib/corePlugin/SelectionPlugin.ts
```diff
@@ -142,6 +143,7 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
     }
 
     private selectImage(editor: IEditor, image: HTMLImageElement) {
+        this.suspendSelectionChangeEvent = true;
         editor.setDOMSelection({
             type: 'image',
             image: image,
```
JiuqingSong, line 149: Can we set suspendSelectionChangeEvent = false after here?

juliaroldi: if we set false after here, it the selection will be removed when selectionchanged happened

packages-content-model/roosterjs-content-model-core/lib/corePlugin/SelectionPlugin.ts
```diff
@@ -221,6 +223,18 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
             if (newSelection?.type == 'range') {
                 this.state.selection = newSelection;
             }
+
```
JiuqingSong, line 226: actually, I mean for line 223-225, it should only be run for Safari. Then for 227-235, it should be run for all

JiuqingSong: Please refer to comment at line 219

JiuqingSong: And this part should also not run when suspendSelectionChangeEvent.

Better add this check at beginning of this function

PR #2458, opened 2024-02-29
Remove Bg Color from block on Merge Paste

BryanValverdeU:
Remove the background color of the source blocks when the mergeType is keepSourceEmphasisFormat.

### DIFF
--- a/packages-content-model/roosterjs-content-model-core/lib/publicApi/model/mergeModel.ts
+++ b/packages-content-model/roosterjs-content-model-core/lib/publicApi/model/mergeModel.ts
@@ -343,6 +343,8 @@ function applyDefaultFormat(
     applyDefaultFormatOption: 'mergeAll' | 'keepSourceEmphasisFormat'
 ) {
     group.blocks.forEach(block => {
+        mergeBlockFormat(applyDefaultFormatOption, block);
+
         switch (block.blockType) {
             case 'BlockGroup':
                 if (block.blockGroupType == 'ListItem') {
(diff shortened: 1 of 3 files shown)


packages-content-model/roosterjs-content-model-core/lib/publicApi/model/mergeModel.ts
```diff
@@ -384,6 +386,16 @@ function applyDefaultFormat(
     });
 }
 
+function mergeBlockFormat(applyDefaultFormatOption: string, block: ContentModelBlock) {
+    if (
+        applyDefaultFormatOption == 'keepSourceEmphasisFormat' &&
+        (block.blockType != 'BlockGroup' || block.blockGroupType != 'ListItem') &&
```
JiuqingSong, line 392: why keep background color for list item?

BryanValverdeU: For list item we do not have background color in the format interface. 
I think we never parse it, but let be double check.

BryanValverdeU: Yes for ContentModelListItemFormat, we are not extending BackgroundColorFormat.

JiuqingSong: How about we add BackgroundColorFormat into ContentModelListItemFormat?

BryanValverdeU: Added BackgroundColor Format to ContentModelListItemFormat as suggested.

JiuqingSong (approved):

PR #2461, opened 2024-03-01
Add Quote features to Content Model

juliaroldi:
Add `deleteEmptyQuote` function, it will delete an empty quote when press enter or backspace key. 
emptyQuoteDelete


JiuqingSong:
Use the snapshot below, press ENTER, it should quit the quote (please compare to old behavior)

```html
<blockquote style="padding-left: 10px; border-left: 3px solid rgb(200, 200, 200); color: rgb(102, 102, 102);"><div>aaa</div><div><br></div></blockquote><!--{"type":"range","start":[0,1,0],"end":[0,1,0],"isReverted":false,"isDarkMode":false}-->
```

JiuqingSong (approved):

PR #2463, opened 2024-03-01
Add Outdent and Indent with Arrow + Shift + Alt key  

juliaroldi:
Before we are using only Tab key to set indentation to a list, with this change will be possible to indent/outdent a list pressing 
Arrow + Shift + Alt.

Changes:
1. Rename keyboardTab file to keyboardIndentation and all other files, replace key with indentation. 
2. When Arrow Left + Shift + Alt, indent the list 
3. When Arrow Right + Shift + Alt, outdent the list 
ArrowShiftAlt


packages-content-model/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
```diff
@@ -87,7 +87,9 @@ export class EditPlugin implements EditorPlugin {
                     break;
 
                 case 'Tab':
-                    keyboardTab(editor, rawEvent);
+                case 'ArrowRight':
+                case 'ArrowLeft':
```
JiuqingSong, line 91: Can we put this into the new ShortcutPlugin instead?

PR #2471, opened 2024-03-04
Stop using default export

JiuqingSong:
The problem of `export default` is, when import, if the imported name is not the same with export name, it is valid, so when we want to do full text search, we may not be able to find those imports with different names. So let's stop using export default in Content Model code.


juliaroldi (approved):
Are we doing this changes for internal functions too?

JiuqingSong:
> Are we doing this changes for internal functions too?

For all exported functions, internal and external

PR #2472, opened 2024-03-05
When setDOMSelection is called with Table or image selection also call scrollIntoView

BryanValverdeU:
When setDOMSelection is called, also call scrollIntoView HTMLelement function to scroll to the selected element. Otherwise in some scenarios, like copying a image or a table selection, the scroll is lost and returns to the top of the editor.

### DIFF
--- a/packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection.ts
+++ b/packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection.ts
@@ -44,6 +44,7 @@ export const setDOMSelection: SetDOMSelection = (core, selection, skipSelectionC
                 core.selection.selection = selection;
 
                 setRangeSelection(doc, image);
+                image.scrollIntoView();
                 break;
             case 'table':
                 const { table, firstColumn, firstRow } = selection;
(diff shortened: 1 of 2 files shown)


packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection.ts
```diff
@@ -44,6 +44,7 @@ export const setDOMSelection: SetDOMSelection = (core, selection, skipSelectionC
                 core.selection.selection = selection;
 
                 setRangeSelection(doc, image);
+                image.scrollIntoView();
```
JiuqingSong, line 47: Can this work for docking scenario? 

We may need to do our own implementation to work together with getVisibleViewPort() function

BryanValverdeU: I checked the change and I do not see a particular problem with the scroll functionality in the Docking Plugin.

About implementing getVisibleViewPort, you mean that if we scroll to the element but it is still hidden by any sticky element in the docking functionality, we should still scroll the Image/Table into the visible viewport of the editor?

JiuqingSong: yes. That is what I mean.

BryanValverdeU: Good point, I agree I will add this too.

BryanValverdeU:
Closing this for now. Will re open later

PR #2473, opened 2024-03-05
Add API to switch logical root

florian-msft:
Add API `setLogicalRoot` to set logical root and send `snapshotLogicalRoot` event to plugins when this happens and allow then store entity states.


packages/roosterjs-content-model-core/lib/coreApi/setLogicalRoot.ts
```diff
@@ -0,0 +1,36 @@
...
+        if (logicalRoot !== core.logicalRoot) {
+            // make sure the old logical root is not content editable and the new one is
+            core.logicalRoot.contentEditable = 'false';
+            logicalRoot.contentEditable = 'true';
+
+            // update the logical root
+            core.logicalRoot = logicalRoot;
+
+            // tell plugins in case they need to clear their caches
+            const event: ContentChangedEvent = {
```
JiuqingSong, line 27: Suggest we have a separate event type for logical root changed. e.g. LogicalRootChangedEvent

packages/roosterjs-content-model-core/lib/coreApi/addUndoSnapshot.ts
```diff
@@ -11,14 +15,25 @@ import type { AddUndoSnapshot, Snapshot } from 'roosterjs-content-model-types';
...
-    const { lifecycle, physicalRoot, undo } = core;
+    const { lifecycle, physicalRoot, logicalRoot, undo } = core;
     let snapshot: Snapshot | null = null;
 
     if (!lifecycle.shadowEditFragment) {
         // Need to create snapshot selection before retrieve innerHTML since HTML can be changed during creating selection when normalize table
         const selection = createSnapshotSelection(core);
         const html = physicalRoot.innerHTML;
 
+        if (!entityStates && core.physicalRoot !== core.logicalRoot) {
```
JiuqingSong, line 26: I noticed a problem here. When we take snapshot, the logical root is a different element. Then user move logical root back to physical root, then undo. How can we know if we should change the logical root back when restore undo snapshot, and where to change to?

JiuqingSong: Or maybe we don't need to know, and let plugin handle it?

JiuqingSong: And I think we only need to trigger this event when logical root is under an entity (or it is entity), right? If is not under entity, the default undo behavior should be enough.

In that case, we can merge this event into EntityOperation event with a new EntityOperation type.

packages/roosterjs-editor-types/lib/enum/PluginEventType.ts
```diff
@@ -129,4 +129,11 @@ export const enum PluginEventType {
      * This is only used by Content Model editing
      */
     BeforeKeyboardEditing = 23,
+
+    /**
+     * EXPERIMENTAL FEATURE
+     * A snapshot of logical root is about to be taken. Entities should return entity states they wish to include in the snapshot.
+     * This is only used by Content Model editing
+     */
+    SnapshotLogicalRoot = 24,
```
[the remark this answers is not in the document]
florian-msft: Hmmm, if I do not add the type here, I get an error in `EventTypeMap` for `ContentModelEventViewPane`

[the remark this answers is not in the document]
JiuqingSong: ContentModelEventViewPane will be deleted here https://github.com/microsoft/roosterjs/pull/2468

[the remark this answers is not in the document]
florian-msft: Ah, thank you for linking, I was working on some other demo related issues and will wait until that PR is merged in to make the changes

florian-msft:
Replaced by #2492

Issue #2474, opened 2024-03-05
Test suite tableResizerTest is unstable


JiuqingSong:
Got different kinds of failure randomly:

![image](https://github.com/microsoft/roosterjs/assets/23065085/0ebd54eb-0121-474b-961d-3053e14d5817)

```
Chrome 122.0.0.0 (Windows 10) Table Resizer tests increases the width of the table FAILED
        Uncaught Error: Editor is already disposed thrown
        Error: Editor is already disposed
            at Editor.getCore (webpack://roosterjs/packages/roosterjs-content-model-core/lib/editor/Editor.ts:381:19 <- karma.test.all.1493455491.js:37466:19)
            at Editor.getVisibleViewport (webpack://roosterjs/packages/roosterjs-content-model-core/lib/editor/Editor.ts:372:21 <- karma.test.all.1493455491.js:37458:21)
            at isTableBottomVisible (webpack://roosterjs/packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableResizer.ts:237:36 <- karma.test.all.1493455491.js:107224:34)
            at Object.onDragEnd (webpack://roosterjs/packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableResizer.ts:203:9 <- karma.test.all.1493455491.js:107203:9)
            at DragAndDropHelper.onMouseUp (webpack://roosterjs/packages/roosterjs-content-model-plugins/lib/pluginUtils/DragAndDrop/DragAndDropHelper.ts:145:35 <- karma.test.all.1493455491.js:105733:95)
            at moveAndResize (webpack://roosterjs/packages/roosterjs-content-model-plugins/test/tableEdit/TableEditTestHelper.ts:213:19 <- karma.test.all.1493455491.js:135002:19)
            at resizeWholeTableTest (webpack://roosterjs/packages/roosterjs-content-model-plugins/test/tableEdit/tableResizerTest.ts:105:22 <- karma.test.all.1493455491.js:136594:49)
            at UserContext.<anonymous> (webpack://roosterjs/packages/roosterjs-content-model-plugins/test/tableEdit/tableResizerTest.ts:175:9 <- karma.test.all.1493455491.js:136642:9)
            at <Jasmine>
        Error: Expected false to be true.
            at <Jasmine>
            at webpack://roosterjs/packages/roosterjs-content-model-plugins/test/tableEdit/tableResizerTest.ts:167:21 <- karma.test.all.1493455491.js:136636:97
[...]

PR #2475, opened 2024-03-05
Temporarily disable unstable test suite

JiuqingSong:
The test fails randomly, so disable it for now. Opened #2474 to track the issue.


juliaroldi (approved):

Andres-CT98:
Does it fail randomly on Github too or only locally?

JiuqingSong:
> Does it fail randomly on Github too or only locally?

I saw it happened both in github and locally. Usually rerun it can pass, but still annoying.

JiuqingSong:
> Does it fail randomly on Github too or only locally?

Here is an example how it fails in github:
https://github.com/microsoft/roosterjs/actions/runs/8160318276/job/22306656056?pr=2476

PR #2476, opened 2024-03-05
Add check for selected segments and paragraphs 

juliaroldi:
In case we press space and  `getSelectedSegmentsAndParagraphs` return no selected segments and paragraphs, `getListTypeStyle` function must return undefined. Scenarios for entity selection and nested editor can make `getSelectedSegmentsAndParagraphs` return an empty array [].


packages/roosterjs-content-model-plugins/lib/autoFormat/utils/getListTypeStyle.ts
```diff
@@ -31,6 +31,9 @@ export function getListTypeStyle(
     shouldSearchForNumbering: boolean = true
 ): ListTypeStyle | undefined {
     const selectedSegmentsAndParagraphs = getSelectedSegmentsAndParagraphs(model, true);
+    if (!selectedSegmentsAndParagraphs[0]) {
```
BryanValverdeU, line 34: Should we add a unit test for this scenario?

JiuqingSong:
Can you add more description what issue this PR fix?

JiuqingSong (approved):

JiuqingSong (approved):

PR #2479, opened 2024-03-05
[Part 1] Port Auto Link 

juliaroldi:
Port `AutoLink` and `UnlinkWhenBackspaceAfterLink` content edit features into `AutoFormatPlugin`
After porting `AutoLink`, `AutoFormatPlugin` will handle automatically create link from pasted url segments, and after porting `UnlinkWhenBackspaceAfterLink`, will automatically remove the link of text segment, if backspace key is pressed when the cursor is after a link. 

autoLink

Another PR will be created to handle AutoLink while typing.


packages/roosterjs-content-model-api/lib/modelApi/link/matchLink.ts
```diff
@@ -1,7 +1,7 @@
 import { getObjectKeys } from 'roosterjs-content-model-dom';
 
 /**
- * @internal
+ * Data of the matched link
  */
 export interface LinkData {
```
JiuqingSong, line 6: Move this to roosterjs-content-model-types/lib/parameters

packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
```diff
@@ -27,6 +40,8 @@ export type AutoFormatOptions = {
 const DefaultOptions: Required<AutoFormatOptions> = {
     autoBullet: true,
     autoNumbering: true,
+    autoUnlink: true,
```
JiuqingSong, line 43: Suggest set autoUnlink to false by default since we didn't use it

packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
```diff
@@ -81,21 +96,37 @@ export class AutoFormatPlugin implements EditorPlugin {
...
+                        unlink(editor, rawEvent);
+                    }
+
                     break;
             }
         }
     }
+
+    private handleContentChangedEvent(editor: IEditor, event: ContentChangedEvent) {
+        const { autoLink } = this.options;
```
JiuqingSong, line 127: The original auto link feature also work when press SPACE or ENTER

JiuqingSong: I see you put it in part 2. that is ok.

PR #2485, opened 2024-03-06
[Part 2] Port Auto link 

juliaroldi:
This change is part of the effort to port the Auto Link from the legacy ContentEdit feature into `AutoFormatPlugin`.  After this change `AutoFormatPlugin` will automatically transform a typed url when space key is pressed after the text segment.
AutoLinkTyping


packages/roosterjs-content-model-plugins/lib/autoFormat/link/createLink.ts
```diff
@@ -5,7 +5,10 @@ import type { IEditor } from 'roosterjs-content-model-types';
 /**
  * @internal
  */
-export function createLink(editor: IEditor) {
+export function createLink(editor: IEditor, autoLink: boolean) {
+    if (!autoLink) {
```
JiuqingSong, line 9: same here

packages/roosterjs-content-model-plugins/lib/autoFormat/list/keyboardListTrigger.ts
```diff
@@ -13,26 +13,28 @@ export function keyboardListTrigger(
...
-        if (listStyleType) {
-            const segmentsAndParagraphs = getSelectedSegmentsAndParagraphs(model, false);
-            if (segmentsAndParagraphs[0] && segmentsAndParagraphs[0][1]) {
-                segmentsAndParagraphs[0][1].segments.splice(0, 1);
-            }
-            const { listType, styleType, index } = listStyleType;
-            triggerList(editor, model, listType, styleType, index);
-            rawEvent.preventDefault();
-            normalizeContentModel(model);
+    if (shouldSearchForBullet || shouldSearchForNumbering) {
```
JiuqingSong, line 16: same here, check outside to reduce unnecessary parameters.

juliaroldi: We already use these parameter on line 20 and 21, then we cannot remove them

juliaroldi: Should I return this check to outside the function anyway?

JiuqingSong: I see. In that case we can keep it

JiuqingSong (approved):

PR #2492, opened 2024-03-08
Add setLogicalRoot API

florian-msft:
Add API `setLogicalRoot` to set logical root and send `snapshotLogicalRoot` event to plugins when this happens and allow then store entity states.


demo/scripts/controlsV2/sidePane/snapshot/SnapshotPane.tsx
```diff
@@ -73,6 +73,7 @@ export class SnapshotPane extends React.Component<SnapshotPaneProps, SnapshotPan
             entityStates,
             isDarkMode,
             selection,
+            logicalRootPath: [],
```
florian-msft, line 76: I didn't add functionality to show this in the pane, please let me know if I should

packages/roosterjs-content-model-types/lib/event/SnapshotLogicalRootEvent.ts
```diff
@@ -0,0 +1,20 @@
+import type { EntityState } from '../parameter/FormatContentModelContext';
+import type { BasePluginEvent } from './BasePluginEvent';
+
+/**
+ * Undo snapshot event when a custom logical root is used (logicalRoot != physicalRoot)
+ * This event is triggered when an undo snapshot is taken
+ * Plugins can handle this event to include entity state to include in the snapshot
+ */
+export interface SnapshotLogicalRootEvent extends BasePluginEvent<'snapshotLogicalRoot'> {
```
JiuqingSong, line 9: Merge this event into EntityOperation event.

packages/roosterjs-content-model-types/lib/event/LogicalRootChangedEvent.ts
```diff
@@ -0,0 +1,6 @@
+import type { BasePluginEvent } from './BasePluginEvent';
+
+/**
+ * Fired when the logical root changes
+ */
+export interface LogicalRootChangedEvent extends BasePluginEvent<'logicalRootChanged'> {}
```
JiuqingSong, line 6: Better pass the new logicalRoot with the event object so plugin knows where it has been changed to.

packages/roosterjs-content-model-types/lib/event/SnapshotLogicalRootEvent.ts
JiuqingSong: And use a better name since the event is just to tell plugin we are adding undo snapshot under entity, plugin can generate its own undo snapshot, and can also do other things.

packages/roosterjs-content-model-core/lib/coreApi/addUndoSnapshot.ts
```diff
@@ -11,19 +16,35 @@ import type { AddUndoSnapshot, Snapshot } from 'roosterjs-content-model-types';
...
     let snapshot: Snapshot | null = null;
 
     if (!lifecycle.shadowEditFragment) {
         // Need to create snapshot selection before retrieve innerHTML since HTML can be changed during creating selection when normalize table
         const selection = createSnapshotSelection(core);
         const html = physicalRoot.innerHTML;
 
+        if (!entityStates) {
+            // give plugins the chance to share entity states to include in the snapshot
+            const event = <SnapshotLogicalRootEvent>{
```
JiuqingSong, line 29: only trigger this event when logical root is under an entity, or itself is entity. Because in >99.999% case we don't need to trigger this event

packages/roosterjs-content-model-core/lib/coreApi/restoreUndoSnapshot.ts
```diff
@@ -23,8 +24,28 @@ export const restoreUndoSnapshot: RestoreUndoSnapshot = (core, snapshot) => {
...
+        // restore logical root if needed
+        if (snapshot.logicalRootPath && snapshot.logicalRootPath.length > 0) {
+            const restoredLogicalRoot = getPositionFromPath(
+                core.physicalRoot,
+                snapshot.logicalRootPath
+            ).node as HTMLDivElement;
+            if (restoredLogicalRoot !== core.logicalRoot) {
+                core.api.setLogicalRoot(core, restoredLogicalRoot);
+            }
+        }
```
JiuqingSong, line 42: Move this out to be a new function `restoreLogicalRoot`

packages/roosterjs-content-model-core/lib/coreApi/restoreUndoSnapshot.ts
```diff
@@ -23,8 +24,28 @@ export const restoreUndoSnapshot: RestoreUndoSnapshot = (core, snapshot) => {
...
+
+        // restore selection and colors
+        try {
```
JiuqingSong, line 45: Suggest move the try ... catch into function `restoreSnapshotSelection`

packages/roosterjs-content-model-core/lib/coreApi/restoreUndoSnapshot.ts
```diff
@@ -23,8 +24,28 @@ export const restoreUndoSnapshot: RestoreUndoSnapshot = (core, snapshot) => {
     try {
         core.undo.isRestoring = true;
 
+        // reset logical root to physical root
+        core.api.setLogicalRoot(core, null);
```
JiuqingSong, line 28: I'm reconsidering about this, maybe no need to set to root here, but always set to correct logical root below when restoreLogicalRoot.

For example:

```ts
function restoreLogicalRoot() {
   const newLogicalRoot = snapshot.logicalRootPath?.length > 0 ? getNodeFromPath(snapshot.logicalRootPath) : core.physicalRoot;
   setLogicalRoot(newLogicalRoot);
}
```

demo/scripts/controls/sidePane/eventViewer/EventViewPane.tsx
```diff
@@ -62,6 +62,7 @@ const EntityOperationMap: { [key in EntityOperation]: string } = {
     [EntityOperation.RemoveFromStart]: 'RemoveFromStart',
     [EntityOperation.ReplaceTemporaryContent]: 'ReplaceTemporaryContent',
     [EntityOperation.UpdateEntityState]: 'UpdateEntityState',
+    [EntityOperation.SnapshotEntityState]: 'SnapshotEntityState',
```
JiuqingSong, line 65: As I said, no need to add this new event type for old editor. This should be a new feature for new editor only. Old editor should never receive this event.

demo/scripts/controlsV2/sidePane/snapshot/SnapshotPane.tsx
JiuqingSong: You may need to add a text box to show/add the selection path. But this is low pri compared to the main feature.

packages/roosterjs-content-model-core/lib/coreApi/addUndoSnapshot/addUndoSnapshot.ts
```diff
@@ -11,19 +19,52 @@ import type { AddUndoSnapshot, Snapshot } from 'roosterjs-content-model-types';
...
+    const { lifecycle, physicalRoot, logicalRoot, undo } = core;
     let snapshot: Snapshot | null = null;
 
     if (!lifecycle.shadowEditFragment) {
         // Need to create snapshot selection before retrieve innerHTML since HTML can be changed during creating selection when normalize table
         const selection = createSnapshotSelection(core);
         const html = physicalRoot.innerHTML;
 
+        // give plugins the chance to share entity states to include in the snapshot
+        const logicalRootEntityWrapper = logicalRoot.closest<HTMLElement>(`.${ENTITY_INFO_NAME}`);
```
JiuqingSong, line 31: Use `core.domHelper.findClosestElementAncestor` instead of "closest", it will help you find element within editor only

packages/roosterjs-content-model-core/lib/coreApi/addUndoSnapshot/addUndoSnapshot.ts
```diff
@@ -1,5 +1,13 @@
...
+import type {
+    AddUndoSnapshot,
+    EntityOperationEvent,
+    Snapshot,
+} from 'roosterjs-content-model-types';
 import { createSnapshotSelection } from './createSnapshotSelection';
-import type { AddUndoSnapshot, Snapshot } from 'roosterjs-content-model-types';
+import { getPath } from './getPath';
+
+const ENTITY_INFO_NAME = '_Entity';
```
JiuqingSong, line 10: Consider export this const from entityUtils.ts, so we just have one single source of this const

JiuqingSong: And mark the export as `@internal`

packages/roosterjs-content-model-core/lib/coreApi/addUndoSnapshot/addUndoSnapshot.ts
```diff
@@ -11,19 +19,52 @@ import type { AddUndoSnapshot, Snapshot } from 'roosterjs-content-model-types';
...
+                }
+            }
+        }
+
         snapshot = {
             html,
             entityStates,
             isDarkMode: !!lifecycle.isDarkMode,
             selection,
+            logicalRootPath: getPath(logicalRoot, 0, physicalRoot),
```
JiuqingSong, line 67: What is the value of logicalRootPath when logicalRoot==physicalRoot? I think it will be an empty array? Please double check.

We can make a perf optimization here. If we know they are the same, just directly write "[]" here and no need to call getPath()

packages/roosterjs-content-model-core/lib/coreApi/setLogicalRoot/setLogicalRoot.ts
```diff
@@ -0,0 +1,40 @@
+import type { LogicalRootChangedEvent, SetLogicalRoot } from 'roosterjs-content-model-types';
+
+/**
+ * Change which node is the current logical root
+ * @param core The StandaloneEditorCore object
+ * @param logicalRoot The new logical root (has to be child of physicalRoot), pass null to use physicalRoot as logical root
+ */
+export const setLogicalRoot: SetLogicalRoot = (core, logicalRoot) => {
```
JiuqingSong, line 8: Need unit test for this file, as well as any other new added functions, if any

packages/roosterjs-editor-types/lib/enum/EntityOperation.ts
```diff
@@ -74,4 +74,10 @@ export const enum EntityOperation {
      * This is normally happened when user undo/redo the content with an entity snapshot added by a plugin that handles entity
      */
     UpdateEntityState,
+
+    /**
+     * Notify plugins that an undo snapshot is taken while a custom logical root is used
+     * Plugins can handle this event to include entity state to include in the snapshot
+     */
+    SnapshotEntityState,
```
JiuqingSong, line 82: No need to add this. When we receive the this new event in eventConverter, just ignore it so old plugin won't receive this event.

demo/scripts/controls/sidePane/eventViewer/EventViewPane.tsx
florian-msft: If I do not include it, I get this error:

```
[tsl] ERROR in C:\Users\fjanke\Documents\code\roosterjs\demo\scripts\controls\sidePane\eventViewer\EventViewPane.tsx(52,7)
      TS2741: Property '12' is missing in type '{ 9: string; 10: string; 1: string; 2: string; 3: string; 0: string; 6: string; 7: string; 5: string; 4: string; 8: string; 11: string; }' but required in type '{ 0: string; 1: string; 2: string; 3: string; 4: string; 5: string; 6: string; 7: string; 8: string; 9: string; 10: string; 11: string; 12: string; }'.
```

because `EntityOperation` has `SnapshotEntityState` I need a matching mapping in `EntityOperationMap`

packages/roosterjs-content-model-core/lib/coreApi/addUndoSnapshot/addUndoSnapshot.ts
```diff
@@ -11,19 +17,54 @@ import type { AddUndoSnapshot, Snapshot } from 'roosterjs-content-model-types';
...
+                            id: entityFormat.id,
+                            state: event.state,
+                        },
+                    ];
+                }
+            }
+        }
+
+        const logicalRootPath =
+            logicalRoot === physicalRoot ? [0] : getPath(logicalRoot, 0, physicalRoot);
```
JiuqingSong, line 61: Are you sure it will be "[0]"? I feel it should be empty array. Can you double check?

packages/roosterjs-content-model-core/lib/coreApi/addUndoSnapshot/addUndoSnapshot.ts
```diff
@@ -11,19 +17,54 @@ import type { AddUndoSnapshot, Snapshot } from 'roosterjs-content-model-types';
...
+    const { lifecycle, physicalRoot, logicalRoot, undo } = core;
     let snapshot: Snapshot | null = null;
 
     if (!lifecycle.shadowEditFragment) {
         // Need to create snapshot selection before retrieve innerHTML since HTML can be changed during creating selection when normalize table
         const selection = createSnapshotSelection(core);
         const html = physicalRoot.innerHTML;
 
+        // give plugins the chance to share entity states to include in the snapshot if the logical root is in an entity
+        const entityWrapper = findClosestEntityWrapper(logicalRoot, core.domHelper);
```
JiuqingSong, line 29: We should also check if physicalRoot and logicalRoot are the same first. If they are the same, no need to find entity since it must return null, so save some time

packages/roosterjs-content-model-core/lib/coreApi/addUndoSnapshot/addUndoSnapshot.ts
florian-msft: Yes, I checked and it is `[0]`

packages/roosterjs-content-model-core/lib/coreApi/addUndoSnapshot/addUndoSnapshot.ts
JiuqingSong: So in general this can be optimized to:

```ts
if (phy != logical) {
  const entityWrapper = !entityStates && ...
  if (entityWrapper ) { 
     ... // trigger event
   }

  logicalRootPath = ...
}
```

For the case that two roots are the same (99.999% case), we can even skip logicalRootPath so no need to change it when restore.

packages/roosterjs-content-model-core/lib/coreApi/addUndoSnapshot/addUndoSnapshot.ts
JiuqingSong: I feel there should be a code bug in existing code. When child and parent node are the same, we should return [] for selection path. 

I think it is because this is not a normal case, so we have never seen it cause issue before. But anyway, let's just skip the path here if logical root is physical root to make sure when restore, the correct root can be set.

JiuqingSong: Ok, I just verified the selection path is correct. But still, let's do not set selection path for logical root if it is the same with physical root since we don't need to.

JiuqingSong (approved):

Issue #1652, opened 2023-03-18
RoosterJs does not recognized multiple pasted linmks 


Shtoosha:
**Describe the bug**

**To Reproduce**
1. Copy text with multiple links: http://life.com and https://fb.com and www.koko.com fcsafsdadsada
2. Past to roosterJS demo task editor
4.  Observe that links are  shown as plain text

**Expected behavior**
Links need to be clickable 

**Screenshots**
![image](https://user-images.githubusercontent.com/53274526/226086209-37acfce7-1709-43f2-8331-b145acdf8d34.png)

**Device Information**
 - Browser  Chrome, Edge

**Additional context**

JiuqingSong:
You can write your own plugin to extends this functionality. And you can also contribute back your plugin if you want. Thanks.

gm-al:
We can extend the current implementation and add another case to handle pasting plain text with clickable links. I can submit a PR for this.

PR #2494, opened 2024-03-08
Support paste plain text with clickable links in RoosterJs 

gm-al:
This is a reference for the issue #1652. This PR expands on the current implementation for handling the paste functionality in RoosterJs. This supports pasting content with plain text and clickable links.


packages/roosterjs-editor-dom/lib/clipboard/handleTextPasteWithLinks.ts
```diff
@@ -0,0 +1,115 @@
...
+    });
+}
+
+/**
+ * @internal
+ * Transform \t characters into EN SPACE characters
+ * @param input String NOT containing \n characters
+ * @example t("\thello", 2) => "&ensp;&ensp;&ensp;&ensp;hello"
+ */
+export function transformTabCharacters(input: string, initialOffset: number = 0) {
```
gm-al, line 62: Would it be a good thing to move functions that are used in multiple places like `transformTabCharacters` to the utils folder and make it a common function to be used in `handleTextPasteWithClickableLinks` and `handleTextPaste`? Thank you :).

packages/roosterjs-content-model-types/lib/enum/PasteType.ts
```diff
@@ -20,4 +20,9 @@ export type PasteType =
     /**
      * If there is a image uri in the clipboard, paste the content as image element
      */
-    | 'asImage';
+    | 'asImage'
+
+    /**
+     * If there is plain text with links in the clipboard, paste the content as plain text with clickable links
+     */
```
JiuqingSong, line 27: Ok I think now I understand what you mean. You want to paste as plain text to remove all formats, but only keep links?

I don't know if it is a good idea. But I feel this should be done in your own code, but not as part of roosterjs.

gm-al: Yes, I thought to help submit a PR to add a new option where links can appear in plain text/clickable. If you don't think this is a good idea to continue, I can close this PR?

I thought when you mentioned in the issue that he could write his own plugin and contribute back- this option could be a good thing to have in roosterjs  :).

JiuqingSong: please feel free to close

PR #2497, opened 2024-03-08
[Entity Features] Handle Enter

juliaroldi:
Handle some edge cases for enter key behavior on entity and add `ClickOnEntityFeature` for keyboard event

Edge cases fixed: 
1) Before enitity when is the first element of segment 
2) After enitity when is the last element of segment 

TEST_ENTITY

| Feature    | State |
| -------- | ------- |
| `ClickOnEntityFeature` | Added    |
| `EscapeFromEntityFeature` | Removed     |
| `EnterBeforeReadonlyEntityFeature`    |  Feature already handled by Entity Plugin, only edge case was fixed   |
| `DeleteBeforeEntityFeature`    | Feature already handled by Entity Plugin, only edge case was fixed     |
| `MoveBetweenDelimitersFeature`    | Feature already handled by Entity Plugin    |
| `RemoveEntityBetweenDelimitersFeature`    | Feature already handled by Entity Plugin    |


JiuqingSong:
What content edit features can be turned off after this change?

packages/roosterjs-content-model-core/lib/corePlugin/entity/entityDelimiterUtils.ts
```diff
@@ -241,6 +243,14 @@ export function handleDelimiterKeyDownEvent(editor: IEditor, event: KeyDownEvent
                 }
             }
         }
+    } else {
+        const entity = getSelectedEntity(selection);
+        if (entity && isNodeOfType(entity, 'ELEMENT_NODE') && helper.isNodeInEditor(entity)) {
+            selection.range.selectNode(entity);
+            if (isEnter) {
+                triggerEntityEventOnEnter(editor, entity, rawEvent);
```
JiuqingSong, line 251: If this is only for ENTER key, check ENTER key first, before searching the entity. That will have better performance

juliaroldi:
> What content edit features can be turned off after this change?

After that change, we only need to port markdown features

packages/roosterjs-content-model-core/lib/corePlugin/entity/entityDelimiterUtils.ts
juliaroldi: The entity is being searched to be selected when the range is expanded, then when we pressed any key, the key pressed can be inserted in the range of the entity. This is was being handle by `EnterBeforeReadonlyEntityFeature`, but was not handle by the EntityPlugin.

JiuqingSong: hmm. ok. Is this moved from some old code? If so, where is the old code?

juliaroldi: it is not moved from old code, but here is the code to remove the expanded selection 
Screenshot 2024-03-19 at 14 21 58

juliaroldi: Maybe we should try only delete the selection content and do not prevent default behavior. I'll try this.

packages/roosterjs-content-model-core/lib/corePlugin/entity/entityDelimiterUtils.ts
```diff
@@ -303,3 +320,31 @@ export const handleEnterInlineEntity: ContentModelFormatter = model => {
...
+                type: format.entityType,
+                isReadonly: !!format.isReadonly,
+                wrapper,
+            },
+            rawEvent: rawEvent,
+        });
+    }
+};
+
+const getSelectedEntity = (selection: RangeSelection) => {
```
JiuqingSong, line 344: In PR https://github.com/microsoft/roosterjs/pull/2492, @florian-msft  is adding a method to get current entity from a given node. I think you can wait and leverage the function

JiuqingSong: image

JiuqingSong: @juliaroldi the PR is merged, you can use the new API now.

packages/roosterjs-content-model-dom/lib/domUtils/entityUtils.ts
```diff
@@ -131,6 +132,15 @@ export function addDelimiters(
     return [delimiterAfter, delimiterBefore];
 }
 
+/**
+ * Find the closest entity wrapper element from a given DOM node
+ * @param node The node to start looking for entity wrapper
+ * @param domHelper The DOM helper to use
+ */
+export function findClosestEntityWrapper(node: Node, domHelper: DOMHelper): HTMLElement | null {
```
JiuqingSong, line 140: there is already such function, please pull to latest and resolve the conflict

JiuqingSong (approved):

PR #2503, opened 2024-03-12
FHL: Port WatermarkPlugin

JiuqingSong:
Port WatermarkPlugin to use Content Model editor, also change its behavior to show watermark as long as editor is empty, even editor has focus.

image

In order to show watermark, add a new core API `setEditorStyle`, it manages global STYLE node for editor DIV and its child nodes. We can now use this API to add/remove global CSS rule for editor. For example:

```ts
editor.setEditorStyle('StyleKey', 'color: red');
```
This will add a global style "color:red" for editor DIV

```ts
editor.setEditorStyle('StyleKey', 'color: red', ['#childNode1', '#childNode2']);
```
This will add a global style "color:red" for child node with id "childNode1" and "childNode2" under editor

```ts
editor.setEditorStyle('StyleKey', 'color: red', 'before');
```
This will add a global style "color:red" for editor DIV with pseudo class "::before".

Also change existing code `setDOMSelection` and `FormatPainterPlugin` to reuse this new API.

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/watermark/WatermarkPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/watermark/WatermarkPlugin.ts
@@ -1,0 +1,101 @@
+import { getObjectKeys } from 'roosterjs-content-model-dom';
+import { isModelEmptyFast } from './isModelEmptyFast';
+import type { WatermarkFormat } from './WatermarkFormat';
+import type { EditorPlugin, IEditor, PluginEvent } from 'roosterjs-content-model-types';
+
+const WATERMARK_CONTENT_KEY = '_WatermarkContent';
+const styleMap: Record<keyof WatermarkFormat, string> = {
+    fontFamily: 'font-family',
+    fontSize: 'font-size',
+    textColor: 'color',
+};
+
+/**
+ * A watermark plugin to manage watermark string for roosterjs
+ */
+export class WatermarkPlugin implements EditorPlugin {
+    private editor: IEditor | null = null;
+    private format: WatermarkFormat;
+    private isShowing = false;
+
+    /**
+     * Create an instance of Watermark plugin
+     * @param watermark The watermark string
+     */
+    constructor(private watermark: string | (() => string), format?: WatermarkFormat) {
+        this.format = format || {
+            fontSize: '14px',
+            textColor: '#AAAAAA',
+        };
+    }
+
+    /**
+     * Get a friendly name of  this plugin
+     */
+    getName() {
+        return 'Watermark';
+    }
+
+    /**
+     * Initialize this plugin. This should only be called from Editor
+     * @param editor Editor instance
+     */
+    initialize(editor: IEditor) {
+        this.editor = editor;
+    }
+
+    /**
+     * Dispose this plugin
+     */
+    dispose() {
+        this.editor = null;
+    }
+
+    /**
+     * Handle events triggered from editor
+     * @param event PluginEvent object
+     */
+    onPluginEvent(event: PluginEvent) {
+        const editor = this.editor;
+
+        if (
+            editor &&
+            (event.eventType == 'editorReady' ||
+                event.eventType == 'contentChanged' ||
+                event.eventType == 'input' ||
+                event.eventType == 'beforeDispose')
+        ) {
+            editor.formatContentModel(model => {
+                const isEmpty = isModelEmptyFast(model);
+
+                if (this.isShowing && !isEmpty) {
+                    this.hide(editor);
+                } else if (!this.isShowing && isEmpty) {
+                    this.show(editor);
+                }
+                return false;
+            });
+        }
+    }
+
+    private show(editor: IEditor) {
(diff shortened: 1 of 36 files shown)


packages/roosterjs-content-model-plugins/lib/watermark/WatermarkPlugin.ts
```diff
@@ -0,0 +1,101 @@
...
+                    this.show(editor);
+                }
+                return false;
+            });
+        }
+    }
+
+    private show(editor: IEditor) {
+        let rule = `position: absolute; pointer-events: none; content: "${
+            typeof this.watermark == 'function' ? this.watermark() : this.watermark
```
vhuseinova-msft, line 83: It looks like with this implementation, there is no way to update it when watermark is already shown as the watermark will be updated with a new value only when it will be shown next time

JiuqingSong: What is the reason we want to update the text when it is already shown?

vhuseinova-msft: For example when user changes locale

JiuqingSong: In that case, can we unmount and remount the whole editor?

Given the local is changed, I assume a lot of things can be changed, such as RTL. So we can just remount the whole editor then it will pick up the new settings.

Similarly, if user changes default format, we should also remount editor.

vhuseinova-msft: I don't think that's a good idea (because of the performance) to remount large editor component fully every time when locale is changed only for Watermark plugin...
By the way, buttons for RoosterJS react package are already updated with the text when `strings` prop is provided without a need to remound it by Contoso developer.

JiuqingSong: According how you use editor, the perf is not a big problem here. Because roosterjs itself is built with an existing DIV, so mount/unmount roosterjs does not mean you need to remove/readd the DIV, but just just remove the editor javascript object and create a new one one the original DIV. Of cause, if you are using the react wrapper `<Rooster />`, that will be different. And you can write your own wrapper component to reuse the existing DIV when language changes.

During editing, the Content Model recreate/rewrite is actually much heavier than unmount/remount editor. If the performance of that is acceptable, the unmount/remount should also be acceptable.

vhuseinova-msft: Continuation of the previous message: Also, what's a point of the callback for setting watermark text if in case when watermark needs to be updated, rooster js wants to be remounted fully.

vhuseinova-msft: `Of cause, if you are using the react wrapper <Rooster />, that will be different. And you can write your own wrapper component to reuse the existing DIV when language changes.`
Yeah, we do use RoosterJS react package and wanted to reuse it as much as possible without additional wrappers on our side.

JiuqingSong: As I mentioned before, the roosterjs-react package is supposed to used for demo and build prototype. So we didn't provide too much customization functionalities. Since our major focus is build the editor engine, but not to maintain a complex react component (actually we don't want to bind roosterjs with react too much since roosterjs should be a framework-independent editor) with all kinds of CSS and theme things (that is the strength of FluentUI). So we encourage every client to build their own UI wrapper to satisfy their special requirement, but not keep adding features into roosterjs-react since we don't have too much time to maintain it. (We have actually deprecated roosterjs-react for some time and later added it back since some client still wants it). Each client usually has very different customization to the UI component which is not our major focus.

About the callback in Watermark plugin, it just provides a way to allow calculating the text after plugin is created. I can remove it if no one really need it.

JiuqingSong: One thing I can do is to make the function show/hide watermark to be "protected" so you can create your own sub class to override some behavior and call the base class function to reuse the behavior.

vhuseinova-msft: `About the callback in Watermark plugin, it just provides a way to allow calculating the text after plugin is created. I can remove it if no one really need it.`
Yeah, it looks like it might  not cover all cases and might be more confusing than useful

`One thing I can do is to make the function show/hide watermark to be "protected" so you can create your own sub class to override some behavior and call the base class function to reuse the behavior.`
It may be useful... 
Could you also add a similar function for the previous version of the Watermark plugin? 
I have code with `update` function for the previous version but it would be nice if they will be similar for easier migration later

BryanValverdeU:
In the old watermark if the default format had bold/italic/underline, that style would be added to the watermark. Just curious if we are removing that old behavior.

BryanValverdeU (approved):

PR #2513, opened 2024-03-19
[Part 1] Port Markdown feature

juliaroldi:
Port the legacy markdown to the content model plugin, MarkdownPlugin: 
1. Set bold when type text between * 
2. Set italic when type text between _ 
3. Set strikethrough when type between ~ 

mardownPlugin

This feature will not work for segments that received format that and multiple paragraphs.


packages/roosterjs-content-model-plugins/lib/markdown/utils/setFormat.ts
```diff
@@ -0,0 +1,90 @@
...
+                                return false;
+                            }
+
+                            previousCharacterSegment.text = previousCharacterSegment.text.substring(
+                                0,
+                                previousCharacterSegment.text.length - formattedText.length - 2
+                            );
+
+                            const formattedSegment = createText(formattedText, marker.format);
+                            paragraph.segments.splice(index + 1, 0, formattedSegment);
```
JiuqingSong, line 69: If you can create a common function to help split the text segment (as I suggested in another PR), you can leverage here to make this easier

packages/roosterjs-content-model-plugins/lib/markdown/utils/setFormat.ts
```diff
@@ -0,0 +1,90 @@
...
+                                        previousSegment.text.length - 1
+                                )
+                            ) {
+                                previousCharacterSegment = previousSegment;
+                                index = i;
+                                break;
+                            }
+                        }
+                        if (previousCharacterSegment && index < markerIndex) {
+                            const splittedSegments = previousCharacterSegment.text.split(character);
```
JiuqingSong, line 56: what if the result has more than 2 segments?

For example "a*b*c*d", it will return an array of 4.

juliaroldi: It will search for the segment that has * and it will iterate through all the segments until the the second *. The for in line 38 search for the segments and the for in line 71 will apply the formatting.

packages/roosterjs-content-model-plugins/lib/markdown/utils/setFormat.ts
juliaroldi: I marked this PR as draft, since I'll will this one to merged with the `splitTextSegment` https://github.com/microsoft/roosterjs/pull/2522

packages/roosterjs-content-model-plugins/lib/markdown/utils/setFormat.ts
juliaroldi: I remove the code to handle multiple segments, since the legacy markdown do not handle multiple styles ate the same time.

JiuqingSong:
1. Type "*"
2. Backsapce
3. Type "a"
4. Type "*"
Expect: show "a*"
Actual: everything is gone.

JiuqingSong:
1. Type "*"
2. Backspace
3. Type "a b"
4. Type "*"

Expect: show "a b*"
Actual: show "a" in bold

juliaroldi:
> 1. Type "*"
> 2. Backspace
> 3. Type "a b"
> 4. Type "*"
> 
> Expect: show "a b*" Actual: show "a" in bold

This bug were happening because if we press backspace before the character the keyboard event returns default prevented as true. I create function just to handle `backspace` and it will execute even when default prevented  is true. Also and another check in setFormat to ensure the text is between the characters.

JiuqingSong (approved):

packages/roosterjs-content-model-plugins/lib/markdown/MarkdownPlugin.ts
```diff
@@ -0,0 +1,192 @@
...
+    IEditor,
+    KeyDownEvent,
+    PluginEvent,
+} from 'roosterjs-content-model-types';
+
+/**
+ * Options for Markdown plugin
+ */
+export interface MarkdownOptions {
+    strikethrough?: boolean;
```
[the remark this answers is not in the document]
juliaroldi: I added at the comment in the constructor, but I will also add in the the interface in the PR for part 2

PR #2522, opened 2024-03-24
Undo Auto link by backspace

juliaroldi:
Add `canUndoByBackspace` to `createLinkAfterSpace`. This will make possible undo the link pressing backspace. 
undoByBackspace


JiuqingSong:
The SPACE input should still take effect after convert to link.

Similar issue for auto list. Try  with the snapshot below, and press SPACE, It should become "1. [SPACE]test", but the actual result is "1. test". The existing SPACE is gone. I think this is caused by trim()

```html
<div>1. test</div><!--{"type":"range","start":[0,0,2],"end":[0,0,2],"isReverted":false,"isDarkMode":false}-->
```

packages/roosterjs-content-model-plugins/lib/autoFormat/link/createLinkAfterSpace.ts
```diff
@@ -15,22 +15,30 @@ export function createLinkAfterSpace(editor: IEditor) {
...
             if (
                 marker.segmentType == 'SelectionMarker' &&
                 textSegment.segmentType == 'Text' &&
                 !textSegment.link
             ) {
                 const link = textSegment.text.split(' ').pop();
-                if (link && matchLink(link)) {
+                const url = link?.trimRight();
+                if (url && link && matchLink(url)) {
                     textSegment.text = textSegment.text.replace(link, '');
```
JiuqingSong, line 27: In a corner case, for example, the text contains two same url strings:
www.bing.com www.bing.com

This will only replace the first one. Actually we need to replace the last one.

packages/roosterjs-content-model-plugins/lib/autoFormat/link/createLinkAfterSpace.ts
```diff
@@ -15,22 +15,30 @@ export function createLinkAfterSpace(editor: IEditor) {
...
+                    const linkSegment = createText(url, marker.format, {
                         format: {
-                            href: link,
+                            href: url,
                             underline: true,
                         },
                         dataset: {},
                     });
+
                     selectedSegmentsAndParagraphs[0][1].segments.splice(length - 1, 0, linkSegment);
```
JiuqingSong, line 36: What if cursor is not at the end of paragraph? We should use index of current selection, but not the whole length

packages/roosterjs-content-model-plugins/lib/autoFormat/link/createLinkAfterSpace.ts
```diff
@@ -15,22 +15,30 @@ export function createLinkAfterSpace(editor: IEditor) {
...
+                    const spaceText = createText(link.substring(url.length), marker.format);
+                    selectedSegmentsAndParagraphs[0][1].segments.splice(length, 0, spaceText);
```
JiuqingSong, line 38: same here

packages/roosterjs-content-model-plugins/lib/autoFormat/link/createLinkAfterSpace.ts
```diff
@@ -15,22 +15,30 @@ export function createLinkAfterSpace(editor: IEditor) {
...
+
```
JiuqingSong, line 39: can we create a common function to split an existing text segment, for example

function splitText(text: ContentModelTextSegment, start: number, end: number) {... }

and return the middle one, the we just need to add a Link decorator into the middle one.

This common function may also be used is other places in the future.

packages/roosterjs-content-model-plugins/lib/pluginUtils/splitTextSegment.ts
```diff
@@ -0,0 +1,42 @@
...
+        text.substring(end),
+        textSegment.format,
+        textSegment.link,
+        textSegment.code
+    );
+    textBefore.isSelected = textSegment.isSelected;
+    middleText.isSelected = textSegment.isSelected;
+    textAfter.isSelected = textSegment.isSelected;
+
+    parent.segments.splice(index, 1, textBefore);
```
JiuqingSong, line 37: better check if the new segment is empty
and `splice` support adding multiple item together. So you can do this:

```ts
const newSegments: ContentModelText = [ 
createText(...) // the middle segment here
];

if (start > 0) {
  newSegments.unshift(createText(...) // first segment);
}

if (end < length - 1) {
  newSegment.push(createText(...) // last segment);
}

newSegments.forEach(x => x.isSelected = text.isSelected);
parent.segments.splice(index, 1, ...newSegments);
```

packages/roosterjs-content-model-plugins/lib/autoFormat/link/createLinkAfterSpace.ts
```diff
@@ -1,37 +1,53 @@
...
+                    if (url && link && matchLink(url)) {
+                        const linkSegment = splitTextSegment(
+                            textSegment,
+                            paragraph,
+                            textSegment.text.length - link.trimLeft().length,
+                            textSegment.text.trimRight().length
+                        );
+                        linkSegment.link = {
+                            format: {
+                                href: url,
```
JiuqingSong, line 41: Actually, this may generate a link that is not workable. Because the protocol part is necessary in the link. You need to retrieve url string from the result of matchLink().

To repro, type something "www.bing.com", space, then export the content and try to click the link, it won't open.

packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
```diff
@@ -113,8 +113,11 @@ export class AutoFormatPlugin implements EditorPlugin {
         if (rawEvent.inputType === 'insertText') {
             switch (rawEvent.data) {
                 case ' ':
-                    const { autoBullet, autoNumbering } = this.options;
+                    const { autoBullet, autoNumbering, autoLink } = this.options;
```
JiuqingSong, line 116: let make these features off by default for now.
and consider add options to turn them on in demo site

JiuqingSong (approved):

PR #2529, opened 2024-03-26
Implement movement with Tab key inside Table

Andres-CT98:
### Description:

It is standard for accessibility and usability usage, to have the possibility to navigate a table using the tab key. The direction of the navigation can be inverted using the Shift key.

### Fix

Implement as a part of SelectionPlugin navigation with Tab key inside Tables. If the key is pressed within the table, the selection marker will be positioned at the start of the next cell. In case there is no further cell to navigate to, the selection marker will be positioned before or after the whole table.

The selectBeforeImage method was changed to selectBeforeOrAfterElement to accommodate both images and tables, and both before and after positioning.

### Warnings and implications

- This changes should not overwrite any Tab behaviour on EditPlugin.


JiuqingSong:
I tried with your change but when I press TAB, instead of jumping to next cell, it will indent the paragraph

Andres-CT98:
> I tried with your change but when I press TAB, instead of jumping to next cell, it will indent the paragraph

Did you disable the Editplugin? Currently that plugin will override the Tab change.

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -132,8 +133,7 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
 
             return;
         } else if (selection?.type == 'image' && selection.image !== rawEvent.target) {
-            this.selectBeforeImage(editor, selection.image);
-
+            this.selectBeforeOrAfterElement(editor, selection.image);
```
juliaroldi, line 136: Is this change from Table features? If yes, could we make this change part of edit plugin?

Andres-CT98: It could be done but I feel it is easier to implement in SelectionPlugin, since Arrow key movement is also implemented there. It is a good behaviour to have as a core feature.

JiuqingSong:
> > I tried with your change but when I press TAB, instead of jumping to next cell, it will indent the paragraph
> 
> Did you disable the Editplugin? Currently that plugin will override the Tab change.

I see. In that case, We also need to make the change also work with EditPlugin since we can't disable it in production.

You can try to add a `willHandleEventExclusively` method in SelectionPlugin and if we know it is a TAB key under table (but not under list inside table), return true, then EditPlugin won't receive this event.

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -359,24 +367,42 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
...
-                                range,
-                                isReverted: false,
-                            },
-                            null /*tableSelection*/
+                            key == Up ? td.childNodes.length : 0,
+                            this.editor
                         );
                     }
+                } else if (key == 'TabLeft' || key == 'TabRight') {
+                    const revrse = key == 'TabLeft';
```
juliaroldi, line 377: nit: just small typo

Andres-CT98: Good question, It works the same both ways. In an HTML Table, the order of the cells is the same in RTL than LTR. This tab change works on a 2D representation of the HTML table where the cursor is.

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -359,24 +366,42 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
                     }
 
                     if (collapsed && td) {
-                        const { node, offset } = normalizePos(
```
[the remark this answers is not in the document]
Andres-CT98: Yes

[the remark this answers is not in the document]
JiuqingSong: why?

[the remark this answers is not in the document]
Andres-CT98: After discussion it was determined to restore normalizePos. It was removed to mitigate the underlying issue fixed here: #2551 . Its removal caused other issues with positioning.

JiuqingSong (approved):

PR #2537, opened 2024-03-27
Fix 262677: Allow insert entity to a specified position without changing selection

JiuqingSong:
When insert entity, we allow passing in a DOM position. However, today we are using selectionOverride for this position which changes current selection to this position. So after insert entity, the selection is also changed.

To fix this, first we need to allow a separate selection marker than the existing selection. I made this happen by creating a new common API `formatInsertPointWithContentModel` and override several element processors which reads the DOM position then insert SelectionMarker to that position and mark it as "not selected" then return this InsertPoint to caller. After that, we can use this InsertPoint to insert new content, and finally remove this marker.

To make selection marker "not selected", we now allow SelectionMarker.isSelected to be false

In order to reduce duplicated code when override processors, I also did some refactor to existing processor code, including `textProcessor` and `addSelectionMarker`. I also added a new processor type `textWithSelectionProcessor` which works as a child processor of `textProcessor` so we now only need to override this one and keep `textProcessor` not overridden.


packages/roosterjs-content-model-core/lib/editor/Editor.ts
```diff
@@ -104,23 +105,25 @@ export class Editor implements IEditor {
...
+                return core.api.createContentModel(core);
 
             case 'disconnected':
-                return cloneModel(core.api.createContentModel(core), {
-                    includeCachedElement: this.cloneOptionCallback,
-                });
+                return cloneModel(
+                    core.api.createContentModel(core, {
+                        processorOverride: {
+                            table: tableProcessor,
```
juliaroldi, line 114: why are we adding this tableProcessor?

juliaroldi:
Should we also add entry point to pass the position when we insertEntity for the entity playground in demo site?

juliaroldi (approved):

packages/roosterjs-content-model-core/lib/editor/Editor.ts
JiuqingSong: I moved this from the "connected" part to "disconnected"

For fidelity and performance reason, today we are using an overridden table processor by default in editor to treat table as a fake entity if there is no selection in it. But here when we want to create a disconnected model, we should set it back to original table processor to really create model for table, but not an entity.

I moved it because with this change, if there is any override option passed in, we will not reuse existing cached model. So for better performance, I removed the override from "connected" part.

PR #2540, opened 2024-03-28
Set isSelected to false for formatHolder of list item

JiuqingSong:
Now that we allow SelectionMarker to have isSelected: false, we can set this false value to formatHolder under list item to better match its meaning.


juliaroldi (approved):

packages/roosterjs-content-model-dom/lib/modelApi/creators/createListItem.ts
```diff
@@ -15,14 +15,18 @@ export function createListItem(
     levels: ContentModelListLevel[],
     format?: ContentModelSegmentFormat
 ): ContentModelListItem {
+    const formatHolder = createSelectionMarker(format);
+
+    formatHolder.isSelected = false;
```
juliaroldi, line 20: if the marker is in the list item, should it be true?

JiuqingSong: Actually we never read that value, so it can be false.

PR #2552, opened 2024-04-02
Port AutoFormat into content model AutoFormatPlugin

juliaroldi:
AutoFormat plugin is responsible for formatting two dashes into a hyphen. This feature was ported into content model AutoFormatPlugin. 
autoHyphen

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
@@ -1,6 +1,7 @@
 import { createLink } from './link/createLink';
 import { createLinkAfterSpace } from './link/createLinkAfterSpace';
 import { keyboardListTrigger } from './list/keyboardListTrigger';
+import { transformHyphen } from './hyphen/transformHyphen';
 import { unlink } from './link/unlink';
 import type {
     ContentChangedEvent,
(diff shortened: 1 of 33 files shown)


JiuqingSong:
In old plugin, auto hyphen happens when we type anything after "--", but here it only happens when we type space. Is that expected?

juliaroldi:
> In old plugin, auto hyphen happens when we type anything after "--", but here it only happens when we type space. Is that expected?
I changed the behavior to match word behavior, that only transforms the  hyphen after space.

packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
```diff
@@ -112,14 +119,23 @@ export class AutoFormatPlugin implements EditorPlugin {
...
                 case ' ':
-                    const { autoBullet, autoNumbering, autoLink } = this.options;
+                    const { autoBullet, autoNumbering, autoLink, autoHyphen } = this.options;
                     keyboardListTrigger(editor, autoBullet, autoNumbering);
                     if (autoLink) {
                         createLinkAfterSpace(editor);
                     }
+                    if (autoHyphen) {
+                        transformHyphen(editor);
+                    }
```
JiuqingSong, line 138: Can we call `formatTextSegmentBeforeSelectionMarker` here, then call `keyboardListTrigger`, `createLinkAfterSpace`, `transformHyphen` inside it? So that we only need to format with Content Model once.

juliaroldi: I created a separated PR: https://github.com/microsoft/roosterjs/pull/2557 for it, since I needed to change a lot of files.

PR #2554, opened 2024-04-03
Enable selecting image when the only element in the range is an Image

BryanValverdeU:
For users that are only able to use Keyboard, they are not able to select image to set alternate text, 
To fix when the range selection is only surrounding a single image change the DOMSelection to Image


BryanValverdeU:
> Restore the snapshot below, press Shift+Right, it should select image and the next character, but it actually only select the image.
> 
> ```
> <div>aabb</div>
> ```

Good catch taking a look.

JiuqingSong (approved):

JiuqingSong:
What about Ctrl+Home and Ctrl+End?

JiuqingSong:
> What about Ctrl+Home and Ctrl+End?

Actually, because there is a hidden selection before/after the image, those Up/Down/Left/Right/Home/End/... key will update this hidden selection. Can we rely on selectionchange event to detect if the real selection need to be changed, then unselect image? So we don't need to handle all keys. Similarly, I remember @juliaroldi  has a bug to deal with Ctrl+A for image selection, it can also be fixed by this way.

BryanValverdeU:
> > What about Ctrl+Home and Ctrl+End?
> 
> Actually, because there is a hidden selection before/after the image, those Up/Down/Left/Right/Home/End/... key will update this hidden selection. Can we rely on selectionchange event to detect if the real selection need to be changed, then unselect image? So we don't need to handle all keys. Similarly, I remember @juliaroldi has a bug to deal with Ctrl+A for image selection, it can also be fixed by this way.
> 
> > I'm working on fix for this bug now and soon, I'll open PR for it.

What other keys can modify selection other than Ctrl + A, and shift + (Arrows/End/Home/PageUp/PageDown)? 

I guess another solution is that when we have ImageSelection always add a Range selecting the image + Adding a style to the editor to hide :selection. Because right now when we select the image we also collapse the selection

JiuqingSong:
> > > What about Ctrl+Home and Ctrl+End?
> > 
> > 
> > Actually, because there is a hidden selection before/after the image, those Up/Down/Left/Right/Home/End/... key will update this hidden selection. Can we rely on selectionchange event to detect if the real selection need to be changed, then unselect image? So we don't need to handle all keys. Similarly, I remember @juliaroldi has a bug to deal with Ctrl+A for image selection, it can also be fixed by this way.
> > > I'm working on fix for this bug now and soon, I'll open PR for it.
> 
> What other keys can modify selection other than Ctrl + A, and shift + (Arrows/End/Home/PageUp/PageDown)?
> 
> I guess another solution is that when we have ImageSelection always add a Range selecting the image + Adding a style to the editor to hide :selection. Because right now when we select the image we also collapse the selection and would also trigger a selectionchanged event

I think that is a good idea.

JiuqingSong (approved):

PR #2555, opened 2024-04-03
Improve backspace on list

JiuqingSong:
When cursor is at beginning of a paragraph under a list, press Backspace, we first change current list item to be a dummy list item (no list number or bullet). Press Backspace again, it will outdent to upper level, or become a regular paragraph if there is only one list level. Press Backspace again, go back to previous list item if exists.

When press ENTER on a dummy list item, create a new list item.

listbackspace

This is to keep the same behavior with Word.

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/edit/deleteSteps/deleteList.ts
+++ b/packages/roosterjs-content-model-plugins/lib/edit/deleteSteps/deleteList.ts
@@ -1,22 +1,32 @@
 import { getClosestAncestorBlockGroupIndex } from 'roosterjs-content-model-dom';
-import type { DeleteSelectionStep } from 'roosterjs-content-model-types';
+import type { DeleteSelectionStep, ContentModelListItem } from 'roosterjs-content-model-types';
 
 /**
  * @internal
  */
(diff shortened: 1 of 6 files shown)


packages/roosterjs-content-model-plugins/lib/edit/deleteSteps/deleteList.ts
```diff
@@ -1,22 +1,32 @@
...
-            item.blockGroupType == 'ListItem'
-        ) {
-            item.levels = [];
+    if (paragraph.segments[0] == marker) {
+        const index = getClosestAncestorBlockGroupIndex<ContentModelListItem>(
+            path,
+            ['ListItem'],
+            ['TableCell', 'FormatContainer']
+        );
+        const item = path[index] as ContentModelListItem | undefined;
```
juliaroldi, line 20: Is possible to be other thing than a list?

JiuqingSong: From line 17, we only query ListItem. so here it should be either ListItem or undefined (then index is -1)

juliaroldi (approved):

PR #2556, opened 2024-04-03
Fix selection with ctrl+a

juliaroldi:
Use  browser's  `selectionChange` event to fix the image selection, when the range is changed by a shortcut, such as CTRL+A. 
 
ctrl+A

### DIFF
--- a/packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
+++ b/packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
@@ -59,9 +59,8 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
 
         this.isSafari = !!env.isSafari;
         this.isMac = !!env.isMac;
-
+        document.addEventListener('selectionchange', this.onSelectionChange);
         if (this.isSafari) {
-            document.addEventListener('selectionchange', this.onSelectionChangeSafari);
             this.disposer = this.editor.attachDomEvent({
                 focus: { beforeDispatch: this.onFocus },
                 drop: { beforeDispatch: this.onDrop },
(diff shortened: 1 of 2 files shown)


packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -522,13 +519,27 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
         }
     };
 
-    private onSelectionChangeSafari = () => {
+    private onSelectionChange = () => {
         if (this.editor?.hasFocus() && !this.editor.isInShadowEdit()) {
```
BryanValverdeU, line 523: I think this if case was only executed when we were using Safari. Should we add a check for this.isSafari here?

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -522,13 +519,27 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
...
+                this.editor.setDOMSelection({
+                    type: 'range',
+                    range: selection.getRangeAt(0),
+                    isReverted: false,
+                });
+            }
+
+            // Safari has problem to handle onBlur event. When blur, we cannot get the original selection from editor.
+            // So we always save a selection whenever editor has focus. Then after blur, we can still use this cached selection.
+            if (newSelection?.type == 'range' && this.isSafari) {
```
juliaroldi, line 542: @BryanValverdeU  I added the check here.

BryanValverdeU (approved):

JiuqingSong (approved):

PR #2571, opened 2024-04-10
fix: on webkit-based applications, the selection will become empty after copy, and focus will cause the window to scroll to the top

miku1958:
selection in focus() become null after this call
https://github.com/microsoft/roosterjs/blob/a4457db8c9806400b27a1567c206b2bc5bf40290/packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts#L236C9-L236C21
image
and focus with null selection make the window scroll to the top

Oddly enough I wasn't able to reproduce the problem in the web demo

### DIFF
--- a/packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
+++ b/packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
@@ -160,8 +160,8 @@ class CopyPastePlugin implements PluginWithState<CopyPastePluginState> {
                     }
 
                     cleanUpAndRestoreSelection(tempDiv);
-                    this.editor.focus();
                     this.editor.setDOMSelection(selection);
+                    this.editor.focus();
 
                     if (isCut) {
                         this.editor.formatContentModel(


packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
```diff
@@ -160,8 +160,8 @@ class CopyPastePlugin implements PluginWithState<CopyPastePluginState> {
                     }
 
                     cleanUpAndRestoreSelection(tempDiv);
-                    this.editor.focus();
                     this.editor.setDOMSelection(selection);
+                    this.editor.focus();
```
JiuqingSong, line 164: @juliaroldi would you please help test this change, in Chrome, Firefox and Safari. This is mainly about the cut/copy behavior, and see if the copied content is right, and if focus is back to correct place, for range/image/table selections.

If all work well, you please approve and merge the change.

Thanks.

BryanValverdeU:
@miku1958 this looks really similar to the issue https://github.com/microsoft/roosterjs/pull/2564 fixed. 

Just curious, could you please try to repro in the master branch and then revert https://github.com/microsoft/roosterjs/pull/2564 (53f1f4dbe3f708b9e24e706af149b49cadcf7232) locally and see if the issue happens after reverting? Just to confirm since you mentioned that the issue is not repro'ing in the demo site.

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
juliaroldi: @miku1958 @JiuqingSong @BryanValverdeU this change works fine in Chrome/Firefox/Edge/Safari, the only bug I found was related to cut behavior on tables, but I can also reproduce the bug without this change.

juliaroldi:
> @miku1958 this looks really similar to the issue #2564 fixed.
> 
> Just curious, could you please try to repro in the master branch and then revert #2564 ([53f1f4d](https://github.com/microsoft/roosterjs/commit/53f1f4dbe3f708b9e24e706af149b49cadcf7232)) locally and see if the issue happens after reverting? Just to confirm since you mentioned that the issue is not repro'ing in the demo site.

I'll wait for @miku1958 input on this before approve and merge.

miku1958:
I was only able to test on the cm0.28 branch because the master branch is so different from it I had to change a lot of things. 

After I copied the changes from https://github.com/microsoft/roosterjs/commit/53f1f4dbe3f708b9e24e706af149b49cadcf7232, I can still reproduce the problem

miku1958:
The problem I'm having is caused by focus(), and it has focused, so the check here won't work:
image

JiuqingSong:
Since the change can fix the original issue, and we have tested there's no regression, I  think we can just merge this one first. About #2564, I think we can do another change to revert it if need. @juliaroldi @BryanValverdeU thoughts?

JiuqingSong (approved):

miku1958:
@JiuqingSong  I think they're two issues, but the root cause maybe the same one, the selection becomes null.

BryanValverdeU:
> Since the change can fix the original issue, and we have tested there's no regression, I think we can just merge this one first. About #2564, I think we can do another change to revert it if need. @juliaroldi @BryanValverdeU thoughts?

This PR looks good for the copy side issue. But we may need to fix the other issues like inserting Row/Column before reverting the change.

BryanValverdeU (approved):

PR #2579, opened 2024-04-16
Adjust Image size 

juliaroldi:
When inserting an image, add max-width to editable portion of the editor div, i.e the client width - ( paddings left and right). 
Use `width: 100%` can cause a scroll bar when resize large images.

### DIFF
--- a/packages/roosterjs-content-model-core/lib/editor/core/DOMHelperImpl.ts
+++ b/packages/roosterjs-content-model-core/lib/editor/core/DOMHelperImpl.ts
@@ -70,6 +70,17 @@ class DOMHelperImpl implements DOMHelper {
 
         return style?.direction == 'rtl';
     }
+
+    /**
+     * Get the width of the editable area of the editor content div
+     */
+    getEditorDivWidth(): number {
+        const contentDiv = this.contentDiv;
+        const style = contentDiv.ownerDocument.defaultView?.getComputedStyle(contentDiv);
+        const paddingLeft = parseInt(style?.paddingLeft || '0');
+        const paddingRight = parseInt(style?.paddingRight || '0');
+        return this.contentDiv.clientWidth - (paddingLeft + paddingRight);
+    }
 }
 
 /**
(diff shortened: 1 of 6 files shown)


packages/roosterjs-content-model-core/lib/editor/core/DOMHelperImpl.ts
```diff
@@ -70,6 +70,17 @@ class DOMHelperImpl implements DOMHelper {
...
     }
+
+    /**
+     * Get the width of the editable area of the editor content div
+     */
+    getEditorDivWidth(): number {
+        const contentDiv = this.contentDiv;
+        const style = contentDiv.ownerDocument.defaultView?.getComputedStyle(contentDiv);
+        const paddingLeft = parseInt(style?.paddingLeft || '0');
+        const paddingRight = parseInt(style?.paddingRight || '0');
```
JiuqingSong, line 81: In case of using other unit than pixel, let's use `parseValueWithUnit` instead of parseInt

packages/roosterjs-content-model-core/lib/editor/core/DOMHelperImpl.ts
```diff
@@ -70,6 +70,17 @@ class DOMHelperImpl implements DOMHelper {
 
         return style?.direction == 'rtl';
     }
+
+    /**
+     * Get the width of the editable area of the editor content div
+     */
+    getEditorDivWidth(): number {
```
JiuqingSong, line 77: suggest name `getClientWidth`

packages/roosterjs-content-model-core/lib/editor/core/DOMHelperImpl.ts
```diff
@@ -70,6 +70,17 @@ class DOMHelperImpl implements DOMHelper {
...
+        const contentDiv = this.contentDiv;
+        const style = contentDiv.ownerDocument.defaultView?.getComputedStyle(contentDiv);
+        const paddingLeft = parseInt(style?.paddingLeft || '0');
+        const paddingRight = parseInt(style?.paddingRight || '0');
+        return this.contentDiv.clientWidth - (paddingLeft + paddingRight);
```
JiuqingSong, line 82: Does it also work well when zoom?

juliaroldi: When we insert a new image, it works fine, but if change the zoom after a image is inserted, the image is not resized.

JiuqingSong (approved):

PR #2583, opened 2024-04-17
Trigger auto format events

juliaroldi:
We need the info that auto format list and auto format happened to trigger the AutoFormat Settings UI in OWA side, so when auto hyphen or auto list happens, trigger event.

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
@@ -129,6 +129,7 @@ export class AutoFormatPlugin implements EditorPlugin {
         ) {
             switch (rawEvent.data) {
                 case ' ':
+                    let apiName = '';
                     formatTextSegmentBeforeSelectionMarker(
                         editor,
                         (model, previousSegment, paragraph, _markerFormat, context) => {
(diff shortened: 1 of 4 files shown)


packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
```diff
@@ -153,19 +165,18 @@ export class AutoFormatPlugin implements EditorPlugin {
...
-                                shouldHyphen ||
-                                shouldLink
-                            );
+                            apiName = getApiName(shouldList, shouldHyphen);
+
+                            return shouldList || shouldHyphen || shouldLink;
                         }
                     );
+                    if (apiName) {
+                        editor.triggerEvent('contentChanged', {
```
JiuqingSong, line 174: We should let `formatTextSegmentBeforeSelectionMarker` trigger the event for us, we just need to pass in the change source. You can pass in from the 3rd parameter 
image

JiuqingSong: Since the apiName can only be retrieved after callback, we can use a trick here:

```ts
const option: FormatContentModelOptions = {
  source: 'autoFormat' // BTW, Can we add this into `ChangeSource` in `roosterjs-content-model-dom`?
}

formatTextSegmentBeforeSelectionMarker (editor, model=> {
  ...
  option.formatApiName = apiName;
}, option);
```

JiuqingSong (approved):

PR #2593, opened 2024-04-19
Select image after inserting it

BryanValverdeU:
After insertting a new image to the editor, select it. So we can see the Top Ribbon contextual tab by default.

### DIFF
--- a/packages/roosterjs-content-model-api/lib/publicApi/image/insertImage.ts
+++ b/packages/roosterjs-content-model-api/lib/publicApi/image/insertImage.ts
@@ -4,6 +4,8 @@ import {
     addSegment,
     createContentModelDocument,
     createImage,
+    isNodeOfType,
+    isElementOfType,
 } from 'roosterjs-content-model-dom';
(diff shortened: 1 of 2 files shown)


packages/roosterjs-content-model-api/lib/publicApi/image/insertImage.ts
```diff
@@ -41,6 +45,22 @@ function insertImageWithSrc(editor: IEditor, src: string) {
         },
         {
             apiName: 'insertImage',
+            onNodeCreated: (model, node) => {
```
JiuqingSong, line 48: Actually, you can do:
```ts
 const image = createImage(...);
 mergeModel(...);
 image.isSelected = true;
 setSelection(model, image) // This will clear all other existing selections and only select this image
 
 return true;
});
```

BryanValverdeU: I see, sounds good. Changed to use this util.
Thanks

JiuqingSong (approved):

PR #2595, opened 2024-04-22
Port Auto fraction and ordinals

juliaroldi:
Port the features to auto format fractions and ordinals numbers. This feature was originally in Custom Replace, but the formatting number should only be triggered after space, then it was moved to AutoFormat Plugin to match the expected behavior. 

autoOrdinals
autofractions


JiuqingSong:
I think this should be part of CustomReplacePlugin

JiuqingSong:
> > I think this should be part of CustomReplacePlugin
> > I OWA these two feature are part of the AutoFormat feature, so I thought would be nice to port these feature to this plugin, since they follow the pattern of type something and then press space and the formatting is applied. For auto ordinals, there is no replace, we only apply superscript format. For fractions, we only replace because we do not have format to change the fraction style.

I see, that is ok.

JiuqingSong (approved):

PR #2597, opened 2024-04-22
Add function to check if the node given is the editor node

Andres-CT98:
### Description:

Modify isNodeInEditor on DOMHelper to exclude root node from search.

### DIFF
--- a/packages/roosterjs-content-model-core/lib/editor/core/DOMHelperImpl.ts
+++ b/packages/roosterjs-content-model-core/lib/editor/core/DOMHelperImpl.ts
@@ -12,12 +12,8 @@ class DOMHelperImpl implements DOMHelper {
         return this.contentDiv.textContent || '';
     }
 
-    isNodeInEditor(node: Node): boolean {
-        return this.contentDiv.contains(node);
-    }
-
-    isNodeEditor(node: Node): boolean {
-        return node == this.contentDiv;
+    isNodeInEditor(node: Node, excludeRoot?: boolean): boolean {
+        return this.contentDiv.contains(node) && (excludeRoot ? node != this.contentDiv : true);
     }
 
     calculateZoomScale(): number {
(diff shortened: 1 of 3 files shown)


packages/roosterjs-content-model-core/lib/editor/core/DOMHelperImpl.ts
```diff
@@ -16,6 +16,10 @@ class DOMHelperImpl implements DOMHelper {
         return this.contentDiv.contains(node);
     }
 
+    isNodeEditor(node: Node): boolean {
```
JiuqingSong, line 19: Actually we were trying to not let developer know the root node. Because when it is very easy to get root DIV, developer may do all kinds of direct operation on the root DIV (e.g. remove the DIV) which can cause unpredictable result.

Here I'd like to explore if there is an alternative way to achieve your goal.

Andres-CT98: isNodeInEditor can also currently be used to get the root node. Just traverse upward any node known to be in the editor and eventually when isNodeInEditor returns false, the root was the previous one.

JiuqingSong: Yes, that is true. It seems we can't 100% hide the root DIV, just want to make it harder to get.
Let me think about this. If this is the only way, it is ok. But before that I still want to explore if there is other way.

JiuqingSong: BTW, have you tried using browser's native ondragstart/ondragover/ondrop event? With those events we don't need to know the editor root because when drop out of editor, browser won't trigger the ondrop event to us.

JiuqingSong: As an alternative way, can we add a parameter to isNodeInEditor() to let it return false when the node is root DIV, and by default it should keep the original behavior?

JiuqingSong (approved):

PR #2599, opened 2024-04-23
Implement Table Movement

Andres-CT98:
### Features

* Allows movement of tables from one point of the editor to another.
* Change of cursor while moving table.
* Live preview of table drop position wile dragging.
* Rectangle outline of table shown while dragging.

### Drag and drop

* At drag start:
    * Create a rectangle representing the outline of the table to move.
    * Add a css rule to change the cursor to the 'move' one.
    * Get the content model table representation of the table to move and the initial selection for later.
* While dragging:
    * Get the insertion position using getNodePositionFromEvent and apply that selection for a live focus drag.
    * Move the rectangle of the table along with the cursor.
* On drag end:
   * Remove the rectangle and restore the cursor.
   * Handle special cases of table selection: Not dropping the table on itself, Not dropping outside editor, and select whole table if cursor never left the Table mover handle.
   * Get the insertion point to do formatInsertPointWithContentModel
   * Remove the old table
   * Use mergeModel to insert the content model table. After, select the first cell on table if successful.
   * If the move is unsuccessful or cancelled, restore initial selection to editor.
   
### Other changes

* Added getNodePositionFromEvent function to obtain the node where the cursor inline position is.
* Exported formatInsertPointWithContentModel function.

### Warnings and implications

* This implementation relies on the DragAndDropHelper utility. In the future, a method relying on the native browser's drag and drop handling should be implemented. This would be an implementation that could handle text, tables, and images with the same code.
* Table copy was briefly implemented but was removed due to cache issues.


packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
```diff
@@ -129,3 +157,214 @@ function isTableTopVisible(editor: IEditor, rect: Rect | null, contentDiv?: Node
...
+            editor.formatContentModel(
+                (model, context) => {
+                    const SPArray = getSelectedSegmentsAndParagraphs(model, false);
+                    if (
+                        SPArray.length == 1 &&
+                        SPArray[0][0].segmentType == 'SelectionMarker' &&
+                        SPArray[0][1] != null
+                    ) {
+                        const paragraph = SPArray[0][1];
+                        paragraph.segments.forEach(segment => {
```
JiuqingSong, line 308: can we use mergeModel here?

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
```diff
@@ -129,3 +157,214 @@ function isTableTopVisible(editor: IEditor, rect: Rect | null, contentDiv?: Node
...
+                                    if (markerParagraph?.blockType == 'Paragraph') {
+                                        const marker = createSelectionMarker(model.format);
+
+                                        markerParagraph.segments.unshift(marker);
+                                        setParagraphNotImplicit(markerParagraph);
+                                        setSelection(FirstCell, marker);
+                                    }
+                                } else {
+                                    // Restore initial selection
+                                    editor.setDOMSelection(initValue.initialSelection);
```
JiuqingSong, line 333: do not call setDOMSelection inside formatContentModel, it will be overwrite when write back

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
```diff
@@ -129,3 +157,214 @@ function isTableTopVisible(editor: IEditor, rect: Rect | null, contentDiv?: Node
...
+                        type: 'range',
+                        range: finalRange.range,
+                        isReverted: false,
+                    },
+                }
+            );
+            // Remove old table only if insertion was successful
+            if (insertionSuccess) {
+                if (!(event.ctrlKey || event.metaKey)) {
+                    table.remove();
```
JiuqingSong, line 352: This is risky since the original table element can be overwritten when write back

JiuqingSong: I think we can use `formatInsertPointWithContentModel` to help do the merging. It allows pass in an additional insert point to mark the model together with the existing one.

For example:

```ts
const domTable = ... // The HTML Table that is dragging
const domIP = getNodePositionFromEvent(editor, x, y); // Get the insert point

formatInsertPointWithContentModel(editor, (model,context,ip) => {
  const [table, path] = getFirstSelectedTable(model);
  const index = path[0].blocks.indexOf(table);
  // Add necessary null check here
  path[0].blocks.splice(index, 1); // Table is removed here

 ip.marker.isSelected = true // ip is the insert point generated from the addtional selection passed below, by default it is marked as isSelected =false, we set it to true here to make it a regular selection marker
 
 const newTableModel = ....
 mergeModel(model, newTableModel); // Now this will be merged into the marker we marked above
},
 domIP, // Pass in the target ip here, it will be marked as ip with isSelected= false
 {
   selectionOverride: { // Override the original selection with the dragging table, so it we can use getFirstSelectedTable to get this table in content model
     type: 'table',
     table: domTable,
     ...
   }
});
 
```

Andres-CT98: > This is risky since the original table element can be overwritten when write back

I need more information about this, because I have no encountered any overwriting issues.

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
Andres-CT98: Yes, I think I was trying to check for the exact segment to do the insertion but it was unnecessary.

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
JiuqingSong: For example:
```ts
const element = document.querySelector(...) // or any other way that you get a HTML element from editor

editor.formatContentModel( .... );

anyOperationOn(element); // Risky here
```

Because after formatContentModel() is called, it is possible the whole document is regenerated, so it is possible the original element does not exist in editor any more.

Today it works for you possibly because we treat the table that is not in selection as a fake entity, so the TABLE element will be preserved. But this behavior can be changed in the future. We treat table as entity just because of content fidelity issue. Given most of the known content fidelity bug is already fixed, we may remove this hacky behavior. So that any other feature should not take dependency to this behavior.

In general, with content model, we should avoid accessing DOM tree directly as much as possible. We should only access DOM when handling DOM event (in which case we only get DOM info from the event) or for performance reason. Content Model already provides ways to do this without accessing DOM tree after formatContentModel.

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
```diff
@@ -129,3 +157,212 @@ function isTableTopVisible(editor: IEditor, rect: Rect | null, contentDiv?: Node
...
+    tableRect.style.height = `${trect.height}px`;
+    tableRect.style.top = `${trect.top}px`;
+    tableRect.style.left = `${trect.left}px`;
+    div.parentNode?.appendChild(tableRect);
+
+    // Get current selection
+    const initialSelection = editor.getDOMSelection();
+
+    // Select first cell of the table
+    editor.setDOMSelection({
```
JiuqingSong, line 189: Curious, instead of just select the first cell, how about we select the whole table?

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
```diff
@@ -129,3 +157,212 @@ function isTableTopVisible(editor: IEditor, rect: Rect | null, contentDiv?: Node
...
+        lastColumn: 0,
+        lastRow: 0,
+        table: table,
+    });
+
+    // Get the table content model
+    const [cmTable] = getFirstSelectedTable(editor.getContentModelCopy('disconnected'));
+
+    // Restore selection
+    editor.setDOMSelection(initialSelection);
```
JiuqingSong, line 202: Instead of set selection then move it back, you can do:

```ts
let cmTable: ContentModelTable|undefined;

  cmTable = getFirstSelectedTable(model);
  return false;
}, {
  selectionOverride: { type: 'table',... }
});
```

This allows you use a different selection to create content model and not impact the original one. So no need to set selection away then set back.

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
JiuqingSong: Oh I see you just want to get the selected table.

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
```diff
@@ -129,3 +157,212 @@ function isTableTopVisible(editor: IEditor, rect: Rect | null, contentDiv?: Node
...
+
+function onDragging(context: TableMoverContext, event: MouseEvent, initValue: TableMoverInitValue) {
+    const { tableRect } = initValue;
+    const { editor } = context;
+
+    // Move table outline rectangle
+    tableRect.style.top = `${event.clientY + TABLE_MOVER_LENGTH}px`;
+    tableRect.style.left = `${event.clientX + TABLE_MOVER_LENGTH}px`;
+
+    event.ctrlKey || event.metaKey
```
JiuqingSong, line 251: Actually, Mac also has Ctrl key. So do we want to allow copy on Mac using Ctrl key or Meta key or both?
If we only want to do copy with Meta key here, we should also check if it is running on Mac then Ctrl key does not count.

JiuqingSong: I just tried on Mac and it seems Mac never allows copy, even with Ctrl or Meta key. So I think we can keep it consistent here, only do copy when !isMac and ctrlKey

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
```diff
@@ -129,3 +157,212 @@ function isTableTopVisible(editor: IEditor, rect: Rect | null, contentDiv?: Node
...
+        // Get position to insert table
+        const insertPosition = getNodePositionFromEvent(editor, event.clientX, event.clientY);
+        if (insertPosition) {
+            // Move table to new position
+            formatInsertPointWithContentModel(
+                editor,
+                insertPosition,
+                (model, context, ip) => {
+                    // Remove old table if not copying
+                    if (!(event.ctrlKey || event.metaKey)) {
```
JiuqingSong, line 307: same here

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
```diff
@@ -129,3 +157,212 @@ function isTableTopVisible(editor: IEditor, rect: Rect | null, contentDiv?: Node
...
+                        const doc = createContentModelDocument();
+                        doc.blocks.push(initValue.cmTable);
+                        insertionSuccess = !!mergeModel(model, doc, context, {
+                            mergeFormat: 'none',
+                            insertPosition: ip,
+                        });
+
+                        if (insertionSuccess) {
+                            // After mergeModel, the new table should be selected
+                            const [finalTable] = getFirstSelectedTable(model);
```
JiuqingSong, line 328: The final table will be the same javascript object with initValue.cmTable, so you can use it directly here.

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
```diff
@@ -129,3 +157,212 @@ function isTableTopVisible(editor: IEditor, rect: Rect | null, contentDiv?: Node
...
+                            if (finalTable) {
+                                // Clear current selection
+                                setSelection(model);
```
JiuqingSong, line 331: setSelection only need to be called once, so no need to clear the selection here.

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
```diff
@@ -129,3 +157,212 @@ function isTableTopVisible(editor: IEditor, rect: Rect | null, contentDiv?: Node
...
+                    if (!(event.ctrlKey || event.metaKey)) {
+                        const [oldTable, path] = getFirstSelectedTable(model);
+                        if (oldTable) {
+                            const index = path[0].blocks.indexOf(oldTable);
+                            path[0].blocks.splice(index, 1);
+                        }
+                    }
+                    if (ip && initValue?.cmTable) {
+                        // Guarantee the cm Insertion point is selected
+                        ip.marker.isSelected = true;
```
JiuqingSong, line 316: Since you are passing in insertPosition into mergeModel at line 323 (which is good), no need to set it to be selected here.

JiuqingSong:
What is the plan of adding test?

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
Andres-CT98: So this means not allowing the feature on Mac?

JiuqingSong: I mean on Mac we only allow move, but not copy. At least that is how normal drag and drop work on Mac. You can also confirm with @jvillalobos

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
Andres-CT98: Seeing as this method is implemented to obtain the cmTable on all the other Table plugins, I prefer to make this optimisation on a separate PR for all plugins.

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
Andres-CT98: I tried to make the change but it was not working while testing.

JiuqingSong: It is ok not change here, but I'd like to understand more about the issue, since it may expose some underlying issue. Can you double check if the table is the same object in content model? If not, why?

JiuqingSong: Any update here? Can you confirm if the cmTable object is the same one after merge?

JiuqingSong:
Per our discussion, please add parameter to allow turn on/off table move

JiuqingSong (approved):

JiuqingSong (approved):

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableMover.ts
Andres-CT98: Here is that PR #2656

PR #2607, opened 2024-04-26
Also allow using Ctrl-Shift-Z on Windows

florian-msft:


packages/roosterjs-content-model-plugins/lib/shortcut/shortcuts.ts
```diff
@@ -129,17 +129,16 @@ export const ShortcutRedo: ShortcutCommand = {
 
 /**
  * Shortcut command for Redo 2
- * Windows: N/A
+ * Windows: Ctrl + Shift + Z
  * MacOS: Meta + Shift + Z
  */
-export const ShortcutRedoMacOS: ShortcutCommand = {
+export const ShortcutRedoAlt: ShortcutCommand = {
```
JiuqingSong, line 135: This makes it a breaking change.

Can you do this:

export const ShortcutRedoAlt = {.... }

/**
 * @deprecated
 */
const export ShortcutRedoMacOS = ShortcutRedoAlt;

So that old one still work, and new one also has better name

JiuqingSong (approved):

PR #2608, opened 2024-04-26
Bump main version to 9.2.0

ianeli1:
{
    "legacy": "8.62.0",
    "react": "8.56.0",
    "main": "9.2.0",
    "legacyAdapter": "8.62.0"
}


JiuqingSong:
When I diff with master, I see some difference, some may be caused by EOL, some seems to be merging issue. Can you try to directly copy those impacted files from master to replace directly?

ianeli1:
> When I diff with master, I see some difference, some may be caused by EOL, some seems to be merging issue. Can you try to directly copy those impacted files from master to replace directly?

I'm still seeing a diff in `packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts` despite it being the exact same file, and committing the master raw does nothing. Odd.

JiuqingSong (approved):

PR #2609, opened 2024-04-27
Do not remove image span

juliaroldi:
When writing model to DOM, do not remove the span element, if it has only an image as a child. 
Also, when select a single image ensure an image has span as parent, if it do not have, add it. 
Screenshot 2024-04-29 114107

### DIFF
--- a/packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection/setDOMSelection.ts
+++ b/packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection/setDOMSelection.ts
@@ -7,6 +7,7 @@ import {
     isNodeOfType,
     parseTableCells,
     toArray,
+    wrap,
 } from 'roosterjs-content-model-dom';
 import type {
     ParsedTable,
@@ -44,14 +45,12 @@ export const setDOMSelection: SetDOMSelection = (core, selection, skipSelectionC
     try {
         switch (selection?.type) {
             case 'image':
-                const image = selection.image;
+                const image = ensureImageHasSpanParent(selection.image);
 
-                const spanImage = ensureImageHasSpanParent(image);
-                if (!spanImage) {
-                    return;
-                }
-
-                core.selection.selection = selection;
+                core.selection.selection = {
+                    type: 'image',
+                    image,
+                };
                 core.api.setEditorStyle(
                     core,
                     DOM_SELECTION_CSS_KEY,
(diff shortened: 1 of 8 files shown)


packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -673,6 +675,24 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
     }
 }
 
+function ensureImageHasSpanParent(image: HTMLImageElement) {
```
JiuqingSong, line 678: Let's move this function into setDOMSelection.ts and call it there.

Because SelectionPlugin is not the only entry point to select image, it can also be selected by code.

packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection/setDOMSelection.ts
```diff
@@ -231,3 +241,23 @@ function setRangeSelection(doc: Document, element: HTMLElement | undefined, coll
...
+        parent.firstElementChild == image &&
+        parent.lastElementChild == image &&
+        !parent.textContent
+    ) {
+        return parent;
+    }
+
+    const span = image.ownerDocument.createElement('span');
+    span.appendChild(image);
+    parent?.appendChild(span);
```
JiuqingSong, line 261: this is not right. the IMAGE may not be the last element of parent, so this will change the order of image.

You can use existing util function `wrap`

packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection/setDOMSelection.ts
```diff
@@ -231,3 +241,23 @@ function setRangeSelection(doc: Document, element: HTMLElement | undefined, coll
...
+function ensureImageHasSpanParent(image: HTMLImageElement) {
+    const parent = image.parentElement;
+
+    if (
+        parent &&
+        isNodeOfType(parent, 'ELEMENT_NODE') &&
+        isElementOfType(parent, 'span') &&
+        parent.firstElementChild == image &&
+        parent.lastElementChild == image &&
+        !parent.textContent
```
JiuqingSong, line 254: line 252,253 already guaranteed there is no textContent

packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection/setDOMSelection.ts
```diff
@@ -41,14 +46,19 @@ export const setDOMSelection: SetDOMSelection = (core, selection, skipSelectionC
             case 'image':
                 const image = selection.image;
 
+                const spanImage = ensureImageHasSpanParent(image);
+                if (!spanImage) {
```
JiuqingSong, line 50: This should never be true. You can declare the function `ensureImageHasSpanParent` to let it return HTMLElement so no need to check here

JiuqingSong (approved):

Issue #2586, opened 2024-04-18
Enter on expanded selection under list should create new list item


JiuqingSong:
```html
<ol start="1" data-editing-info="{&quot;orderedStyleType&quot;:1}" style="list-style-type: decimal;"><li style="font-family: Calibri; font-size: 11pt; color: rgb(0, 0, 0);"><div style="font-family: Calibri; font-size: 11pt; color: rgb(0, 0, 0);">abcd</div></li></ol><!--{"type":"range","start":[0,0,0,0,1],"end":[0,0,0,0,3],"isReverted":true,"isDarkMode":false}-->
```

Press Enter

Expect: 
```
1. a
2. [CURSOR]d
```

Actual:
```
1. a[CURSOR]d
```

JiuqingSong:
Fixed by https://github.com/microsoft/roosterjs/pull/2610

Issue #2510, opened 2024-03-16
Perf: when content is super long, it takes long time to take undo snapshot


JiuqingSong:

JiuqingSong:
This is now solved with the new cache solution

PR #2610, opened 2024-05-01
Let Content Model handle ENTER key

JiuqingSong:
This covers the following scenarios:
1. Enter on normal paragraph, split the paragraph and bring focus to the new line
2. Enter on list item, split the list item
3. Enter on empty list item, reduce list level
4. Enter on empty quote line, unquote this line
5. Enter on expanded selection, first delete the selected content, then do the action above
6. Shift+Enter, always split current paragraph


JiuqingSong:
Known issue: After ENTER, need to bring cursor into view

JiuqingSong:
#2586  #2510

juliaroldi (approved):

juliaroldi (approved):

PR #2613, opened 2024-05-02
Content Model cache improvement: Let Content model update cache for child list change

JiuqingSong:
As a part of cache improvement, when user type in a new line, browser will change the HTML from 
```html
<div><br></div>
```
into 
```html
<div>[Typed text]</div>
````

This will generate 3 mutations:
1. Insert a new text node
2. Remove original `BR` node
3. Modify the text content to the typed content

In this change, I'm making `TextMutationObserver` and `DomIndexer` to be able to handle these changes without regenerating the whole content model. In general, it will first check if the mutations can be handled, then let domIndexer add/remove/modify existing content model to reflect these mutations:
1. Insert a new text node => Create a new ContentModelText
2. Remove original `BR` node => Remove existing ContentModelBr
3. Modify the text content to the typed content => Update model text

Now with this change, in most normal editing operations, we don't need to regenerate content model. The remaining ones are move about entity. We will handle them later.


JiuqingSong:
#2510

BryanValverdeU (approved):

BryanValverdeU (approved):

PR #2638, opened 2024-05-16
Fix cursor jump issue when applying Gboard suggestions

Rain-Zheng:
When applying a Gboard suggestion (such as auto-correction or the double-tap-space-for-period), the cursor will jump backwards. The root cause is https://github.com/microsoft/roosterjs/pull/2577 restored the selection in next animation frame. 

To fix this regression, we need to restore the selection on next `keyup` event, because IME can trigger other input events and change the cursor position before next frame.

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
@@ -22,6 +22,7 @@ export class EditPlugin implements EditorPlugin {
     private editor: IEditor | null = null;
     private disposer: (() => void) | null = null;
     private shouldHandleNextInputEvent = false;
+    private restoreSelectionCallback: (() => void) | null = null;
 
     /**
      * Get name of this plugin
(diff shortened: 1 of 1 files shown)


packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
```diff
@@ -150,15 +155,12 @@ export class EditPlugin implements EditorPlugin {
...
             // See: https://issues.chromium.org/issues/330596261
             const selection = editor.getDOMSelection();
-            const doc = this.editor?.getDocument();
-            doc?.defaultView?.requestAnimationFrame(() => {
-                if (this.editor) {
-                    this.editor.setDOMSelection(selection);
-                }
-            });
+            this.restoreSelectionCallback = () => {
+                this.editor?.setDOMSelection(selection);
```
JiuqingSong, line 162: instead of storing a function, can we just store the selection object instead?

JiuqingSong (approved):

Rain-Zheng:
Hi @JiuqingSong, will rooster 9.4.0 be available this week? If not, could you help release a hot fix version for this issue? We have received several reports about this issue so we want to fix it sooner. Thank you!

JiuqingSong:
Yes, it will be.

PR #2647, opened 2024-05-17
Add dark color handling for table and image selection

vhuseinova-msft:
When setting s selection color, add a check if in dark mode and use a color for dark mode instead of light one.
Also, added tests for this update

### DIFF
--- a/packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection/setDOMSelection.ts
+++ b/packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection/setDOMSelection.ts
@@ -51,16 +51,13 @@ export const setDOMSelection: SetDOMSelection = (core, selection, skipSelectionC
                     type: 'image',
                     image,
                 };
-                let imageSelectionColor =
-                    core.selection.imageSelectionBorderColor || DEFAULT_SELECTION_BORDER_COLOR;
-                if (imageSelectionColor !== DEFAULT_SELECTION_BORDER_COLOR && isDarkMode === true) {
-                    imageSelectionColor = core.darkColorHandler.getDarkColor(
-                        imageSelectionColor,
-                        undefined,
-                        'border',
-                        image
-                    );
+                let imageSelectionColor = DEFAULT_SELECTION_BORDER_COLOR;
+                if (isDarkMode === true && core.selection.imageSelectionBorderColorDark) {
+                    imageSelectionColor = core.selection.imageSelectionBorderColorDark;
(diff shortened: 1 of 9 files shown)


packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection/setDOMSelection.ts
```diff
@@ -51,12 +51,20 @@ export const setDOMSelection: SetDOMSelection = (core, selection, skipSelectionC
                     type: 'image',
                     image,
                 };
+                let imageSelectionColor =
+                    core.selection.imageSelectionBorderColor || DEFAULT_SELECTION_BORDER_COLOR;
+                if (imageSelectionColor !== DEFAULT_SELECTION_BORDER_COLOR && isDarkMode === true) {
```
JiuqingSong, line 56: Instead of calculating the color every time, can we just calculate it once, and save the result?

For example, when initialize selection plugin, we do
state.imageSelectionBorderColor = option.imageSelectionBorderColor ?? default_color;
state.imageSelectionBorderColorDark = option.imageSelectionBorderColor ? this.editor.getDarkColor(option.imageSelectionBorderColor) : default_color;

vhuseinova-msft: Updated!

packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection/setDOMSelection.ts
```diff
@@ -51,12 +49,14 @@ export const setDOMSelection: SetDOMSelection = (core, selection, skipSelectionC
                     type: 'image',
                     image,
                 };
+                let imageSelectionColor = isDarkMode
```
JiuqingSong, line 52: Use "const" instead of "let"

packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection/setDOMSelection.ts
```diff
@@ -112,13 +112,14 @@ export const setDOMSelection: SetDOMSelection = (core, selection, skipSelectionC
                         : handleTableSelected(parsedTable, tableId, table, firstCell, lastCell);
 
                 core.selection.selection = selection;
+
+                let tableSelectionColor = isDarkMode
```
JiuqingSong, line 116: same

JiuqingSong (approved):

PR #2652, opened 2024-05-21
Fix Table first column

Andres-CT98:
### Description:

Applying table borders on the first column of a table where the property of 'hasFirstColumn' is set should not apply borders in some places. The First Column style should always be imposed over any border changes. Making changes to the text alignment should also not add borders.

### Fix

* Remove the check for background color override as a requirement to apply First column style and rework how borders are applied. The color is now applied on the main formatCells function.
* setFirstColumnFormat was renamed to setFirstColumnFormatBorders and exported to use after applyTableBorderFormat. This way, First column style will be enforced even after any border changes.
* Optimisations were made to setFirstColumnFormatBorders and setHeaderRowFormat to remove redundant checking. The functions now exit if their respective property is unset or false.
* Remove the Table Header button from demo
* Add a new Options button to modify the table properties: Header Row, First Column, Banded Columns, and Banded Rows 

### Cases tested:

1. Insert a table
2. Set Shading to whole table
3. Toggle First Column
4a. Change alignment of the second cell of the first column
4b. Apply inner borders to first column

**Before:** Observe as the inner borders appear, contradicting the First Column style

**After:** No inner borders appear


packages/roosterjs-content-model-api/lib/publicApi/table/applyTableBorderFormat.ts
```diff
@@ -366,6 +369,12 @@ export function applyTableBorderFormat(
                     modifyPerimeter(tableModel, sel, borderFormat, perimeter, isRtl);
                 }
 
+                const tableMeta = hasMetadata(tableModel) ? updateTableMetadata(tableModel) : {};
```
JiuqingSong, line 372: From my PR https://github.com/microsoft/roosterjs/pull/2641, I just checked in a new API `getTableMetadata`.

Can you use this new API instead?

packages/roosterjs-content-model-api/test/publicApi/table/applyTableBorderFormatTest.ts
```diff
@@ -26,7 +26,7 @@ describe('applyTableBorderFormat', () => {
         format?: ContentModelTableCell['format']
     ) {
         // Create a table with all cells selected except the first and last row and column
-        const table = createTable(rows);
+        const table: ContentModelTable = createTable(rows);
```
JiuqingSong, line 29: Curious, any special reason to explicitly declare the type here?
(It is ok to add type declaration, just want to understand if anything is broken without this type here)

packages/roosterjs-content-model-api/test/publicApi/table/applyTableBorderFormatTest.ts
```diff
@@ -55,6 +55,7 @@ describe('applyTableBorderFormat', () => {
         const model = createContentModelDocument();
         model.blocks.push(table);
 
+        console.log('FAILEDXXX', table.dataset);
```
JiuqingSong, line 58: Is it intentional to add this log here?

packages/roosterjs-content-model-api/test/publicApi/table/applyTableBorderFormatTest.ts
```diff
@@ -77,6 +78,8 @@ describe('applyTableBorderFormat', () => {
             blockGroupType: 'Document',
             blocks: [expectedTable],
         });
+
+        console.log('FAILED,', model, expectedTable);
```
JiuqingSong, line 82: same

packages/roosterjs-content-model-dom/lib/modelApi/editing/applyTableFormat.ts
```diff
@@ -236,28 +240,36 @@ function formatCells(
...
+ */
+export function setFirstColumnFormatBorders(
     rows: ContentModelTableRow[],
-    format: Partial<TableMetadataFormat>,
-    metaOverrides: MetaOverrides
+    format: Partial<TableMetadataFormat>
 ) {
+    // Exit early hasFirstColumn is not set
+    if (!format.hasFirstColumn) {
+        return;
```
JiuqingSong, line 254: Question, what if the table had first column before, then we want to use this function to remove the first column? If it just return here, will its border style remains no change?

packages/roosterjs-content-model-dom/lib/modelApi/editing/applyTableFormat.ts
```diff
@@ -271,12 +283,17 @@ function setHeaderRowFormat(
     format: TableMetadataFormat,
     metaOverrides: MetaOverrides
 ) {
+    // Exit early if hasHeaderRow is not set
+    if (!format.hasHeaderRow) {
+        return;
```
JiuqingSong, line 288: Same here. If we want to remove header row, how should it work?

packages/roosterjs-react/lib/contextMenu/types/ContextMenuItemStringKeys.ts
```diff
@@ -70,6 +70,15 @@ export type TableEditDeleteMenuItemStringKey =
     | 'menuNameTableDeleteColumn'
     | 'menuNameTableDeleteRow';
 
+/**
+ * Key of localized strings of Table Options menu items
+ */
+export type TableOptionsMenuItemStringKey =
```
JiuqingSong, line 76: Suggest to add this type in demo site only, but no need to update roosterjs-react for now.

packages/roosterjs-content-model-api/test/publicApi/table/applyTableBorderFormatTest.ts
Andres-CT98: Just stronger typing.

packages/roosterjs-content-model-dom/lib/modelApi/editing/applyTableFormat.ts
Andres-CT98: The function is only intended to set the style, not remove it. The borders will go back to normal after any call to applyTableFormat. There is however an issue with isHeader, so that if 'First Column' is set, then unset, the cells will remain as headers. So I think change should be made in formatCells to make isHeader false for every cell, and then reapply it if hasFirstColumn or hasHeaderRow is true.

packages/roosterjs-react/lib/contextMenu/types/ContextMenuItemStringKeys.ts
Andres-CT98: Ok, but is there a reason those types are a part of roosterjs-react when they are only used on the demo site?

JiuqingSong: For the other types, they are used by table context menu in roosterjs-react.

packages/roosterjs-content-model-dom/lib/modelApi/editing/applyTableFormat.ts
Andres-CT98: The issue is fixed on latest build.

JiuqingSong (approved):

PR #2656, opened 2024-05-23
Optimise content model table fetching for Table Edit Plugins 

Andres-CT98:
Optimisation of the way the content model table is obtained at the onDragStart event by leveraging formatContentModel over setting selection directly on DOM twice. The utility function getCMTableFromTable was created.


packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/CellResizer.ts
```diff
@@ -107,24 +107,25 @@ function onDragStart(context: DragAndDropContext, event: MouseEvent): DragAndDro
...
-
-    // Get the table content model
-    const cmTable = getFirstSelectedTable(editor.getContentModelCopy('disconnected'))[0];
-
-    // Restore selection
-    editor.setDOMSelection(selection);
+    // Get Table block in content model
+    let cmTable: ContentModelTable | undefined;
+
+    editor.formatContentModel(
```
JiuqingSong, line 113: it seems the 3 calls are the same. Can you extract them to be a shared function?

Just do internal export should be good enough.

JiuqingSong (approved):

PR #2657, opened 2024-05-24
Fix normalisation and First Column issues

Andres-CT98:
* normalizeTable should allow table headers on the first column and row, as this are valid table layouts.
* setTableCellShade should not normalise tables as normalisation is unrelated to cell shading.
* setFirstColumnFormatBorders should only apply isHeader to cells in the first column except the first row.
* formatCells should remove color to cells in the first column except the first row.


packages/roosterjs-content-model-dom/lib/modelApi/editing/applyTableFormat.ts
```diff
@@ -261,6 +261,7 @@ export function setFirstColumnFormatBorders(
 
                 switch (rowIndex) {
                     case 0:
+                        cell.isHeader = false;
```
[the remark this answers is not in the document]
Andres-CT98: Yes, the idea is that First Column does not apply to the first cell, at least that is the behaviour we implemented and I am just restoring it.

However I noticed that this can conflict with hasHeaderRow, so I made a change to apply the header status accordingly.

`cell.isHeader = !!format.hasHeaderRow;`

packages/roosterjs-content-model-dom/lib/modelApi/editing/applyTableFormat.ts
```diff
@@ -261,6 +261,7 @@ export function setFirstColumnFormatBorders(
 
                 switch (rowIndex) {
                     case 0:
+                        cell.isHeader = !!format.hasHeaderRow;
                         break;
                     case 1:
                         setBorderColor(cell.format, 'borderBottom');
```
JiuqingSong, line 267: Not relate to this change, but I realize that it is possible there are only 2 rows, in that case, "1" is the same with "rows.length-1", but then there are different handling. Is that ok?

JiuqingSong (approved):

packages/roosterjs-content-model-dom/lib/modelApi/editing/applyTableFormat.ts
Andres-CT98: Good catch, they should be flipped in order. The order should be row 0, then last row, then row 1, and then any other row.

JiuqingSong (approved):

PR #2673, opened 2024-05-30
[Content Model] Add skipKnownColorsWhenGetDarkColor in DarkColorHandler

haven2world:
We need an approach to allow `DarkColorHandler` to skip some elemnets and let clients decide its color. I added two changes in this PR.

1. Added a new option `skipKnownColorsWhenGetDarkColor`, when it's enabled, we will always skip the color in `knownColors` and call `getDarkColor`.
2. Allow `getDarkColor` to return `null`, when it return `null`, we won't update the color in `knownColors` and use the current color for the element.

I'm trying to not break the existing CSS variable logic, so we still update the `knownColors` but won't use it when we need a color for dark mode. It'll be useful when we need a light color for a dark mode color sometimes.
I also use a `COLOR_VAR_ENFORCE_FALLBACK_KEY` to construct those CSS variables of elements' color which we want to skip handling its dark mode, so the CSS variables will be always undefined and fallback to its light color.

### DIFF
--- a/packages/roosterjs-content-model-core/lib/editor/core/DarkColorHandlerImpl.ts
+++ b/packages/roosterjs-content-model-core/lib/editor/core/DarkColorHandlerImpl.ts
@@ -8,7 +8,8 @@ class DarkColorHandlerImpl implements DarkColorHandler {
     constructor(
         private readonly root: HTMLElement,
         public getDarkColor: ColorTransformFunction,
-        public readonly knownColors: Record<string, Colors>
+        public readonly knownColors: Record<string, Colors>,
+        public readonly skipKnownColorsWhenGetDarkColor: boolean
     ) {}
 
     updateKnownColor(isDarkMode: boolean, key?: string, colorPair?: Colors): void {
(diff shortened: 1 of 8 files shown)


JiuqingSong:
In general, what I was thinking is:
1. add an option in EditorOptions
2. pass it to DarkColorHandler
3. After calculate dark color, check if this option is true, then skip setting into knownColors

No other changes is needed.

haven2world:
> In general, what I was thinking is:
> 
> 1. add an option in EditorOptions
> 2. pass it to DarkColorHandler
> 3. After calculate dark color, check if this option is true, then skip setting into knownColors
> 
> No other changes is needed.

That's what I tried at first, but it didn't work well.
1. In `getColor`, if an element's color is a color string rather than a CSS variable under dark mode, we'll try to find it's corresponding light color by lookup `knownColors` in `findLightColorFromDarkColor`. In this case, if we don't update `knownColors`, the logic won't work.
2. Based on 1, if we always update known color but doesn't use it in `setColor`, then based on current `ColorTransformFunction`, a color string will always be returned, and if we want to keep the color of an element, we need to return its current color, and the color will be recored into the `knownColors`. So `ColorTransformFunction` need to be able to return `null` to differentiate this case.

PR #2675, opened 2024-05-30
Fix Table Edit plugin tests

Andres-CT98:
Revamped tests for Table Edit Plugin to not use mouse events and reorganised some tests.

The following changes were made for testing: 
* Made onMouseMove from TableEditor public
* Exported the onDragStart and onDragging functions of CellResizer and TableResizer
* Exported the TableInsertHandler class

Also renamed the Context and Initvalue interfaces for CellResizer and TableResizer.


packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/CellResizer.ts
```diff
@@ -68,7 +68,11 @@ export function createCellResizer(
     return { node: td, div, featureHandler };
 }
 
-interface DragAndDropContext {
+/**
+ * @internal
+ * Exported for testing
+ */
+export interface cellResizerContext {
```
JiuqingSong, line 75: Please use upper case for the first letter for type name

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/CellResizer.ts
```diff
@@ -77,15 +81,23 @@ interface DragAndDropContext {
     onStart: () => void;
 }
 
-interface DragAndDropInitValue {
+/**
+ * @internal
+ * Exported for testing
+ */
+export interface cellResizerInitValue {
```
JiuqingSong, line 88: same

packages/roosterjs-content-model-plugins/test/tableEdit/tableInserterTest.ts
```diff
@@ -1,124 +1,108 @@
...
-
-            table = getCurrentTable(editor);
-            const newRows = getTableRows(table);
-            const newCols = getTableColumns(table);
-            expect(newRows).toBe(inserterType == VERTICAL_INSERTER_ID ? rows : rows + 1);
-            expect(newCols).toBe(inserterType == HORIZONTAL_INSERTER_ID ? cols : cols + 1);
-        }
-        return !!inserter ? 'found' : 'not found';
+        //Act
+        div.click();
```
JiuqingSong, line 85: nit: Not something we must fix but want to bring up. We can also consider not use something like "element.click()" but instead we directly call the click handler function so we don't need to rely on the browser's event routing.

JiuqingSong (approved):

PR #2678, opened 2024-05-31
Fix cell resizer normalisation

Andres-CT98:
### Description:

There have been instances of the following error occurring on CellResizer.ts and TableResizer.ts: 
**`Error Message: Uncaught TypeError: Cannot read properties of undefined (reading 'height')`**

The specific property call happens on these lines:
CellResizer.ts: `td.style.height = cmTable.rows[anchorRow].height + 'px';`
TableResizer.ts: `td.style.height = cmTable.rows[row].height + 'px';`

The error is caused after the call to normalizeTable, as the same height property is accessed before normalisation while throwing no issues. In some cases, if a table has never been normalised it is possible for it to lose rows after normalisation, so an error like this one could happen. This happens when an HTML Table contains any row which all its cells have the property rowspan greater than 1, causing a "hidden" row.

### Fix

The call to normalise the whole table is unnecessary, because the onDragging methods are only intended to modify the cell's properties width and height. So the call to normalise was replaced with a single step to guarantee the new row height, and column width, is not less than the minimum allowed.

As an abundance of caution, a check to verify that `cmTable.rows[anchorRow]` is not undefined was added in CellResizer.ts.

### Cases tested:

1. Open Inspect in browser
2. Using the demo site, restore the following snapshot: `<table style="border-collapse: collapse; border-spacing: 0px; box-sizing: border-box;"><tbody><tr><td rowspan="2"><div >a</div></td><td rowspan="2" ><div >b</div></td><td rowspan="2" ><div >c</div></td></tr><tr></tr><tr><td ><div >a</div></td><td ><div >b</div></td><td ><div >c</div></td></tr><tr><td rowspan="2" colspan="3" ><div >a</div></td></tr><tr></tr></tbody></table>`
3. Attempt to resize the whole table or just the last row
4. Observe the console

**Before:** Error message is thrown

**After:** No error message

### Warnings and implications

- A similar error is unlikely to occur on operations involving columns, because the in case of "hidden" columns, a resize will not be applied or will be applied to the previous "real" column.


packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/CellResizer.ts
```diff
@@ -159,18 +159,21 @@ export function onDraggingHorizontal(
...
     // Assign new widths and heights to the CM table
-    if (cmTable && anchorRow != undefined) {
+    if (cmTable && anchorRow != undefined && cmTable.rows[anchorRow] != undefined) {
         // Modify the CM Table size
         mutateBlock(cmTable).rows[anchorRow].height = (anchorRowHeight ?? 0) + deltaY;
 
-        // Normalize the table
-        normalizeTable(cmTable);
+        // Normalize the new height value
+        const newHeight =
```
JiuqingSong, line 167: nit: for this kind of operation, you can always use Math.max()

newHeight = Math.max(cmTable.rows[anchorRow].height, MIN_ALLOWED_TABLE_CELL_HEIGHT)

JiuqingSong (approved):

JiuqingSong (approved):

PR #2679, opened 2024-06-03
Increase the probability that removeUnnecessarySpan and mergeNode will be triggered.

miku1958:
After customizing the handler it may appear that duplicate attributes and styles are set on the child, add these two methods to remove them so that removeUnnecessarySpan and mergeNode can work properly


packages/roosterjs-editor-types/lib/Constants.ts
```diff
@@ -0,0 +1,11 @@
+/**
```
JiuqingSong, line 1: Do not modify any code from old packages (roosterjs-editor-*), we will delete those old code files soon.

If need, add them into new packages.

And do not add consts to type package. Type package is suppose to generate 0 js code. It is just type declaration.

packages/roosterjs-editor-dom/lib/htmlSanitizer/getInheritableStyles.ts
```diff
@@ -1,15 +1,6 @@
+import { INHERITABLE_PROPERTIES } from 'roosterjs-editor-types';
```
JiuqingSong, line 1: do not change files from old packages. We will delete them soon

JiuqingSong: See https://github.com/microsoft/roosterjs/pull/2660 for reference.

packages/roosterjs-content-model-dom/package.json
```diff
@@ -3,7 +3,8 @@
     "description": "Content Model for roosterjs",
     "dependencies": {
         "tslib": "^2.3.1",
-        "roosterjs-content-model-types": ""
+        "roosterjs-content-model-types": "",
+        "roosterjs-editor-types": ""
```
JiuqingSong, line 7: do not add dependency to old packages. Content model code is brand new, it has no dependency to old code. Old code will be deleted

packages/roosterjs-content-model-dom/lib/modelToDom/optimizers/optimize.ts
```diff
@@ -1,22 +1,38 @@
...
  */
-export function optimize(root: Node) {
+export function optimize(root: Node, isRecursive: boolean = false) {
     /**
      * Do no do any optimization to entity
      */
     if (isEntityElement(root)) {
         return;
     }
+    if (!isRecursive && root.parentElement != null) {
```
JiuqingSong, line 17: Actually, instead of adding those attributes then remove them, it is better not to add them at all.

We can do this when generate HTML element, which is earlier than here. So no need to remove them here.

Another concern is perf. This function will be run to a lot of elements, so we need to be careful here, not adding too much loops.

If your requirement is about LI element, we can do this optimization when create LI in handleListItem.ts by adding those inline styles into context styles so styles with the same value will not be generated at all when create texts.

Let me see if I can make a PR.

JiuqingSong:
I published the PR https://github.com/microsoft/roosterjs/pull/2683 so close this one.

Feel free to reopen if that one can't solve your issue.

PR #2681, opened 2024-06-04
Merge Link & Image Format when using MergeModel

BryanValverdeU:
Add functionality to merge the link format. So we can successfully merge the target and source styles on paste if the option to merge style is chosen.

Also fix for images, as some formatting like width was getting lost when pasting with mergeFormat option.

Copy the link in this pen and paste and merge format:
https://codepen.io/Bryanvalverdeu/pen/ExzmOKb

Before
image

After
image


JiuqingSong (approved):

packages/roosterjs-content-model-dom/lib/modelApi/editing/mergeModel.ts
```diff
@@ -405,3 +436,29 @@ function getSemanticFormat(segmentFormat: ContentModelSegmentFormat): ContentMod
...
 }
+
+/**
+ * Segment format can also contain other type of metadata, for example in Images/Hyperlink,
+ * we want to preserve these properties when merging format
+ */
+function getFormatWithoutSegmentFormat(
+    sourceFormat: ContentModelSegmentFormat
+): ContentModelSegmentFormat {
+    const resultFormat = {
```
JiuqingSong, line 447: Can we just simply copy the format that we want to copy, but not copy all then delete others?

I feel the format we want to keep is less than the ones we want to remove

JiuqingSong: Or, if we need to copy everything other than segment format, you can do like this:

1. add a object to contains all properties of segment:
```ts
// An object to provide keys of required properties of segment format, the do NOT use any of its values
const requiredEmptySegmentFormat: Required<ContentModelSegmentFormat> = {
  fontFamily: null!,
  ...
}
```
2. When copy format, delete format with keys of this object
```ts
const format = { ... }
getObjectKeys(requiredEmptySegmentFormat).forEach(key => delete format[key]);
```

So that if we add some new property to segment format in the future, we can still take care of them here since it will break the build.

BryanValverdeU: > Or, if we need to copy everything other than segment format, you can do like this:
> 
> 1. add a object to contains all properties of segment:
> 
> ```ts
> // An object to provide keys of required properties of segment format, the do NOT use any of its values
> const requiredEmptySegmentFormat: Required<ContentModelSegmentFormat> = {
>   fontFamily: null!,
>   ...
> }
> ```
> 
> 2. When copy format, delete format with keys of this object
> 
> ```ts
> const format = { ... }
> getObjectKeys(requiredEmptySegmentFormat).forEach(key => delete format[key]);
> ```
> 
> So that if we add some new property to segment format in the future, we can still take care of them here since it will break the build.

This looks better, I agree. Changing this.

JiuqingSong: Ok, makes sense

JiuqingSong (approved):

PR #2683, opened 2024-06-04
Skip same format from LI

JiuqingSong:
When `<LI>` and segments both have some inline format, we don't need to render the same format for segment. For example:

```html
<ul>
  <li style="font-family: Arial; font-size: 10pt; color: red;">
     <div>
        <span style="font-size: 12pt; font-family: Arial; font-weight: bold">test</span>
     </div>
   </li>
</ul>
```

Here both LI and SPAN have same format for font-family, so we can skip font-family for SPAN.

The expected render result is

```html
<ul>
   <li style="font-family: Arial; font-size: 10pt; color: red;">
      <div style="font-size: 12pt;">
         <b>test</b>
      </div>
   </li>
</ul>
```


juliaroldi (approved):

miku1958:
Is it possible to use stackFormat for the custom handler?

JiuqingSong:
> Is it possible to use stackFormat for the custom handler?

We didn't make this function public yet. But you need it, we can export it. Feel free to submit a PR.

Please note that there are two "stackFormat", one for dom=>model, one for model=>dom. If we export them, we'd better rename them otherwise they will have same name.

PR #2684, opened 2024-06-05
Auto format trigger

juliaroldi:
Trigger event Auto format event in AutoFormatPlugin, so we can trigger the AutoFormat UI on OWA.
Screenshot 2024-06-05 at 19 52 23

autoFormatUI

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
@@ -220,14 +220,9 @@ export class AutoFormatPlugin implements EditorPlugin {
                                 shouldFraction ||
                                 shouldOrdinals
                             );
-                        }
+                        },
+                        formatOptions
                     );
-                    if (formatOptions.changeSource) {
-                        editor.triggerEvent('contentChanged', {
-                            source: formatOptions.changeSource,
-                            formatApiName: formatOptions.apiName,
-                        });
-                    }
 
                     break;
             }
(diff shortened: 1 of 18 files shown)


packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
```diff
@@ -220,9 +220,14 @@ export class AutoFormatPlugin implements EditorPlugin {
                                 shouldFraction ||
                                 shouldOrdinals
                             );
-                        },
-                        formatOptions
+                        }
                     );
+                    if (formatOptions.changeSource) {
```
BryanValverdeU, line 225: What if instead of triggering another plugin event here we change the formatWithContentModel implementation by moving [line 27](https://github.com/microsoft/roosterjs/blob/f8a23bb492e1d08165f182dd27d6f9a3327d9cd4/packages/roosterjs-content-model-core/lib/coreApi/formatContentModel/formatContentModel.ts#L27) tbefore [line 81](https://github.com/microsoft/roosterjs/blob/f8a23bb492e1d08165f182dd27d6f9a3327d9cd4/packages/roosterjs-content-model-core/lib/coreApi/formatContentModel/formatContentModel.ts#L81C1-L89C15)

Since the option object is destructed to get the values and we modify the options object in the callback the AutoFormat Change source is not correctly set. 

@JiuqingSong to check if this is okay.

juliaroldi: Thanks, @BryanValverdeU!! That worked out better that my solution.

JiuqingSong (approved):

PR #2692, opened 2024-06-12
Do not merge Hyperlink text color

BryanValverdeU:
Do not merge Hyperlink text color to follow Word's behavior when using MergePaste.


JiuqingSong (approved):

JiuqingSong:
One little concern is, the file mergeModel.ts becomes longer and longer. We need to refactor it some time.

BryanValverdeU:
> One little concern is, the file mergeModel.ts becomes longer and longer. We need to refactor it some time.

Agree, I think the merge formatting functions can be moved to another file.
I will do this in a follow up PR.

PR #2694, opened 2024-06-12
Fix insert entity issue

JiuqingSong:
This is a little bit hard to explain.

Suppose we have the DOM structure:

```html
<div><span id=entity>...</span>[CURSOR]<div>
```

Then we call `insertEntity` to insert an entity and put the existing SPAN into the new entity, we set position to be right before this SPAN so the new entity should be inserted to replace the SPAN, and cursor is not moved.

However, inside `insertEntity` we will move the SPAN into entity wrapper, then call `formatContentModel` to insert entity. Since the SPAN is removed from the DOM tree, we need to recreate a content model, and in that case the real cursor and override cursor will be at the same place. Then our code will put the real selection marker before the override selection marker, which causes the entity is inserted after focus.

To fix this, we can check if the override selection marker is right after real one, then use the real one instead in that case.

### DIFF
--- a/packages/roosterjs-content-model-api/test/modelApi/entity/insertEntityModelTest.ts
+++ b/packages/roosterjs-content-model-api/test/modelApi/entity/insertEntityModelTest.ts
@@ -2620,4 +2620,51 @@ describe('insertEntityModel, use insert point', () => {
             ],
         });
     });
+
+    it('Block entity, Has insert point override which is right after real marker', () => {
+        const model = createContentModelDocument();
+        const para1 = createParagraph();
+        const para2 = createParagraph();
+        const text1 = createText('test');
+        const markerReal = createSelectionMarker();
+        const markerOverride = createSelectionMarker();
+
+        text1.isSelected = true;
+        para1.segments.push(text1);
+
+        markerOverride.isSelected = false;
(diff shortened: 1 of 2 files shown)


packages/roosterjs-content-model-api/test/modelApi/entity/insertEntityModelTest.ts
```diff
@@ -2620,4 +2620,51 @@ describe('insertEntityModel, use insert point', () => {
...
+
+        const ip: InsertPoint = {
+            path: [model],
+            paragraph: para2,
+            marker: markerOverride,
+        };
+
+        insertEntityModel(model, Entity, 'focus', false, true, undefined, ip);
+
+        console.log(JSON.stringify(model));
```
BryanValverdeU, line 2648: nit: console log

BryanValverdeU (approved):

packages/roosterjs-content-model-api/test/modelApi/entity/insertEntityModelTest.ts
JiuqingSong: good catch. Fixed

BryanValverdeU (approved):

PR #2696, opened 2024-06-12
(prototype, do not review)

JiuqingSong:


packages/roosterjs-content-model-core/lib/editor/core/createEditorCore.ts
```diff
@@ -18,6 +20,21 @@ import type {
...
     const corePlugins = createEditorCorePlugins(options, contentDiv);
+    const plugins = (options.plugins ?? []).filter(x => !!x);
+    const domToModelOptions: DomToModelOption[] = [];
+    const modelToDomOptions: ModelToDomOption[] = [];
+
+    plugins.forEach(x => {
+        const config = x.getContentModelConfig?.();
+
+        if (config?.domToModelOption) {
+            domToModelOptions.push(config.domToModelOption);
```
juliaroldi, line 31: If two plugins override the same element handler (such as image), both of the handlers will take effect on the element?

JiuqingSong: If that really happens, the second one wins

JiuqingSong: Here in ImageEditPlugin I only add an additional format parser and applier, I won't really override handlers.

juliaroldi: should we make these additional parses and applier coexist? Otherwise, a plugin parser/applier can break a functionality of the plugin.

JiuqingSong: We don't hard request them coexist. Sometimes we only need one of them, although here we need both.

PR #2701, opened 2024-06-14
Fix auto link event in create link

juliaroldi:
When trigger a create link event, only add the anchor element in content data.


packages/roosterjs-content-model-plugins/lib/autoFormat/link/createLink.ts
```diff
@@ -30,7 +30,11 @@ export function createLink(editor: IEditor) {
         {
             changeSource: ChangeSource.AutoLink,
             onNodeCreated: (_modelElement, node) => {
-                if (!anchorNode) {
+                if (
+                    !anchorNode &&
+                    isNodeOfType(node, 'ELEMENT_NODE') &&
+                    isElementOfType(node, 'a')
+                ) {
```
JiuqingSong, line 37: actually, this will just add the first "A". If there are multiple and you are not operating on the first one, this will still be wrong.

You need to check if modelElement is the one you want too

JiuqingSong (approved):

PR #2713, opened 2024-06-18
Edit Image with image selection

juliaroldi:
This adds the functionality of edit an image using drag and drop rotate/resize handles. To achieve this the following changes were added
1. Selection Plugin was modified to select the image on the left click of mouse down event. Also, when the cursor should be moved to the after or before the image, the scenario when the image is in shadow down is considered. 
2. Image Edit was  modified to add wrapper when the user click on image or click in the crop button. This will add isEditing datataset to indicate the image is in editing mode. 

ImageEdit


packages/roosterjs-content-model-types/lib/editor/EditorPlugin.ts
```diff
@@ -1,6 +1,16 @@
+import type { DomToModelOption } from '../context/DomToModelOption';
+import type { ModelToDomOption } from '../context/ModelToDomOption';
 import type { PluginEvent } from '../event/PluginEvent';
 import type { IEditor } from './IEditor';
 
+/**
+ * Configuration for content model of a plugin
+ */
+export interface PluginContentModelConfig {
+    domToModelOption?: DomToModelOption;
```
JiuqingSong, line 10: Add comment for each member

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -525,7 +497,11 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
 
     private selectBeforeOrAfterElement(editor: IEditor, element: HTMLElement, after?: boolean) {
         const doc = editor.getDocument();
-        const parent = element.parentNode;
+        let parent = element.parentNode;
+        if (isNodeOfType(parent, 'ELEMENT_NODE') && parent.shadowRoot && parent.parentNode) {
+            element = parent;
+            parent = parent.parentNode;
+        }
```
JiuqingSong, line 504: why need this?

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -546,7 +522,6 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
 
     private getClickingImage(event: UIEvent): HTMLImageElement | null {
         const target = event.target as Node;
-
```
JiuqingSong, line 549: let's keep this empty line

packages/roosterjs-content-model-types/lib/editor/EditorPlugin.ts
```diff
@@ -42,4 +52,9 @@ export interface EditorPlugin {
      * @param event The event to handle:
      */
     onPluginEvent?: (event: PluginEvent) => void;
+
+    /**
+     * @returns The content model configuration for this plugin
+     */
+    getContentModelConfig?: () => PluginContentModelConfig;
```
JiuqingSong, line 59: Add comment to explain what it does. And need to say this method will be called before initialize(), so we can't use editor object in this method.

packages/roosterjs-content-model-types/lib/parameter/ImageEditor.ts
```diff
@@ -20,7 +20,12 @@ export type ImageEditOperation =
     /**
      * Flip an image
      */
-    | 'flip';
+    | 'flip'
+
+    /**
+     * Resize and rotate an image
+     */
+    | 'resizeAndRotate';
```
JiuqingSong, line 28: Curious, if we want to start resize and rotate, can we let the function accepts an array and pass in ['resize', 'rotate']? So that no need to have a separate operation 'reiszeAndRotate'

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -118,7 +132,120 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
...
+        }
+    }
+
+    private keyDownHandler(editor: IEditor, event: KeyDownEvent) {
+        if (this.isEditing) {
+            const selection = editor.getDOMSelection();
+            if (!isModifierKey(event.rawEvent)) {
+                this.selectionChangeHandler(editor, selection);
+            } else if (selection?.type == 'image') {
+                this.formatImageWithContentModel(
```
JiuqingSong, line 183: what is this for?

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -118,7 +132,120 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
...
+                );
+            }
+        }
+    }
+
+    private selectionChangeHandler(editor: IEditor, selection: DOMSelection | null) {
+        editor.formatContentModel(model => {
+            const previousSelectedImage = findEditingImage(model);
+            const editingImage = getSelectedImage(model);
+            const format = editingImage?.image.format as EditableImageFormat;
```
JiuqingSong, line 196: as EditableImageFormat|undefined.

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
JiuqingSong: Is this for applying change when moving cursor?

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
juliaroldi: when type over an image, the cursor was going into the shadow dom, to ensure the cursor is being placed out of the shadow dom, I added this code. it makes copy/cut possible.

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
juliaroldi: If a modifier key is not pressed, the cursor will change the position, so we consider as selection change. If a modifier key is a pressed, we apply the change and leave the selection in the image.

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
JiuqingSong: Hmm in my test branch, I'm selecting the whole image (or whole wrapper SPAN), then we don't need to do any special handling for keyboard event since the image is really selected. Can we still do that?

Or at least, can we handle shadow dom in ImageEdit plugin since SelectionPlugin should not know the detail about shadow DOM

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
JiuqingSong: add comment to explain this.

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -38,10 +46,10 @@ const DefaultOptions: Partial<ImageEditOptions> = {
     preserveRatio: true,
     disableRotate: false,
     disableSideResize: false,
-    onSelectState: 'resize',
+    onSelectState: ['resize', 'rotate'],
 };
 
-const IMAGE_EDIT_CHANGE_SOURCE = 'ImageEdit';
+//const LEFT_MOUSE_BUTTON = 0;
```
JiuqingSong, line 52: Remove this

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -118,14 +130,170 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
...
+                    this.isCropMode,
+                    isModifierKey(event.rawEvent) && isImageSelection //if it's a modifier key over a image, the image should select the image
+                );
+            }
+        }
+    }
+
+    private applyFormatWithContentModel(
+        editor: IEditor,
+        selection: DOMSelection | null,
```
JiuqingSong, line 202: If the selection is always retrieved from editor, no need to pass it in, but get it from editor

JiuqingSong (approved):

PR #2718, opened 2024-06-20
Using Tab key on table selects the whole next cell

Andres-CT98:
### Description:

Using the Tab key (or Shift+Tab) moves the selection marker to an adjacent cell, or the previous/next row. The position is always at the start of the cell.

### Fix

Modify the behaviour of using the Tab key in a table so that instead of collapsing the selection at the start, it selects the whole contents of the cell.

### Cases tested:

1. Create a Table and populate it with content
2. Use the tab key to move across cells

**Before:** After movement, the selection is always at the start of the cell

**After:** After movement, the selection encompasses all the contents of the cell

### DIFF
--- a/packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
+++ b/packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
@@ -411,6 +411,7 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
             }
 
             let lastCo = findCoordinate(tableSel?.parsedTable, end, domHelper);
+            let tabMove = false;
             const { parsedTable, firstCo: oldCo, table } = this.state.tableSelection;
 
             if (lastCo && tableSel.table == table) {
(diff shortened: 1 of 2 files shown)


packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -465,7 +466,8 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
                         const cell = parsedTable[row][col];
 
                         if (typeof cell != 'string') {
-                            this.setRangeSelectionInTable(cell, 0, this.editor);
+                            tabMove = true;
+                            this.setRangeSelectionInTable(cell, 0, this.editor, true);
```
JiuqingSong, line 470: Add inline comment for "true"

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -486,20 +488,29 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
                 }
             }
 
-            if (!collapsed && lastCo) {
+            if (!collapsed && lastCo && tabMove == false) {
```
JiuqingSong, line 491: Just check "!tabMove"

JiuqingSong (approved):

PR #2724, opened 2024-06-21
Ensure focusNode exists when selectionChanges

juliaroldi:
When selectionChange event happens, ensure the focus exist before setting the selection in the new position, otherwise
getRangeSelection(0) will return an error, since nothing is selected.


packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -595,7 +595,7 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
             //If am image selection changed to a wider range due a keyboard event, we should update the selection
             const selection = this.editor.getDocument().getSelection();
 
-            if (newSelection?.type == 'image' && selection) {
+            if (newSelection?.type == 'image' && selection && selection.focusNode) {
                 if (selection && !isSingleImageInSelection(selection)) {
```
JiuqingSong, line 599: Here we checked "selection" twice, in line 598 and 599. These two lines can be merged

PR #2726, opened 2024-06-24
Fix context menu for Image Edit menu

juliaroldi:
To make context menu work with image edit plugin, when the user right clicks on an image, the editing operations will be applied, and the image will be selected. Also, if the image was selected and right click occurs in an element that is not a image, remove image selection. 
ContextMenuOnImages

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
@@ -28,7 +28,6 @@ import type { ImageHtmlOptions } from './types/ImageHtmlOptions';
 import type { ImageEditOptions } from './types/ImageEditOptions';
 import type {
     ContentModelImage,
-    DOMSelection,
     EditorPlugin,
     IEditor,
     ImageEditOperation,
@@ -48,6 +47,8 @@ const DefaultOptions: Partial<ImageEditOptions> = {
     disableSideResize: false,
     onSelectState: ['resize', 'rotate'],
 };
+
+const MouseRightButton = 2;
 
 /**
  * ImageEdit plugin handles the following image editing features:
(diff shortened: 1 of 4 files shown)


BryanValverdeU:
Hey Julia, I think this looks good. is Shift + F10 to open the context menu should also be supported?

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -143,14 +144,31 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
...
+            }
+            if (
+                isElementOfType(target, 'span') &&
+                target.firstElementChild &&
+                isNodeOfType(target.firstElementChild, 'ELEMENT_NODE') &&
+                isElementOfType(target.firstElementChild, 'img')
+            ) {
+                return true;
+            }
+            return false;
```
BryanValverdeU, line 160: nit: I think you can replace this with: 

```
           return (isElementOfType(target, 'img')) || (
                isElementOfType(target, 'span') &&
                target.firstElementChild &&
                isNodeOfType(target.firstElementChild, 'ELEMENT_NODE') &&
                isElementOfType(target.firstElementChild, 'img')
            )
```

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -97,11 +98,11 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
...
-                        this.applyFormatWithContentModel(
-                            this.editor,
-                            this.isCropMode,
-                            true /* shouldSelectImage */
-                        );
+                        // this.applyFormatWithContentModel(
+                        //     this.editor,
+                        //     this.isCropMode,
+                        //     true /* shouldSelectImage */
+                        // );
```
BryanValverdeU, line 105: Should we remove this?

juliaroldi: I was using to debug and forgot to uncomment it!

juliaroldi:
> Hey Julia, I think this looks good. is Shift + F10 to open the context menu should also be supported?

I didn't test it, at least for mac, this shortcut controls volume

BryanValverdeU:
> > Hey Julia, I think this looks good. is Shift + F10 to open the context menu should also be supported?
> 
> I didn't test it, at least for mac, this shortcut controls volume

I see for Window it also triggers the context menu.

juliaroldi:
> > > Hey Julia, I think this looks good. is Shift + F10 to open the context menu should also be supported?
> > 
> > 
> > I didn't test it, at least for mac, this shortcut controls volume
> 
> I see for Window it also triggers the context menu.

It will not trigger, because only modifier key selects the image, let me add shift to select the image too

BryanValverdeU (approved):

BryanValverdeU:
Shift + F10 is working now:
Fixed

BryanValverdeU (approved):

PR #2727, opened 2024-06-24
Only dismiss the Table Mover if the end of the drag is not in the Table Mover div

BryanValverdeU:
The Table Mover feature is incorrectly dismissed when only clicking it. To fix do not dismiss the table mover if the target on drag end target is not the div containing the Table mover.

Before (Notice that on click the mouse cursor is set to arrow because the table mover element got detached.)
TableMoverBefore

After
TableMoverAfter


Andres-CT98 (approved):
Everything looks fine. I was worried that maybe not disposing it would cause a potential double Table Mover, but we did already have a check preventing that.

juliaroldi (approved):

PR #2728, opened 2024-06-24
Expose splittext api

FrancisMengx:


packages/roosterjs-content-model-plugins/test/pluginUtils/splitTextSegmentTest.ts
```diff
@@ -1,5 +1,5 @@
 import { ContentModelParagraph, ContentModelText } from 'roosterjs-content-model-types';
-import { splitTextSegment } from '../../lib/pluginUtils/splitTextSegment';
+import { splitTextSegment } from 'roosterjs-content-model-api';
```
juliaroldi, line 2: Could you move this test case to right content-model-api package?

packages/roosterjs-content-model-api/lib/publicApi/utils/splitTextSegment.ts
```diff
@@ -5,7 +5,12 @@ import type {
 } from 'roosterjs-content-model-types';
 
 /**
- * @internal
```
juliaroldi, line 8: Please move it to segments folder, since it related only to text segments.

flyingbee2012 (approved):

PR #2729, opened 2024-06-25
Add option for Tab Key handling in EditPlugin

Rain-Zheng:
EditPlugin overrides the default behavior of Tab key by increasing the margin-left value of current line. As a result, user can't navigate to next focusable element by pressing Tab key, so we need an option for disabling this behavior.


BryanValverdeU:
LGTM, could you please add a unit test for this scenario?

BryanValverdeU:
#2729

BryanValverdeU (approved):

PR #2738, opened 2024-06-28
Bump Rooster main version to 9.7

romanisa:
Bump Rooster main version to 9.7.0


versions.json
```diff
@@ -1,6 +1,6 @@
 {
     "react": "9.0.0",
-    "main": "9.6.0",
+    "main": "9.7.0",
     "legacyAdapter": "8.62.1",
     "overrides": {
         "roosterjs-content-model-plugins": "9.6.1"
```
juliaroldi, line 6: I think we can remove this override, since we want all packages in version 9.7.0

juliaroldi (approved):

BryanValverdeU (approved):

Issue #2735, opened 2024-06-28
[A11y] Add annoncements when a bulleted and numbered lists are created


vhuseinova-msft:
**Describe the bug**
Currently RoosterJS allow to customize strings for announcing `announceListItemBullet`, `announceListItemNumbering` and `announceOnFocusLastCell`. This GitHub issue is for adding a few new options when:
- a bulleted list is added 
- a numbered list is added.
The announcement should be used when bulleted/numbered lists are added as part of autoformatiing and when  `toggleBullet`/`toggleNumbering` apis are called.

**To Reproduce**
1. Open RoosterJS demo app
2. Add a numbered/bulleted list
3. Observe that there is no announcements that lists are created (and there is no options to set this in code)

**Expected behavior**
Developers are able to add/customize strings that will be used when bulleted/numbered lists are created.
 

**Device Information**
 - OS: [e.g. iOS] MacOs
 - Browser [e.g. chrome, safari] Edge

vhuseinova-msft:
cc @BryanValverdeU

PR #2743, opened 2024-07-11
Fix #2734 by Setting List Metadata `applyListStyleFromLevel: true` when toggling a list

### LINKED ISSUE #2734: [A11y] Value returned for `announceListItemNumbering` in `announcerStringGetter` is not used when numbered list is created with `toggleNumbering` api
**Describe the bug**
RoosterJS provides a way to set a custom announcements for new items in numbered list (`announcerStringGetter`). But it looks like value returned for `announceListItemNumbering` in `announcerStringGetter` is not used when numbered list is created with `toggleNumbering` api. 
Please note that numbered lists that are created with auto formatting, use `announceListItemNumbering` as expected.

**To Reproduce**
1. Open Rooster JS demo web site
2. Click add numbered list button
3. type some text
4. Press enter
5. Observe that Narrator/VoiceOver doesn't announce a custom announcement for  `announceListItemNumbering`

**Expected behavior**
The `announceListItemNumbering` string returned in `announcerStringGetter` should be used to announce when a new bullet is added (in autoformatted list and when list is added with `toggleNumbering` api)

**Device Information**
 - OS: [e.g. iOS] MacOS
 - Browser [e.g. chrome, safari] Edge

BryanValverdeU:
This will fix #2734.

Right now, we do not set metadata when toggling a numbering or bullet list.
This causes  [getListAnnounceData](https://github.com/microsoft/roosterjs/blob/f57a68c5014b404a1e41a977ad685db01a2d47b0/packages/roosterjs-content-model-api/lib/modelApi/list/getListAnnounceData.ts#L32C31-L32C51 ) to return undefined, so when using Enter on a List we do not announce anything.

To fix, always set the applyListStyleFromLevel metadata so we announce the new list item marker.

To test:
1. Start Narrator
2. Toggle List item using buttons in ribbon
3. Add text to first item
4. Press enter to create a new list item

Before
No announcing of the new list item

After
We announce the new list item marker.

### DIFF
--- a/packages/roosterjs-content-model-api/lib/modelApi/list/setListType.ts
+++ b/packages/roosterjs-content-model-api/lib/modelApi/list/setListType.ts
@@ -6,6 +6,7 @@ import {
     mutateBlock,
     normalizeContentModel,
     setParagraphNotImplicit,
+    updateListMetadata,
 } from 'roosterjs-content-model-dom';
 import type {
     ContentModelListItem,
@@ -121,6 +122,17 @@ export function setListType(
 
                     mutateBlock(parent).blocks.splice(index, 1, newListItem);
                     existingListItems.push(newListItem);
+
+                    const levelIndex = newListItem.levels.length - 1;
+                    const level = mutateBlock(newListItem).levels[levelIndex];
+
+                    if (level) {
+                        updateListMetadata(level, metadata =>
+                            Object.assign({}, metadata, {
+                                applyListStyleFromLevel: true,
+                            })
+                        );
+                    }
                 } else {
                     existingListItems.forEach(
                         x => (mutateBlock(x).levels[0].format.marginBottom = '0px')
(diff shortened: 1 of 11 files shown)


packages/roosterjs-content-model-api/lib/modelApi/list/setListType.ts
```diff
@@ -121,6 +122,17 @@ export function setListType(
...
                     mutateBlock(parent).blocks.splice(index, 1, newListItem);
                     existingListItems.push(newListItem);
+
+                    const levelIndex = newListItem.levels.length - 1;
+                    const level = mutateBlock(newListItem).levels[levelIndex];
+
+                    if (level) {
+                        updateListMetadata(level, metadata =>
+                            Object.assign({}, metadata, {
+                                applyListStyleFromLevel: true,
```
juliaroldi, line 132: besides here, when do we set this metadata to true?

BryanValverdeU: The other place we set this metadata is when indenting a List. In setModelIndentation.ts

juliaroldi: Ok, I tested, and I did not find any regressions.

juliaroldi (approved):

BryanValverdeU:
This will also fix #2735

PR #2744, opened 2024-07-11
Fix crop after selection

juliaroldi:
When a selection is made directly on the image, it results in a segment of the image being selected. If the selection occurs while the image is within a wrapper, the resulting selection will be the span that wraps the image, leading to a selection sequence of <---SelectionMarker---><---Image---><---SelectionMarker--->. This scenario was not considered before, then a change was added to getSelectedImage function to recognize image in the wrapper. 
Bug repro steps: 
1. Select the image
2. Right click 
3. Select crop in context menu
4. Crop layer does not appear. 
BUG-ON_RIGHT-CLICK

Expected:
FIX-BUG-ON_RIGHT-CLICK


BryanValverdeU (approved):
Tested and it looks good to me, 
Quick question, any way we can prevent that <Marker><image><Marker> selection when using the wrapper?

Something like 
```
if (wrapperIsPresent) { return image } else { normalPath }
```

juliaroldi (commented):
I changed the approach to fix the selection after the wrapper is removed.  @BryanValverdeU could you please review again?

BryanValverdeU (approved):

Issue #2751, opened 2024-07-22
Original Classes of Entity Wrapper are Threw Away


Echo-az:
**Describe the bug**
The classes of entity wrapper are set to `generateEntityClassNames`. As a result, the original classes are threw away.

**To Reproduce**
1. Create an entity with customized classes
2. Trigger model to dom

**Expected behavior**
Original classed are kept

**Device Information**
 - OS: macOS
 - Browser: WebKit/Safari
 - Version: latest

**Additional context**
I know we can workaround this problem by adding a child container div with classes we want. But I think throwing away original classes may not be reasonable.

Echo-az:
Tried #2752 to fix this problem

JiuqingSong:
close old ones

ghost:
Can you close them all for me please

PR #2752, opened 2024-07-22
Entity: Add class list rather than replace directly

Echo-az:
Use `classList.add` rather than setting `className` for entity wrapper to avoid original class loss.

### DIFF
--- a/packages/roosterjs-content-model-dom/lib/domUtils/entityUtils.ts
+++ b/packages/roosterjs-content-model-dom/lib/domUtils/entityUtils.ts
@@ -107,12 +107,11 @@ function parseEntityClassName(
  * @param format The source entity format object
  * @returns A combined CSS class name string for entity wrapper
  */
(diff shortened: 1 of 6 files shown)


Echo-az:
Related issue: #2751

packages/roosterjs-content-model-dom/lib/domUtils/entityUtils.ts
```diff
@@ -107,12 +107,11 @@ function parseEntityClassName(
  * @param format The source entity format object
  * @returns A combined CSS class name string for entity wrapper
  */
-export function generateEntityClassNames(format: ContentModelEntityFormat): string {
+export function generateEntityClassList(format: ContentModelEntityFormat): string[] {
```
BryanValverdeU, line 110: Changing the function name may be a breaking change as it is already exported in the index file, suggest create a new function.

JiuqingSong:
Close this one since no update for half a year.

PR #2759, opened 2024-08-06
Fix ordinal trigger

juliaroldi:
Trigger ordinal superscript only after numbers. 
Bug: 
TransformOrdinalBug

Fix: 
TransformOrdinalFix

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/autoFormat/numbers/transformOrdinals.ts
+++ b/packages/roosterjs-content-model-plugins/lib/autoFormat/numbers/transformOrdinals.ts
@@ -16,8 +16,7 @@ const getOrdinal = (value: number) => {
 
 /**
  * @internal
- */
-export function transformOrdinals(
+ */ export function transformOrdinals(
     previousSegment: ContentModelText,
     paragraph: ShallowMutableContentModelParagraph,
     context: FormatContentModelContext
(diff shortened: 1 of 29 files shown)


packages/roosterjs-content-model-plugins/lib/autoFormat/numbers/transformOrdinals.ts
```diff
@@ -41,3 +40,12 @@ export function transformOrdinals(
     }
     return false;
 }
+
+function getValue(text: string) {
+    const number = text.substring(0, text.length - 2);
```
JiuqingSong, line 45: Suggest do not use hard coded number "2" here, but use a const and explain it with a comment.

Same for line 26

packages/roosterjs-content-model-plugins/lib/autoFormat/numbers/transformOrdinals.ts
```diff
@@ -16,16 +16,15 @@ const getOrdinal = (value: number) => {
...
+ */ export function transformOrdinals(
     previousSegment: ContentModelText,
     paragraph: ShallowMutableContentModelParagraph,
     context: FormatContentModelContext
 ): boolean {
     const value = previousSegment.text.split(' ').pop()?.trim();
     if (value) {
         const ordinal = value.substring(value.length - 2);
-        const ordinalValue = parseInt(value);
+        const ordinalValue = getValue(value);
```
JiuqingSong, line 27: You can pass in `ordinal` instead, so no need to substring again

juliaroldi: One substring is to get the st, nd, rd, and th and the other is for the numeric part.

JiuqingSong (approved):

PR #2760, opened 2024-08-06
Apply first segment format to new cell

Andres-CT98:
### Description:

New rows/columns do not copy the segment format.

### Fix

When inserting a new row/column, the segmentFormat of the first segment will be copied from the anchor row/column.

### Cases tested:

**Before:** 
noFormatNewRow

**After:** 
formatNewRow

### Warnings and implications

- If a segment has no format, the new row/column will not copy this empty format, instead it will add the model default one


packages/roosterjs-content-model-api/lib/modelApi/table/insertTableColumn.ts
```diff
@@ -23,11 +23,18 @@ export function insertTableColumn(
             table.rows.forEach(row => {
                 const cell = row.cells[insertLeft ? sel.firstColumn : sel.lastColumn];
 
+                const cellFormat =
+                    cell.blocks.length > 0 &&
+                    cell.blocks[0].blockType == 'Paragraph' &&
+                    cell.blocks[0].segments.length > 0
+                        ? { ...cell.format, ...cell.blocks[0].segments[0].format }
```
JiuqingSong, line 30: There is code to help only copy segment format, https://github.com/microsoft/roosterjs/blob/84ac4284cde532cb3952a3af02e7a2afcc2a5ced/packages/roosterjs-content-model-core/lib/corePlugin/cache/domIndexerImpl.ts#L470

You can extract it to be a public function so call it from both places. This is to avoid copy additional format, for example, image.

packages/roosterjs-content-model-api/lib/modelApi/table/insertTableRow.ts
```diff
@@ -26,11 +26,17 @@ export function insertTableRow(
             table.rows.splice(insertAbove ? sel.firstRow : sel.lastRow + 1, 0, {
                 format: { ...sourceRow.format },
                 cells: sourceRow.cells.map(cell => {
+                    const cellFormat =
```
JiuqingSong, line 29: This code is same with the previous one, try extract them into a public function

packages/roosterjs-content-model-dom/lib/modelApi/editing/normalizeTable.ts
```diff
@@ -52,12 +54,17 @@ export function normalizeTable(
...
             if (cell.blocks.length == 0) {
-                const format = cell.format.textColor
-                    ? {
-                          ...defaultSegmentFormat,
-                          textColor: cell.format.textColor,
-                      }
-                    : defaultSegmentFormat;
+                const cellFormat = { ...cell.format };
+                // Remove all non-segment related properties from cell format
+                getObjectKeys(cellFormat).forEach(key => {
```
JiuqingSong, line 59: This will delete format that exists in ContentModelTableCellFormat but not in ContentModelSegmentFormat.

Actually ContentModelTableCellFormat  is not a sub class of ContentModelSegmentFormat, so better not delete like this

packages/roosterjs-content-model-types/lib/contentModel/format/ContentModelTableCellFormat.ts
```diff
@@ -12,5 +12,5 @@ export type ContentModelTableCellFormat = ContentModelBlockFormat &
     BorderBoxFormat &
     VerticalAlignFormat &
     WordBreakFormat &
-    TextColorFormat &
+    ContentModelSegmentFormat &
```
JiuqingSong, line 15: I prefer not do this since it will allow moving all segment format into cell. We should just keep them in segment level.

When insert new row/column, we can also create new empty paragraph with the segment format.

Andres-CT98: I am just extending the pattern set by TextColorFormat. Previously there was no issue to add segment formatting to the cell.

packages/roosterjs-content-model-dom/lib/modelApi/editing/normalizeTable.ts
Andres-CT98: That is the purpose of the function. Why would ContentModelTableCellFormat not being a sub class of ContentModelSegmentFormat affect this? This change adds the properties of ContentModelSegmentFormat to ContentModelTableCellFormat.

JiuqingSong: We expect segment format to be applied to segment only. For example, if we want to apply font-family, do it on text but not on table cell:

Do:
```html
<td><span style="font-family: xxx">text</span></td>
```

Do not do:
```html
<td style="font-family: xxx">text</td>
```

So that each level has its only responsibility. And when we process table, no need to handle those segment formats. It is not table's responsibility.

That is why ContentModelTableCellFormat does not contain ContentModelSegmentFormat.

It only contains text color, given that we need to apply a default text color when set table cell's shade color. It is just a default color and it can always be overwritten by color from segment.

PR #2764, opened 2024-08-08
Pressing Tab inside a table should select all node contents of the next cell

Andres-CT98:
### Description:
When Pressing tab to get to the next cell in a table, #2718, in the case of an empty cell the selection should not delete the segment format.

### Fix

The selection should not be done using the method [selectNodeContents](https://developer.mozilla.org/en-US/docs/Web/API/Range/selectNodeContents), instead a custom range using the node's first and last child should be done.

### Cases tested:

1. Create a table bigger than 2x2
2. Populate the first cell with text
3. Go to the first cell and press Tab, this should select the next cell
4. Input some text
5. Inspect the elements of both the first and second cell.

**Before:** The texts on each cell have different fonts, the first one being the editor default and the second the browser default.

**After:** The texts of both cells have the same font, being the editor default.


packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -536,12 +536,12 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
         selectAll?: boolean
     ) {
         const range = editor.getDocument().createRange();
+        // Get deepest editable position in the cell
+        const { node, offset } = normalizePos(cell, nodeOffset);
+
         if (selectAll) {
-            range.selectNodeContents(cell);
+            range.selectNodeContents(node);
```
JiuqingSong, line 543: So if there are multiple nodes in next cell, it will only select the first one. Is that expected?

image

JiuqingSong: image

In this example, there are Bold style in middle of the text, which causes 3 text nodes existing. This change will cause only the first part is selected.

Andres-CT98: Good point. The specific scenario this is trying to fix is one where the cell is empty and only has a div with br inside. I could just look for that scenario specifically, but I am not sure if it is reliable to hardcode it that way.

Could this be an issue with text input and not selection?

JiuqingSong: If you expect different behavior between empty cell and cell with content, then yes you need to have different code for them.

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -559,8 +559,16 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
         selectAll?: boolean
     ) {
         const range = editor.getDocument().createRange();
-        if (selectAll) {
-            range.selectNodeContents(cell);
+        if (selectAll && cell.firstChild) {
+            const cellStart = cell.firstChild;
+            const cellEnd = cell.lastChild || cell.firstChild;
```
JiuqingSong, line 564: actually, if lastChild is null, firstChild must also be null. So just use lastChild should be fine

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -559,8 +559,16 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
...
+            // Get first deepest editable position in the cell
+            const posStart = normalizePos(cellStart, 0);
```
JiuqingSong, line 566: need null check, in case cellStart/cellEnd is null.

If that happens, maybe we need to create a paragraph? Or just simple put focus into the cell.

Andres-CT98: They can not be null.

JiuqingSong: We cannot make such assumption. At least we need a null check here or it can throw at runtime.

Andres-CT98: The null check is above, after passing it cellStart/cellEnd can not longer be null.

JiuqingSong: Oh I see, sorry I missed it.

JiuqingSong (approved):

PR #2768, opened 2024-08-16
Fix Image range selection

juliaroldi:
Fix 1: when select image for copy/cut, the image range selection must appear. 
Fix 2: on firefox delete the image using delete/backspace to remove the image.

SelectImageWithCopyCut

Warning: this fix for image range selection does not work on Safari, selecting image range on safari is breaking the cut behavior.

### DIFF
--- a/packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
+++ b/packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
@@ -362,7 +362,11 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
                     }
                 }
 
-                if ((isModifierKey(rawEvent) || rawEvent.shiftKey) && selection.image) {
+                if (
+                    (isModifierKey(rawEvent) || rawEvent.shiftKey) &&
+                    selection.image &&
+                    !this.isSafari
+                ) {
                     const range = selection.image.ownerDocument.createRange();
                     range.selectNode(selection.image);
                     this.setDOMSelection(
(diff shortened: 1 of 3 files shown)


packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -768,21 +783,6 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
             this.state.mouseDisposer = undefined;
         }
     }
-
-    private trySelectSingleImage(selection: RangeSelection) {
```
BryanValverdeU, line 772: When selecting the image with Ctrl + Arrow Keys now we do not show image selection and instead we show range selection. Is this expected? 

I believe it is different than Word, in Word as soon as the image is selected with either Shift + Arrow, Word adds the wrapper for resizing the image.

JiuqingSong:
About the Safari warning, does it regress anything, or just not a complete fix?

juliaroldi:
> About the Safari warning, does it regress anything, or just not a complete fix?

The fix will not work for Safari, if the image is selected with range when cut/copy, the cut will not work, so for Safari, we do not change anything, just leave the behavior we have now.

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -362,6 +361,23 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
                         this.selectBeforeOrAfterElement(editor, selection.image);
                     }
                 }
+
+                if (
+                    (isModifierKey(rawEvent) || rawEvent.shiftKey) &&
+                    selection.image &&
+                    !this.isSafari
```
juliaroldi, line 368: @JiuqingSong Safari will ignore range selection

JiuqingSong (approved):

PR #2770, opened 2024-08-16
Add `<del>` elements to default processors and use knownElementProcessor for this type of element.

BryanValverdeU:
When pasting content with Elements of [del ](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/del)tag. Since the element is unknown we process it with generalProcessor. But this cause that the result of the paste is different. This is because when we call the [ensureParagraph.ts](https://github.com/microsoft/roosterjs/blob/024877f79ec1c3955be6796afe784fcb5149a396/packages/roosterjs-content-model-dom/lib/modelApi/common/ensureParagraph.ts#L19) function and the current group is general and not paragraph, ensure Paragraph will create a new paragraph and it will be inserted inside of the del element. Causing the issue seen in the image below for the Before,

 To fix add this type of tag that adds strikethrough style to the known element processor.

Source:
image

Before
image

After
image

### DIFF
--- a/packages/roosterjs-content-model-dom/lib/domToModel/context/defaultProcessors.ts
+++ b/packages/roosterjs-content-model-dom/lib/domToModel/context/defaultProcessors.ts
@@ -55,6 +55,7 @@ export const defaultProcessorMap: ElementProcessorMap = {
     table: tableProcessor,
     u: knownElementProcessor,
     ul: listProcessor,
+    del: knownElementProcessor,
 
     '*': generalProcessor,
     '#text': textProcessor,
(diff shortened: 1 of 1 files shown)


juliaroldi (approved):

packages/roosterjs-content-model-dom/lib/domToModel/context/defaultProcessors.ts
```diff
@@ -55,6 +55,7 @@ export const defaultProcessorMap: ElementProcessorMap = {
     table: tableProcessor,
     u: knownElementProcessor,
     ul: listProcessor,
+    del: knownElementProcessor,
```
JiuqingSong, line 58: please do alphabet order

BryanValverdeU: Sounds good. Addressed this in latest commit.

JiuqingSong (approved):

PR #2773, opened 2024-08-22
Set segmentFormat text color to black when creating the model of the clipboard content and using Keep source formatting paste type

BryanValverdeU:
Right now, if we paste some content in the editor and we use the Keep source formatting or normal paste type. 
If the default format is different that black, that color will be set to the elements that do not have any text color in the HTML.

Which causes inconsistencies with the content.
To fix, if the paste type is of type normal or keep source formatting. Set the text color in the default format that is going to be used to generate the model from the clipboard content to black. That way if the element does not have a text color set we just default to black.

If the paste type is different than Keep source format, there is no need to do this, as we will merge all the format anyway.

Source
image

Before
image

After
image

https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/288926/

### DIFF
--- a/packages/roosterjs-content-model-core/lib/command/paste/mergePasteContent.ts
+++ b/packages/roosterjs-content-model-core/lib/command/paste/mergePasteContent.ts
@@ -13,10 +13,15 @@ import type {
     ClipboardData,
     CloneModelOptions,
     ContentModelDocument,
+    ContentModelSegmentFormat,
     IEditor,
     MergeModelOption,
+    PasteType,
     ReadonlyContentModelDocument,
+    ShallowMutableContentModelDocument,
 } from 'roosterjs-content-model-types';
+
+const BlackColor = 'rgb(0,0,0)';
 
 const CloneOption: CloneModelOptions = {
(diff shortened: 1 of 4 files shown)


packages/roosterjs-content-model-core/lib/command/paste/mergePasteContent.ts
```diff
@@ -87,6 +89,27 @@ export function mergePasteContent(
...
+): ContentModelSegmentFormat {
+    const selectedSegment = getSelectedSegments(model, true /*includeFormatHolder*/)[0];
+
+    if (selectedSegment) {
+        const result = getSegmentTextFormat(selectedSegment);
+        if (pasteType == 'normal') {
+            // When using normal paste (Keep source formatting) set the default text color to black when creating the
+            // Model from the clipboard content, so the elements that do not contain any text color in their style
+            // Are set to black. Otherwise, These segments would get the selected segments format or the default text set in the content.
+            result.textColor = BlackColor;
```
JiuqingSong, line 104: what if result already has textColor, will it be overwritten here?

JiuqingSong (approved):

JiuqingSong (approved):

PR #2774, opened 2024-08-22
Insert Image wrapper when insert image

juliaroldi:
To ensure the image editing tools appear correctly, please initiate the image wrapper as soon as the image is loaded into the editor. This process will enable the rotate and resize handles to display immediately after the 'insertImage' API triggers and the image is successfully loaded.

Before: 
InsertImageNotWrapped

After: 
InsertImageWrapped

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
@@ -17,10 +17,8 @@ import { updateWrapper } from './utils/updateWrapper';
 import {
     getSafeIdSelector,
     isElementOfType,
-    isModifierKey,
     isNodeOfType,
     mutateSegment,
-    toArray,
     unwrap,
 } from 'roosterjs-content-model-dom';
 import type { DragAndDropHelper } from '../pluginUtils/DragAndDrop/DragAndDropHelper';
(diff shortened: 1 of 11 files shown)


packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -140,6 +141,11 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
             case 'keyDown':
                 this.keyDownHandler(this.editor, event);
                 break;
+            case 'contentChanged':
+                if (event.source === ChangeSource.Format && event.formatApiName == 'insertImage') {
```
BryanValverdeU, line 145: Can we also add the option to allow different format sources? I think InsertImage in OWA uses 'InsertImage' ChangeSource

juliaroldi: @JiuqingSong opinions on this?

BryanValverdeU: I think there is no need to add it to the ChangeSource file nor change current InsertImage API, I think adding an additional parameter to the class for additional ChangeSources to select image could enough to support the OWA Insert Image scenario.

But I will let Jason chime in.

juliaroldi: I had another idea, instead of additional parameter for ChangeSource we can pass a parameter for formatApiName, because to insert the image, we usually use the formatContentModel which triggers Format change source, then we use the formatApiName to identify if an image is inserted

JiuqingSong:
In general, I don't think this should be done by the plugin, but by the code to insert image.

It should be:
when insert image, directly select it. After image is inserted, selection will be on the image, so an SelectionChanged event will be triggered, then ImageEditPlugin start to edit it.

insertImage code should have the right to decide whether the image should be selected. So I think for this issue it should be fixed in OWA side instead.

PR #2784, opened 2024-09-03
Resolve null sheet in convertInlineCss

BryanValverdeU:
There is an error when initializing the editor, and seems to be related to a type cast in convertInlineCss, to fix add a null check so we do not have a runtime error.

Callstack
```
TypeError: Cannot read properties of null (reading 'cssRules')

roosterjs-content-model-core/lib/command/createModelFromHtml/convertInlineCss.ts cssRules 33
Native Browser Function(Array.forEach)
roosterjs-content-model-core/lib/command/createModelFromHtml/convertInlineCss.ts forEach 30
roosterjs-content-model-core/lib/command/createModelFromHtml/createModelFromHtml.ts retrieveCssRules 35
roosterjs-editor-adapter/lib/editor/EditorAdapter.ts initContent 137
```

### DIFF
--- a/packages/roosterjs-content-model-core/lib/command/createModelFromHtml/convertInlineCss.ts
+++ b/packages/roosterjs-content-model-core/lib/command/createModelFromHtml/convertInlineCss.ts
@@ -28,7 +28,11 @@ export function retrieveCssRules(doc: Document): CssRule[] {
     const result: CssRule[] = [];
 
     styles.forEach(styleNode => {
-        const sheet = styleNode.sheet as CSSStyleSheet;
+        const sheet = styleNode.sheet;
+
+        if (!sheet) {
+            return;
+        }
 
         for (let ruleIndex = 0; ruleIndex < sheet.cssRules.length; ruleIndex++) {
             const rule = sheet.cssRules[ruleIndex] as CSSStyleRule;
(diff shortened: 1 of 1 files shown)


packages/roosterjs-content-model-core/lib/command/createModelFromHtml/convertInlineCss.ts
```diff
@@ -28,16 +28,18 @@ export function retrieveCssRules(doc: Document): CssRule[] {
     const result: CssRule[] = [];
 
     styles.forEach(styleNode => {
-        const sheet = styleNode.sheet as CSSStyleSheet;
+        const sheet = styleNode.sheet;
```
JiuqingSong, line 31: do we have a local repro? I wonder in what case this is null

BryanValverdeU: I do not have a local repro, but checking the telemetry, the error rate is really low. 

According to mozilla, this value can be null if there is no sheet associated with the element: 
A [StyleSheet](https://developer.mozilla.org/en-US/docs/Web/API/StyleSheet) object, or null if none is associated with the element.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLStyleElement/sheet#value

BryanValverdeU: Looks like this is a way to repro:
image

JiuqingSong (approved):

PR #2785, opened 2024-09-04
Bump Rooster to patch (#2784)

BryanValverdeU:
Patch #2784


JiuqingSong (approved):

juliaroldi:
@BryanValverdeU @JiuqingSong  can this changed be included in this patch https://github.com/microsoft/roosterjs/pull/2781?

juliaroldi:
> > @BryanValverdeU @JiuqingSong can this changed be included in this patch #2781?
> 
> @juliaroldi Do you feel it is safe? I am planning to backfix this to PROD as it is causing the app to crash.

This change does not need to go to PROD right now. I feel it is better to test in SDF before.

PR #2796, opened 2024-09-12
Bump RoosterJs Plugins to 9.10.1

juliaroldi:
Add patch fix for the bug https://github.com/microsoft/roosterjs/pull/2794

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -1,6 +1,8 @@
 {
     "react": "9.0.0",
-    "main": "9.10.1",
+    "main": "9.10.0",
     "legacyAdapter": "8.62.1",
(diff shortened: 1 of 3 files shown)


versions.json
```diff
@@ -1,6 +1,6 @@
 {
     "react": "9.0.0",
-    "main": "9.10.0",
+    "main": "9.10.1",
```
BryanValverdeU, line 3: Hmm, I think we only need to bump the plugins package. And not all the other packages.

juliaroldi: Good point!

BryanValverdeU (approved):

PR #2806, opened 2024-09-24
Pressing Tab inside a table should only select the next cell's content, not anything beyond that

Andres-CT98:
### Description:
When Pressing tab to get to the next cell in a table, #2718 & #2764, the selection should be limited to only the contents of the adjacent cell, or in the case of last cell, the next line.

### Fix

When obtaining the cells' first and last child, their positions should be normalised using noramlizePos. After that, the offset values for the range should always be 0 for the starting node and the length of the final node for the end node. This asures that the new range is a valid one that always exists inside the cell.

### Cases tested:

1. Create a table bigger than 2x2
2. Add text after the table
3. Go to the first cell and press Tab, this should select the next cell
4. Type a letter

**Before:** The text below the table is removed

**After:** The text below the table is intact


JiuqingSong:
We can close this one now.

PR #2820, opened 2024-10-03
Remove style "white-space" from empty paragraph

JiuqingSong:
When normalize a paragraph (e.g. triggered from ENTER key), if it is empty (has BR or selection marker only), we should not keep its "white-space" style if any. So that the new empty paragraph can be just a normal paragraph.


juliaroldi (approved):

juliaroldi:
> When normalize a paragraph (e.g. triggered from ENTER key), if it is empty (has BR or selection marker only), we should not keep its "white-space" style if any. So that the new empty paragraph can be just a normal paragraph.

Is this changing preventing a bug?

JiuqingSong:
> > When normalize a paragraph (e.g. triggered from ENTER key), if it is empty (has BR or selection marker only), we should not keep its "white-space" style if any. So that the new empty paragraph can be just a normal paragraph.
> 
> Is this changing preventing a bug?

Yes. User pasted something with such style, then press ENTER, this style is copied to the new paragraph. If the new paragraph is empty, the layout will be strictly related to the HTML content. A hidden line break will cause extra empty when layout.

PR #2822, opened 2024-10-08
Make image edit features work with touch

juliaroldi:
Enable resize/rotate/crop images using touch events. 
TouchWithImages


packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -420,6 +420,7 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
         if (this.imageEditInfo) {
             this.startEditing(editor, image, ['resize', 'rotate']);
             if (this.selectedImage && this.imageEditInfo && this.wrapper && this.clonedImage) {
+                const isMobileOrTable = !!editor.getEnvironment().isMobileOrTablet;
```
BryanValverdeU, line 423: Should we also support it for normal PC? Since there are some with touch screens

BryanValverdeU (approved):

PR #2826, opened 2024-10-11
Bump Core, Plugins and Dom packages to 9.11.2

BryanValverdeU:


juliaroldi (approved):

JiuqingSong:
Is this a regular version bump, or a patch? If it is regular bump, just change the main version value in versions.json, no need to use override.

JiuqingSong (approved):

PR #2834, opened 2024-10-18
Revert Shortcut command for Clear Format on mac

miku1958:
This shortcut is the same as the default global spotlight shortcut, so it is invalid if the user keeps spotlight‘s. Some users prefer to keep it the same as Windows by disable spotlight

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/shortcut/ShortcutPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/shortcut/ShortcutPlugin.ts
@@ -121,7 +121,7 @@ export class ShortcutPlugin implements EditorPlugin {
             const editor = this.editor;
             const { ctrlKey, metaKey } = event.rawEvent;
             if (ctrlKey && metaKey) {
-                // We don't support both Ctrl and Meta key pressed at the same time
+                // We don't support both Ctrl and Meta key pressed at the same time.
                 return null;
             }
             return (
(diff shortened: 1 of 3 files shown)


JiuqingSong:
@juliaroldi would you please comment here?
Thanks.

juliaroldi (commented):
We need also check if the Control key is not pressed, otherwise the shortcut meta + control + space will not work to show the emoji panel

miku1958:
> We need also check if the Control key is not pressed, otherwise the shortcut meta + control + space will not work to show the emoji panel

@juliaroldi Sorry for late response, busy last week.

I don't think keystroke conflicts are a concern for this PR, and it's not just a problem with this shortcut, ShortcutPlugin should take care of additional keystroke triggers. 

On the other hang, emoji panel is using the browser shortcut, which has highest priority in browser(like the spotlight in system).

juliaroldi:
> > We need also check if the Control key is not pressed, otherwise the shortcut meta + control + space will not work to show the emoji panel
> 
> @juliaroldi Sorry for late response, busy last week.
> 
> I don't think keystroke conflicts are a concern for this PR, and it's not just a problem with this shortcut, ShortcutPlugin should take care of additional keystroke triggers.
> 
> On the other hand, emoji panel is using the browser shortcut, which has highest priority in browser(like the spotlight in system).

Hi @miku1958, with the current implementation of the ShortcutPlugin, the meta + control + space will not trigger the emoji panel. Could please test this scenario on your branch? I did the test in Chrome Browser.

miku1958:
@juliaroldi I do some tests and make some fix, please review it again thanks.

juliaroldi (approved):

JiuqingSong:
@miku1958 For some unknown reason the required check "license/cla" is not running. This happened before and it seems like a github issue. I don't have a way to manually trigger it. So to workaround, you can try to push some change again (say modify a comment, or any minor change), to trigger it.

Thanks.

miku1958:
@JiuqingSong it seems work now

JiuqingSong (approved):

PR #2845, opened 2024-10-24
Fix setSelectionToBlockGroup by wrapping setSelectionToBlock in handleSelection

wisaulni:
The existing implementation of handleSelection assumed that you'd only be setting selections from start/end on individual elements within blocks, and not on entire blocks themselves. Therefore, attempting to select from block to block (which succeeds in the type system) fails with no selection because no comparison is done between blocks and the start / end values.

This PR fixes this, and adds a test to guard against regression.


wisaulni:
@microsoft-github-policy-service agree company="Microsoft"

JiuqingSong:
Because Paragraph is not a selectable object, so I'd like to say the original behavior is expected

wisaulni:
Is there a reason why it shouldn't work when passed in a block? I'd think there's at least some case to be made that you might want to select an entire block at once.

JiuqingSong:
Actually, this can be interpreted into two questions.
1. Whether `Paragraph` should implement `Selectable` interface
2. Whether we should allow select the whole paragraph

For 1, if we allow `Paragraph` to be `Selectable`, consider the following case:
```ts
{
  blockType: 'Paragraph',
  isSelected: true,
  segments: [{
    segmentType: 'Text',
    text: 'text 1',
    isSelected: true,
  }, {
    segmentType: 'Text',
    text: 'text 2',
    isSelected: fals,;
  }],
}
```

In this example, paragraph is marked as selected, and it has two text segments, one of them is selected, the other one is marked as not selected. Then there is a conflict. Since the whole paragraph is selected, then all its segments should be selected. But we have explicitly mark text 2 as not selected, should we treat it as selected or not selected?

The root cause is we allow both paragraph and segments as selectable. So the solution is do not let Paragraph implements Selectable, so we can never mark a whole paragraph as selected.

For 2, Yes, we should allow select the whole paragraph, by selecting all its segments. However, the API `setSelection` is a lower level API, it only accepts Selectable object. So in order to select a paragraph, you can create another utility function to write setSelection like:
```ts
function selectParagraph(model: ContentModelDocument, paragraph: ContentModelParagraph) {
  if (paragraph.segments.length > 0) {
     setSelection(model, paragraph.segments[0], paragraph.segments[paragraph.segments.length - 1]);
  }
}
```

With this way, all above problems are solved.

wisaulni:
Makes sense. I think then the typing on the function needs to be tightened up. I think it might be doable by marking blocks themselves with a type that has isSelected: never, which will exclude them from the selectable type.

JiuqingSong:
> Makes sense. I think then the typing on the function needs to be tightened up. I think it might be doable by marking blocks themselves with a type that has isSelected: never, which will exclude them from the selectable type.

I tried to add `isSelected: never` to `ContentModelParagraph` but that causes build time error since we didn't assign a value to `isSelected` when create paragraph. I also tried to add `isSelected?: never`, that can pass the build, but even I pass paragraph to `setSelection`, build still not fail.

In theory the best way should be we make `isSelected` required in `Selectable` interface, but that will break so many existing code. So for now I'll leave it as it is.

PR #2848, opened 2024-10-25
Fix test for AutoFormat and ImageEdit 

juliaroldi:
Fix unit tests for AutoFormat and ImageEdit.

### DIFF
--- a/packages/roosterjs-content-model-plugins/test/autoFormat/AutoFormatPluginTest.ts
+++ b/packages/roosterjs-content-model-plugins/test/autoFormat/AutoFormatPluginTest.ts
@@ -1044,12 +1044,17 @@ describe('Content Model Auto Format Plugin Test', () => {
 
             const segment: ContentModelText = {
                 segmentType: 'Text',
-                text: 'test--test',
+                text: 'test--test ',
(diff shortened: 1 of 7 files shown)


packages/roosterjs-content-model-plugins/test/autoFormat/AutoFormatPluginTest.ts
```diff
@@ -1092,7 +1092,13 @@ describe('Content Model Auto Format Plugin Test', () => {
                             segments: [
                                 {
                                     segmentType: 'Text',
-                                    text: 'test—test',
+                                    text: 'test—tes',
```
JiuqingSong, line 1095: I wonder why the result is like this. I think the result should just be a single text segment

JiuqingSong (approved):

packages/roosterjs-content-model-plugins/test/autoFormat/AutoFormatPluginTest.ts
juliaroldi: This test shouldn't pass, but I think it passed, because the selection marker was not in the model.

PR #2849, opened 2024-10-26
Remove tablePreProcessor

JiuqingSong:
tabePreProcessor was added from earlier days when our content fidelity was not good enough, we were using this processor to treat table as entity if no selection is in it. With content fidelity got keep improved, and other issue found due to this behavior (https://outlookweb.visualstudio.com/Outlook%20Web/_queries/edit/313176/?triage=true), I think it is time to remove this processor, but always use regular table processor for tables.


juliaroldi (commented):
Can old bugs solved by this preprocessor surface again?

JiuqingSong:
> Can old bugs solved by this preprocessor surface again?

It is possible to see some content fidelity issue come back. Actually, those issues are always there, just the chance was reduced since user may not put focus into table. And when we see them happen, we should fix them.

juliaroldi (approved):

PR #2851, opened 2024-10-29
Query content model blocks. 

juliaroldi:
This changed introduced a new API, `queryContentModel`, to the `roosterjs-content-model-api`. Its role is to retrieve either blocks of a similar type, or a particular block specified by a selector. This enhancement builds upon the `findEditingImage` function, enabling its logic to be applied in additional features like the OWA Accessibility Checker. Moreover, it addresses the issue of how images inserted into tables from Word Online are handled, as `queryContentModel` also checks `formatContainer` blocks during image searches.

Test case: 
1. Copy a table from Word Online 
2. Paste in the editor 
3. Insert an image in a table cell
4. Select the image and click away
5. the image must be deselected 

imageOnWordTables


packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModel.ts
```diff
@@ -0,0 +1,116 @@
...
+    ReadonlyContentModelBlockGroup,
+    ReadonlyContentModelParagraph,
+    ReadonlyContentModelSegment,
+    ReadonlyContentModelTable,
+} from 'roosterjs-content-model-types';
+
+/**
+ * Options for queryContentModel
+ */
+export interface QueryContentModelOptions<T> {
```
JiuqingSong, line 14: add type constraint here for T.

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModel.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * The type of segment to query
+     */
+    segmentType?: ContentModelSegmentType;
+
+    /**
+     * Optional selector to filter the blocks/segments
+     */
+    selector?: (element: T) => boolean;
```
JiuqingSong, line 28: How about name it "filter"?

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModel.ts
```diff
@@ -0,0 +1,116 @@
...
+        switch (block.blockType) {
+            case 'BlockGroup':
+                if (type == block.blockType && (!selector || selector(block as T))) {
+                    elements.push(block as T);
+                }
+                const blockGroupsResults = queryContentModel<T>(block, options);
+                elements.push(...(blockGroupsResults as T[]));
+                break;
+            case 'Table':
+                if (type == block.blockType && (!selector || selector(block as T))) {
```
JiuqingSong, line 62: what if T is a segment type, are we converting block to segment?

JiuqingSong: Actually we should make sure only the correct type of object can trigger selector()

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModel.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+function searchInTables<T extends ReadonlyContentModelBlock | ReadonlyContentModelSegment>(
+    table: ReadonlyContentModelTable,
+    options: QueryContentModelOptions<T>
+): T[] {
+    const blocks: T[] = [];
+    for (const row of table.rows) {
+        for (const cell of row.cells) {
+            const items = queryContentModel<T>(cell, options);
```
JiuqingSong, line 97: Better create a `internalQueryContentModel` which does not need to check value of options then let both here and `queryContentModel` call it

JiuqingSong (commented):
In general, it is a perfect idea to have a common query function.

Once done, this function will be used widely, so we need to be careful on how it is designed and worked. We need to make sure if it is misused, it can fail the build.

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModel.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/**
+ * Query content model blocks or segments
+ * @param group The block group to query
+ * @param options The query option
+ */
+export function queryContentModel<
+    T extends ReadonlyContentModelBlock | ReadonlyContentModelSegment
+>(group: ReadonlyContentModelBlockGroup, options: QueryContentModelOptions<T>): T[] {
```
haven2world, line 43: I think it will be helpful if we can also return the path to the found element, since in CM we don't have a property to access the parent element. That's how we implement it on OM's repo.

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModel.ts
haven2world, line 28: `(element: ContentModelType) => element is Target` this is the type of the selector we used in our implementation, I'm not sure if it's a good one though. Just an idea.

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModel.ts
juliaroldi: That seems a good idea, do you have an example of the implementation?

haven2world: This is basically what we're using now
```typescript
export function findElement<Target extends ReadonlyContentModelType>(
  element: ReadonlyContentModelType,
  path: (ReadonlyContentModelBlockGroup | ReadonlyContentModelParagraph)[],
  match: (element: ReadonlyContentModelType) => element is Target,
  matchAll: boolean,
): ModelElementAndPath<Target>[] {
  const result: ModelElementAndPath<Target>[] = [];
  internalFindElement(element, path, match, matchAll, result);
  return result;
}

function internalFindElement<Target extends ReadonlyContentModelType>(
  element: ReadonlyContentModelType,
  path: (ReadonlyContentModelBlockGroup | ReadonlyContentModelParagraph)[],
  match: (element: ReadonlyContentModelType) => element is Target,
  matchAll: boolean,
  result: ModelElementAndPath<Target>[],
): void {
  if (!matchAll && result.length > 0) {
    return;
  }

  if (match(element)) {
    result.push({ element, path });
    if (!matchAll) {
      return;
    }
  }

  if (isContentModelBlockGroup(element)) {
    element.blocks.forEach((child) => {
      internalFindElement(child, [element, ...path], match, matchAll, result);
    });
    return;
  }

  if (isContentModelBlock(element)) {
    switch (element.blockType) {
      case 'Paragraph':
        element.segments.forEach((segment) => {
          internalFindElement(segment, [element, ...path], match, matchAll, result);
        });
        break;
      case 'Table':
        element.rows.forEach((row) => {
          row.cells.forEach((cell) => {
            internalFindElement(cell, path, match, matchAll, result);
          });
        });
        break;
      default:
        break;
    }
  }
}
```

juliaroldi: I think that is a nice idea, since sometimes we need the parent block, but I don't want to add more complexity for this function, and I don't want to create a specific return type just this function. For now, I think we can use the filter callback to get the parent and the segment as I did in the findImageEditing function. Also, the user can always create a more customized function to query the need content model blocks/segments. @JiuqingSong any thoughts?

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModelBlocks.ts
```diff
@@ -0,0 +1,91 @@
...
+            return elements;
+        }
+        const block = group.blocks[i];
+        const results = queryContentModelBlocksInternal<T>(block, type, filter, findFirstOnly);
+        elements.push(...results);
+    }
+    return elements;
+}
+
+function queryContentModelBlocksInternal<T extends ReadonlyContentModelBlock>(
```
JiuqingSong, line 33: since you don't need the parameter `options` now, you can merge these two functions together, so no need to duplicate the for loop for block group

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModelBlocks.ts
```diff
@@ -0,0 +1,91 @@
...
+ * @param blockType The type of block to query @default 'Paragraph'
+ * @param filter Optional selector to filter the blocks
+ * @param findFirstOnly True to return the first block only, false to return all blocks
+ */
+export function queryContentModelBlocks<T extends ReadonlyContentModelBlock>(
+    group: ReadonlyContentModelBlockGroup,
+    blockType?: ContentModelBlockType,
+    filter?: (element: T) => element is T,
+    findFirstOnly?: boolean
+): T[] {
```
JiuqingSong, line 19: Do you also want to return the path?

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModelBlocks.ts
```diff
@@ -0,0 +1,91 @@
...
+    filter?: (element: T) => element is T
+): block is T {
+    return isBlockType<T>(block, type) && (!filter || filter(block));
+}
+
+function isBlockType<T extends ReadonlyContentModelBlock>(
+    block: ReadonlyContentModelBlock,
+    type: string
+): block is T {
+    return block.blockType == type;
```
JiuqingSong, line 90: No need to have this function. Just directly check blockType == type should work

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModelBlocks.ts
```diff
@@ -0,0 +1,91 @@
...
+    type: ContentModelBlockType,
+    filter?: (element: T) => element is T,
+    findFirstOnly?: boolean
+): T[] {
+    const elements: T[] = [];
+    if (isExpectedBlockType(block, type, filter)) {
+        elements.push(block);
+    }
+
+    if (block.blockType == 'BlockGroup') {
```
JiuqingSong, line 44: Use switch...case instead of multiple if

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModelBlocks.ts
```diff
@@ -0,0 +1,91 @@
...
+        for (const childBlock of block.blocks) {
```
JiuqingSong, line 45: What if you are querying block group? Then you should return directly here

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModelBlocks.ts
```diff
@@ -0,0 +1,91 @@
...
+            );
+            elements.push(...results);
+        }
+    }
+
+    if (block.blockType == 'Table') {
+        const table = block;
+        for (const row of table.rows) {
+            for (const cell of row.cells) {
+                for (const cellBlock of cell.blocks) {
```
JiuqingSong, line 63: cell is BlockGroup, so you can reuse the code for blockGroup here

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModelBlocks.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Query content model blocks
+ * @param group The block group to query
+ * @param blockType The type of block to query @default 'Paragraph'
+ * @param filter Optional selector to filter the blocks
+ * @param findFirstOnly True to return the first block only, false to return all blocks
+ */
+export function queryContentModelBlocks<T extends ReadonlyContentModelBlock>(
+    group: ReadonlyContentModelBlockGroup,
+    blockType?: ContentModelBlockType,
```
JiuqingSong, line 16: This blockType needs to be coupled with T. Otherwise, you can pass two different types, for example:
```ts
queryContentModelBlock<ContentModelParagraph>(group, 'Table');
```

In theory you should make this fail the build.

JiuqingSong: You can declare like this.
```ts
export function queryContentModelBlocks<T extends ReadonlyContentModelBlock>(
  group: ReadonlyContentModelBlockGroup,
  blockType: T extends ReadonlyContentModelBlockBase<infer U> ? U : never,
 ...) {
   ...
}
```

JiuqingSong: Then blockType and T must match, otherwise build will fail.

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModelBlocks.ts
juliaroldi: No, all the elements of type or filtered by the filter function

packages/roosterjs-content-model-api/lib/modelApi/common/queryContentModelBlocks.ts
```diff
@@ -0,0 +1,77 @@
...
+        if (findFirstOnly && elements.length > 0) {
+            return elements;
+        }
+        const block = group.blocks[i];
+
+        switch (block.blockType) {
+            case 'Paragraph':
+            case 'Divider':
+            case 'Entity':
+                if (isExpectedBlockType(block, type, filter)) {
```
juliaroldi, line 32: @JiuqingSong, I tried not verifying this in every, but not do it is causing build errors

JiuqingSong (approved):

PR #2854, opened 2024-10-31
Add change data and apiName to ContentChangedEvent when handle keyboard input

JiuqingSong:
In order to allow knowing which key is pressed when handling ContentChangedEvent with ChangeSource=Keyboard, we add apiName and change data into the event object. Then plugin can check event.data or event.formatApiName to know the key:

image


BryanValverdeU (approved):

packages/roosterjs-content-model-plugins/lib/edit/keyboardEnter.ts
```diff
@@ -58,6 +58,8 @@ export function keyboardEnter(
             rawEvent,
             scrollCaretIntoView: true,
             changeSource: ChangeSource.Keyboard,
+            getChangeData: () => rawEvent.which,
```
juliaroldi, line 61: Can we use the key property instead of which? I think which is deprecated.

PR #2865, opened 2024-11-08
Customize the behavior of merging format values

Rain-Zheng:
We are working on a feature that displays a prompt in the formatting bar when the selected content contains multiple formatting values. For example, if the selected content includes text elements with both "Aptos" and "Arial" fonts, a "Multiple fonts" prompt will be shown to the user. So we are adding a parameter to `getFormatState` to allow customization of the merge behavior.


JiuqingSong:
Instead of providing a callback, I prefer to add a parameter to indicate how to solve conflict format, and default value is to "remove" (current behavior)

For example:

```ts
type ConflictFormatSolution = 'remove' | 'keepFirst' | 'returnMultiple';

export function getFormatState(editor: IEditor, conflictSolution: ConflictFormatSolution = 'remove') {
 ...
}
```

packages/roosterjs-content-model-dom/lib/modelApi/editing/retrieveModelFormatState.ts
```diff
@@ -241,14 +252,20 @@ function mergeValue<K extends keyof ContentModelFormatState>(
     key: K,
     newValue: ContentModelFormatState[K] | undefined,
     isFirst: boolean,
-    parseFn: (val: ContentModelFormatState[K]) => ContentModelFormatState[K] = val => val
+    parseFn: (val: ContentModelFormatState[K]) => ContentModelFormatState[K] = val => val,
+    callbacks?: MergeFormatValueCallbacks
```
JiuqingSong, line 256: It seems the parameter is always passed, so you can make it a required parameter, and put it before the optional parameter `parseFn`, so no need to always add `undefined` when call this function

Rain-Zheng: Updated, thanks!

JiuqingSong (approved):

Issue #2864, opened 2024-11-07
Splitted text node causes Content Model cache inconsistent


JiuqingSong:
When splitted text node is added into editor by plugin, and `normalizeContentModel` is called, it is possible that two text segments got merged but DOM nodes are still separated, which causes DOM node and cached model not match. When keep typing, new typed text can be lost since we are upgrading an orphan text segment

PR #2871, opened 2024-11-14
Fix merge segment issue

JiuqingSong:
In my PR #2846, I add a process to merge text segments when `normalizeContentModel`. However this causes a problem, especially when `normalizeContentModel` is called from `formatContentModel` which returns false, so content model and DOM nodes are not matching.

In this change, I removed this process from `normalizeContentModel`, and only do it from `formatSegmentWithContentModel` since that is the place to cause original problem (text node got merged but content model is not). I also made this function exported so others can also call it when necessary.

Note that the new API `mergeTextSegments` should only be called on a mutable paragraph which means after it is called, a rewrite of this paragraph is required.

I also added a check in domIndexerImpl to make sure a DOM node is correct indexed, when the indexed segment is orphan segment, treat it as not indexed so in case the same problem happens, we will clear cache.


JiuqingSong:
#2864

BryanValverdeU (approved):

Issue #2860, opened 2024-11-04
Provide a way to plugin to know added and deleted DOM elements for blocks when rewrite with Content Model


JiuqingSong:
We can find a way to pass the added and deleted DOM element for blocks when rewrite with Content Model.

For example, in editor we have:
<div>1</div>
<div>2</div>
<div>3</div>

And in Content Model we have:
Paragraph1: cachedElement = div 1
Paragraph2: cachedElement = undefined
Paragraph3: cachedElement = div 3

When rewrite, div1 and div3 will not be changed, we need to delete div2 and create a new DIV. After that, we can pass the old div2 and new DIV to plugins so they can do necessary actions if they have some DOM elements cached in the plugins.

PR #2873, opened 2024-11-14
#2860 Expose added and removed block elements during Content Model rewrite

JiuqingSong:
The goal of this PR is to let plugins know what is added/removed when we rewrite with Content Model. This is done by:
1. Add a new event type `rewriteFromModel`
2. Add properties in `ModelToDomContext` to store those added and removed block elements
3. In model handlers, put newly created block elements into `modelToDomContext.addedBlockElements`
4. In `reuseCachedElements`, put removed elements into `modelToDomContext.removedBlockElements`
5. When create editor, pass these elements via `EditorReadyEvent`
6. Every time editor DOM is changed from `formatContentModel`, trigger `RewriteFromModelEvent` with these elements so plugins can handle these 2 events to know what are added and removed

Note that for removed block elements, if there is list, we will only pass OL/UL element but not the elements under it. So if you want to check removed element and do proper handling, you may also need to go deep to these elements to find the elements you are interested in.

Also note that this event only works for content changed triggered from `formatContentModel` / `setContentModel`. If content change is done by other ways such as direct DOM access, we don't trigger this event. To be compatible with old plugins (v8) using `EditorAdapter`, you can handle `ContentChangedEvent` and check if `event.contentModel` exists. If it exists, it means this is from `formatContentModel` and there will be a `RewriteFromModelEvent` following so you may not need to handle that `ContentChangedEvent`. Otherwise, it is from old code and there will not be a realted `RewriteFromModelEvent`.


juliaroldi (approved):
Which scenario do we need this change?

PR #2897, opened 2024-12-02
[Hot fix] Roosterjs Plugins 9.15.1

juliaroldi:
Add fix https://github.com/microsoft/roosterjs/pull/2895.


BryanValverdeU:
I think we are missing the version.json update

juliaroldi:
> I think we are missing the version.json update

Thanks!! I forgot to push the commit.

PR #2898, opened 2024-12-03
Use DOMCreator instead of TrustedHTMLHandler

juliaroldi:
The current trustedHTMLHandler converts a string into a trusted string, and if necessary, transforms the trusted string into a DOM. To optimize this process, introduce a new DOMCreator that directly converts a string into a trusted DOM. Consequently, the trustedHTMLHandler editor option will accept two types of handlers: one that transforms strings into trusted strings and another that converts strings into Document objects.


packages/roosterjs-content-model-types/lib/parameter/TrustedHTMLHandler.ts
```diff
@@ -1,4 +1,16 @@
 /**
  * A handler type to convert HTML string to a trust HTML string
  */
-export type TrustedHTMLHandler = (html: string) => string;
+export type LegacyTrustedHTMLHandler = (html: string) => string;
```
JiuqingSong, line 4: Let's mark this interface as "@deprecated" in comment. So next time we do major version bump, we will remove it.

packages/roosterjs-content-model-types/lib/editor/IEditor.ts
```diff
@@ -193,12 +193,21 @@ export interface IEditor {
     hasFocus(): boolean;
 
     /**
+     * @deprecated use getDOMCreator instead
```
JiuqingSong, line 196: Very good!

packages/roosterjs-content-model-core/lib/editor/core/createEditorCore.ts
```diff
@@ -43,7 +43,11 @@ export function createEditorCore(contentDiv: HTMLDivElement, options: EditorOpti
             options.knownColors,
             options.generateColorKey
         ),
-        trustedHTMLHandler: options.trustedHTMLHandler || defaultTrustHtmlHandler,
+        trustedHTMLHandler:
+            options.trustedHTMLHandler && !isDOMCreator(options.trustedHTMLHandler)
+                ? options.trustedHTMLHandler
+                : defaultTrustHtmlHandler,
```
JiuqingSong, line 49: If options.trustedHTMLHandler is DOMCreator, we still need to create a valid trustedHTMLHandler, to satisfy the following case:
1. Pass in a DOMCreator when init editor
2. call editor.getTrustedHTMLHandler() and use it to sanitize HTML

You can create a function here to use DOMCreator() to return a DOM tree then get its innerHTML and return

JiuqingSong: Once we integrate this into OWA, we can search all places where getTrustedHTMLHandler is used and replace with getDOMCreator(), so this new function should not really be used in OWA but other roosterjs customers may still need it

packages/roosterjs-content-model-core/lib/utils/domCreator.ts
```diff
@@ -0,0 +1,37 @@
+import type {
+    DOMCreator,
+    LegacyTrustedHTMLHandler,
+    TrustedHTMLHandler,
+} from 'roosterjs-content-model-types';
+
+/**
+ * @internal
+ */
+export function domCreator(trustedHTMLHandler?: TrustedHTMLHandler): DOMCreator {
```
JiuqingSong, line 10: Suggest name it `createDOMCreator`

JiuqingSong (approved):

PR #2900, opened 2024-12-07
Fix issue that paste will remove cache

JiuqingSong:
I found that when paste, we will always use cloned model as target model, although the model content is the same, but it will clear all caches so that after paste we need to rerender the whole content. 

It is because of a code bug that we always assign `modelBeforePaste` to clipboardData even for the first paste, so when we merge paste model, it is always treated as second paste.

To fix it, add a parameter to tell `mergePasteContent` API that it is a first paste although there is `modelBeforePaste` in clipboardData. Then we will not clone model again, so caches can be preserved.


BryanValverdeU (approved):

JiuqingSong:
@BryanValverdeU  can you help approve again? Thanks

juliaroldi (approved):

PR #2901, opened 2024-12-09
fix Safari responds to the EnterKey from IM

miku1958:
How to repro: Press Enter when IM shows candidates

https://github.com/user-attachments/assets/7e379977-67a6-4c16-91cf-7b412d2f5558

And its keycode is 229
image

Which is dead code
image


packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
```diff
@@ -181,7 +181,8 @@ export class EditPlugin implements EditorPlugin {
                     break;
 
                 case 'Enter':
-                    if (!hasCtrlOrMetaKey) {
+                    if (!hasCtrlOrMetaKey && event.rawEvent.keyCode == 13) {
```
JiuqingSong, line 184: Interesting. If we are sure that keyCode will be 229 for IME enter, I prefer to check keyCode != 229 here which makes it more clear what it is doing.

Please add a const declaration for this key name and add comment

miku1958: Added

JiuqingSong:
I created a PR to fix it from core plugin. https://github.com/microsoft/roosterjs/pull/2905

packages/roosterjs-content-model-plugins/lib/edit/keyCode.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Key code enumeration
+ */
+export const enum KeyCode {
```
JiuqingSong, line 4: Just put it back in EditPlugin.ts and no need to export it.
And use a regular const instead of const enum since const enum has a lot of problem.

JiuqingSong (approved):

PR #2923, opened 2025-01-16
Remove margins of auto formatted list. 

juliaroldi:
To make the list format consistent between list triggered by characters and list triggered by ribbon control, add the option to remove the list margins when starting a new list using AutoFormat Plugin.

### DIFF
--- a/demo/scripts/controlsV2/sidePane/editorOptions/EditorOptionsPlugin.ts
+++ b/demo/scripts/controlsV2/sidePane/editorOptions/EditorOptionsPlugin.ts
@@ -47,6 +47,7 @@ const initialState: OptionState = {
         autoOrdinals: true,
         autoMailto: true,
         autoTel: true,
+        removeListMargins: false,
     },
     markdownOptions: {
         bold: true,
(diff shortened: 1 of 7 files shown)


JiuqingSong:
question, I tried to insert list from "1." and ribbon, but I didn't see any difference here. What is this PR fixing?

image

juliaroldi:
> question, I tried to insert list from "1." and ribbon, but I didn't see any difference here. What is this PR fixing?
> 
> image

In the `setListType` function we have removeMargins parameter, I just added an option to remove the margin to the auto format options, so if we set the removeMargins to true in the ribbon control, we can also set to true in the AutoFormat options plugin.

JiuqingSong:
> > question, I tried to insert list from "1." and ribbon, but I didn't see any difference here. What is this PR fixing?
> > image
> 
> In the `setListType` function we have removeMargins parameter, I just added an option to remove the margin to the auto format options, so if we set the removeMargins to true in the ribbon control, we can also set to true in the AutoFormat options plugin.

Got it. Thanks.

JiuqingSong (approved):

PR #2924, opened 2025-01-17
Remove Table Header Row font weight

juliaroldi:
The default font-weight of the TH element overrides the font-weight of the segment. To resolve this, set the TH font-weight to normal when it has no predefined style. This way, when adding a header row to a table, the bold font-weight will be applied directly to the text segments.

Before the change the bold button was showing the wrong state. 
image

After the change the bold button was showing the text is bold, as expected.
image


JiuqingSong:
So in this way you moved bold format down to segments.

What if user turn off header after that?

PR #2931, opened 2025-01-28
Stop editing when content changed

juliaroldi:
Before setting new content into the editor, check if it has the `'data-is-editing="true"'` dataset and remove it. Also stop editing the image when the content changed.

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
@@ -55,6 +55,7 @@ const MouseRightButton = 2;
 const DRAG_ID = '_dragging';
 const IMAGE_EDIT_CLASS = 'imageEdit';
 const IMAGE_EDIT_CLASS_CARET = 'imageEditCaretColor';
+const IMAGE_EDIT_DATASET = 'data-is-editing="true"';
 
 /**
  * ImageEdit plugin handles the following image editing features:
@@ -177,6 +178,9 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
             case 'extractContentWithDom':
                 this.removeImageEditing(event.clonedRoot);
                 break;
+            case 'beforeSetContent':
+                this.beforeSetContentHandler(this.editor, event.newContent);
+                break;
         }
     }
 
@@ -277,6 +281,11 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
                 );
             }
         }
+    }
+
+    private beforeSetContentHandler(editor: IEditor, newContent: string) {
+        newContent.replace(IMAGE_EDIT_DATASET, '');
+        editor?.setEditorStyle(IMAGE_EDIT_CLASS_CARET, null);
     }
 
     /**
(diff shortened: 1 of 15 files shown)


packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -279,6 +283,11 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
         }
     }
 
+    private beforeSetContentHandler(editor: IEditor, newContent: string) {
+        newContent.replace(IMAGE_EDIT_DATASET, '');
```
JiuqingSong, line 287: are you sure this can work since you didn't use the return value?

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -279,6 +283,11 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
...
+        editor?.setEditorStyle(IMAGE_EDIT_CLASS_CARET, null);
```
JiuqingSong, line 288: "beforeSetContent" is a legacy event from v8. Since there is no API "setContent" in v9, I prefer not use this event. Can we do this in ContentChangedEvent? We just need to make sure do not reset cursor style if images is still selected.

Actually we should do that in SelectionPlugin since this problem can still exist if ImageEditPlugin is not added

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
JiuqingSong: Instead of changing the source HTML, I prefer to do it in DOM tree, when handling ContentChangedEvent and check change source is setContent

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
juliaroldi: I was just testing in OWA right now and remove the dataset in new content worked, but the remove caret styled persisted, I'm testing using ContentChangedEvent right now. I think we should remove the caret style in ImageEdit, since we add this style when we add the image edit wrapper.

JiuqingSong: Isn't the style set from core API setDOMSelection? https://github.com/microsoft/roosterjs/blob/11c9e3edb04c647981b35e82198d58f01149c2b9/packages/roosterjs-content-model-core/lib/coreApi/setDOMSelection/setDOMSelection.ts#L61

juliaroldi: We add the style here  [
 editor.setEditorStyle(IMAGE_EDIT_CLASS_CARET, `caret-color: transparent;`);](https://github.com/microsoft/roosterjs/blob/11c9e3edb04c647981b35e82198d58f01149c2b9/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts#L441)

JiuqingSong (approved):

PR #2934, opened 2025-01-31
Send anchor event in Auto Link

juliaroldi:
After a text is automatically transformed into a link by typing and pressing space, send the anchor element when the ContentChanged event is triggered. Also add more unit tests.

### DIFF
--- a/packages/roosterjs-content-model-core/lib/coreApi/formatContentModel/formatContentModel.ts
+++ b/packages/roosterjs-content-model-core/lib/coreApi/formatContentModel/formatContentModel.ts
@@ -25,13 +25,8 @@ export const formatContentModel: FormatContentModel = (
     options,
     domToModelOptions
 ) => {
-    const {
-        onNodeCreated,
-        getChangeData,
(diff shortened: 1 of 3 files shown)


packages/roosterjs-content-model-core/lib/coreApi/formatContentModel/formatContentModel.ts
```diff
@@ -81,7 +81,7 @@ export const formatContentModel: FormatContentModel = (
                 contentModel: clearModelCache ? undefined : model,
                 selection: clearModelCache ? undefined : selection,
                 source: options?.changeSource || ChangeSource.Format,
-                data: getChangeData?.(),
+                data: options?.getChangeData?.() || getChangeData?.(),
```
JiuqingSong, line 84: hmm I don't understand, getChangedData() is from options, so options.getChangedData should be always same with getChangedData.

JiuqingSong (approved):

BryanValverdeU (approved):

PR #2937, opened 2025-02-05
Markdown Package 

juliaroldi:
This change introduces the `rooosterjs-content-model-markdown package`, which is responsible for converting plain text written in Markdown into a content model structure. All the [basic syntax](https://www.markdownguide.org/basic-syntax/#overview) except for alternative heading, HTML and code are supported.  The package provides a single public function, `convertMarkdownToContentModel`, which handles the conversion of plain text and inserts it into the editor.

The conversion process involves dividing the plain text by line breaks and identifying the appropriate content model blocks based on the Markdown syntax. The text is divided by the regex `/\r\n|\r|\\n|\n/`, but the splitLinesPattern parameter allows to replace this regex with another one. 

image

All the created content model blocks are inserted into a new model document in the order they appear in the plain text, and after all the plain text is processed, the new content model document is merged in the editor.

MardownTranformation


packages/roosterjs-content-model-markdown/lib/markdownToModel/creators/createBlockQuoteFromMarkdown.ts
```diff
@@ -0,0 +1,30 @@
...
+};
+
+/**
+ * @internal
+ */
+export function createBlockQuoteFromMarkdown(
+    text: string,
+    blockquote?: ContentModelFormatContainer
+): ContentModelFormatContainer {
+    text = text.replace('>', '');
```
JiuqingSong, line 25: I think we should only replace the first `>`

juliaroldi: replace function only replace the first element found, to replace all we need to use />/g

packages/roosterjs-content-model-markdown/lib/markdownToModel/convertMarkdownToContentModel.ts
```diff
@@ -0,0 +1,33 @@
...
+ * @param text The markdown text
+ * @param splitLinesPattern The pattern to split lines. Default is /\r\n|\r|\\n|\n/
+ * @returns The ContentModelDocument
+ */
+export function convertMarkdownToContentModel(
+    editor: IEditor,
+    text: string,
+    splitLinesPattern?: string
+) {
+    editor.formatContentModel(
```
JiuqingSong, line 17: Suggest this API does not have dependency to editor. It should just take string as input, and ContentModelDocument as output

JiuqingSong: This whole package should not have dependency to editor

JiuqingSong (approved):

Issue #2942, opened 2025-02-10
text/uri-list not supported when pasting


haven2world:
**Describe the bug**
If a user long-press the address bar on safari for iOS to copy the address, the pasteboard will only contain two types:

![Image](https://github.com/user-attachments/assets/ff6524f1-f315-427b-b28f-b43e84011d59)
On the web side, it will be recognized as "text/uri-list".
And it won't contain "text/plain" or "text/html" type, so roosterjs won't paste anything.
Now we need to support it due to users' report. But I'd like to know if roosterjs wants to support it internally.
If not, I think I should add `uri-list` to the `allowedCustomPasteType`, but could you let me know where I should handle the data and paste it?
Maybe the `beforePaste` event?

**To Reproduce**
1. Open RoosterJS demo in safari on iOS
2. Long press the address bar to copy it
3. Try to paste in the editor

**Expected behavior**
The URL should be pasted like other web page.

**Screenshots**

https://github.com/user-attachments/assets/ec443622-76f9-4160-985b-e8d20757b0bd

haven2world:
Great! I'll raise a PR

PR #2943, opened 2025-02-11
Support text/uri-list when pasting

haven2world:
Lont press safari's address bar to copy will get the web link in the type text/uri-list. Handling it in roosterjs to support pasting it.
To solve the issue #2942


BryanValverdeU:
#2943

BryanValverdeU (approved):

JiuqingSong (approved):

PR #2950, opened 2025-02-18
Implement Excel non-native paste event handling and related utilities

BryanValverdeU:
When pasting from Excel but retrieving the clipboard content programatically, such as in the Top Ribbon Paste button, The native APIs are returning the content with a table and removes all the attributes we use to determine whether the content is from Excel.

This causes an inconsistency when pasting the content from Ctrl + V and the paste button, as when using Ctrl + V we add the border colors, if the tables. But since with the paste button we don't have the border colors, we need to handle this case.

This PR adds a new paste source for excel non native so we can add parsers to add the borders to the tables when pasting excel content and not using ctrl + v

Fix 272712


packages/roosterjs-content-model-plugins/lib/paste/pasteSourceValidations/getPasteSource.ts
```diff
@@ -29,7 +30,8 @@ export type KnownPasteSourceType =
     | 'googleSheets'
     | 'wacComponents'
     | 'default'
-    | 'singleImage';
+    | 'singleImage'
+    | 'excelNonNativeEvent';
```
JiuqingSong, line 34: Does this exist in all cases when copy from Excel? If so, do we still need to keep 'excelDesktop'? Maybe we just need to replace 'excelDesktop' and always use the new one.

BryanValverdeU: Yes, the shadow workbook type exists when copying from Excel.

The main difference is that when pasting wIth Ctrl + V we need the current approach, (clipboard fragment does not have a table element, so we do an special case).
But when getting the clipboard programatically, the fragment does contain a table but we need to add borders to the table.

Another option is to merge both logic in excel handling, but if fragment contains a table we do not need to build the table from the rawHtml

JiuqingSong: Ok, then let's keep the current one to keep it simple.

JiuqingSong (approved):

packages/roosterjs-content-model-plugins/lib/paste/pasteSourceValidations/getPasteSource.ts
BryanValverdeU: Oops, just modified the logic. Now we just keep all the excel handling in a single block of the switch. I think it looks cleaner. Let me know if it is okay

JiuqingSong (approved):

PR #2951, opened 2025-02-18
Select image with keyboard

juliaroldi:
If there were no image selected, and the selection changed to a single image, select the image as image selection.
 
ImageKeyboardSelection

### DIFF
--- a/packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
+++ b/packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
@@ -646,21 +646,23 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
 
             //If am image selection changed to a wider range due a keyboard event, we should update the selection
             const selection = this.editor.getDocument().getSelection();
-
-            if (
-                newSelection?.type == 'image' &&
-                selection &&
-                selection.focusNode &&
-                !isSingleImageInSelection(selection)
-            ) {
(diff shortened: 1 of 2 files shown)


packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -646,21 +646,23 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
...
+                    const range = selection.getRangeAt(0);
+                    this.editor.setDOMSelection({
+                        type: 'range',
+                        range,
+                        isReverted:
+                            selection.focusNode != range.endContainer ||
+                            selection.focusOffset != range.endOffset,
+                    });
+                } else if (newSelection?.type !== 'image' && image) {
+                    this.editor.setDOMSelection({
```
JiuqingSong, line 661: Question, if I select text+image, will it also trigger the event?

juliaroldi: It should not trigger, the isSingleImageInSelection function verifies if there is only an image in the range selection. So, the image will be selected only if there only one image and no text or other element.

JiuqingSong: ok. Sounds good.

JiuqingSong (approved):

BryanValverdeU (approved):

PR #2953, opened 2025-02-20
Markdown Package - Export model to markdown

juliaroldi:
This PR is the second PR of https://github.com/microsoft/roosterjs/pull/2937. 
This change adds the ability to convert the content model to markdown. The `convertContentModelToMarkdown` is a public API that takes a model document as a parameter and returns a string in markdown format with basic syntax.


JiuqingSong:
In general, we should not create markdown from DOM, but should be from content model instead. Otherwise, no need to have this change, there should be existing library to do so.

So we need to have 2 separate steps for this work:
1. Create content model for selection
2. Create markdown from content model

These two steps should be fully decoupled, so when generate mark down, we don't know selection at all, but just generate from full content model.

We already have some code to generate content model from selection, that is used when we do cut/copy. So you can extract the code from there to be common, then reuse it.

And you can even skip step 1, just do step 2 only. Since step 1 is a common step, it is not specifically for mark down.

juliaroldi:
> In general, we should not create markdown from DOM, but should be from content model instead. Otherwise, no need to have this change, there should be existing library to do so.
> 
> So we need to have 2 separate steps for this work:
> 
> 1. Create content model for selection
> 2. Create markdown from content model
> 
> These two steps should be fully decoupled, so when generate mark down, we don't know selection at all, but just generate from full content model.
> 
> We already have some code to generate content model from selection, that is used when we do cut/copy. So you can extract the code from there to be common, then reuse it.
> 
> And you can even skip step 1, just do step 2 only. Since step 1 is a common step, it is not specifically for mark down.
I refactored the code to use model object to create markdown.

packages/roosterjs-content-model-markdown/lib/modelToMarkdown/processor/modelProcessor.ts
```diff
@@ -0,0 +1,44 @@
...
+    const listCounter: ListCounter = {
+        listItemCount: 0,
+        subListItemCount: 0,
+    };
+    for (const block of model.blocks) {
+        const isListItem = block.blockType === 'BlockGroup' && block.blockGroupType === 'ListItem';
+        if (!isListItem && (listCounter.listItemCount > 0 || listCounter.subListItemCount > 0)) {
+            listCounter.listItemCount = 0;
+            listCounter.subListItemCount = 0;
+            markdown += '\n';
```
JiuqingSong, line 21: Suggest make the  NewLine char as a parameter, in case user want to use \r\n instead

packages/roosterjs-content-model-markdown/lib/modelToMarkdown/creators/createMarkdownParagraph.ts
```diff
@@ -0,0 +1,47 @@
...
+    for (const segment of segments) {
+        switch (segment.segmentType) {
+            case 'Text':
+                markdownString += textProcessor(segment);
+                break;
+            case 'Image':
+                markdownString += `![${segment.alt}](${segment.src})`;
+                break;
+            case 'Br':
+                markdownString += '\n';
```
JiuqingSong, line 19: For paragraph under list, we need to indent
For paragraph under block quote, we need to add "|"
For paragraph under table, we should not break table structure

JiuqingSong: So I think we may need a context object to remember current state, and when create new line, what kind of related things to add, especially these things combined together.

packages/roosterjs-content-model-markdown/lib/modelToMarkdown/processor/modelProcessor.ts
```diff
@@ -0,0 +1,44 @@
...
+            case 'Table':
+                markdown += createMarkdownTable(block, listCounter) + '\n';
+                break;
+            case 'Divider':
+                markdown += '---\n\n';
+                break;
+            default:
+                // Ignore other block types
+                break;
+        }
```
JiuqingSong, line 40: You already have `createMarkdownBlock`, you can reuse

JiuqingSong (approved):

BryanValverdeU:
[like]  Bryan Valverde reacted to your message:

PR #2956, opened 2025-02-26
Trigger only one auto format at time

juliaroldi:
Refactor the `AutoFormatPlugin` class to trigger only one auto format at a time to avoid conflict between the formatting.  
Before: 
Bug Hyphen list

After: 
FixHyphenList

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/autoFormat/hyphen/transformHyphen.ts
+++ b/packages/roosterjs-content-model-plugins/lib/autoFormat/hyphen/transformHyphen.ts
@@ -13,11 +13,6 @@ export function transformHyphen(
     paragraph: ShallowMutableContentModelParagraph,
     context: FormatContentModelContext
 ): boolean {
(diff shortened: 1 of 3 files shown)


packages/roosterjs-content-model-plugins/lib/autoFormat/hyphen/transformHyphen.ts
```diff
@@ -13,12 +13,16 @@ export function transformHyphen(
     paragraph: ShallowMutableContentModelParagraph,
     context: FormatContentModelContext
 ): boolean {
+    // If the selection marker is at the beginning of a paragraph, there is no text to be transformed before it
+    if (paragraph.segments[0].segmentType == 'SelectionMarker') {
```
JiuqingSong, line 17: I think instead of checking where is the selection marker, we should check if the event is already handled by some other feature (here it is switch to bullet list), an event should only be handled by one feature.

On the other hand, if we disabled the auto bullet feature, it is ok to transform hyphen here.

PR #2961, opened 2025-03-07
[FHL] Hint Text API and plugin

JiuqingSong:


JiuqingSong:
No plan to check it in recently. So close.

PR #2966, opened 2025-03-12
Apply formatting to document block groups in handleBlockGroupChildren

BryanValverdeU:
Currently if the format of the ContentModelDocument is modified the change is not reflected when calling formatContentModel, the reason seems to be that we do not apply the style to the parent element where the contentModelToDom is being handled.

This PR adds an applyFormat when the model being handled is a ContentModelDocument, so changes to the default format of the ContentModelDocument are applied correctly in the root where the contentModelToDom is being handled..

### DIFF
--- a/packages/roosterjs-content-model-dom/lib/modelToDom/handlers/handleBlockGroupChildren.ts
+++ b/packages/roosterjs-content-model-dom/lib/modelToDom/handlers/handleBlockGroupChildren.ts
@@ -1,3 +1,4 @@
+import { applyFormat } from '../utils/applyFormat';
 import { isNodeOfType } from '../../domUtils/isNodeOfType';
 import type {
     ContentModelBlockGroup,
@@ -38,6 +39,14 @@ export const handleBlockGroupChildren: ContentModelHandler<ContentModelBlockGrou
                 context.domIndexer?.onBlockEntity(childBlock, group);
             }
         });
+
+        if (
+            group.blockGroupType == 'Document' &&
+            group.format &&
+            isNodeOfType(parent, 'ELEMENT_NODE')
+        ) {
+            applyFormat(parent, context.formatAppliers.segment, group.format, context);
+        }
 
         // Remove all rest node if any since they don't appear in content model
         while (refNode) {
--- a/packages/roosterjs-content-model-dom/test/modelToDom/handlers/handleBlockGroupChildrenTest.ts
+++ b/packages/roosterjs-content-model-dom/test/modelToDom/handlers/handleBlockGroupChildrenTest.ts
@@ -456,4 +456,57 @@ describe('handleBlockGroupChildren', () => {
             removedBlockElements: [],
         });
     });
+
+    it('handle document and update the parent format based on the document format', () => {
+        const div1 = document.createElement('div');
+        const div2 = document.createElement('div');
+
+        div1.textContent = 'test1';
+        div2.textContent = 'test2';
+
+        parent.appendChild(div1);
+        parent.appendChild(div2);
+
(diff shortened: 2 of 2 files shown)


juliaroldi (approved):

packages/roosterjs-content-model-dom/lib/modelToDom/handlers/handleBlockGroupChildren.ts
```diff
@@ -39,6 +40,14 @@ export const handleBlockGroupChildren: ContentModelHandler<ContentModelBlockGrou
             }
         });
 
+        if (
+            group.blockGroupType == 'Document' &&
+            group.format &&
+            isNodeOfType(parent, 'ELEMENT_NODE')
+        ) {
+            applyFormat(parent, context.formatAppliers.segment, group.format, context);
```
JiuqingSong, line 48: format on Document is not used for applying to the root, but it is used for default segment format

packages/roosterjs-content-model-dom/lib/modelToDom/handlers/handleBlockGroupChildren.ts
```diff
@@ -39,6 +40,14 @@ export const handleBlockGroupChildren: ContentModelHandler<ContentModelBlockGrou
             }
         });
 
+        if (
+            group.blockGroupType == 'Document' &&
```
juliaroldi, line 44: by the name of the function, handleBlockGroupChildren,  it only should handle only the children of a blockGroup. Should we create handler just to handle the parent?

packages/roosterjs-content-model-dom/lib/modelToDom/handlers/handleBlockGroupChildren.ts
JiuqingSong: What is the scenario to change the default format after content model is created?

packages/roosterjs-content-model-dom/lib/modelToDom/handlers/handleBlockGroupChildren.ts
BryanValverdeU: I think we can just move this to contentModelToDom, good catch.

packages/roosterjs-content-model-dom/lib/modelToDom/handlers/handleBlockGroupChildren.ts
JiuqingSong: In that case, you need to reformat the whole doc where user typed. Because this format is applied everytime user is typing, but not used for formatting the doc.

BryanValverdeU: Hmm, good point. Let me confirm with Jorge.
I guess another option is to let the content just before the user changed the mailbox, but when they try to Clear format, use the new account default format.

JiuqingSong: anyway, I think the fix should be in OWA side

BryanValverdeU: Hmm, if that is the case, sure I can do the change in OWA, but since there are no capabilities from Content Model to do it, might need to do it modifying the DOM directly, which defeats the purpose of having content model.

Also, from formatContentModel API perspective, modifying the format of the model is not doing anything. So I thought that was a bug. Maybe we can set that property as readonly so users are not able to modify it.

PR #2967, opened 2025-03-12
New getEditorHTMLContent api

juliaroldi:
Add `getEditorHTMLContent` API to `roosterjs-content-model-api`. This API uses getClonnedRoot to retrieve the inner HTML content of the editor.

### DIFF
--- a/packages/roosterjs-content-model-api/lib/publicApi/utils/getEditorHTMLContent.ts
+++ b/packages/roosterjs-content-model-api/lib/publicApi/utils/getEditorHTMLContent.ts
@@ -1,0 +1,25 @@
+import { transformColor } from 'roosterjs-content-model-dom';
+import type { IEditor } from 'roosterjs-content-model-types';
+
+/**
+ * Return the inner HTML content of the editor
+ * @param editor The editor object
+ * @returns
+ */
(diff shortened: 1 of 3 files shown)


packages/roosterjs-content-model-api/lib/publicApi/utils/getEditorHTMLContent.ts
```diff
@@ -0,0 +1,25 @@
+import { transformColor } from 'roosterjs-content-model-dom';
+import type { IEditor } from 'roosterjs-content-model-types';
+
+/**
+ * Return the inner HTML content of the editor
+ * @param editor The editor object
+ * @returns
+ */
+export function getEditorHTMLContent(editor: IEditor): string {
```
JiuqingSong, line 9: Instead of adding a new API, how about just replace the existing code of exportContent when it requests cleanHTML?

JiuqingSong: And @BryanValverdeU I think we just need to use this one from roosterjs directly, and in OWA side we only do the content model output length check and logging, then later when we solve the problem, we can easily remove the override

PR #2968, opened 2025-03-20
Biwu/321release

flyingbee2012:


packages/roosterjs-content-model-plugins/lib/autoFormat/interface/AutoFormatOptions.ts
```diff
@@ -33,4 +33,9 @@ export interface AutoFormatOptions extends AutoLinkOptions {
      * Remove the margins of auto triggered list
      */
     removeListMargins?: boolean;
+
+    /**
+     * Auto Horizontal line
+     */
+    autoHorizontalLine?: boolean;
```
BryanValverdeU, line 40: Since we updated this interface, I think we need to bump to "9.22.0" instead.

JiuqingSong (approved):

PR #2972, opened 2025-03-28
Support undeletable anchor Step 3: Support anchor

JiuqingSong:
This is the first step to support undeletable anchor.

The final goal is to allow adding anchor such as and specify what kind of anchor is not deletable, then all editing function will not delete it.

In this PR I'm adding support to anchor -- an `<A>` element with name but no href.

In order to support anchor, we need to let element parser recognize A element that has no href, also need to support empty anchor: A element without child. When that happens, add an empty text segment as the holder of anchor.

We also need to modify some utility functions such as `normalizeParagraph` and `isSegmentEmpty` to let them keep those empty text segment which has anchor link.


BryanValverdeU:
Just curious, what happens if user with DevTools delete the element or with third party code the body of the editor is edited? Maybe just a corner case, but was curious what happens in that scenario

BryanValverdeU (approved):

JiuqingSong:
> Just curious, what happens if user with DevTools delete the element or with third party code the body of the editor is edited? Maybe just a corner case, but was curious what happens in that scenario

Then we will lose those info.
But since user opens debugger, they can already copy everything :P

PR #2973, opened 2025-03-28
Support undeletable anchor Step 4: Prevent deleting undeletable anchor

JiuqingSong:
This is the first step to support undeletable anchor.

The final goal is to allow adding anchor such as and specify what kind of anchor is not deletable, then all editing function will not delete it.

This is the last step, I'm modifying code for deleting content to avoid deleting undeletable anchors. This includes:
1. When delete segment, make sure undeletable anchor is not removed from model
2. When delete selection, we can delete text content of anchor, but need to leave the text segment not deleted
3. When delete multiple lines, if there is undeletable anchor in middle of them, add them back after delete
4. When BACKSPACE between regular text and undeletable anchor, use content model code handle this deletion, because if delete using browser's behavior, if the anchor is empty, it will also be removed.

For 1-3, we read undeletable property from content model. For 4, in order to make it fast, we read undeletable property from DOM directly. This is the reason why we need to keep undeletable property into hidden property.

### DIFF
--- a/packages/roosterjs-content-model-dom/lib/modelApi/editing/deleteExpandedSelection.ts
+++ b/packages/roosterjs-content-model-dom/lib/modelApi/editing/deleteExpandedSelection.ts
@@ -38,6 +38,7 @@ export function deleteExpandedSelection(
         deleteResult: 'notDeleted',
         insertPoint: null,
         formatContext,
+        undeletableSegments: [],
     };
 
     iterateSelections(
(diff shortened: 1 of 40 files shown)


packages/roosterjs-content-model-dom/lib/modelApi/editing/deleteExpandedSelection.ts
```diff
@@ -81,7 +84,15 @@ export function deleteExpandedSelection(
...
                                 tableContext
                             );
-                        } else if (deleteSegment(block, segment, context.formatContext)) {
+                        } else if (
+                            deleteSegment(
+                                block,
+                                segment,
+                                context.formatContext,
+                                undefined /*direction*/,
+                                isFirstDeletingParagraph ? undefined : context.undeletableSegments // For first paragraph we can keep undeletable segments so ono need to merge it later
```
BryanValverdeU, line 93: nit: typo

BryanValverdeU (approved):

PR #2975, opened 2025-04-01
Find previously marked paragraph

JiuqingSong:
Sometimes we want to find a paragraph and then call some service or any other async call, then update the paragraph using the result. At this point there is no way to find back the paragraph since content model could be recreated.

So in this PR I'm adding a marker to paragraph in both content model and DOM, then allow caller to find latest paragraph model using a previous one. Here is an example how to use it:

```ts
let para: ReadonlyContentModelParagraph;

   para = getSelectedParagraphs(model)[0]; // Now store the paragraph here
   return false;
});

await someAsyncCall(); // This may take a long time, and content can be changed by user while it is waiting

editor.formatContentModel((model, context) => {
   if (para) {
      // use paragraphIndexer under `FormatContentModelContext` to get the latest paragraph
      const updatedPara = context.paragraphIndexer?.getParagraphFromMarker(para);

      // Now you can use updatedPara to do whatever you want
      const mutablePara = updatedPara && mutateBlock(updatePara);
     
     if (mutablePara) {
       // ...
       return true;
     }
   }

   // Updated paragraph not found, could be deleted by user, so no op
   return false;
});
```

### DIFF
--- a/packages/roosterjs-content-model-dom/lib/domUtils/hiddenProperties/hiddenProperty.ts
+++ b/packages/roosterjs-content-model-dom/lib/domUtils/hiddenProperties/hiddenProperty.ts
@@ -2,7 +2,11 @@
  * @internal
  */
 export interface HiddenProperty {
-    dummy?: {}; // Temp used by test, will be removed later
+    /**
+     * A marker string that can be used to identify a specific paragraph in the DOM.
+     * This is useful for scenarios where you need to track or manipulate specific paragraphs
+     */
+    paragraphMarker?: string;
 
     // TODO: Add more properties as needed
 }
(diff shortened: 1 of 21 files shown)


juliaroldi (dismissed):

packages/roosterjs-content-model-dom/lib/domUtils/hiddenProperties/hiddenProperty.ts
```diff
@@ -2,10 +2,18 @@
...
+     * This is useful for scenarios where you need to track or manipulate specific paragraphs
+     */
+    paragraphMarker?: string;
+
     /**
      * Specify we should not delete this element when delete/backspace key is pressed
      */
     undeletable?: boolean;
+
+    // Add more properties as needed
```
BryanValverdeU, line 16: Is this comment needed?

JiuqingSong: it is optional, just let others know they key keep adding things here

BryanValverdeU (approved):

PR #3000, opened 2025-04-15
Add 'CustomCopyCut' experimental feature to fix some copy cut bugs

BryanValverdeU:
Browser is adding or removing content from the selection.
To fix this, we are adding a new experimental feature called 'CustomCopyCut', when this experimental feature is enabled, we prevent the default behavior of the browser and instead we set the clipboard content ourselves.

Before
AfterPaste`

After
AfterPaste


packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
```diff
@@ -154,7 +160,12 @@ class CopyPastePlugin implements PluginWithState<CopyPastePluginState> {
                     isCut,
                 }).range;
 
-                if (newRange) {
+                if (this.customCopyCutEnabled && isClipboardEvent(event)) {
+                    event.preventDefault();
+                    const text = contentModelToText(pasteModel);
+                    event.clipboardData?.setData('text/html', tempDiv.innerHTML);
+                    event.clipboardData?.setData('text/plain', text);
```
JiuqingSong, line 167: What if right click on inline image and choose "copy image"?

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
```diff
@@ -131,7 +137,7 @@ class CopyPastePlugin implements PluginWithState<CopyPastePluginState> {
             }
 
             const tempDiv = this.getTempDiv(this.editor.getDocument());
-            const context = createModelToDomContext();
+            const context = createModelToDomContext(undefined /* editorContext */);
```
JiuqingSong, line 140: no need to change this

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/utils/pruneUnselectedModel.ts
```diff
@@ -0,0 +1,101 @@
...
+} from 'roosterjs-content-model-types';
+
+/**
+ * @internal
+ */
+export function pruneUnselectedModel(
+    model: ContentModelBlockGroupBase<ContentModelBlockGroupType, HTMLElement>,
+    isSelectionAfterElement: boolean = false
+) {
+    for (let index = model.blocks.length - 1; index >= 0; index--) {
```
JiuqingSong, line 16: after this, we should do a unwrap process if all selections are in the same block group. For example

```html
<ol start="1" data-editing-info="{&quot;applyListStyleFromLevel&quot;:false,&quot;orderedStyleType&quot;:1}" style="list-style-type: decimal;"><li style="font-family: Calibri; font-size: 11pt; color: rgb(0, 0, 0);"><div>asdfasdfasdf</div></li></ol><!--{"type":"range","start":[0,0,0,0,2],"end":[0,0,0,0,8],"isReverted":false,"isDarkMode":false}-->
```

Since all selection is in the same paragraph, I expect we only copy a paragraph rather than a list item

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
BryanValverdeU: Copy Image does not trigger the event the ClipboardEvent to the editor. So it will work normally, based on my testing

But if selecting the image and copying it with the context menu item, will just add the image to the HTML element.

JiuqingSong: ok

JiuqingSong:
If I copy some text from one single line, and paste, it will add a new line after the pasted content and move cursor to the new line. It should not add new line actually.

When only one line is copied, we need to mark the copied line as an implicit paragraph, and remove its block format. If there is any `segmentFormat` in block, also need to move it down into segments (but do not overwrite existing format with the same name in segment if any), so the generated html does not contain DIV.

JiuqingSong (approved):

PR #3012, opened 2025-04-29
Support List Pasting from PowerPoint Desktop

BryanValverdeU:
When copying lists from PowerPoint Desktop, the lists added to the clipboard are not UL/OL elements, which causes that the end result when pasting when user press Enter the lists do not continue.

So to fix, transform the  clipboard contents to lists by using the available metadata:

image
 
In the above image, is an example of a list from PowerPoint.

1. The class 0# represents the depth of the list item, if the list is in the first level, the class is not present. So we use this to determine the level of the list. For example, if the class is not present, the level of the list item should be only 1. But if the class is present for example 0n, the list level should be n + 1.
2. The element with the style mso-special-format is the element with the bullet. The value of that style also represents the type of list and the start number of the list. for example, `numbullet6\,1` means that the list is LowerAlpha and that the list starts with 1. The available numbered list are shown in the below image: 
image
3. SInce the span with mso-special-format is the representation of the bullet, when processing that element, we just return as the list marker will be added to the LI element instead.

For unordered lists, this is the enum of the by default lists markers:
image

Power point sets different type of markers, using the [Wingdings](https://learn.microsoft.com/en-us/typography/font-list/wingdings) font, we could add all those markers to be supported in Rooster, but for this PR I only added support for the default markers in PowerPoint, which are the below:
image


packages/roosterjs-content-model-dom/lib/constants/BulletListType.ts
```diff
@@ -42,8 +42,21 @@ export const BulletListType = {
...
+     * Box Shadow bullet type
+     */
+    BoxShadow: 10,
+    /**
+     * Rhombus with a cross inside
+     */
+    Xrhombus: 11,
+    /**
+     * Check mark bullet type
+
```
JiuqingSong, line 55: remove this empty line

packages/roosterjs-content-model-plugins/lib/paste/WordDesktop/processPastedContentFromWordDesktop.ts
```diff
@@ -71,7 +71,7 @@ function adjustPercentileLineHeight(format: ContentModelBlockFormat, element: HT
 function listLevelParser(
```
JiuqingSong, line 71: since it is a parser, let's declare it as a parser, following the same pattern with line 84

JiuqingSong (approved):

PR #3015, opened 2025-04-29
Clean image edit when undo

juliaroldi:
When setContent event happens and the selection is an image, clean the image edit state. Therefore, when the undo action happens the styles for hide the cursor will be removed. 
Fix in OWA: 
UndoImage


JiuqingSong:
The PR is fine.
I'm thinking that since we have the hiddenProperty feature now, maybe we can leverage this feature instead of setting the data-editing property so after undo it will be disappeared automatically, then we just need to clean up `isEditing` and `isCropMode`.

juliaroldi:
> The PR is fine. I'm thinking that since we have the hiddenProperty feature now, maybe we can leverage this feature instead of setting the data-editing property so after undo it will be disappeared automatically, then we just need to clean up `isEditing` and `isCropMode`.

I'll do this in another change.

PR #3018, opened 2025-04-30
Retain Excel Table styles if styles get removed due to html comments.

BryanValverdeU:
In OWA, the Excel styles are getting removed due to a custom DOMCreator The reason is that Styles in HEAD are wrapped in html comment tags

So we may need to remove them and add a way to add CSS styles to the BeforePasteEvent. 

Source: 
image

Beforeimage

Afterimage


packages/roosterjs-content-model-core/lib/command/createModelFromHtml/createModelFromHtml.ts
```diff
@@ -1,7 +1,8 @@
-import { convertInlineCss, retrieveCssRules } from './convertInlineCss';
+import { convertInlineCss } from './convertInlineCss';
 import { createDOMCreator } from '../../utils/domCreator';
 import { createDomToModelContextForSanitizing } from './createDomToModelContextForSanitizing';
 import { createEmptyModel, domToContentModel, parseFormat } from 'roosterjs-content-model-dom';
+import { retrieveCssRules } from '../../../../roosterjs-content-model-dom/lib/domUtils/style/retrieveCssRules';
```
JiuqingSong, line 5: Fix this import. If necessary, export this function from dom package.

PR #3021, opened 2025-05-05
Add image hidden marker

juliaroldi:


packages/roosterjs-content-model-plugins/lib/imageEdit/utils/findEditingImage.ts
```diff
@@ -20,7 +20,7 @@ export function findEditingImage(
             for (const segment of paragraph.segments) {
                 if (
                     segment.segmentType == 'Image' &&
-                    ((imageId && segment.format.id == imageId) || segment.dataset.isEditing)
+                    ((imageId && segment.format.id == imageId) || segment.format.imageMarker)
```
JiuqingSong, line 23: So this means as long as image has a marker, no matter what value it is, we treat it as editing?

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -174,24 +175,9 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
...
     }
 
-    private removeImageEditing(clonedRoot: HTMLElement) {
-        const images = clonedRoot.querySelectorAll('img');
-        images.forEach(image => {
-            if (image.dataset.isEditing) {
-                delete image.dataset.isEditing;
-            }
-            if (image.dataset.editingInfo) {
-                delete image.dataset.editingInfo;
```
JiuqingSong, line 190: I think we still need to keep this code to remove editing info.

packages/roosterjs-content-model-plugins/lib/imageEdit/utils/findEditingImage.ts
JiuqingSong: In that case, you can declare it as a boolean. So we treat it as editing when its value is true.

juliaroldi: I changed to verify if the marker equal to isEditing, maybe as string we can reuse this marker to point order states than editing.

JiuqingSong: That is a good idea. So let's rename it to "state" instead of "marker" since marker is too common and not easy to understand what it is used for.

JiuqingSong (approved):

PR #3025, opened 2025-05-08
Bump main version to 9.27.0

BryanValverdeU:
#3003
#3007 
#3006 
#3010 
#3015 
#3000 
#3014 
#3019
#3011 
#3021 
#3023 
#3012 
#3024 
#3027 
#3029 
#3028 
#3026
#3031


packages/roosterjs-content-model-types/lib/contentModel/format/FormatHandlerTypeMap.ts
```diff
@@ -114,6 +115,11 @@ export interface FormatHandlerTypeMap {
      */
     id: IdFormat;
 
+    /**
+     * Format of ImageStateFormat
+     */
+    imageState: ImageStateFormat;
```
BryanValverdeU, line 121: @JiuqingSong could this be considered as breaking change?

JiuqingSong: technically it is. But it should be fine because we don't expect developers create object of this type directly.

BryanValverdeU: I see, thanks.

JiuqingSong (dismissed):

JiuqingSong (approved):

PR #3031, opened 2025-05-08
Add margin-inline-start to watermark styles to fix #3030

### LINKED ISSUE #3030: [A11y] Using Windows high contrast themes, cursor does not appear
**Describe the bug**
As a user: Turn on Aquatic theme (Navigation Path: System Settings> Accessibility > Contrast theme>Aquatic theme) 
Navigate to RoosterJS Demo site
https://microsoft.github.io/roosterjs/index.html
Notice that by default the when the textfield is focused, the cursor in the textfield does not display properly. 
Once the user begins typing, the textfield cursor shows up again. 

This is observed in all high contrast themes not specific to Aquatic

**Expected behavior**
Textfield cursor should always be visible in high contrast themes.

**Screenshots**

Image

Example of it being used in ACS UI Chat Composite
Image

BryanValverdeU:
Fix #3030 

Looks like in Contrast themes, the watermark ::before element content hides the cursor. So add margin at the start of the line so it is visible:

Before
image

After
image

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/watermark/WatermarkPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/watermark/WatermarkPlugin.ts
@@ -135,7 +135,7 @@ export class WatermarkPlugin implements EditorPlugin {
     }
 
     private applyWatermarkStyle(editor: IEditor) {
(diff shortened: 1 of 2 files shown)


JiuqingSong (approved):

JiuqingSong:
nice fix

PR #3032, opened 2025-05-12
Allow queryContentModelBlocks to query blocks in entities

florian-msft:
This PR updates `queryContentModelBlocks` to optionally take a callback `shouldExpandEntity` which should return a `EditorContext` or `null`. If an `EditorContext` is returned, `queryContentModelBlocks` will "expand" the entity and continue recursively checking its children. If `null` is returned, the existing behavior is preserved and the entity's children are ignored.

The PR also exports `createEditorContextForEntity` for external consumption as it is helpful for consumers to create an `EditorContext` for a given entity for usage in `queryContentModelBlocks`.


packages/roosterjs-content-model-api/lib/publicApi/utils/formatSegmentWithContentModel.ts
```diff
@@ -135,7 +135,16 @@ export function formatSegmentWithContentModel(
...
 }
 
-function createEditorContextForEntity(editor: IEditor, entity: ContentModelEntity): EditorContext {
+/**
+ * Create EditorContext for an entity
+ * @param editor The editor object
+ * @param entity The entity to create the context for
+ * @returns The EditorContext for the entity
+ */
+export function createEditorContextForEntity(
```
JiuqingSong, line 144: Move this function to a separate file.

florian-msft: Updated with changes to move it into a new file and added tests!

JiuqingSong (approved):

PR #3041, opened 2025-05-20
Fix A11y bug with table selection

BryanValverdeU:
Issue definition:
1. Start table selection
2. Format
3. Press Arrow Down key

Expect: 
Table Selection to be removed.

Actual:
Table Selection is kept but the cursor is moved.

The root cause of the issue is that when handling the contentChanged event, we invalidate the tableSelection state. So, when handling the keyDown event and there is no table Selection but the getDOMSelection returned value is a table selection, try to create a new tableSelection state to be able to handle the event correctly.


packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -350,6 +350,24 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
                 break;
 
             case 'table':
+                if (this.state.tableSelection == null) {
```
JiuqingSong, line 353: please add a comment to explain why tableSelection can be null.

JiuqingSong (approved):

PR #3047, opened 2025-05-22
Fix Word Desktop pasting when using Safari

### LINKED ISSUE #3016: Paste a list form WordDesktop to Safari will lose the format
**Describe the bug**
There are two bugs actually.
### 1. Safari misidentified the pasta source:
In `PastePlugin.ts`, in function `onPluginEvent(event)`, the pasteSource should be 'wordDesktop', but it use 'default' instead. This is because, when run `createDOMFromHtml` in `paste.ts`, the document.children is missing some attributes. And this lead to the misidentifying of the pasteSource:

Image

rawHtml:
Image

Expected:
Image

Actually:

Image

### 2. list format is not retrieved correctly in safari and all fallback to `orderedStyleType:1`.

This is because the clipboardData.rawHtml in safari doesn't have:
Image

Edge rawHtml:
Image

Safari rawHtml:

Image

So in `paste.ts`, the `htmlFromClipboard.htmlBefore` and `htmlFromClipboard.htmlAfter` will be empty:
Image

And this make the metadata in `processWordList.ts` is empty, and all list format fallback to `orderedStyleType:1` in `processWordList.ts`. No matter what type of the list is in Word.

Image

**To Reproduce**
1. Copy a multi-level ordered/unordered list from Word to roosterJS demo site on Safari.
2. See the format of the list

**Device Information**
 - OS: macOS
 - Browser: safari

**Additional context**

BryanValverdeU:
Safari removes the html attributes with their DOMParser implementation. Which causes that the current checks to verify if the Clipboard content is from Word to fail.

To fix add an additional check to confirm if the pasted content is from Word Desktop by checking if the rawHtml contains the attributes using indexOf in the string.

Additionally, the check to get the Style metadata for lists failed in Safari, the reason is that the style tag is like: 
`<style someAttribute>`, so fix that as well.

This is to fix #3016


packages/roosterjs-content-model-plugins/lib/paste/pasteSourceValidations/isWordDesktopDocument.ts
```diff
@@ -12,9 +12,17 @@ const WORD_PROG_ID = 'Word.Document';
...
-    const { htmlAttributes } = props;
+    const { htmlAttributes, clipboardData } = props;
+
     return (
         htmlAttributes[WORD_ATTRIBUTE_NAME] == WORD_ATTRIBUTE_VALUE ||
-        htmlAttributes[PastePropertyNames.PROG_ID_NAME] == WORD_PROG_ID
+        htmlAttributes[PastePropertyNames.PROG_ID_NAME] == WORD_PROG_ID ||
+        // Safari removes the metadata from the clipboard html, so we need to do this check.
+        !!(
+            clipboardData.rawHtml &&
```
JiuqingSong, line 22: Should we add a check to Mac here? (editor.getEnvironment().isMac)

JiuqingSong: or isSafari

BryanValverdeU: Sounds good, adding it back.

JiuqingSong (approved):

PR #3049, opened 2025-05-22
Prevent multiple event attachments for mousemove in SelectionPlugin

BryanValverdeU:
This PR ensures only one `mousemove` handler is attached when dragging selections and adds tests to verify this behavior.  
- Added a guard in `SelectionPlugin` to prevent multiple `mousemove` attachments  
- Updated tests to cover duplicate `mouseDown` calls and assert single attachment


packages/roosterjs-content-model-core/test/corePlugin/selection/SelectionPluginTest.ts
```diff
@@ -876,6 +876,83 @@ describe('SelectionPlugin handle table selection', () => {
...
+            mouseDisposer: mouseMoveDisposer,
+            imageSelectionBorderColor: DEFAULT_SELECTION_BORDER_COLOR,
+            imageSelectionBorderColorDark: DEFAULT_SELECTION_BORDER_COLOR,
+            tableCellSelectionBackgroundColor: DEFAULT_TABLE_CELL_SELECTION_BACKGROUND_COLOR,
+            tableCellSelectionBackgroundColorDark: DEFAULT_TABLE_CELL_SELECTION_BACKGROUND_COLOR,
+        });
+        expect(mouseDispatcher).toBeDefined();
+    });
+
+    it('MouseDown - save a table selection when left click', () => {
```
BryanValverdeU, line 904:
```suggestion
    it('MouseDown - only attach event handler once even when double mousedown event', () => {
```

JiuqingSong (approved):

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -219,11 +219,13 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
                 }
             }
 
-            this.state.mouseDisposer = editor.attachDomEvent({
-                mousemove: {
-                    beforeDispatch: this.onMouseMove,
-                },
-            });
+            if (!this.state.mouseDisposer) {
+                this.state.mouseDisposer = editor.attachDomEvent({
```
JiuqingSong, line 223: how about we just dispose the previous one and then attach a new one? I'm afraid if in a corner case this may cause the event can never be attached.

BryanValverdeU: That works for me, changed in latest commit.

JiuqingSong (approved):

PR #3050, opened 2025-05-23
Only paste text content of button

BryanValverdeU:
Add a processor on paste to handle button elements, and instead just paste the text content of the button vs handling as a generalElement

https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/338590/


packages/roosterjs-content-model-plugins/lib/paste/PastePlugin.ts
```diff
@@ -219,3 +224,15 @@ function tableBorderParser(format: ContentModelTableCellFormat, element: HTMLEle
...
+
+/**
+ * @internal exported only for unit test
+ */
+export const pasteButtonProcessor: ElementProcessor<HTMLButtonElement> = (
+    group: ContentModelBlockGroup,
+    element: HTMLButtonElement,
+    context: DomToModelContext
+): void => {
+    const text = createText(element.textContent || '', context.segmentFormat);
```
JiuqingSong, line 236: I think we should keep the format of the text on button. So if it has some special format, user can easily distinguish them

BryanValverdeU: Sounds good, looks like button can contain blocks and segments but still be inline, so we need to check all child elements. 
Addressed this in latest commit

JiuqingSong (approved):

Issue #2909, opened 2024-12-11
Apply format issue with Japanese keyboard on mobile


haven2world:
**Describe the bug**
When a user use Japanese keyboard on mobile the toggle format doesn't work correctly.

**To Reproduce**
1. Go to demo site on mobile (iOS or Android)
2. Click on toggle bold button
3. type some characters but don't click the confirm button
4. click the toggle bold button again
5. type some more characters and click the confirm button

**Expected behavior**
The text shouldn't be bold

**Screenshots**
https://github.com/user-attachments/assets/48ad5165-24b0-4c78-8568-5b20b37f5ef7

**Device Information**
on both iOS and Android

haven2world:
bump this issue

JiuqingSong:
I tried but as soon as I press the button, IME is dismissed so English letter will be applied to editor.

Any special way to make the IME not dismissed?

haven2world:
Were you using Japanese keyboard? Feels like it's a Japanese keyboard specific issue

JiuqingSong:
Yes:

![Image](https://github.com/user-attachments/assets/9dbd6fc4-d3b2-497b-a083-8019681a8dcc)

haven2world:
Probably you need to run it on mobile

JiuqingSong:
probably. In that case can you help debug and see why? Or maybe an easier fix is just to dismiss the IME when click out.

haven2world:
Sure, but I'm not familiar with this part, do you have any clue or where should I start to debug?

JiuqingSong:
In `formatSegmentWithContentModel.ts`, if selection is collapsed, we will pass the pending format back:

![Image](https://github.com/user-attachments/assets/ed3f5eb2-9ad0-4c45-be82-cef04c430132)

Then in `formatContentModel.ts`, we apply this pending format into editor core:

![Image](https://github.com/user-attachments/assets/0e4b7085-7a21-4dac-8636-7f56c88de0f5)

Finally in FormatPlugin, while handling compositionEnd event, we call `checkAndApplyPendingFormat` to put it into model:

![Image](https://github.com/user-attachments/assets/436ba775-8edc-4e65-a1fb-f53f291b280c)

That is the whole flow of how pending format works.

haven2world:
Hi @JiuqingSong 
I finally got some time to debug this issue,
I just followed the same repro steps and debugged at the position you pointed out:
This is what I observed:
[...]

PR #3064, opened 2025-06-04
Fix toggling format issue at composition stage for CJK

haven2world:
To fix issue #2909


packages/roosterjs-content-model-core/lib/corePlugin/cache/domIndexerImpl.ts
```diff
@@ -232,6 +233,20 @@ export class DomIndexerImpl implements DomIndexer {
             ) {
                 this.reconcileTextSelection(oldSelection.start.node);
             } else {
+                if (
+                    oldSelection.type == 'range' &&
```
JiuqingSong: so many conditions, please extract them to be a separate function, and add comment to explain what it is doing.

JiuqingSong: and please add test case.

JiuqingSong (approved):

PR #3068, opened 2025-06-06
Version bump 9.29.3

flyingbee2012:
{
    "react": "9.0.2",
    "main": "9.29.3",
    "legacyAdapter": "8.63.2",
    "overrides": {}
}


versions.json
```diff
@@ -1,6 +1,6 @@
 {
```
BryanValverdeU, line 1: I dont see interface changes, I think we can do 9.29.3 and remove the override.

```
{
    "react": "9.0.2",
    "main": "9.29.3",
    "legacyAdapter": "8.63.2",
    "overrides": {}
}
```

PR #3074, opened 2025-06-17
Adjust List Indentation

juliaroldi:
To match Word behavior, when toggling a list to an indented segment, add the indentation to the list. 
Before:
IndentListBug

After: 
IndentList


packages/roosterjs-content-model-api/lib/modelApi/list/setListType.ts
```diff
@@ -156,3 +159,37 @@ function shouldIgnoreBlock(block: ReadonlyContentModelBlock) {
...
+        });
+
+        if (tabSpaces) {
+            listItem.levels[0].format.marginLeft = tabSpaces * 40 + 'px';
+        }
+    }
+}
+
+function countSpacesBeforeText(str: string) {
+    const space = ' ';
```
BryanValverdeU: We can move this outside of the function as a const

packages/roosterjs-content-model-api/lib/modelApi/list/setListType.ts
```diff
@@ -156,3 +159,37 @@ function shouldIgnoreBlock(block: ReadonlyContentModelBlock) {
...
+        block.segments[0].segmentType == 'Text'
+    ) {
+        const spaces = countSpacesBeforeText(block.segments[0].text);
+        const tabSpaces = Math.floor(spaces / 4);
+        mutateSegment(block, block.segments[0], textSegment => {
+            textSegment.text = textSegment.text.substring(tabSpaces * 4);
+        });
+
+        if (tabSpaces) {
+            listItem.levels[0].format.marginLeft = tabSpaces * 40 + 'px';
```
BryanValverdeU: nit: Add 40 as a const. 
Also how did we found that we need to multiply by 40?

juliaroldi: In the `setModelIndentation` file. We use setModelIndentation to handle indentation by TAB and we do not pass the length, so it fallback to 40px. 
image

packages/roosterjs-content-model-api/lib/modelApi/list/setListType.ts
```diff
@@ -156,3 +165,37 @@ function shouldIgnoreBlock(block: ReadonlyContentModelBlock) {
...
+            }
+        }
+    }
+}
+
+function countSpacesBeforeText(str: string) {
+    let count = 0;
+
+    for (const char of str) {
+        if (char === SPACE) {
```
JiuqingSong, line 193: what about `&nbsp;`?

juliaroldi: We count the spaces of the text of text segment. Do we need to check `&nbsp`; when not handling HTML directly?

JiuqingSong: I mean, space (\u20) and `&nbsp;` (\ua0) are different characters. We need to handle both of them.

juliaroldi: When we indent a segment, we add 4 (\u20) before the text, that is why I'm only considering the (\u20). If we add 4 (\ua0) spaces, should we consider the indented too?

JiuqingSong: Actually, the space we added is `&ensp;`, or say \u2002. So that your code can work since you are also checking \u2002.

My question is, do we want to support the spaces added by user manually, in that case it could be \u0020 (ASCII space) or \u00a0 (`&nbsp;`). If not, then current code change is fine. Otherwise, we need to check all of them.

juliaroldi: I got it now, I'll reach Jorge and ask this question.

PR #3076, opened 2025-06-19
Disable isComposing check in DOM event propagation for Android

Rain-Zheng:
The `isComposing` property is not reliable in some Android IME (such as Samsung keyboard). As a result, the plugins can't receive keyboard and input event normally when typing.

### DIFF
--- a/packages/roosterjs-content-model-core/lib/corePlugin/domEvent/DOMEventPlugin.ts
+++ b/packages/roosterjs-content-model-core/lib/corePlugin/domEvent/DOMEventPlugin.ts
@@ -156,7 +156,10 @@ class DOMEventPlugin implements PluginWithState<DOMEventPluginState> {
                 event.stopPropagation();
             }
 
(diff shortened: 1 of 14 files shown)


BryanValverdeU (approved):

packages/roosterjs-content-model-core/lib/corePlugin/domEvent/DOMEventPlugin.ts
```diff
@@ -156,7 +156,10 @@ class DOMEventPlugin implements PluginWithState<DOMEventPluginState> {
                 event.stopPropagation();
             }
 
-            if (this.editor && eventType && !event.isComposing && !this.state.isInIME) {
+            const isAndroid = this.editor?.getEnvironment()?.isAndroid ?? false;
+            const isComposing = !isAndroid && (event.isComposing || this.state.isInIME);
+
+            if (this.editor && eventType && !isComposing) {
```
BryanValverdeU, line 162: Could we add some unit tests for this change?

PR #3079, opened 2025-06-25
Allow plugins to store state in snapshot

florian-msft:
This PR adds a new property `additionalState` to the undo snapshots. Before the snapshot is created, the new event `beforeAddUndoSnapshot` is fired which allows plugins to push onto a shared array of custom state. This state is stored with the snapshot. When restoring, the custom state is broadcast to plugins as part of the existing `contentChanged` event.

### DIFF
--- a/packages/roosterjs-content-model-core/lib/editor/Editor.ts
+++ b/packages/roosterjs-content-model-core/lib/editor/Editor.ts
@@ -200,14 +200,13 @@ export class Editor implements IEditor {
      * Add a single undo snapshot to undo stack
      * @param entityState @optional State for entity if we want to add entity state for this snapshot
      */
-    takeSnapshot(entityState?: EntityState, additionalState?: string[]): Snapshot | null {
+    takeSnapshot(entityState?: EntityState): Snapshot | null {
         const core = this.getCore();
 
         return core.api.addUndoSnapshot(
             core,
             false /*canUndoByBackspace*/,
-            entityState ? [entityState] : undefined,
-            additionalState
+            entityState ? [entityState] : undefined
         );
     }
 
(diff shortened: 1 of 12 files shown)


packages/roosterjs-content-model-types/lib/parameter/Snapshot.ts
```diff
@@ -70,6 +70,12 @@ export interface Snapshot {
      */
     html: string;
 
+    /**
+     * Additional state supplied by plugins. When doing an undo/redo to this snapshot, this state will be added to the
+     * content model context as additional state.
+     */
+    additionalState?: string[];
```
BryanValverdeU: Do you think an array is enough? 
If there are multiple plugins in different logic adding additional state, could be hard to get the desired state from the array when restoring the data.

Maybe a Map<string, string> or index signature { [key: string]: string } could make the retrival of the data easier.

packages/roosterjs-content-model-core/lib/editor/Editor.ts
```diff
@@ -200,13 +200,14 @@ export class Editor implements IEditor {
...
-    takeSnapshot(entityState?: EntityState): Snapshot | null {
+    takeSnapshot(entityState?: EntityState, additionalState?: string[]): Snapshot | null {
         const core = this.getCore();
 
         return core.api.addUndoSnapshot(
             core,
             false /*canUndoByBackspace*/,
-            entityState ? [entityState] : undefined
+            entityState ? [entityState] : undefined,
+            additionalState
```
BryanValverdeU: Do we need both allowing additional state when calling the Editor API and also let Plugins handle the event to add more state?

I feel only adding snapshot state from the editor API would make the logic easier. If multiple plugins add different states could make handling of the state a bit harder to predict.

packages/roosterjs-content-model-types/lib/parameter/Snapshot.ts
BryanValverdeU: Or maybe adding a key string to the additionalData, so we can filter the array could be useful
```
{ key: string, state: string }
```

florian-msft: That's a good point! I will go with the first approach (but can also change it to the array of objects with a key if needed)!

BryanValverdeU (approved):

PR #3082, opened 2025-06-30
Add auto direction to setDirection

wisaulni:
Adds 'auto' to the setDirection API, and builds in support for auto-detection of a text's directionality.

Unicode defines characters are 'strong' RTL indicators, and we use a manual list of those in order to determine directionality of a paragraph. It's a simple matter to count the characters and see which one has more.

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/edit/keyboardEnter.ts
+++ b/packages/roosterjs-content-model-plugins/lib/edit/keyboardEnter.ts
@@ -35,7 +35,7 @@ export function keyboardEnter(
                     ? []
                     : [handleAutoLink, handleEnterOnList, deleteEmptyQuote];
 
-                if (handleNormalEnter || hasEnterForEntity(result.insertPoint?.paragraph)) {
+                if (handleNormalEnter || handleEnterForEntity(result.insertPoint?.paragraph)) {
                     steps.push(handleEnterOnParagraph);
                 }
 
@@ -64,9 +64,9 @@ export function keyboardEnter(
     );
 }
 
-function hasEnterForEntity(paragraph: ReadonlyContentModelParagraph | undefined) {
+function handleEnterForEntity(paragraph: ReadonlyContentModelParagraph | undefined) {
     return (
         paragraph &&
-        (paragraph.isImplicit || paragraph.segments.some(x => x.segmentType == 'SelectionMarker'))
+        (paragraph.isImplicit || paragraph.segments.some(x => x.segmentType == 'Entity'))
     );
 }
--- a/packages/roosterjs-content-model-plugins/test/edit/keyboardEnterTest.ts
+++ b/packages/roosterjs-content-model-plugins/test/edit/keyboardEnterTest.ts
@@ -1366,7 +1366,7 @@ describe('keyboardEnter', () => {
         );
     });
 
-    it('selection cover table', () => {
+    it('Handle enter when there is entity', () => {
         const model: ContentModelDocument = {
             blockGroupType: 'Document',
             blocks: [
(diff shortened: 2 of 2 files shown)


packages/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
```diff
@@ -99,3 +109,40 @@ function setProperty(
...
+    if (block.blockType === 'Paragraph') {
+        const findTextSegements: ReadonlyContentModelText[] = block.segments.filter(
+            (seg): seg is ReadonlyContentModelText => seg.segmentType === 'Text'
+        );
+        let innerText =
+            findTextSegements.length > 0
+                ? findTextSegements.reduce((prev, seg) => prev + seg.text, '')
+                : undefined;
+        if (!!innerText) {
+            // Strongly typed RTL character ranges. Referenced unicode's DerivedBidiClass.txt, excluding things in the 2 bit range.
```
BryanValverdeU: Can you provide more context on this comment?

packages/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
```diff
@@ -99,3 +109,40 @@ function setProperty(
...
+            const rtlPattern = /[\u0590-\u05FF\u0600-\u08FF\uFB1D-\uFDFF\uFE70-\uFEFF]/g;
```
BryanValverdeU: Can we move this as a const in module scope?

packages/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
```diff
@@ -99,3 +109,40 @@ function setProperty(
...
+
+            // Remove links
+            innerText = innerText.replace(
+                /http\S+|www\S+|https\S+|<a\s+(?:[^>]*?\s+)?href=(["']).*?\1.*?>.*?<\/a>/g,
```
BryanValverdeU: Same for this

packages/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
```diff
@@ -99,3 +109,40 @@ function setProperty(
...
+                ''
+            );
+
+            // Remove whitespace
+            innerText = innerText.replace(/\s/g, '');
```
BryanValverdeU: Same for this regex

packages/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
wisaulni: [This is the unicode listing of bidi (bidirectional) character ranges.](https://www.unicode.org/Public/UNIDATA/extracted/DerivedBidiClass.txt)

I took everything that was not in the 2b range (basically, everything that isn't archaic/ancient languages) and put it in this regex. In the future those could be included but you'd need to target a version of ECMAscript that includes support for 2 bit unicode in regexes.

packages/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
wisaulni: Fixed in next push.

packages/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
wisaulni: Fixed in next push.

packages/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
wisaulni: Fixed in next push.

BryanValverdeU (approved):

PR #3089, opened 2025-07-16
test

juliaroldi:


packages/roosterjs-content-model-dom/lib/domToModel/processors/childProcessor.ts
```diff
@@ -12,19 +12,21 @@ import type {
...
  * @param context DOM to Content Model context
+ * @param processNonVisibleElements If true elements that has display:none style will be processed @default false
  */
-export const childProcessor: ElementProcessor<ParentNode> = (
+export const childProcessor: ElementProcessor<ParentNode, boolean> = (
     group: ContentModelBlockGroup,
     parent: ParentNode,
-    context: DomToModelContext
+    context: DomToModelContext,
+    processNonVisibleElements: boolean = false
```
JiuqingSong: No need to add a new parameter, but you can integrate the parameter into `context`

packages/roosterjs-content-model-dom/lib/domToModel/processors/childProcessor.ts
```diff
@@ -37,14 +37,19 @@ export const childProcessor: ElementProcessor<ParentNode> = (
...
  * @param parent Parent DOM node to process
  * @param context DOM to Content Model context
+ * @param processNonVisibleElements If true elements that has display:none style will be processed @default false
  *
  */
 export function processChildNode(
     group: ContentModelBlockGroup,
     child: Node,
-    context: DomToModelContext
+    context: DomToModelContext,
```
JiuqingSong: we already passed in context here, so no need to add another parameter

packages/roosterjs-content-model-dom/lib/domToModel/processors/childProcessor.ts
```diff
@@ -37,14 +37,19 @@ export const childProcessor: ElementProcessor<ParentNode> = (
...
+    processNonVisibleElements: boolean = false
 ) {
-    if (isNodeOfType(child, 'ELEMENT_NODE') && child.style.display != 'none') {
+    if (
+        isNodeOfType(child, 'ELEMENT_NODE') &&
+        (child.style.display != 'none' || processNonVisibleElements)
+    ) {
         context.elementProcessors.element(group, child, context);
```
JiuqingSong, line 50: do we save the display value today? if not, it will be rendered without "display:none" so user can see it

juliaroldi: We save 
image

juliaroldi:
> Now that we have a new parameter in context, how can we pass it in when use editor?
I'm think in something like this, I'll test this change in OWA, before merging this PR
image

JiuqingSong:
> > Now that we have a new parameter in context, how can we pass it in when use editor?
> > I'm think in something like this, I'll test this change in OWA, before merging this PR
> 
> image

Since this is a new parameter, I expect to have a better way to pass it in, for example, a parameter in EditorOptions, then no need to do processor override here

packages/roosterjs-content-model-core/lib/command/paste/generatePasteOptionFromPlugins.ts
```diff
@@ -25,6 +25,7 @@ export function generatePasteOptionFromPlugins(
         processorOverride: {},
         styleSanitizers: {},
         attributeSanitizers: {},
+        processNonVisibleElements: false,
```
JiuqingSong: Do we need to get this value from editor? You can get the settings from `editor.getEnvironment().domToModelSettings`

packages/roosterjs-content-model-dom/lib/domToModel/context/createDomToModelContext.ts
```diff
@@ -103,6 +103,7 @@ export function createDomToModelConfig(
         ),
         defaultElementProcessors: defaultProcessorMap,
         defaultFormatParsers,
+        processNonVisibleElements: options.some(x => !!x?.processorOverride),
```
JiuqingSong: when check processorOverride?

packages/roosterjs-content-model-dom/lib/domToModel/utils/isBlockElement.ts
```diff
@@ -1,6 +1,6 @@
 import { getDefaultStyle } from './getDefaultStyle';
 
-const BLOCK_DISPLAY_STYLES = ['block', 'list-item', 'table', 'table-cell', 'flex'];
+const BLOCK_DISPLAY_STYLES = ['block', 'list-item', 'table', 'table-cell', 'flex', 'none'];
```
JiuqingSong: For display:none, I think we should still rely on its default display of the element.

For example, `<div style="display:none">` should still be treated as a block, but SPAN should be treated as inline.

So when you see display is none, fallback to its default display:

```ts
const display = element.style.display == 'none' ? null : element.style.display;
const effectiveDisplay = display || getDefaultStyle(element).display || '';

return ...
```

packages/roosterjs-editor-adapter/lib/editor/utils/eventConverter.ts
```diff
@@ -152,6 +152,7 @@ export function oldEventToNewEvent<TOldEvent extends OldEvent>(
                         processorOverride: {},
                         styleSanitizers: {},
                         attributeSanitizers: {},
+                        processNonVisibleElements: false,
```
JiuqingSong: Same here, get it from editor.

packages/roosterjs-editor-adapter/lib/editor/utils/eventConverter.ts
```diff
@@ -108,7 +109,8 @@ export const OldEventTypeToNewEventType: Record<PluginEventType, NewPluginEventT
  */
 export function oldEventToNewEvent<TOldEvent extends OldEvent>(
     input: TOldEvent,
-    refEvent?: NewEvent
+    refEvent?: NewEvent,
+    domToModelSettings?: DomToModelSettings
```
JiuqingSong: when to pass in this parameter?

juliaroldi: We call it from the adapter library. I thought this setting would not be useful for this legacy => new transformation

JiuqingSong (approved):

PR #3099, opened 2025-07-22
Adjust Block Indentation

juliaroldi:
When inserting a list or table in an indented segment, add the indentation to the block. 
IndentationListTable


packages/roosterjs-content-model-api/lib/modelApi/common/adjustIndentation.ts
```diff
@@ -0,0 +1,118 @@
...
+        } else {
+            table.format.marginLeft = indentationMargin * IndentStepInPixel + 'px';
+        }
+    }
+}
+
+const getTableIndentation = (paragraph: ShallowMutableContentModelParagraph) => {
+    let margin = 0;
+    const segments = paragraph.segments;
+    if (segments.length < 2) {
```
JiuqingSong: Add comment to explain what "2" means here

packages/roosterjs-content-model-api/lib/modelApi/common/adjustIndentation.ts
```diff
@@ -0,0 +1,118 @@
...
+        return;
+    }
+
+    const isEmptyLine = paragraph.segments.every(
+        s => s.segmentType == 'Text' || s.segmentType == 'SelectionMarker' || s.segmentType == 'Br'
```
JiuqingSong: why text segment means it is an empty line?

packages/roosterjs-content-model-api/lib/modelApi/common/adjustIndentation.ts
```diff
@@ -0,0 +1,118 @@
...
+    );
+    if (!isEmptyLine) {
+        return;
+    }
+    const lastSegment = paragraph.segments[paragraph.segments.length - 1];
```
JiuqingSong: what if current selection is in middle of a paragraph?

JiuqingSong:
I have a question about the whole flow. When press TAB at beginning of a paragraph, why not just set margin of this paragraph instead of inserting spaces? In that case we don't need to adjust margin for list and table any more.

juliaroldi:
> I have a question about the whole flow. When press TAB at beginning of a paragraph, why not just set margin of this paragraph instead of inserting spaces? In that case we don't need to adjust margin for list and table anymore.

Indent the whole paragraph was the former behavior. But some users complained, then we changed.  The bug related to this change is the 359969

packages/roosterjs-content-model-api/lib/modelApi/common/adjustIndentation.ts
juliaroldi: I'll add verification if the text segment has only spaces.

JiuqingSong (approved):

PR #3103, opened 2025-07-25
Fix 366527- Support RTL on Table 

juliaroldi:
The style `direction: RTL` does move the table to the right, so use `justifySelf:flex-end`.


JiuqingSong:
Or, in the format API `setModelDirection`, when we see a table, we can also set its alignment.

juliaroldi:
The style direction: RTL does move the table to the right, so use justifySelf:flex-end.

PR #3106, opened 2025-07-29
Fix superscript rule 

juliaroldi:
In English the suffix for ordinal numbers -st, -nd, -rd, and -th are applied based on the last digit of the number, except for 11,12,13. 

- Numbers ending in 1 → -st (except 11 → 11th)
- Numbers ending in 2 → -nd (except 12 → 12th)
- Numbers ending in 3 → -rd (except 13 → 13th)
- All others →-th

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/autoFormat/numbers/transformOrdinals.ts
+++ b/packages/roosterjs-content-model-plugins/lib/autoFormat/numbers/transformOrdinals.ts
@@ -72,7 +72,8 @@ function getNumericValue(text: string, checkFullText: boolean = false): number |
     const number = checkFullText ? text : text.substring(0, text.length - ORDINAL_LENGTH);
     const isNumber = /^-?\d+$/.test(number);
     if (isNumber) {
-        return parseInt(text);
+        const numericValue = parseInt(number);
+        return numericValue < 20 ? numericValue : parseInt(number.substring(number.length - 1));
     }
     return null;
 }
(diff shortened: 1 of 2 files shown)


packages/roosterjs-content-model-plugins/lib/autoFormat/numbers/transformOrdinals.ts
```diff
@@ -72,7 +72,8 @@ function getNumericValue(text: string, checkFullText: boolean = false): number |
     const number = checkFullText ? text : text.substring(0, text.length - ORDINAL_LENGTH);
     const isNumber = /^-?\d+$/.test(number);
     if (isNumber) {
-        return parseInt(text);
+        const numericValue = parseInt(number);
+        return numericValue < 20 ? numericValue : parseInt(number.substring(number.length - 1));
```
JiuqingSong: nit: Is it possible to be negative number? For example, -21 is less than 20, but it should be -21st (I think?)

juliaroldi: I fixed to use absolute values.

juliaroldi: But I think negatives are not used for ordinal numbers, but I'm not sure.

JiuqingSong (approved):

PR #3113, opened 2025-08-06
Fix #3110

### LINKED ISSUE #3110: When switching the selected image in a nested logical root, the selected image becomes size 0
**Describe the bug**
When using a custom logical root with multiple images and switching the selected image inside of it, the selected image collapses to size 0.

**To Reproduce**
1. Open the Rooster demo page
2. Open the side pane, then undo snapshots
3. Create a custom snapshot with the HTML snippet below, logical root: `[4,0]`, then restore snapshot
4. Click one image, then the other, and back again
5. Note how the selected image collapses to width and height 0

```html
<div>Outside</div>
<div><br></div>
<div>Inside
<img src="https://res.df.onecdn.static.microsoft/assets/eventify/themeImages/16_9/000.jpg" width="696" style="width: 696px; height: auto; max-width: 100%;"><img src="https://res.df.onecdn.static.microsoft/assets/eventify/themeImages/16_9/000.jpg" width="696" style="width: 696px; height: auto; max-width: 100%;">
</div>
```

**Expected behavior**
The image retains its size even when selected

**Screenshots**

https://github.com/user-attachments/assets/b44cb4ee-c0f6-4247-979e-43c0949f6d08

**Device Information**
 - OS: Windows
 - Browser: Edge
 - Version: 138.0.3351.109 (64-Bit)

**Additional context**
When selected and collapsed, the image has the correct width and height attributed but CSS styles cause it to be size 0: ``

BryanValverdeU:
Fix (#3110) logical root change causing selected image size to be 0. Also add a logical root preset to be able to test this scenario easier.

Before
BeforeImageSelection

After
Recording 2025-08-06 095847

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
@@ -469,15 +469,7 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
         ) {
             // Image dimensions are zero and loading is incomplete, wait for image to load.
             image.onload = () => {
-                this.imageEditInfo = {
-                    ...this.imageEditInfo,
-                    widthPx: image.clientWidth,
-                    heightPx: image.clientHeight,
-                    naturalWidth: image.naturalWidth,
-                    naturalHeight: image.naturalHeight,
-                };
-                image.style.width = this.imageEditInfo.widthPx + 'px';
-                image.style.height = this.imageEditInfo.heightPx + 'px';
+                this.updateImageDimensionsIfZero(image);
                 this.startEditingInternal(editor, image, apiOperation);
                 image.onload = null;
                 image.onerror = null;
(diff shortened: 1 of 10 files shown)


packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -487,10 +479,18 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
...
             };
         } else {
+            this.updateImageDimensionsIfZero(image);
             this.startEditingInternal(editor, image, apiOperation);
         }
     }
 
+    private updateImageDimensionsIfZero(image: HTMLImageElement) {
+        if (this.imageEditInfo?.widthPx === 0 || this.imageEditInfo?.heightPx === 0) {
+            this.imageEditInfo.widthPx = image.clientWidth;
```
juliaroldi, line 489: if the image has not loaded yet, can clientWidth/clientHeight be zero?

juliaroldi: Whe get the width/height from clientWidth/clientHeight when creating the imageEditInfo object. Is it possibleclientWidth/clientHeight are equal to zero here?
image

JiuqingSong (approved):

packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
BryanValverdeU: Yes, looks like we can hit this scenario sometimes. Which causes the image to dissapear.

PR #3119, opened 2025-08-14
add support for nested style markdowns

FrancisMengx:


juliaroldi (commented):
Hi @FrancisMengx, I find a small regression in this change, if you transform `**test** ~~ *test*` it will result in 

Image the ~~ is causing the next word to have strikethrough without the closing characters.

packages/roosterjs-content-model-markdown/lib/markdownToModel/appliers/applyTextFormatting.ts
```diff
@@ -5,44 +5,134 @@ import type {
...
+
     return textSegments;
 }
 
-function splitSegments(text: string): string[] {
-    return text.split(SPLIT_PATTERN).filter(s => s.trim().length > 0);
+function parseMarkerAt(text: string, index: number): FormatMarker | null {
+    const remaining = text.substring(index);
+
+    // Check for strikethrough (~~)
```
juliaroldi: nit: can we remove these comments?

PR #3126, opened 2025-08-20
Toggle blockquote in Format Container

juliaroldi:
FormatContainer can be used to other elements other than blockquotes. When toggling blockquote in format containers that are not blockquotes, search for the selected elements inside the container and only apply blockquote to theses selected elements. 

Before: 
ToggleBlockquoteBug

After: 
ToggleBlockquoteFixed

### DIFF
--- a/packages/roosterjs-content-model-api/lib/modelApi/block/toggleModelBlockQuote.ts
+++ b/packages/roosterjs-content-model-api/lib/modelApi/block/toggleModelBlockQuote.ts
@@ -50,40 +50,29 @@ export function toggleModelBlockQuote(
             current?: ContentModelFormatContainer
         ): target is ContentModelFormatContainer =>
             canMergeQuote(target, current?.format || (isRtl ? formatRtl : formatLtr));
-        if (isCustomContainer(paragraphOfQuote)) {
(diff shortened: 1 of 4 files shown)


packages/roosterjs-content-model-api/lib/modelApi/block/toggleModelBlockQuote.ts
```diff
@@ -52,8 +52,16 @@ export function toggleModelBlockQuote(
             canMergeQuote(target, current?.format || (isRtl ? formatRtl : formatLtr));
 
         paragraphOfQuote.forEach(({ block, parent }) => {
-            if (isQuote(block)) {
-                // Already in quote, no op
+            if (isBlockGroupOfType<ContentModelFormatContainer>(block, 'FormatContainer')) {
```
JiuqingSong: I think the problem is getOperationalBlocks (line 34) should not return format container if it is not a quote.

So instead of filtering them here, you can add a new callback parameter to this function to allow doing additional check for a block. For example

getOperationalBlocks(..., (block) => {
  if (block.blockType == 'FormatContainer') {
    return block.tagName == 'blockquote';
  } else {
    return true;
 }

PR #3127, opened 2025-08-22
Do not add undo snapshot from ImageEdit plugin if image is not changed

JiuqingSong:
1. Add a new property `skipUndo` in `ContentChangedEvent`
2. When this property is true, UndoPlugin will not add undo snapshot / set has new content for this event
3. When ContentChangedEvent is triggered due to the following reasons, set this property to true
   - Switch color mode
   - Keyboard input (to follow the existing behavior)
4. When add image editing border, set this property to true, so no need to add undo snapshot for image editing border
5. When remove image editing border, check if image is edited since current border is added. If not, do not add undo snapshot

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
@@ -1,6 +1,5 @@
 import { applyChange } from './utils/applyChange';
 import { canRegenerateImage } from './utils/canRegenerateImage';
-import { checkEditInfoState } from './utils/checkEditInfoState';
 import { checkIfImageWasResized, isASmallImage } from './utils/imageEditUtils';
 import { createImageWrapper } from './utils/createImageWrapper';
 import { Cropper } from './Cropper/cropperContext';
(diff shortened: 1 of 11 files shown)


packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
```diff
@@ -402,6 +407,18 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
                             normalizeImageSelection(previousSelectedImage);
                         }
 
+                        if (this.initialEditingInfo && this.imageEditInfo) {
+                            // Finish editing, check if image is changed and skip undo snapshot if not
+                            const changeState = checkEditInfoState(
```
juliaroldi: We already call checkEditInfoState inside applyChange, can we make applyChange return the the value of checkEditInfoState? then we would not need to call checkEditInfoState again, just use the result of applyChange.

JiuqingSong: Good point

juliaroldi (approved):

Issue #3111, opened 2025-08-05
iOS: Keyboard auto-capitalization is broken due to CM backspace key handling


haven2world:
**Describe the bug**
This is an iOS only issue.
When user tap delete key to delete the first letter of one paragraph, Content Model will handle this key down event and the browser won't receive it.
iOS's system keyboard implements the auto-capitalization feature based on this event.
Is it necessary to handle the delete event? Is it possible to not preventDefault?

JiuqingSong:
Is this same with https://github.com/microsoft/roosterjs/issues/3081?

On desktop Chrome, when delete the last character of a paragraph, it is possible to lose all the formats, so when user type again, the input text will have no format. It happens randomly. That is why we handle delete key with our code.

If you are sure this won't happen in iOS. you can submit a PR to skip handling DELETE/Backspace key in iOS or Safari.

haven2world:
I think it's the same cause. Let me check the behavior on iOS

Issue #3081, opened 2025-06-30
The letters in the soft keyboard are not deleted


haven2world:
**Describe the bug**
On iOS, if I input something and then delete them, these letters cannot be cleared completely in the soft keyboard
I guess it's related to delete key handled in rooster

**To Reproduce**
1. Go to demo
2. Input some letters
3. Type delete to remove them
4. See the soft keyboard

**Expected behavior**
When the content in the editor is deleted, the content in the soft keyboard should be also deleted

**Screenshots**
After entering:

![Image](https://github.com/user-attachments/assets/e153a3a1-183c-4f7e-8b86-b1a51d5b8d8d)

After being deleted:

![Image](https://github.com/user-attachments/assets/1ef6a188-283d-45e8-8d64-562e20930f0f)

**Device Information**
 - OS:  iOS

PR #3131, opened 2025-08-28
Fix auto-capitalization for iOS 

isnine:
Fix auto-capitalization for iOS by allowing browser to handle first character deletion

## Root Cause

The Content Model intercepts delete detection when deleting the first character, which prevents the system browser from applying keyboard auto-capitalization.

### Solution
When deleting the first character (i.e., offset = 1) on iOS, allow the browser to handle the deletion.
When offset = 0, delete detection is still handled by the Content Model. In my testing, other styles can still be cleared correctly by the Content Model at that point.

### Preview
Screen Recording 2025-08-29 at 14 53 10

#3111 #3081

### DIFF
--- a/packages/roosterjs-content-model-core/lib/editor/core/createEditorCore.ts
+++ b/packages/roosterjs-content-model-core/lib/editor/core/createEditorCore.ts
@@ -71,7 +71,7 @@ function createEditorEnvironment(
         modelToDomSettings: createModelToDomSettings(options),
         isMac: appVersion.indexOf('Mac') != -1,
         isAndroid: /android/i.test(userAgent),
-        isIOS: /iPad|iPhone|iPod/.test(userAgent),
+        isIOS: /iPad|iPhone/.test(userAgent),
         isSafari:
             userAgent.indexOf('Safari') >= 0 &&
             userAgent.indexOf('Chrome') < 0 &&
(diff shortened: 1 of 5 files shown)


packages/roosterjs-content-model-plugins/lib/edit/keyboardDelete.ts
```diff
@@ -121,16 +122,21 @@ function shouldDeleteWithContentModel(
...
 }
 
-function canDeleteBefore(rawEvent: KeyboardEvent, text: Text, offset: number) {
-    if (rawEvent.key != 'Backspace' || offset <= 1) {
+function canDeleteBefore(rawEvent: KeyboardEvent, text: Text, offset: number, editor: IEditor) {
+    if (rawEvent.key != 'Backspace') {
         return false;
     }
+    if (offset <= 1) {
+        const { isAndroid, isMobileOrTablet } = editor.getEnvironment();
```
JiuqingSong: Instead of passing in the whole editor, you can just pass in EditorEnvironment

packages/roosterjs-content-model-core/lib/editor/core/createEditorCore.ts
```diff
@@ -71,6 +71,7 @@ function createEditorEnvironment(
         modelToDomSettings: createModelToDomSettings(options),
         isMac: appVersion.indexOf('Mac') != -1,
         isAndroid: /android/i.test(userAgent),
+        isIOS: /iPad|iPhone|iPod/.test(userAgent),
```
haven2world: I guess we don't need to consider iPod now

JiuqingSong (approved):

Issue #3135, opened 2025-09-01
Copy broken on v9.35.0 without initialModel


Durss:
**Describe the bug**
Copy behavior is broken on latest Rooster version.

**To Reproduce**
1. Go to [this sandbox](https://codesandbox.io/p/devbox/epic-feynman-c65kf9)
2. wait it little so everything loads fine
3. write text on the field
4. select some texte
5. press ctrl/cmd+C
6. text is unselected and nothing is copied to clipboard. 

Now, if a`initialModel` is given on rooster init _(uncomment line 25 of `RoosterEditor.vue`)_, the issue disapears.
Everything works fine under v9.34.0

**Expected behavior**
The text should remain selected and be copied to clipboard

**Device Information**
 - OS: MacOS 15
 - Browser Vivaldi
 - Version rooster 9.35.4

BryanValverdeU:
Thanks for reporting @Durss.
@JiuqingSong this change might be related to this issue:  (#3130).

<img width="2642" height="922" alt="Image" src="https://github.com/user-attachments/assets/a3049ba1-b2f8-4f3b-b9da-d60a6595f1d5" />

since we do not insert container now, container.parentNode is undefined when setting the value of the selection in the context and the paragraph is implicit. Then contentModelToDom does no return a range.

PR #3137, opened 2025-09-02
Revert "Improve handleParagraph function"

JiuqingSong:
Reverts microsoft/roosterjs#3130

This change causes the following issues:

1. For implicit paragraph, the parent block is not correctly assigned into context.regularSelection when convert model to dom
2. Because we delay appending paragraph container after handling all segments, in onNodeCreated for segment, if it wants to read computed styles, it will not work correctly

Given the above issues, I decide to revert this change.


juliaroldi (approved):

JiuqingSong:
#3135

PR #3142, opened 2025-09-08
Fix js error in keyboardInput

JiuqingSong:
I have no idea why this can happen. But since we see it in the log, let's fix it by adding a null check

```
Error Message: Uncaught TypeError: Cannot read properties of undefined (reading 'length')

at webpack://owa/packages/roosterjs-content-model-plugins/lib/edit/keyboardInput.ts:72
at webpack://owa/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts:242
at webpack://owa/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts:142
at webpack://owa/packages/common/controls/packages/controls/owa-editor-lazy-plugin/src/utils/createLazyContentModelPlugin.ts:135
at webpack://owa/packages/roosterjs-content-model-core/lib/coreApi/triggerEvent/triggerEvent.ts:30
at Native Browser Function(Array.forEach)
at webpack://owa/packages/roosterjs-content-model-core/lib/coreApi/triggerEvent/triggerEvent.ts:28
at webpack://owa/packages/roosterjs-content-model-core/lib/editor/Editor.ts:267
at webpack://owa/packages/roosterjs-content-model-core/lib/corePlugin/domEvent/DOMEventPlugin.ts:163
at webpack://owa/packages/roosterjs-content-model-core/lib/coreApi/attachDomEvent/attachDomEvent.ts:18
```

https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/384304

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/imageEdit/ImageEditPlugin.ts
@@ -240,11 +240,7 @@ export class ImageEditPlugin implements ImageEditor, EditorPlugin {
             event.rawEvent.button !== MouseRightButton &&
             !this.isCropMode
         ) {
-            this.applyFormatWithContentModel(
-                editor,
-                this.isCropMode,
-                this.shadowSpan === event.rawEvent.target
-            );
+            this.applyFormatWithContentModel(editor, this.isCropMode);
         }
     }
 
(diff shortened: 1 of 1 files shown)


haven2world:
BTW, I'm curious about why we don't enable the ts config "strictNullChecks"

JiuqingSong:
> BTW, I'm curious about why we don't enable the ts config "strictNullChecks"

Actually we are in strict mode, and it can do strict null check. And this property is not supposed to be null in its type declaration. But we still see this error stack, so just add a simple protection

PR #3144, opened 2025-09-08
Reset logical root before restore snapshot

JiuqingSong:
When restore undo snapshot, if logical root is not the same with physical root, browser may want to maintain the scroll position of current editable area (logical root), so it can cause editor scroll to an unexpected position. so restore logical root to physical root before restoring.

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/edit/keyboardInput.ts
+++ b/packages/roosterjs-content-model-plugins/lib/edit/keyboardInput.ts
@@ -69,7 +69,11 @@ export function keyboardInput(editor: IEditor, rawEvent: KeyboardEvent) {
 function shouldInputWithContentModel(selection: DOMSelection | null, rawEvent: KeyboardEvent) {
     if (!selection) {
         return false; // Nothing to delete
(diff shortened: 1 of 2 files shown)


packages/roosterjs-content-model-core/lib/coreApi/restoreUndoSnapshot/restoreUndoSnapshot.ts
```diff
@@ -24,6 +24,11 @@ export const restoreUndoSnapshot: RestoreUndoSnapshot = (core, snapshot) => {
     try {
         core.undo.isRestoring = true;
 
+        // Force clear logical root before restoring snapshot, so the whole editor is editable.
```
BryanValverdeU, line 27: nit: Should we add unit test for this?

BryanValverdeU (approved):

packages/roosterjs-content-model-core/lib/coreApi/restoreUndoSnapshot/restoreUndoSnapshot.ts
JiuqingSong: there is test cases verifying it is called, but not verifying it is called before restoring html. I can add one later.

PR #3150, opened 2025-09-12
Version bump to 9.36.1

flyingbee2012:
Version bump to 9.37.0

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -1,6 +1,6 @@
 {
     "react": "9.0.2",
-    "main": "9.37.0",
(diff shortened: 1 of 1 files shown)


versions.json
```diff
@@ -1,6 +1,6 @@
 {
     "react": "9.0.2",
-    "main": "9.36.0",
+    "main": "9.37.0",
```
JiuqingSong: There isn't any public interface change, so we can use 9.36.1

flyingbee2012: Done!

PR #3154, opened 2025-09-16
Restore iOS auto-capitalization by delegating Backspace to the browser

isnine:
# Summary
We’ve observed that iOS maintains internal text-input state (e.g., keyboard auto-capitalization). When we intercept Backspace with preventDefault(), that state can incorrectly. 

# Code changes
1. Adding a callback to delegate Backspace to the browser.

# Issue
Our previous PR already fixed the issue where the system didn’t handle backspace when 'offset == 1'. However, since 'offset == 0' is still handled by the Content Model, the system’s internal state becomes inconsistent, which leads to auto-capitalization being triggered incorrectly.

Simulator Screen Recording - 15PRO(MailUIKIT Test) - 2025-09-16 at 15 12 54

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
@@ -275,6 +275,7 @@ export class EditPlugin implements EditorPlugin {
         switch (rawEvent.inputType) {
             case 'deleteContentBackward':
                 if (!this.shouldBrowserHandleBackspace(editor)) {
+                    // This logic is Android specific. It's because some Android keyboard doesn't support key and keycode, the value of them is always Unidentified, so we have to manually create a new one.
                     handled = keyboardDelete(
                         editor,
                         new KeyboardEvent('keydown', {
--- a/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
@@ -313,7 +313,7 @@ export class EditPlugin implements EditorPlugin {
         const opt = this.options.shouldHandleBackspaceKey;
         switch (typeof opt) {
             case 'function':
-                return Boolean(opt(editor));
+                return opt(editor);
             case 'boolean':
                 return opt;
             default:
--- a/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
@@ -320,4 +320,3 @@ export class EditPlugin implements EditorPlugin {
         }
     }
 }
-
--- a/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
@@ -33,6 +33,14 @@ export type EditOptions = {
      * @returns A boolean
      */
     shouldHandleEnterKey?: ((editor: IEditor) => boolean) | boolean;
+
+    /**
+     * Callback or boolean to determine whether the browser (not Content Model) should handle the Backspace key press.
+     * If the value/callback returns true, Rooster will NOT handle Backspace and will defer to the browser's native behavior.
+     * @param editor - The editor instance (when using callback).
+     * @returns A boolean
+     */
+    shouldHandleBackspaceKey?: ((editor: IEditor) => boolean) | boolean;
 };
 
 const BACKSPACE_KEY = 8;
@@ -201,7 +209,13 @@ export class EditPlugin implements EditorPlugin {
                 case 'Backspace':
                     // Use our API to handle BACKSPACE/DELETE key.
                     // No need to clear cache here since if we rely on browser's behavior, there will be Input event and its handler will reconcile cache
-                    keyboardDelete(editor, rawEvent, this.options.handleExpandedSelectionOnDelete);
+                    if (!this.shouldBrowserHandleBackspace(editor)) {
+                        keyboardDelete(
+                            editor,
+                            rawEvent,
+                            this.options.handleExpandedSelectionOnDelete
+                        );
+                    }
                     break;
 
                 case 'Delete':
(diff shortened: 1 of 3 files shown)


packages/roosterjs-content-model-plugins/lib/edit/keyboardDelete.ts
```diff
@@ -36,7 +36,8 @@ import type { DOMSelection, DeleteSelectionStep, IEditor } from 'roosterjs-conte
 export function keyboardDelete(
     editor: IEditor,
     rawEvent: KeyboardEvent,
-    handleExpandedSelection: boolean = true
+    handleExpandedSelection: boolean = true,
+    shouldBrowserHandleDelete: boolean = false
```
haven2world: We need to also update the comments above for the new paramter

isnine: thanks for the reminder, updated.

isnine:
@JiuqingSong Could you take a look when you have time?

packages/roosterjs-content-model-plugins/lib/edit/keyboardDelete.ts
```diff
@@ -31,12 +31,14 @@ import type { DOMSelection, DeleteSelectionStep, IEditor } from 'roosterjs-conte
...
  * @param handleExpandedSelection Whether to handle expanded selection within a text node by CM
+ * @param shouldBrowserHandleBackspace True to let browser handle backspace key instead of content model
  * @returns True if the event is handled by content model, otherwise false
  */
 export function keyboardDelete(
     editor: IEditor,
     rawEvent: KeyboardEvent,
-    handleExpandedSelection: boolean = true
+    handleExpandedSelection: boolean = true,
+    shouldBrowserHandleBackspace: boolean = false
```
JiuqingSong: This introduced the following pattern:

```ts
function f(shouldUse: boolean) {
  if (shouldUse) {
     ...
  }
}
```

If we already know we won't do anything in a function, then don't call it at all. So please use the following pattern instead:

```ts
function f() {... }

if (shouldUse) {
   f();
}
```

packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
```diff
@@ -260,25 +274,29 @@ export class EditPlugin implements EditorPlugin {
         let handled = false;
         switch (rawEvent.inputType) {
             case 'deleteContentBackward':
+                const backspaceEvent = new KeyboardEvent('keydown', {
```
JiuqingSong: question (and to the old code as well): why need to create a new event here, but not using rawEvent?

packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
```diff
@@ -260,25 +274,29 @@ export class EditPlugin implements EditorPlugin {
...
-                        which: BACKSPACE_KEY,
-                    }),
-                    this.options.handleExpandedSelectionOnDelete
+                    backspaceEvent,
+                    this.options.handleExpandedSelectionOnDelete,
+                    this.shouldBrowserHandleBackspace(editor)
                 );
                 break;
             case 'deleteContentForward':
+                const deleteEvent = new KeyboardEvent('keydown', {
```
JiuqingSong: same here

isnine:
> There was a similar pr #3131, can't that one solve this issue?

Although both issues involve auto-capitalization, they differ.
#3131 fixes the offset==1 case, which works unless offset==0 triggers CM handling.
When that happens, iOS’s keyboard state becomes inconsistent, causing auto-capitalization to appear incorrectly later.

When I submitted #3131, I didn’t realize iOS keeps its own keyboard state. I’ve created this PR to let callers choose to delegate backspace handling to the browser.

packages/roosterjs-content-model-plugins/lib/edit/keyboardDelete.ts
isnine: Good point! updated.

packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
isnine: This part of the code is from PR [#2402](https://github.com/microsoft/roosterjs/pull/2402), but I don’t have enough context about the change. cc @Rain-Zheng

Rain-Zheng: This logic is Android specific. It's because some Android keyboard doesn't support `key` and `keycode`, the value of them is always `Unidentified`, so we have to manually create a new one.

JiuqingSong: Ok. @isnine since you are touching the code, can you add comment here so that other people can understand.

isnine: updated.

haven2world:
If we decide to rely on the same callback, should we expose more context in the callback's parameters? Like the range, or even the current ContentModel? I'm not sure if we can get the up-to-date CM at this moment though.

isnine:
> If we decide to rely on the same callback, should we expose more context in the callback's parameters? Like the range, or even the current ContentModel? I'm not sure if we can get the up-to-date CM at this moment though.

@haven2world We already have the range.Do you need any additional information?

In my approach, I don't plan t o determine whether to let CM handle backspace based on the current CM node. We can discuss it offline if you have different opinion.  

CleanShot 2025-09-18 at 11 19 21

packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
```diff
@@ -291,4 +308,16 @@ export class EditPlugin implements EditorPlugin {
             this.selectionAfterDelete = editor.getDOMSelection();
         }
     }
-}
+
+    private shouldBrowserHandleBackspace(editor: IEditor): boolean {
+        const opt = this.options.shouldHandleBackspaceKey;
+        switch (typeof opt) {
+            case 'function':
+                return Boolean(opt(editor));
```
JiuqingSong: Why need `Boolean` here? function opt already returns boolean

isnine: Updated.

PR #3156, opened 2025-09-16
Touch Selection - handle double tab

vinguyen12:
**Spec:**
Double tap on a word: Highlight the word or closest word if the user tapped on a space in between 2 words and open the Floatie Ribbon **(-> browser handles this)**
- note: if the user double taps on the open space after a word and there is no word on the right-side of the same, then highlight the first space of the wide gap
- note: if a user double taps on a character like a comma, period, colon, or semi-colon, then highlight that character
- note: if a user double taps on a bracket [,{,(,),},] that is next to a word, then highlight the word and not the bracket **(-> browser handles this)**

**Changes:**
 - Fix:  
      + reset `this.pointerEvent` to be `null `after trigger plugin event
      + use `setTimeout` to delay plugin event triggered for 200s to wait for new selection to be updated properly before defining the reposition (and also wait for `dblclick` event to check if it is double tab or single tab).
 - Listen to `dblclick` native event to trigger `pointerDoubleClick` if there is pointer event stored
 - Add new `pointerDoubleClick` plugin event and add handler in Touch Plugin
 - Add handler for 2 scenarios:
     + double clicked character is a punctuation mark: select that char only
     + double clicked character is a white space: check if right side has word, if yes, let browser handle it; if no, traverse to left and select the first white space char of the selected open space

### DIFF
--- a/packages/roosterjs-content-model-core/lib/corePlugin/domEvent/DOMEventPlugin.ts
+++ b/packages/roosterjs-content-model-core/lib/corePlugin/domEvent/DOMEventPlugin.ts
@@ -17,7 +17,7 @@ const EventTypeMap: Record<string, 'keyDown' | 'keyUp' | 'keyPress'> = {
     keyup: 'keyUp',
     keypress: 'keyPress',
 };
-const DELAY_UPDATE_TIME = 150;
+const POINTER_DETECTION_DELAY = 200; // Delay time to wait for selection to be updated and also detect if pointerup is a tap or part of double tap
 
 /**
  * DOMEventPlugin handles customized DOM events, including:
@@ -253,7 +253,7 @@ class DOMEventPlugin implements PluginWithState<DOMEventPluginState> {
                     }
                     this.pointerEvent = null;
                     this.isDblClicked = false;
-                }, DELAY_UPDATE_TIME);
+                }, POINTER_DETECTION_DELAY);
             }
         }
     };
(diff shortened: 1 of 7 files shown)


packages/roosterjs-content-model-core/lib/corePlugin/domEvent/DOMEventPlugin.ts
```diff
@@ -220,13 +224,40 @@ class DOMEventPlugin implements PluginWithState<DOMEventPluginState> {
...
+
+                if (!window) {
+                    return;
+                }
+
+                if (this.timer) {
+                    window.clearTimeout(this.timer);
+                }
+
+                this.timer = window.setTimeout(() => {
```
juliaroldi, line 241: Should we also clear this timer after its execution?

vinguyen12: Since the timer is already fired after execution (browser automatically cleans up the timer internally after execution), I think setting `this.timer = 0` is sufficient to clean up our reference. Please let me know if you have further suggestions, Julia.

juliaroldi: I would clean up in case something prevents the browser garbage collection, at least on dispose function.

packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
```diff
@@ -46,6 +46,57 @@ export class TouchPlugin implements EditorPlugin {
...
+                    }
+                } else if (/\s/.test(char)) {
+                    // If the clicked character is an open space with no word of right side
+                    const rightSideOfChar = text.substring(offset, text.length);
+                    const isRightSideAllSpaces =
+                        rightSideOfChar.length > 0 && !/\S/.test(rightSideOfChar);
+                    if (isRightSideAllSpaces) {
+                        // select the first space only
+                        let start = offset - 1;
+                        while (start > 0 && /\s/.test(text.charAt(start - 1))) {
```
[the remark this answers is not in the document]
vinguyen12: Not applicable

juliaroldi (approved):

PR #3157, opened 2025-09-17
Touch Plugin - Handle the entire flow of touch selection (not rely on browser)

vinguyen12:
If we let browser define selection first, the cursor will be placed in the touched position then being moved by our touch plugin handler. This would cause cursor flashing and moving.

Therefore, to avoid that issue, we have to handle the entire flow of touch selection but not rely on browser.

Changes in PR:
 - No longer using `adjustWordSelection` since it is relied on selection after browser moving the cursor
 - Trigger plugin event in `mouseDown` handler instead of `mouseUp` handler
 - move `setTimeout` to recognize double click and prevent default event behavior inside Touch Plugin -> to avoid impact other scenarios
 - double click event is triggered by touch but attached with a `MouseEvent` so better define it is in `MouseEvent` instead of `PointerEvent`
 -  reuse `getNodePositionFromEvent` to get the node at `x`, `y `value provided by event
 - to find the starting and ending of word at selection, traverse to left until get to a white space, and traverse to right until get to a white space.

Previous PRs:
 - [Add selection handler for pointer down event #3148](https://github.com/microsoft/roosterjs/pull/3148)
 - [Add TouchPlugin to handle text selection by touch #3151](https://github.com/microsoft/roosterjs/pull/3151)
 - [Touch Selection - handle double tab #3156](https://github.com/microsoft/roosterjs/pull/3156)


packages/roosterjs-content-model-plugins/lib/utils/getNodePositionFromEvent.ts
```diff
@@ -0,0 +1,37 @@
...
+    if ('caretPositionFromPoint' in doc) {
+        // Firefox, Chrome, Edge, Safari, Opera
+        const pos = (doc as any).caretPositionFromPoint(x, y);
+        if (pos && domHelper.isNodeInEditor(pos.offsetNode)) {
+            return { node: pos.offsetNode, offset: pos.offset };
+        }
+    }
+
+    if (doc.caretRangeFromPoint) {
+        // Safari
```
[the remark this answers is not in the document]
vinguyen12: image

packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
```diff
@@ -25,13 +34,19 @@ export class TouchPlugin implements EditorPlugin {
...
     }
 
     /**
      * Dispose this plugin
      */
     dispose() {
         this.editor = null;
+
+        if (this.timer) {
+            document?.defaultView?.clearTimeout(this.timer);
```
BryanValverdeU: May need to use the same document as the editor:
```
this.editor.getDocument()
```

packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
```diff
@@ -43,60 +58,183 @@ export class TouchPlugin implements EditorPlugin {
...
+            case 'pointerDown':
+                this.isDblClicked = false;
+                this.isTouchPenPointerEvent = true;
+                event.originalEvent.preventDefault();
 
-                const node = selection.focusNode;
-                if (node?.nodeType !== Node.TEXT_NODE) {
-                    return;
+                if (this.timer) {
+                    window.clearTimeout(this.timer);
```
BryanValverdeU: Need to use the editor window here too: this.editor.getDocument()?.defaultView || window

packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
```diff
@@ -43,60 +58,183 @@ export class TouchPlugin implements EditorPlugin {
...
-                        // select the first space only
-                        let start = offset;
-                        while (start > 0 && /\s/.test(text.charAt(start - 1))) {
-                            start--;
-                        }
-                        const newRange = this.editor.getDocument()?.createRange();
-                        if (newRange) {
-                            newRange.setStart(node, start);
-                            newRange.setEnd(node, start + 1);
+                this.timer = window.setTimeout(() => {
```
BryanValverdeU: Same here

packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
```diff
@@ -43,60 +58,183 @@ export class TouchPlugin implements EditorPlugin {
...
+                                    leftCursorWordLength >= rightCursorWordLength
+                                        ? rightCursorWordLength
+                                        : -leftCursorWordLength;
+                                movingOffset =
+                                    Math.abs(movingOffset) > MAX_TOUCH_MOVE_DISTANCE
+                                        ? 0
+                                        : movingOffset;
+                                if (movingOffset !== 0) {
+                                    const newRange = this.editor.getDocument().createRange();
+                                    newRange.setStart(node, offset + movingOffset);
```
BryanValverdeU: Consider adding a unit test, if the offset is not valid, this could throw an error.

vinguyen12: thanks Bryan! great point. I added a check in the if statement for the offset for now. I will add unit test (possibly in different PR).

BryanValverdeU (approved):

packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
```diff
@@ -43,60 +57,193 @@ export class TouchPlugin implements EditorPlugin {
...
+
+                        if (caretPosition) {
+                            const { node, offset } = caretPosition;
+
+                            const nodeTextContent = node.textContent || '';
+                            const charAtSelection = nodeTextContent[offset];
+                            if (
+                                node.nodeType === Node.TEXT_NODE &&
+                                charAtSelection &&
+                                WORD_MATCHING_REGEX.test(charAtSelection)
```
juliaroldi: Why we do this verification?

vinguyen12: we want to make sure it is a text node, and the selection is character, not a space or other elements?

vinguyen12: actually I think we can change this to check  node.nodeType === Node.TEXT_NODE && not spacing characters

vinguyen12: I changed this to check text not space and not punctuation instead

```
node.nodeType === Node.TEXT_NODE &&
                                charAtSelection &&
                                !SPACE_MATCHING_REGEX.test(charAtSelection) &&
                                !PUNCTUATION_MATCHING_REGEX.test(charAtSelection)
```

BryanValverdeU (approved):

PR #3158, opened 2025-09-18
Revert "Fix auto-capitalization for iOS "

isnine:
Reverts Microsoft/roosterjs#3131

https://github.com/microsoft/roosterjs/pull/3154#issuecomment-3303682909

### DIFF
--- a/packages/roosterjs-content-model-core/lib/editor/core/createEditorCore.ts
+++ b/packages/roosterjs-content-model-core/lib/editor/core/createEditorCore.ts
@@ -71,7 +71,6 @@ function createEditorEnvironment(
         modelToDomSettings: createModelToDomSettings(options),
         isMac: appVersion.indexOf('Mac') != -1,
         isAndroid: /android/i.test(userAgent),
-        isIOS: /iPad|iPhone/.test(userAgent),
         isSafari:
             userAgent.indexOf('Safari') >= 0 &&
             userAgent.indexOf('Chrome') < 0 &&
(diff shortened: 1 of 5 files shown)


packages/roosterjs-content-model-types/lib/parameter/EditorEnvironment.ts
```diff
@@ -39,11 +39,6 @@ export interface EditorEnvironment {
      */
     readonly isAndroid?: boolean;
 
-    /**
-     * Whether editor is running on iOS
-     */
-    readonly isIOS?: boolean;
```
JiuqingSong, line 45: Because this property is already checked in and already published a official version, we need to keep it here (as well as the code used for calculate its value), otherwise if some one is using this property, this change will break their code.

packages/roosterjs-content-model-core/lib/editor/core/createEditorCore.ts
```diff
@@ -71,7 +71,6 @@ function createEditorEnvironment(
         modelToDomSettings: createModelToDomSettings(options),
         isMac: appVersion.indexOf('Mac') != -1,
         isAndroid: /android/i.test(userAgent),
-        isIOS: /iPad|iPhone/.test(userAgent),
```
JiuqingSong, line 74: we need to keep this.

isnine:
Sure, I will keep it

isnine:
I will include the changes in the PR #3154

PR #3162, opened 2025-09-26
[AutoFormat] Handle input exclusively

juliaroldi:
After inputting a link URL or characters that triggers a list, handle the input event exclusively to avoid auto capitalization, when AutoCorrect Plugin is enabled.

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
@@ -2,6 +2,8 @@ import { ChangeSource } from 'roosterjs-content-model-dom';
 import { checkAndInsertHorizontalLine } from './horizontalLine/checkAndInsertHorizontalLine';
 import { createLink } from './link/createLink';
 import { formatTextSegmentBeforeSelectionMarker, promoteLink } from 'roosterjs-content-model-api';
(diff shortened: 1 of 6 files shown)


packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
```diff
@@ -100,6 +102,36 @@ export class AutoFormatPlugin implements EditorPlugin {
...
+            selection.type === 'range' &&
+            selection.range.collapsed &&
+            rawEvent.data == ' '
+        ) {
+            const previousText = selection.range.startContainer.textContent?.trim();
+            if (!previousText) {
+                return false;
+            }
+
+            return isLastWordUrl(previousText) || isLetterFollowedByMarker(previousText);
```
JiuqingSong: Ideally, we should have a same check when checking exclusiveness and really handle the event, rather than having separate functions. 

But if that is too hard to refactor, or have bad perf impact, then it is ok to keep them separate. 

Can you double check?

juliaroldi: I changed the code to use the getListTypeStyle and promoteLink, if these functions do have return, we should handle the event exclusively.

JiuqingSong (approved):

PR #3163, opened 2025-09-29
Patch fix to 9.37.1

vinguyen12:
Include this change in release, https://github.com/microsoft/roosterjs/pull/3159

### DIFF
--- a/versions.json
+++ b/versions.json
@@ -1,6 +1,6 @@
 {
     "react": "9.0.2",
-    "main": "9.37.1",
+    "main": "9.37.0",
     "legacyAdapter": "8.65.0",
-    "overrides": {}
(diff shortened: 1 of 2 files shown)


versions.json
```diff
@@ -1,6 +1,6 @@
 {
     "react": "9.0.2",
-    "main": "9.37.0",
+    "main": "9.37.1",
```
JiuqingSong: since this only modified the core plugin (roosterjs-content-model-core), let's add a override  in the overrides section below:

"overrides": {
   "roosterjs-content-model-core": "9.37.1"
}

vinguyen12: thanks Jason. I fixed this

PR #3169, opened 2025-10-02
Fix touch issue when re-clicking inside editor

vinguyen12:
### Touch Plugin Issue:
 - when clicking outside of editor and then clicking back inside editor **at the same previous spot**, focus not to be placed back to editor.
 - **Explain root cause of issue:** when clicking back inside editor, `setDOMSelection` is called and `getDOMSelection` returns the previous spot which causes `exisistingSelection` and selection are at the same spot, `setDOMSelection` will return right away and not continue to move focus part. 
image

 - **Why it doesn't repro on mouse click but touch click?**: for mouse click `getDOMSelection` return `null ` which makes it not falling into the above if statement. 

 - **Why `getDOMSelection`  returns null for mouse but not for touch?: in** `getDOMSelection`, `core.domHelper.hasFocus()` is returning different for touch and mouse. Since touch we haven't let browser move focus and active element not changed, `core.domHelper.hasFocus() `returns false which leads to we're not getting new selection but instead using the old selection. 
image

### Solution:
- Instead of taking all control of selection, we let browser define the selection first and only move cursor when needed. We followed this at first but there was issue that cursor is flashing and moving because browser places cursor first and we move cursor after. The fix for this would be hiding the cursor during the move and display it only after the move happened.


packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
```diff
@@ -69,70 +71,14 @@ export class TouchPlugin implements EditorPlugin {
...
                     this.timer = 0;
 
-                    if (!this.isDblClicked && this.editor) {
-                        const caretPosition = getNodePositionFromEvent(
-                            this.editor,
-                            event.rawEvent.x,
-                            event.rawEvent.y
-                        );
+                    if (this.editor) {
+                        if (!this.isDblClicked) {
```
JiuqingSong, line 86: I remember in MouseEvent there is a property to tell you how many times it is clicked as a single event which can be used by detecting double click, triple click, ... Not sure if that also works in PointerEvent, can you double check?

packages/roosterjs-content-model-plugins/lib/touch/repositionTouchSelection.ts
```diff
@@ -0,0 +1,84 @@
...
+            const isCollapsedSelection =
+                segmentAndParagraphs.length >= 1 &&
+                segmentAndParagraphs.every(x => x[0].segmentType == 'SelectionMarker');
+
+            // 1. adjust selection to a word if selection is collapsed
+            if (isCollapsedSelection) {
+                const para = segmentAndParagraphs[0][1];
+                const path = segmentAndParagraphs[0][2];
+
+                segmentAndParagraphs = adjustWordSelection(
```
JiuqingSong: After this line, it is possible the content model is already changed (by splitting existing segments). So you can just directly write it back to DOM by returning true, then no need to calculate DOMSelection

packages/roosterjs-content-model-plugins/lib/touch/repositionTouchSelection.ts
```diff
@@ -0,0 +1,84 @@
...
+
+const MAX_TOUCH_MOVE_DISTANCE = 6; // the max number of offsets for the touch selection to move
+
+/**
+ * @internal
+ * Touch selection, if tap within 6 characters of the beginning/end. Find the nearest edge of the word and move cursor there.
+ */
+export function repositionTouchSelection(editor: IEditor) {
+    editor.formatContentModel(
+        (model, _context) => {
```
JiuqingSong: You can extract this part as a function and call formatContentModel from the plugin, so it will be easier to add unit test to this function.

packages/roosterjs-content-model-plugins/lib/touch/repositionTouchSelection.ts
vinguyen12: @JiuqingSong how can we directly write the new range back to DOM? Can you explain it more? (I tried with return true after this line, but it didn't work, probably need to define new selection?)

JiuqingSong: If adjustWordSelection already modified the model, returning true here should work. If not, ping me and we can debug it

packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
```diff
@@ -219,6 +172,83 @@ export class TouchPlugin implements EditorPlugin {
                 break;
         }
     }
+
+    repositionTouchSelection = () => {
```
JiuqingSong: Actually, I mean it is ok to have a file for `repositionTouchSelection`, but in this function you don't need to call formatContentModel. Instead, you can accept a content model directly. Then in the plugin you can call formatContentModel in which call this new functioin. For example:

```ts
// plugin
...
this.editor.formatContentModel(model => repositionTouchSelection());

// repositionTouchSelection.ts
export function repositionTouchSelection(model: ReadonlyContentModelDocument) {
 ...
}
```

packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
vinguyen12: yes, but pointer down is also triggered. The order I'm observing is pointerdown then dbclick. But I think we can't tell if dbl click is triggered by touch. Here is the wiki for double click event, https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event.

JiuqingSong: I see. Ok.

packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
vinguyen12: Thanks Jason. I applied this change.

packages/roosterjs-content-model-plugins/lib/touch/repositionTouchSelection.ts
vinguyen12: discussed offline with Jason. Since we want to move cursor to beginning/end of the word, we will keep the current implementation

packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
```diff
@@ -219,6 +174,76 @@ export class TouchPlugin implements EditorPlugin {
...
+                        this.editor.setDOMSelection({
+                            type: 'range',
+                            range: newRange,
+                            isReverted: false,
+                        });
+                    }
+                }
+            }
+        }
+        return false;
```
JiuqingSong: it is still returning false here, and you are still doing DOM selection, this will cause the cached content model mismatch with DOM tree.

vinguyen12: discussed offline and decide to always return true

packages/roosterjs-content-model-plugins/lib/touch/TouchPlugin.ts
```diff
@@ -219,6 +177,89 @@ export class TouchPlugin implements EditorPlugin {
                 break;
         }
     }
+
+    repositionTouchSelection = (model: ReadonlyContentModelDocument) => {
+        if (this.editor) {
```
JiuqingSong: nit: this.editor is not used in this function, so no need to check it

JiuqingSong (approved):

JiuqingSong (approved):

PR #3173, opened 2025-10-06
Add callback for Table CellResizer

vinguyen12:
Add callback for Table CellResizer similar to other TableEditFeature.

With this, client will have more control over how they can define the resizer component and action.


JiuqingSong:
question, what do we want to do with this callback?

vinguyen12:
> question, what do we want to do with this callback?

This callback gives us access to the CellResizer handler component, so we can use it for applying the wanted z-index there (similar to what we're doing for other components of TableResizer and TableInserter)

JiuqingSong (approved):

PR #3179, opened 2025-10-14
Add `test:fast`  for faster unit test runs

BryanValverdeU:
test:fast runs all the unit tests once
test:fast:debug will keep running the test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,10 +27,9 @@
         "test": "node tools/build.js normalize & karma start --chrome",
         "test:chrome": "node tools/build.js normalize & karma start --chrome",
         "test:firefox": "node tools/build.js normalize & karma start --firefox",
-        "test:debug": "node tools/build.js normalize & karma start --no-single-run --chrome",
+        "test:debug": "node tools/build.js normalize & karma start karma.fast.conf.js --no-single-run --chrome",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -28,6 +28,8 @@
         "test:chrome": "node tools/build.js normalize & karma start --chrome",
         "test:firefox": "node tools/build.js normalize & karma start --firefox",
         "test:debug": "node tools/build.js normalize & karma start --no-single-run --chrome",
+        "test:fast": "node tools/build.js normalize & karma start karma.fast.conf.js --chrome",
```
JiuqingSong, line 31: I think for all test commands, we only need to keep the following:

* test, test:chrome: test using chrome with original config so it will do type check
* test:firefox: test using firefox with original config so it will do type check
* test:debug: test using chrome with new config, skip type check  **(Modified)**
* test:debug-firefox: test using firefox with new config, skip type check  **(Modified)**
* test:coverage: test and add coverage report

So no no need to introduce new command
When test with debugging, always using fast config
The original test/test:chrome/test:firefox command is not used very much.

BryanValverdeU: Sounds good, updated the old debug commands instead of using the test:debug:fast. 

Could we keep test:fast? I find this useful to ask Copilot to run tests and fix them in case of a change that happened.

JiuqingSong (approved):

PR #3180, opened 2025-10-15
Add support for retaining formats when splitting paragraphs on Enter/Delete/Backspace keypress

BryanValverdeU:
Pressing the Enter key within an element that includes elementToProof was causing the element to lose its associated class in OWA which caused Spelling issues. 
Further investigation revealed that the splitParagraph function was only preserving BlockFormats, while additional formats introduced via custom handlers and parsers were being lost during the split operation.
This PR resolves the issue by introducing a new property, formatsToKeep, which ensures that all relevant formats—including those added through customization—are retained when splitting paragraphs.

- Introduced `formatsToKeep` option in `EditOptions` to specify which formats to retain.
- Updated `keyboardEnter`, `handleEnterOnParagraph`, and `splitParagraph` functions to handle the new formats.
- Enhanced tests to verify the correct behavior of format retention during paragraph splits.

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
+++ b/packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
@@ -3,6 +3,7 @@ import { keyboardEnter } from './keyboardEnter';
 import { keyboardInput } from './keyboardInput';
 import { keyboardTab } from './keyboardTab';
 import { parseTableCells } from 'roosterjs-content-model-dom';
+import type { EditOptions } from './EditOptions';
 import type {
     DOMSelection,
     EditorPlugin,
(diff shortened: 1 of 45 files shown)


packages/roosterjs-content-model-plugins/lib/edit/EditPlugin.ts
```diff
@@ -41,6 +41,13 @@ export type EditOptions = {
      * @returns A boolean
      */
     shouldHandleBackspaceKey?: ((editor: IEditor) => boolean) | boolean;
+
+    /**
+     * An array of format keys that should be kept when pressing Enter key to split a paragraph.
+     * For example, if the formatToKeep contains 'fontFamily', when pressing Enter in a paragraph with fontFamily='Arial',
+     * the new paragraph will also have fontFamily='Arial'.
+     */
+    formatsToKeep?: string[];
```
JiuqingSong: Name it `formatToKeepWhenEnter`?

BryanValverdeU: Found that the issue is not only when pressing Enter, also with Backspace and Delete. I changed the name of the function to formatsToPreserveOnMerge, let me know if this is okay

JiuqingSong (approved):

PR #3181, opened 2025-10-20
Support rowSpan equal to 0

BryanValverdeU:
RowSpan 0 is supported by HTML:
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/td#rowspan

However looks like we are currently removing any cells with this style. This PR fixes the issue by instead of adding rowSpan=0, add rowSpan=`{# of rows below the cell with rowSpan=0}`. So mimic the same behavior as rosSpan=0 but also being handled correctly by Content Model


packages/roosterjs-content-model-dom/lib/domToModel/processors/tableProcessor.ts
```diff
@@ -200,8 +200,14 @@ export const tableProcessor: ElementProcessor<HTMLTableElement> = (
                                     colSpan <= td.colSpan;
                                     colSpan++, targetCol++
                                 ) {
-                                    for (let rowSpan = 1; rowSpan <= td.rowSpan; rowSpan++) {
-                                        const hasTd = colSpan == 1 && rowSpan == 1;
+                                    for (
+                                        let rowSpan = 1;
+                                        // RowSpan of 0 means it should span to the end of the table
+                                        // https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/td#rowspan
+                                        rowSpan <= (td.rowSpan == 0 ? 1 : td.rowSpan);
```
JiuqingSong: If we treat it as 1, how can it extends until the end of the section?

And do we need to fix colSpan?

Any unit test here?

BryanValverdeU: The write back result will keep the rowSpan as 0, so HTML will handle that. This check here is just to make sure we process the cell and it is added to the model.

For colspan 0 is invalid, so td.colSpan returns 1 when the attribute is set as 0 for Edge, but I think we can still add a check in case other browsers handle the attribute differently.

Adding unit tests.

JiuqingSong: My concern is, when rowspan is 0, how it is represented in content model? And if user format the table, can it be correctly handled?

For example, 
```html
<tr><td rowspan=0></td><td></td></tr>
<tr><td></td></tr>
```

This should be translated to:
```
{table cell              } {table cell}
{table cell spanTop=true} {table cell}
```

BryanValverdeU: ~~Got it, added a new property to table cell to represent this.~~

BryanValverdeU: rowSpan=0 can be represented as rowSpan=#of rows below the cell.

Updated the logic so we can represent the logic better in the model

packages/roosterjs-content-model-dom/lib/domToModel/processors/tableProcessor.ts
```diff
@@ -197,18 +199,29 @@ export const tableProcessor: ElementProcessor<HTMLTableElement> = (
...
+                                        rowSpan++
+                                    ) {
+                                        const hasTd = colSpan >= 1 && rowSpan == 1;
```
JiuqingSong: why change the check of colSpan to be ">=" 1?

packages/roosterjs-content-model-dom/lib/domToModel/processors/tableProcessor.ts
```diff
@@ -282,6 +299,41 @@ export const tableProcessor: ElementProcessor<HTMLTableElement> = (
...
+                const originalCell = belowRow.cells[colIndex];
+
+                // Shift all cells from colIndex onwards to the right
+                for (let shiftCol = belowRow.cells.length; shiftCol > colIndex; shiftCol--) {
+                    belowRow.cells[shiftCol] = belowRow.cells[shiftCol - 1];
+                }
+
+                // Create an empty cell with spanAbove in the original position
+                belowRow.cells[colIndex] = createTableCell(
+                    false, // spanLeft
```
JiuqingSong: I think spanLeft is possibly true, depends on if the TD has colSpan.

packages/roosterjs-content-model-dom/lib/domToModel/processors/tableProcessor.ts
```diff
@@ -282,6 +299,41 @@ export const tableProcessor: ElementProcessor<HTMLTableElement> = (
...
 };
 
+function adjustRowSpanZeroCells(table: ContentModelTable, rowSpanZeroIndexes: number[][]): void {
+    // Process all cells with rowSpan=0 from end to start to avoid conflicts when shifting cells
+    for (let i = rowSpanZeroIndexes.length - 1; i >= 0; i--) {
+        const [rowIndex, colIndex] = rowSpanZeroIndexes[i];
+
+        // This cell should span to the end of the table
+        // Set all cells below this one in the same column to spanAbove
+        for (let belowRowIndex = rowIndex + 1; belowRowIndex < table.rows.length; belowRowIndex++) {
```
JiuqingSong: According to the standard, rowSpan will extend the cell the end of current row group, so if there are THEAD/TBODY/TFOOT, or multiple TBODY, we should not let it extend to the end of table.

JiuqingSong: Maybe you just need to check how many rows are there in this group first, then treat rowSpan has the same value(minus existing rows before this row) with the row count, so it can fall in the the existing code path.

JiuqingSong:
```html
<tbody>
 <tr><td>1</td>...</tr>
 <tr><td rowspan=0>2</td>...
  <tr>...</tr>
 </tbody>
```

Since there are 3 rows in this tbody, and there is one row before, so treat rowspan as 2 here.

JiuqingSong (approved):

PR #3182, opened 2025-10-22
New getContentForCopy API

juliaroldi:
Extract the logic to create the BeforeCutCopyEvent into a new public API called triggerBeforeCutCopyEvent. It adds a new property in BeforeCutCopyEvent called pasteModel, that has the selection adjusted.

### DIFF
--- a/karma.fast.conf.js
+++ b/karma.fast.conf.js
@@ -99,7 +99,7 @@ module.exports = function (config) {
 
         webpack: {
             mode: 'development',
-            devtool: 'eval-source-map', // Faster than inline-source-map
+            devtool: 'inline-source-map', // More accurate source maps for debugging
             module: {
                 rules,
             },
--- a/packages/roosterjs-editor-adapter/lib/corePlugins/BridgePlugin.ts
+++ b/packages/roosterjs-editor-adapter/lib/corePlugins/BridgePlugin.ts
@@ -153,17 +153,28 @@ export class BridgePlugin implements ContextMenuProvider<any> {
      * @param target Target node that triggered a ContextMenu event
      * @returns An array of context menu items, or null means no items needed
      */
-    getContextMenuItems(target: Node): any[] {
+    getContextMenuItems(target: Node, event?: Event): any[] {
         const allItems: any[] = [];
 
         this.contextMenuProviders.forEach(provider => {
-            const items = provider.getContextMenuItems(target) ?? [];
-            if (items?.length > 0) {
-                if (allItems.length > 0) {
-                    allItems.push(null);
-                }
+            if (isMixedPluginProvider(provider)) {
+                const items = provider.getContextMenuItems(target, event) ?? [];
+                if (items?.length > 0) {
+                    if (allItems.length > 0) {
+                        allItems.push(null);
+                    }
 
-                allItems.push(...items);
+                    allItems.push(...items);
+                }
+            } else {
+                const items = provider.getContextMenuItems(target) ?? [];
+                if (items?.length > 0) {
+                    if (allItems.length > 0) {
+                        allItems.push(null);
+                    }
+
+                    allItems.push(...items);
+                }
             }
         });
 
@@ -218,6 +229,15 @@ export class BridgePlugin implements ContextMenuProvider<any> {
             plugin.onPluginEventV9?.(newEvent);
         }
     }
+}
+
+/**
+ * Check if a provider is a V9 context menu provider
+ * @param provider The provider to check
+ * @returns True if the provider is a V9 context menu provider, false otherwise
+ */
+function isMixedPluginProvider(provider: any): provider is ContextMenuProvider<any> {
+    return isMixedPlugin(provider);
 }
 
 /**
--- a/packages/roosterjs-editor-adapter/test/corePlugins/BridgePluginTest.ts
+++ b/packages/roosterjs-editor-adapter/test/corePlugins/BridgePluginTest.ts
@@ -255,7 +255,7 @@ describe('BridgePlugin', () => {
             mockedPlugin2,
         ]);
 
-        spyOn(eventConverter, 'newEventToOldEvent').and.callFake(newEvent => {
+        spyOn(eventConverter, 'newEventToOldEvent').and.callFake((newEvent: any) => {
             return ('NEW_' + newEvent) as any;
         });
 
@@ -277,10 +277,10 @@ describe('BridgePlugin', () => {
 
     it('onPluginEvent without exclusive handling', () => {
         const initializeSpy = jasmine.createSpy('initialize');
-        const onPluginEventSpy1 = jasmine.createSpy('onPluginEvent1').and.callFake(event => {
+        const onPluginEventSpy1 = jasmine.createSpy('onPluginEvent1').and.callFake((event: any) => {
             event.data = 'plugin1';
         });
-        const onPluginEventSpy2 = jasmine.createSpy('onPluginEvent2').and.callFake(event => {
+        const onPluginEventSpy2 = jasmine.createSpy('onPluginEvent2').and.callFake((event: any) => {
             event.data = 'plugin2';
         });
         const disposeSpy = jasmine.createSpy('dispose');
@@ -305,7 +305,7 @@ describe('BridgePlugin', () => {
             mockedPlugin2,
         ]);
 
-        spyOn(eventConverter, 'newEventToOldEvent').and.callFake(newEvent => {
+        spyOn(eventConverter, 'newEventToOldEvent').and.callFake((newEvent: any) => {
             return {
                 eventType: 'old_' + newEvent.eventType,
             } as any;
@@ -390,7 +390,7 @@ describe('BridgePlugin', () => {
             mockedPlugin2,
         ]);
 
-        spyOn(eventConverter, 'newEventToOldEvent').and.callFake(newEvent => {
+        spyOn(eventConverter, 'newEventToOldEvent').and.callFake((newEvent: any) => {
             return {
                 eventType: 'old_' + newEvent.eventType,
                 eventDataCache: newEvent.eventDataCache,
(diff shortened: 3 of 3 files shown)


packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/utils/createCopyRange.ts
```diff
@@ -0,0 +1,90 @@
...
+ * @returns
+ */
+export function createCopyRange(
+    editor: IEditor,
+    pasteModel: ContentModelDocument,
+    selection: DOMSelection,
+    tempDiv: HTMLDivElement
+): Range | null {
+    pruneUnselectedModel(pasteModel);
+    const doc = editor.getDocument();
```
JiuqingSong: since you only need document from editor, pass in document instead

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/utils/createCopyRange.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    if (selection.type === 'table') {
```
JiuqingSong: you only need selection type here, so no need to pass in the whole selection object.

And actually, can we get selection type from model? As a public api, we need to make sure all combination of parameters are valid. If user passes in a selection that does not match the selection in the model, what will happen? So better get selection from model instead.

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/utils/createCopyRange.ts
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Create the range for the "beforeCutCopy" event
+ * @param editor The editor object
+ * @param pasteModel The model of the selection
+ * @param selection The editor selection
+ * @param tempDiv A temporary DIV element used for cut/copy content
+ * @returns
+ */
+export function createCopyRange(
```
JiuqingSong: And as a public API, better put it under a public place, but not corePlugin. You can put it under command folder

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/utils/createCopyRange.ts
```diff
@@ -0,0 +1,90 @@
...
+    editor: IEditor,
+    pasteModel: ContentModelDocument,
+    selection: DOMSelection,
+    tempDiv: HTMLDivElement
```
JiuqingSong: Better also hide the code of creating tempDiv into this function, so caller does not need to know how to pass it in.

We need to make sure each parameter of a public API is easy to understand

JiuqingSong: BTW,where is the code to trigger beforeCutCopy event? Do you want to trigger it manually from your code?

JiuqingSong:
As a public API, it is ok to only take editor as parameter, no need to know other things, because you can get everything from editor. Then it only needs to return HTML and plain text, then let the caller decide how to use it. For example, in the plugin it will put them into clipboard, and in your scenario you will save them as quick part.

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
```diff
@@ -111,92 +92,42 @@ class CopyPastePlugin implements PluginWithState<CopyPastePluginState> {
...
-                }
-
-                doc.defaultView?.requestAnimationFrame(() => {
-                    if (!this.editor) {
-                        return;
-                    }
-
-                    cleanUpAndRestoreSelection(tempDiv);
-                    this.editor.setDOMSelection(selection);
-                    this.editor.focus();
```
juliaroldi, line 177: @JiuqingSong can we remove this?

BryanValverdeU:
Please make sure that Copy and Cut buttons in OWA Ribbon and Context menu keep working with this change.

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
JiuqingSong: I guess we can. Please test.

packages/roosterjs-content-model-types/lib/event/BeforeCutCopyEvent.ts
```diff
@@ -18,4 +19,9 @@ export interface BeforeCutCopyEvent extends BasePluginDomEvent<'beforeCutCopy',
      * Whether this is a cut event
      */
     isCut: boolean;
+
+    /**
+     * The model adjusted by beforeCutCopy event
+     */
+    pasteModel?: ContentModelDocument;
```
JiuqingSong: Is this necessary?

I was thinking of adding this in event. But I finally didn't. This is because in the event we already have `clonedRoot` which contains the DOM elements, and plugin can modify it to change the content to copy. However, when add a model here, plugin developer can be confused that which one I should modify, and why modifying a model won't work.

If we just want to provide a model for plugin to know what is being cut/copy, please add comment that this is not expected to be modified, and change its type to be ReadonlyContentModelDocument.

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
```diff
@@ -111,92 +92,42 @@ class CopyPastePlugin implements PluginWithState<CopyPastePluginState> {
...
-        const doc = this.editor.getDocument();
-        const selection = this.editor.getDOMSelection();
-
-        adjustImageSelectionOnSafari(this.editor, selection);
-
-        if (selection && (selection.type != 'range' || !selection.range.collapsed)) {
-            const pasteModel = this.editor.getContentModelCopy('disconnected');
-
-            pruneUnselectedModel(pasteModel);
+        const beforeCutCopyEvent = triggerBeforeCutCopyEvent(
```
JiuqingSong: I prefer to let this function return html and plain text only, but no the event object. Since you actually only need them

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
```diff
@@ -111,92 +92,42 @@ class CopyPastePlugin implements PluginWithState<CopyPastePluginState> {
...
-            } else if (selection.type === 'range') {
-                adjustSelectionForCopyCut(pasteModel);
+            if (isClipboardEvent(event)) {
+                event.preventDefault();
+                const text = pasteModel ? contentModelToText(pasteModel) : '';
+                event.clipboardData?.setData('text/html', clonedRoot.innerHTML);
+                event.clipboardData?.setData('text/plain', text);
+            } else if (range) {
+                const doc = this.editor.getDocument();
+                addRangeToSelection(doc, range);
```
JiuqingSong: we now do not change focus, so no need to set selection back. If event is not clipboard event (actually t his should never happen), just do nothing

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
```diff
@@ -111,92 +92,42 @@ class CopyPastePlugin implements PluginWithState<CopyPastePluginState> {
...
 
-                        return true;
-                    }
-                    return false;
-                });
-            } else if (selection.type === 'range') {
-                adjustSelectionForCopyCut(pasteModel);
+            if (isClipboardEvent(event)) {
+                event.preventDefault();
+                const text = pasteModel ? contentModelToText(pasteModel) : '';
```
JiuqingSong: this can be done inside the your new function, and return the text only

packages/roosterjs-content-model-core/lib/command/cutCopy/preprocessTable.ts
```diff
@@ -0,0 +1,23 @@
+import { getSelectedCells } from 'roosterjs-content-model-dom';
+import type { ContentModelTable } from 'roosterjs-content-model-types';
+
+/**
+ * @internal
+ */
+export function preprocessTable(table: ContentModelTable) {
```
JiuqingSong, line 7: Since this function is only used by your new function, move it into the same folder with the new function

packages/roosterjs-content-model-core/lib/command/cutCopy/adjustSelectionForCopyCut.ts
```diff
@@ -0,0 +1,34 @@
...
+    ContentModelDocument,
+    ContentModelParagraph,
+    ContentModelSegment,
+    ReadonlyTableSelectionContext,
+} from 'roosterjs-content-model-types';
+
+/**
+ * @internal
+ */
+export function adjustSelectionForCopyCut(pasteModel: ContentModelDocument) {
```
JiuqingSong, line 12: same here

packages/roosterjs-content-model-core/lib/command/cutCopy/triggerBeforeCutCopyEvent.ts
```diff
@@ -0,0 +1,105 @@
...
+ * Trigger the "beforeCutCopy" event
+ * @param editor The editor object
+ * @param isCut If the event to be created is cut. @default false
+ * @param event an optional parameter for when the clipboard event needs to be trigger
+ * @returns
+ */
+export function triggerBeforeCutCopyEvent(
+    editor: IEditor,
+    isCut: boolean = false,
+    event?: ClipboardEvent
```
JiuqingSong: I rethink about this, I feel we should make this event parameter required. So the code to create this object should be outside of this function.

packages/roosterjs-content-model-core/lib/command/cutCopy/triggerBeforeCutCopyEvent.ts
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Trigger the "beforeCutCopy" event
+ * @param editor The editor object
+ * @param isCut If the event to be created is cut. @default false
+ * @param event an optional parameter for when the clipboard event needs to be trigger
+ * @returns
+ */
+export function triggerBeforeCutCopyEvent(
+    editor: IEditor,
+    isCut: boolean = false,
```
JiuqingSong: Make it required as well

packages/roosterjs-content-model-core/lib/command/cutCopy/triggerBeforeCutCopyEvent.ts
```diff
@@ -0,0 +1,105 @@
...
+};
+
+/**
+ * Trigger the "beforeCutCopy" event
+ * @param editor The editor object
+ * @param isCut If the event to be created is cut. @default false
+ * @param event an optional parameter for when the clipboard event needs to be trigger
+ * @returns
+ */
+export function triggerBeforeCutCopyEvent(
```
JiuqingSong: Since this function is mainly used for getting the copy/cut content, triggering event is just a step of it, let name it getContentForCopy.

Since cutting the content is not done in this function, so no need to mention "Cut" in the function name.

packages/roosterjs-content-model-core/lib/command/cutCopy/getContentForCopy.ts
```diff
@@ -0,0 +1,113 @@
...
+    isElementOfType,
+    isNodeOfType,
+    iterateSelections,
+    wrap,
+} from 'roosterjs-content-model-dom';
+
+/**
+ * The html and text content for the copy action
+ */
+export interface CopyContent {
```
JiuqingSong: 1. Move this type to `roosterjs-content-model-types` package
2. Have a better name. For example, ContentForCopy, or TextAndHtmlContent. The name should be like a noun, but not a verb
3. Add comment to each member

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
```diff
@@ -111,92 +90,34 @@ class CopyPastePlugin implements PluginWithState<CopyPastePluginState> {
...
-        const doc = this.editor.getDocument();
-        const selection = this.editor.getDOMSelection();
-
-        adjustImageSelectionOnSafari(this.editor, selection);
-
-        if (selection && (selection.type != 'range' || !selection.range.collapsed)) {
-            const pasteModel = this.editor.getContentModelCopy('disconnected');
-
-            pruneUnselectedModel(pasteModel);
+        const beforeCutCopyEvent = getContentForCopy(this.editor, isCut, event as ClipboardEvent);
```
JiuqingSong: Now it does not return an event, so rename this variable

packages/roosterjs-content-model-core/lib/corePlugin/copyPaste/CopyPastePlugin.ts
```diff
@@ -111,92 +90,34 @@ class CopyPastePlugin implements PluginWithState<CopyPastePluginState> {
...
+        if (beforeCutCopyEvent) {
+            const { htmlContent, textContent } = beforeCutCopyEvent;
 
-                        return true;
-                    }
-                    return false;
-                });
-            } else if (selection.type === 'range') {
-                adjustSelectionForCopyCut(pasteModel);
+            if (isClipboardEvent(event)) {
```
JiuqingSong: We are already using event as a ClipboardEvent, so let's move this check before line 93, to make it safer.

Then no type cast is needed at line 93

packages/roosterjs-content-model-types/lib/event/BeforeCutCopyEvent.ts
JiuqingSong: I don't think you still need this

PR #3188, opened 2025-10-30
Support passing event to getContextMenuItems when plugin is a MixedPlugin

BryanValverdeU:
In previous change I added support to pass in event to getContextMenuItems, now I am also adding the support to the BridgePlugin when the plugin is a MixedPlugin and getContextMenuItems needs 2 parameters.

### DIFF
--- a/packages/roosterjs-editor-adapter/lib/corePlugins/BridgePlugin.ts
+++ b/packages/roosterjs-editor-adapter/lib/corePlugins/BridgePlugin.ts
@@ -153,17 +153,28 @@ export class BridgePlugin implements ContextMenuProvider<any> {
      * @param target Target node that triggered a ContextMenu event
      * @returns An array of context menu items, or null means no items needed
      */
-    getContextMenuItems(target: Node): any[] {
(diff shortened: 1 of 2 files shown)


packages/roosterjs-editor-adapter/lib/corePlugins/BridgePlugin.ts
```diff
@@ -220,6 +231,10 @@ export class BridgePlugin implements ContextMenuProvider<any> {
     }
 }
 
+function isV9ContextMenuProvider(provider: any): provider is ContextMenuProvider<any> {
+    return (provider?.getContextMenuItems?.length || 0) == 2 && isMixedPlugin(provider);
```
JiuqingSong: Can you add a comment to explain what "@" means here?

BryanValverdeU: Thanks, added a comment as suggested.

JiuqingSong (approved):

PR #3189, opened 2025-10-30
Fix model to markdown

FrancisMengx:
Parsing block quote should be happen only for blockquote text or it should be parsed as default blockgroup


packages/roosterjs-content-model-markdown/lib/modelToMarkdown/creators/createMarkdownBlockGroup.ts
```diff
@@ -30,8 +30,10 @@ export function createMarkdownBlockGroup(
             }
             break;
         case 'FormatContainer':
-            markdownString += createMarkdownBlockQuote(blockGroup, newLinePattern, listCounter);
-            break;
+            if (blockGroup.tagName == 'blockquote') {
+                markdownString += createMarkdownBlockQuote(blockGroup, newLinePattern, listCounter);
+                break;
+            }
         default:
```
JiuqingSong, line 43: Although it logically works, but it can't pass lint check.

Also would you please add a test case, thanks

packages/roosterjs-content-model-markdown/lib/modelToMarkdown/creators/createMarkdownBlockGroup.ts
```diff
@@ -30,18 +30,40 @@ export function createMarkdownBlockGroup(
             }
             break;
         case 'FormatContainer':
-            markdownString += createMarkdownBlockQuote(blockGroup, newLinePattern, listCounter);
+            if (blockGroup.tagName == 'blockquote') {
+                markdownString += createMarkdownBlockQuote(blockGroup, newLinePattern, listCounter);
+            } else {
+                markdownString += createDefaultMarkdownString(
```
JiuqingSong, line 36: Why not just use existing function `createMarkdownBlockGroup` recursively?

PR #3193, opened 2025-11-03
Update legacyAdapter version to 8.65.1 in versions.json

BryanValverdeU:


juliaroldi:
Can we also update rooster-content-model-types and rooster-content-model-types? @JiuqingSong

JiuqingSong (approved):

juliaroldi:
> > Can we also update rooster-content-model-types and rooster-content-model-types? @JiuqingSong
> 
> @juliaroldi do you mean you want a patch? Which change you want to patch?
I wanted to add the getContentForCopy api, but I think it is better to bump on Friday, since we will have more time to test.

juliaroldi (approved):

PR #3195, opened 2025-11-04
feat: add announcements for table selection and unselection in the editor

BryanValverdeU:
https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/392041/

Based on the bug above, when table selection changes, we should announce the content of the selected table cells. This PR adds this functionality.


juliaroldi (approved):

packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -766,13 +769,29 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
                     { table, firstCo, lastCo, parsedTable, startNode }
                 );
 
+                if (action) {
```
JiuqingSong: I wonder instead of putting announcing code directly in core plugin, can we do it in a normal plugin by handling SelectionChanged event?

BryanValverdeU: You suggest creating a new plugin for announcing? If we can pass in the ParsedTable in the selectionChanged event, I think it is fine to do. Otherwise we would need to iterate the whole table again.

JiuqingSong: Can we put ParsedTable into table selection object? Then put the announce code in OWA side.

JiuqingSong: I'm just feeling announcing code is not that "core"

BryanValverdeU: Working on the changes.

BryanValverdeU: Just pushed the change to move the announcement to a separate plugin.

JiuqingSong (approved):

juliaroldi (approved):

PR #3196, opened 2025-11-04
Update the Entity's format when applying format on the content inside the Entity

haven2world:
When we support formatting api on entities by setting the `formattableRoots` in the `beforeFormat` event, we cannot get the updated format when we call `getFormatState`

**Expected behavior**
The format state should reflect the format correctly if we allow to change it.


haven2world:
> Change looks good, please just fix test case, and add new case for this change if need

@JiuqingSong I updated the UT and the checks all passed

PR #3197, opened 2025-11-05
feat: add option to remove transparency from Pasted Images

BryanValverdeU:
Add a new feature to remove transparency from Pasted images from Word Desktop:
removeTransparency

### DIFF
--- a/demo/scripts/controlsV2/mainPane/MainPane.tsx
+++ b/demo/scripts/controlsV2/mainPane/MainPane.tsx
@@ -519,6 +519,7 @@ export class MainPane extends React.Component<{}, MainPaneState> {
         const {
             pluginList,
             allowExcelNoBorderTable,
+            pastePluginOptions,
             listMenu,
             tableMenu,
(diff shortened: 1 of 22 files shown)


JiuqingSong:
So this solution is to edit the rendered image from IMG element, but that means image blob in content model or its related object url (if any) is not changed.

Instead of reading image from IMG element after image is loaded (using onload event) and creating a chainOnNodeCreated callback, I think we can edit the pasting DOM tree directly. And I think we only need to do this for datauri image in clipboard. For example, in beforePasteEvent, if we see img with src="data:imag/....", create blob from the data uri and edit it with canvas, then put the result back to the IMG. Then after that other plugins can still create ObjectUrl from this edited image, as if it was always like that.

Then no need to add onNodeCreated, chainOnNodeCreated, and new change source.

Also I think this can be done in OWA side as part of InlineImagePlugin. If someday we see ask from other teams for this feature, we can move it into roosterjs.

PR #3203, opened 2025-11-18
Add format when insert table

juliaroldi:
Add the format parameter in the insertTable API, then it will be possible to insert a table with a predefined format. It can be used to add margin bottom to the table, then when a table is inserted under another their border does not overlay. 
TableMargin


packages/roosterjs-content-model-api/lib/publicApi/table/insertTable.ts
```diff
@@ -35,14 +40,17 @@ export function insertTable(
             if (insertPosition) {
                 const doc = createContentModelDocument();
                 const table = createTableStructure(doc, columns, rows);
+                if (format) {
+                    table.format = format;
```
JiuqingSong: clone the format object (you can use {...format}) but do not directly assign it

PR #3208, opened 2025-11-20
Refactor paste source validation to use getDocumentSource

BryanValverdeU:
We want to be able to run the Word Desktop Processing not only for Paste. This PRs refactors the code so we can do that by:

 1. Move the DocumentSourceValidations to DOM package, so we can use it easier.
 2. Change the parameter signature of processPastedContentFromWordDesktop so we dont need to pass in the BeforePasteEvent and instead use simplified parameters.
 3. Move some utils from core to dom packages that are also required to confirm whether the document is from Word Desktop source


packages/roosterjs-content-model-plugins/lib/paste/parsers/removeNegativeTextIndentParser.ts
```diff
@@ -3,8 +3,11 @@ import type { FormatParser, TextIndentFormat } from 'roosterjs-content-model-typ
 /**
  * @internal
  */
-export const removeNegativeTextIndentParser: FormatParser<TextIndentFormat> = format => {
+export const removeNegativeTextIndentParser: FormatParser<TextIndentFormat> = (format, element) => {
     if (format.textIndent?.startsWith('-')) {
         delete format.textIndent;
     }
+    if (element.style.textIndent.startsWith('-')) {
```
JiuqingSong: Normally we don't modify the element in parsers. Any reason we want to remove the style from element as well?

BryanValverdeU: Due to ordering of the parsers, if we have 2 parsers, like in the order below:
1. Remove negative text indent.
2. Add text indent

The parser to remove negative text indent will be no-op and the second could add the negative text indent.

BryanValverdeU: I decided to remove that change from the current PR and I will be adding it later in another PR.

packages/roosterjs-content-model-plugins/lib/paste/parsers/removeNegativeTextIndentParser.ts
```diff
@@ -3,7 +3,7 @@ import type { FormatParser, TextIndentFormat } from 'roosterjs-content-model-typ
 /**
  * @internal
  */
-export const removeNegativeTextIndentParser: FormatParser<TextIndentFormat> = format => {
+export const removeNegativeTextIndentParser: FormatParser<TextIndentFormat> = (format, element) => {
```
JiuqingSong, line 6: minor: since element is not used, no need to add it

JiuqingSong (approved):

PR #3213, opened 2025-11-24
Use getPromoteLink  to verify if a link can be promoted

juliaroldi:
Add getPromoteLink function to verify if a link can be promoted. 
LinkOnSpace


packages/roosterjs-content-model-plugins/lib/autoFormat/AutoFormatPlugin.ts
```diff
@@ -112,17 +112,29 @@ export class AutoFormatPlugin implements EditorPlugin {
...
+                        autoNumbering,
+                    } = this.options;
+                    const list = getListTypeStyle(model, autoBullet, autoNumbering);
+                    const link = promoteLink(previousSegment, paragraph, {
+                        autoLink,
+                        autoTel,
+                        autoMailto,
+                    });
+                    shouldHandle = !!link || !!list;
+                    if (shouldHandle) {
```
JiuqingSong: Please add comment to explain why need to clear cache here, why cache becomes invalid after this call

JiuqingSong: Actually, it seems function `promoteLink` modifies the model but it did not write back.

Can we just create a function to check if it can promote link, but do not modify model? So we don't need to clear cache.

Issue #3214, opened 2025-11-25
Page jumps to top when inserting a column or row via "+" button


liangdrime:
While inserting a column or row by clicking the table’s **"+"** button, the page jumps to the top if the content is long and the table is near the bottom.

![Image](https://github.com/user-attachments/assets/cd5b07f2-ae47-42c9-910e-7079e925e551)

**Steps to reproduce the behavior**:
1. Insert a table at the bottom of a long document on a Mac.
2. Click inside the table, then hit the **"+"** button to add a column or row.
3. Check what happens next.

**Expected behavior**
The page should stay in the same position, without jumping around.

**Device Information**
 - OS: macOS

**Additional context**

On Mac (WebKit), clicking a button element causes the web view to blur the current editable area, which means the editor loses focus.  

When the **"+"** button is clicked, it calls `formatTableContentModel`, which then triggers `editor.focus()`. However, in `focus.ts`, the cached `selection.selection` is null, so it falls back to `core.logicalRoot.focus()`—and that causes the editor to jump to the top of the page.  

```ts
export const focus: Focus = core => {
    if (!core.lifecycle.shadowEditFragment) {
        const { api, domHelper, selection } = core;

        if (!domHelper.hasFocus() && selection.selection?.type == 'range') {
            api.setDOMSelection(core, selection.selection, true /*skipSelectionChangedEvent*/);
        }

        // fallback, in case editor still have no focus
        if (!domHelper.hasFocus()) {
            core.logicalRoot.focus();
        }
    }
};
```

To fix this:
- One way is to add `preventScroll` to `core.logicalRoot.focus()`.
- Another option is to improve the selection cache logic in `SelectionPlugin` so it works for all WebKit browsers, not just Safari.
   ```ts
   // Safari has problem to handle onBlur event. When blur, we cannot get the original selection from editor.
   // So we always save a selection whenever editor has focus. Then after blur, we can still use this cached selection.
   if (newSelection?.type == 'range') {
        if (this.isSafari) { // 👉 update it to include both iOS and macOS, since they're all using Apple WebKit.
             this.state.selection = newSelection;
        }
   }
   ```

PR #3215, opened 2025-11-25
Update "isSafari" detection to include iOS and macOS

liangdrime:
On Mac, we’ve run into an issue (#3214) where the page jumps back to the top after adding a column or row using the **"+"** button in a table.

This seems to be a common WebKit issue—the `click` event causes the editable area to lose focus, which in turn makes the editor lose its selection.

We noticed there’s already a fix for Safari that caches the selection and restores it when the editor regains focus after `onBlur`.
```ts
// Safari has problem to handle onBlur event. When blur, we cannot get the original selection from editor.
// So we always save a selection whenever editor has focus. Then after blur, we can still use this cached selection.
if (newSelection?.type == 'range') {
   if (this.isSafari) {
      this.state.selection = newSelection;
   }
}
```

Since iOS and macOS both use WebKit and should behave the same way as Safari, this PR updates the scope of `isSafari` environment to include both iOS and macOS so they can leverage the same Safari-specific fixes.


liangdrime:
Cc @haven2world  for awareness.

liangdrime:
Hi @JiuqingSong, what do you think about this change?

PR #3216, opened 2025-11-25
Support color transformation in table borders

juliaroldi:
Introduce a new `TransformTableBorderColor` that enable border color transformation when the editor switch from light to dark mode.
TableBorderColorsFixed

### DIFF
--- a/packages/roosterjs-content-model-dom/lib/formatHandlers/defaultFormatHandlers.ts
+++ b/packages/roosterjs-content-model-dom/lib/formatHandlers/defaultFormatHandlers.ts
@@ -21,6 +21,7 @@ import { legacyTableBorderFormatHandler } from './table/legacyTableBorderFormatH
 import { letterSpacingFormatHandler } from './segment/letterSpacingFormatHandler';
 import { lineHeightFormatHandler } from './block/lineHeightFormatHandler';
 import { linkFormatHandler } from './segment/linkFormatHandler';
+import { listItemAlignFormatHandler } from './list/listItemAlignFormatHandler';
 import { listItemThreadFormatHandler } from './list/listItemThreadFormatHandler';
 import { listLevelThreadFormatHandler } from './list/listLevelThreadFormatHandler';
 import { listStyleFormatHandler } from './list/listStyleFormatHandler';
@@ -79,6 +80,7 @@ const defaultFormatHandlerMap: FormatHandlers = {
     letterSpacing: letterSpacingFormatHandler,
     lineHeight: lineHeightFormatHandler,
     link: linkFormatHandler,
+    listItemAlign: listItemAlignFormatHandler,
     listItemThread: listItemThreadFormatHandler,
     listLevelThread: listLevelThreadFormatHandler,
     listStyle: listStyleFormatHandler,
@@ -137,14 +139,7 @@ export const defaultFormatKeysPerCategory: {
     block: sharedBlockFormats,
     listItemThread: ['listItemThread'],
     listLevelThread: ['listLevelThread'],
-    listItemElement: [
-        ...sharedBlockFormats,
-        'direction',
-        'textAlign',
-        'lineHeight',
-        'margin',
-        'listStyle',
-    ],
+    listItemElement: [...sharedBlockFormats, 'listItemAlign', 'margin', 'listStyle'],
     listLevel: ['direction', 'textAlign', 'margin', 'padding', 'listStyle', 'backgroundColor'],
     styleBasedSegment: [...styleBasedSegmentFormats, 'textColor', 'backgroundColor', 'lineHeight'],
     elementBasedSegment: elementBasedSegmentFormats,
--- a/packages/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
+++ b/packages/roosterjs-content-model-api/lib/modelApi/block/setModelDirection.ts
@@ -55,6 +55,12 @@ export function setModelDirection(
                     level.format.direction = calcDirection;
                 });
 
+                // We already set direction on levels, no need to keep it on list item level
+                delete item.format.direction;
+
+                // Remove textAlign to let it be calculated based on direction change
+                delete item.format.textAlign;
+
                 item.blocks.forEach(block => internalSetDirection(block, calcDirection));
             });
         } else if (block) {
--- a/packages/roosterjs-content-model-api/test/modelApi/block/setModelDirectionTest.ts
+++ b/packages/roosterjs-content-model-api/test/modelApi/block/setModelDirectionTest.ts
@@ -123,7 +123,10 @@ describe('setModelDirection', () => {
                     {
                         blockType: 'BlockGroup',
                         blockGroupType: 'ListItem',
-                        format: {},
+                        format: {
+                            textAlign: 'start',
+                            direction: 'ltr',
+                        },
                         formatHolder: {
                             segmentType: 'SelectionMarker',
                             format: {},
--- a/packages/roosterjs-content-model-dom/lib/formatHandlers/block/textAlignFormatHandler.ts
+++ b/packages/roosterjs-content-model-dom/lib/formatHandlers/block/textAlignFormatHandler.ts
@@ -10,19 +10,9 @@ export const textAlignFormatHandler: FormatHandler<DirectionFormat & TextAlignFo
     parse: (format, element, context, defaultStyle) => {
         directionFormatHandler.parse(format, element, context, defaultStyle);
 
-        let textAlign = element.style.textAlign || defaultStyle.textAlign;
+        const textAlign = element.style.textAlign || defaultStyle.textAlign;
 
-        if (
-            element.tagName == 'LI' &&
-            element.parentElement?.style.display === 'flex' &&
-            element.parentElement.style.flexDirection === 'column' &&
-            element.style.alignSelf
-        ) {
-            // For LI element with flex style applied, we use its "align-self" style value instead since LI has a different implementation for align
-            textAlign = element.style.alignSelf;
-        }
-
-        if (textAlign) {
+        if (textAlign && element.parentElement?.style.display !== 'flex') {
             format.textAlign = calcAlign(textAlign, format.direction);
         }
     },
@@ -30,16 +20,7 @@ export const textAlignFormatHandler: FormatHandler<DirectionFormat & TextAlignFo
         const dir: 'ltr' | 'rtl' = format.direction == 'rtl' ? 'rtl' : 'ltr';
 
         if (format.textAlign) {
-            const parent = element.parentElement;
-            const parentTag = parent?.tagName;
-
-            if (element.tagName == 'LI' && parent && (parentTag == 'OL' || parentTag == 'UL')) {
-                element.style.alignSelf = format.textAlign;
-                element.parentElement.style.flexDirection = 'column';
-                element.parentElement.style.display = 'flex';
-            } else {
-                element.style.textAlign = ResultMap[format.textAlign][dir];
-            }
+            element.style.textAlign = ResultMap[format.textAlign][dir];
         }
     },
 };
--- a/packages/roosterjs-content-model-dom/lib/formatHandlers/list/listItemAlignFormatHandler.ts
+++ b/packages/roosterjs-content-model-dom/lib/formatHandlers/list/listItemAlignFormatHandler.ts
@@ -1,0 +1,39 @@
+import { calcAlign } from '../utils/dir';
+import type { DirectionFormat, TextAlignFormat } from 'roosterjs-content-model-types';
+import type { FormatHandler } from '../FormatHandler';
+
+/**
+ * @internal
+ */
+export const listItemAlignFormatHandler: FormatHandler<TextAlignFormat & DirectionFormat> = {
+    parse: (format, element, context) => {
+        // For list, we usually use align-self to implement alignment
+        if (element.style.alignSelf) {
+            format.textAlign = calcAlign(element.style.alignSelf, context.blockFormat.direction);
+        } else if (element.style.textAlign && element.parentElement?.style.display !== 'flex') {
+            let align = element.style.textAlign;
+
+            // For RTL environment, 'start' and 'end' in textAlign means opposite direction compared to LTR unless parent is using flex display
+            if (context.blockFormat.direction === 'rtl' && (align == 'start' || align == 'end')) {
+                align = align == 'start' ? 'end' : 'start';
+            }
+
+            format.textAlign = calcAlign(align, context.blockFormat.direction);
+        }
+    },
+    apply: (format, element) => {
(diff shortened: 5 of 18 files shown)


packages/roosterjs-content-model-core/lib/corePlugin/lifecycle/LifecyclePlugin.ts
```diff
@@ -156,6 +163,30 @@ class LifecyclePlugin implements PluginWithState<LifecyclePluginState> {
                 isDarkMode,
                 darkColorHandler
             );
+            this.adjustTableColor(contentDiv, isDarkMode, darkColorHandler);
```
JiuqingSong: No, this should not be here. The function `adjustContainerColor` is only for changing color of editor container DIV, not the editor content. And it is only called once when initialize editor.

Border color should be rendered by content model only, no direction DOM modification from other ways.

packages/roosterjs-content-model-dom/lib/formatHandlers/common/borderFormatHandler.ts
```diff
@@ -45,6 +48,24 @@ export const borderFormatHandler: FormatHandler<BorderFormat> = {
             if (value && width != defaultWidth) {
                 format[key] = value == 'none' ? '' : value;
             }
+
+            if (isElementOfType(element, 'th') || isElementOfType(element, 'td')) {
```
JiuqingSong: why only TH and TD?

packages/roosterjs-content-model-dom/lib/formatHandlers/common/borderFormatHandler.ts
```diff
@@ -45,6 +48,24 @@ export const borderFormatHandler: FormatHandler<BorderFormat> = {
...
+                const borderValues = extractBorderValues(format[key]);
```
JiuqingSong: suggest we do a string search first, for example, search for "--var" in the string and only when it exist, we call extractBorderValues and do color transformation.

packages/roosterjs-content-model-dom/lib/formatHandlers/common/borderFormatHandler.ts
```diff
@@ -45,6 +48,24 @@ export const borderFormatHandler: FormatHandler<BorderFormat> = {
...
+
+                if (borderValues.color) {
+                    const color = getColorInternal(
+                        borderValues.color,
+                        false,
```
JiuqingSong: add inline comments

packages/roosterjs-content-model-dom/lib/formatHandlers/common/borderFormatHandler.ts
```diff
@@ -61,15 +82,29 @@ export const borderFormatHandler: FormatHandler<BorderFormat> = {
...
+    apply: (format, element, context) => {
         AllKeys.forEach(key => {
             const value = format[key];
-
             if (value) {
                 element.style[key] = value;
             }
         });
 
+        if (isElementOfType(element, 'th') || isElementOfType(element, 'td')) {
```
JiuqingSong: same here, why only TD and TH

packages/roosterjs-content-model-dom/lib/formatHandlers/common/borderFormatHandler.ts
```diff
@@ -61,15 +82,29 @@ export const borderFormatHandler: FormatHandler<BorderFormat> = {
...
+            for (const key of BorderKeys) {
```
JiuqingSong: Combine this for loop with previous one at line 86

JiuqingSong: And for dark mode, first calculate dark color, then reuse the same line at 89 for setting back styles.

packages/roosterjs-content-model-dom/lib/formatHandlers/utils/color.ts
```diff
@@ -59,14 +61,26 @@ export function getColor(
...
-    if (color && DeprecatedColors.indexOf(color) > -1) {
-        color = isBackground ? undefined : BlackColor;
-    } else if (darkColorHandler && color) {
+    return color ? getColorInternal(color, isBackground, isDarkMode, darkColorHandler) : undefined;
+}
+
+/**
+ * @internal
+ */
+export function getColorInternal(
```
JiuqingSong: It is a good idea to split the code of getting color from DOM and transform to light mode. Actually we can refactor it even better. Let's split it into 3 parts:
Prepare: create a BorderKey type (you can put it in types package) then create some string arrays and maps:
```ts
type BorderKey = 'borderTop' | 'borderRight' | 'borderBottom' | 'borderLeft';

/**
 * Keys of border items
 */
export const BorderKeys: (BorderKey & keyof BorderFormat & keyof CSSStyleDeclaration)[] = [
    'borderTop',
    'borderRight',
    'borderBottom',
    'borderLeft',
];

export const BorderColorKeyMap: {
    [key in BorderKey]: string & keyof CSSStyleDeclaration;
} = {
    borderTop: 'borderTopColor',
    borderRight: 'borderRightColor',
    borderBottom: 'borderBottomColor',
    borderLeft: 'borderLeftColor',
};

```

1. retrieveElementColor(element, isDarkMode, source: 'text' | 'background' | BorderKey, fallback?: string): string {...}. This will retrieve color from element according to its source string. When retrieve from text or background, keep the existing code to get from CSS then HTML attribute, for others, we can have a map from this border keys to the border color property names, for example:
```ts
function retrieveElementColor(
    element: HTMLElement,
    source: 'text' | 'background' | BorderKey,
    fallback?: string
): string | undefined {
    switch (source) {
        case 'text':
            return element.style.color || element.getAttribute('color') || fallback;

        case 'background':
            return element.style.backgroundColor || element.getAttribute('bgcolor') || fallback;

        default:
            return element.style.getPropertyValue(BorderColorKeyMap[source]) || fallback;
    }
}
```
2. getLightModeColor(...), this is to transform dark color to light color (same with your getColorInternal)
3. getColor(), keep the same parameter list and call to this two functions to keep compatibility:
```ts
function getColor(...) {
   const color = retrieveElementColor(element, isDarkMode, isBackground ? 'background' : 'color', fallback);

  return getLightModeColor(color, ...);
}
```

Then, you just need to call retrieveElementColor() and getLightModeColor() for borders in `borderFormatHandler`, no need to call `extractBorderValues` to avoid an extra RegEx call when parse.

Then for the setColor function, we can do a similar refactoring, the difference is when set color we don't need to set border color separately, but reuse the similar way you have to set borderLeft/Top/Right/Bottom in a single string.

packages/roosterjs-content-model-dom/lib/formatHandlers/utils/color.ts
```diff
@@ -96,6 +109,56 @@ export function setColor(
...
+
+/**
+ * Set the given color to the table border
+ * @param element The element to set color to
+ * @param borderKey the style attribute of the border, ex: borderTop
+ * @param border the border values with the color in light mode
+ * @param isDarkMode Whether element is in dark mode now
+ * @param darkColorHandler @optional The dark color handler object to help manager dark mode color
+ */
+export function setColorOnBorder(
```
JiuqingSong: Please refer to my previous comment, no need to have a separate `setColorOnBorder` function, but just set them together with border width and border style.

packages/roosterjs-content-model-dom/lib/formatHandlers/utils/color.ts
```diff
@@ -119,8 +182,7 @@ export function setColor(
         color = isDarkMode ? `${VARIABLE_PREFIX}${key}, ${color}${VARIABLE_POSTFIX}` : color;
     }
 
-    element.removeAttribute(isBackground ? 'bgcolor' : 'color');
```
JiuqingSong, line 122: Where is this line after the refactor?

juliaroldi: This line is in the setColor function, I did not change it.

packages/roosterjs-content-model-dom/lib/formatHandlers/common/borderFormatHandler.ts
juliaroldi: Should we apply this change to all elements? Should image borders be affected by this change?

JiuqingSong: It should be PM's decision if we want to support other elements as well. My personal opinion is we can support all. After all this is just for dark mode, it will not change the final HTML. So think about a table and some other element has the same border color, then switch to dark mode, table border color changed, but others do not, which seems like a bug to me.

About this comment, my point is no matter if we want to support for other elements, this check should not be here. This is just a format handler, it should only care format, but not element. It should not know any detail about the element.

If we only want to apply this behavior to table, then build two separate format handlers, one for table and one for others. They can share some code, but should be two entry points.

packages/roosterjs-content-model-dom/lib/formatHandlers/defaultFormatHandlers.ts
```diff
@@ -159,13 +161,15 @@ export const defaultFormatKeysPerCategory: {
...
         'border',
+        'borderColor',
         'backgroundColor',
         'padding',
         'verticalAlign',
         'wordBreak',
         'textColor',
         'htmlAlign',
         'size',
+        'borderColor',
```
JiuqingSong: There are two "borderColor"

packages/roosterjs-content-model-dom/lib/domUtils/style/transformColor.ts
```diff
@@ -1,31 +1,93 @@
...
+function transformBorderColor(
+    element: HTMLElement,
+    toDarkMode: boolean,
+    darkColorHandler?: DarkColorHandler
+) {
+    if (isElementOfType(element, 'td') || isElementOfType(element, 'th')) {
+        BorderKeys.forEach(key => {
+            const style = element.style[key];
+            if (style) {
+                const border = extractBorderValues(style);
```
JiuqingSong: I believe you can use something like element.style.borderLeftColor and others to get border color directly, no need to call extractBorderValues() here.

JiuqingSong: As long as you are directly operating on DOM element, you can use these attributes, no need to extract from the string.

packages/roosterjs-content-model-dom/lib/domUtils/style/transformColor.ts
```diff
@@ -1,31 +1,93 @@
...
+                    if (lightColor) {
+                        const transformedColor = adaptColor(
+                            element,
+                            lightColor,
+                            'border',
+                            toDarkMode,
+                            darkColorHandler
+                        );
+                        if (transformedColor) {
+                            element.style[key] = combineBorderValue({
```
JiuqingSong: Same here, directly set element.style.border<direction>Color

packages/roosterjs-content-model-dom/lib/formatHandlers/common/borderColorFormatHandler.ts
```diff
@@ -0,0 +1,65 @@
...
+                            borderValues.color,
+                            'border',
+                            !!context.isDarkMode,
+                            context.darkColorHandler
+                        );
+                        const borderStyles = combineBorderValue({
+                            ...borderValues,
+                            color: transformedColor,
+                        });
+                        element.style[key] = borderStyles;
```
JiuqingSong: No need to combine them back, just set border[Direction]Color is enough.

packages/roosterjs-content-model-dom/lib/formatHandlers/utils/borderKeys.ts
```diff
@@ -0,0 +1,23 @@
...
+    'borderLeft',
+];
+
+/**
+ * @internal
+ */
+export const BorderColorKeyMap: {
+    [key in BorderKey]: string & keyof CSSStyleDeclaration;
+} = {
+    borderTop: 'borderTopColor',
```
JiuqingSong: See you already have these attributes, just use them when set/get colors

packages/roosterjs-content-model-dom/lib/formatHandlers/common/borderColorFormatHandler.ts
```diff
@@ -0,0 +1,63 @@
...
+                if (borderColor) {
+                    const lightModeColor = getLightModeColor(
+                        borderColor,
+                        false /*isBackground*/,
+                        !!context.isDarkMode,
+                        context.darkColorHandler
+                    );
+                    const borderValues = extractBorderValues(value);
+                    format[key] = combineBorderValue({
+                        width: borderValues?.width || '1px',
```
JiuqingSong: I don't think we should set a default border width and style here. We should keep its original value.

If borderValues is null, it means the original value is invalid, so just keep width and style empty.

JiuqingSong: Actually, you didn't do null check for borderValues at line 45, so no need to check it here as well.

packages/roosterjs-content-model-dom/lib/formatHandlers/utils/color.ts
```diff
@@ -59,14 +61,23 @@ export function getColor(
...
+
+    return color ? getLightModeColor(color, isBackground, isDarkMode, darkColorHandler) : undefined;
+}
+
+/**
+ * @internal
+ */
+export function getLightModeColor(
+    color: string,
+    isBackground: boolean,
```
JiuqingSong: I suggest we don't pass in `isBackground` to this function. This function should just be used for getting light color, it should not know any knowledge about if it is background or others. For `DeprecatedColors`, you can pass in a fallback color, for example:

function getLightModeColor(color, isDarkMode, darkModeHandler, fallback?: string) {
  if (is deprecated color) {
    return fallback;
  }
 ..
}

Then pass it in from outside:

return color ? getLightModelColor(color, isDarkMode, darkColorHandler, isBackground ? undefined : BackColor);

So this function can be used from any scenario

packages/roosterjs-content-model-dom/lib/formatHandlers/common/borderColorFormatHandler.ts
juliaroldi: if we let the width and style empty this happens: 

TableBug

JiuqingSong: That is weird, in that case, what is the value of the border style?

juliaroldi: The border value is empty for width and styles. I believe is because we have border-style and border-width instead of styles like "borderTop"
image

PR #3221, opened 2025-12-02
Table Row/Column Selector

juliaroldi:
Add to new Table Editor features, `TableRowSelector` and `TableColumnSelector` to select table rows or columns. 
TableSelectorAndResizer

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/CellResizer.ts
+++ b/packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/CellResizer.ts
@@ -3,7 +3,6 @@ import { DragAndDropHelper } from '../../../pluginUtils/DragAndDrop/DragAndDropH
 import { getCMTableFromTable } from '../utils/getTableFromContentModel';
 import type { TableEditFeature } from './TableEditFeature';
 import {
-    isElementOfType,
     normalizeRect,
     MIN_ALLOWED_TABLE_CELL_WIDTH,
     mutateBlock,
@@ -125,6 +124,7 @@ export interface CellResizerContext {
 export interface CellResizerInitValue {
     cmTable: ReadonlyContentModelTable | undefined;
     anchorColumn: number | undefined;
+    nextColumn: number;
     anchorRow: number | undefined;
     anchorRowHeight: number;
     allWidths: number[];
@@ -139,20 +139,9 @@ export function onDragStart(context: CellResizerContext, event: MouseEvent): Cel
     const rect = normalizeRect(td.getBoundingClientRect());
 
     // Get cell coordinates
-    const columnIndex = td.cellIndex;
-    const row =
-        td.parentElement && isElementOfType(td.parentElement, 'tr') ? td.parentElement : undefined;
-    const rowIndex = row?.rowIndex;
-
-    if (rowIndex == undefined) {
-        return {
-            cmTable: undefined,
-            anchorColumn: undefined,
-            anchorRow: undefined,
-            anchorRowHeight: -1,
-            allWidths: [],
-        }; // Just a fallback
-    }
+    let rowIndex: number | undefined;
+    let columnIndex: number | undefined;
+    let nextColumnIndex = -1;
 
     const { editor, table } = context;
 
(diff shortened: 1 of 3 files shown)


packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableRowColumnSelector.ts
```diff
@@ -0,0 +1,306 @@
...
+    table: HTMLTableElement,
+    isRowSelector: boolean,
+    anchorContainer?: HTMLElement,
+    onTableEditorCreated?: OnTableEditorCreatedCallback
+): TableEditFeature | null {
+    const rect = normalizeRect(table.getBoundingClientRect());
+    if (rect) {
+        const zoomScale = editor.getDOMHelper().calculateZoomScale();
+        const createElementData = getInsertElementData(rect, isRowSelector);
+        const div = createElement(createElementData, document) as HTMLDivElement;
```
JiuqingSong: do not use global document

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableRowColumnSelector.ts
```diff
@@ -0,0 +1,306 @@
...
+        div.id = isRowSelector ? ROW_SELECTOR_ID : COLUMN_SELECTOR_ID;
```
JiuqingSong: Be careful, it is allowed to have multiple instances of editor on the same page. So do not use const id, but combine with editor id. We have a utility function to help do that. `ensureUniqueId` You may want to create a util in DOMHelper to help do this.

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableRowColumnSelector.ts
```diff
@@ -0,0 +1,306 @@
...
+        const context: TableRowColumnSelectorContext = {
+            table,
+            zoomScale,
+            editor,
+            div,
+
+            isRow: isRowSelector,
+        };
+
+        (anchorContainer || document.body).appendChild(div);
```
JiuqingSong: same here, do not use global document

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableRowColumnSelector.ts
JiuqingSong: Actually I didn't see you use this ID to query the DIV anywhere else, so you can remove it

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableRowColumnSelector.ts
```diff
@@ -0,0 +1,306 @@
...
+} from 'roosterjs-content-model-types';
+
+/**
+ * @internal
+ */
+export const ROW_SELECTOR_ID = 'rowSelector';
+/**
+ * @internal
+ */
+export const COLUMN_SELECTOR_ID = 'columnSelector';
```
JiuqingSong: remove them if not really used

packages/roosterjs-content-model-plugins/lib/tableEdit/editors/features/TableRowColumnSelector.ts
juliaroldi: This is used for unit test only, it is not needed to make the feature work.

JiuqingSong: In that case, you can use a separate function to create this DIV, then mock it in test case so you can get that DIV in test code.

Adding ID here causes it appears in dev code and in production, others may write code to rely on it. And that may not work if there are multiple editor instances in the same page.

JiuqingSong: For example, (in a separate file)
```ts
export function createHelperDiv() {
   return ...;
}
```

In this file:
```ts
const div = createHelperDiv();
```

In test code:
```ts
const div = document.createElement('div');
spyOn(createHelperDiv, 'createHelperDiv').and.returnValue(div);
```

PR #3227, opened 2025-12-05
Support Trusted publishing for npmjs

JiuqingSong:
Due to a recent change of npmjs.com, it is now recommended to use "Trusted Publishing" when publish packages from CI/CD.

This is to make the change from our side accordingly. I have setup trusted publish in npmjs.com for all active packages that we have in master branch.

To read more about trusted publishing: https://docs.npmjs.com/trusted-publishers

Note that roosterjs 9.44.0 is already published, this change will take effect from next time we publish new version.

### DIFF
--- a/tools/buildTools/publish.js
+++ b/tools/buildTools/publish.js
@@ -5,9 +5,8 @@ const exec = require('child_process').execSync;
 const { distPath, readPackageJson, packages } = require('./common');
 
 const VersionRegex = /\d+\.\d+\.\d+(-([^\.]+)(\.\d+)?)?/;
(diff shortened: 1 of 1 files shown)


tools/buildTools/publish.js
```diff
@@ -1,7 +1,6 @@
 'use strict';
 
 const path = require('path');
-const fs = require('fs');
 const exec = require('child_process').execSync;
 const { distPath, readPackageJson, packages } = require('./common');
```
vinguyen12, line 6: looks like we are not using `NpmrcContent` anywhere and can remove it?

PR #3238, opened 2025-12-16
Export formatContainerProcessor for public API use

### LINKED ISSUE #3236: Feature Request: Export formatContainerProcessor from public API
## Use Case

We're building an email client and need to customize how `<blockquote>` elements are processed during DOM-to-Model conversion. In our application, email citations use blockquotes with zero margins:

```html
<blockquote style="margin-top: 0; margin-bottom: 0;">...</blockquote>
```

We want these blockquotes to always preserve their `FormatContainer` semantics, but the default `knownElementProcessor` only triggers `formatContainerProcessor` when certain style values are `> 0`.

## Request

Please export `formatContainerProcessor` from the public API, so users can use it in `processorOverride` for custom processing needs:

```typescript
// In roosterjs-content-model-dom/lib/index.ts
export { formatContainerProcessor } from './domToModel/processors/formatContainerProcessor';
```

## Current Workarounds

Without the public export, we have to use fragile workarounds:

**Option 1: Require internal module path**
```typescript
const { formatContainerProcessor } = require(
  'roosterjs-content-model-dom/lib/domToModel/processors/formatContainerProcessor'
);
```

**Option 2: Borrow from another tag's default processor**
```typescript
processorOverride: {
  blockquote: (group, element, context) => {
    context.defaultElementProcessors.pre(group, element, context);
  },
},
```

## Rationale

- Other processors (`childProcessor`, `entityProcessor`, `tableProcessor`) are already publicly exported
- `processorOverride` is a public API, but users lack access to reusable processor implementations
- Exporting `formatContainerProcessor` would enable legitimate customization without relying on internal paths

Dennis273:
## Summary

Export `formatContainerProcessor` from the public API of `roosterjs-content-model-dom` to enable users to customize DOM-to-Model conversion for format container elements (e.g., `blockquote`, `div`) via `processorOverride`.

## Problem

Currently, `formatContainerProcessor` is marked as `@internal` and not exported from the public API. Users who need to customize how format container elements are processed during DOM-to-Model conversion have to rely on fragile workarounds:

1. **Require internal module path** (fragile, may break):
   ```typescript
   const { formatContainerProcessor } = require(
     'roosterjs-content-model-dom/lib/domToModel/processors/formatContainerProcessor'
   );
   ```

2. **Borrow from another tag's default processor** (workaround):
   ```typescript
   processorOverride: {
     blockquote: (group, element, context) => {
       context.defaultElementProcessors.pre(group, element, context);
     },
   },
   ```

This is inconsistent with other processors like `childProcessor`, `entityProcessor`, and `tableProcessor`, which are already publicly exported.

## Solution

- Export `formatContainerProcessor` from `roosterjs-content-model-dom/lib/index.ts`
- Update JSDoc comments to remove `@internal` tag and add proper documentation
- Maintain consistency with other publicly exported processors

## Changes

- **Added export** in `packages/roosterjs-content-model-dom/lib/index.ts`
- **Updated documentation** in `packages/roosterjs-content-model-dom/lib/domToModel/processors/formatContainerProcessor.ts`

## Usage Example

After this change, users can now use `formatContainerProcessor` directly:

```typescript
import { formatContainerProcessor } from 'roosterjs-content-model-dom';

// In domToModel config
const config = createDomToModelConfig({
  processorOverride: {
    blockquote: formatContainerProcessor, // Now works!
  },
});
```

This is particularly useful for email clients where blockquotes with zero margins need to preserve their `FormatContainer` semantics:

```html
<blockquote style="margin-top: 0; margin-bottom: 0;">...</blockquote>
```

## Testing

- ✅ Code passes linting
- ✅ Pre-commit hooks passed
- ✅ Follows existing code patterns and conventions

fix #3236

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -151,12 +151,33 @@ module.exports = {
         'no-var': 'error',
         'etc/no-const-enum': ['error', { allowLocal: true }],
         'import/no-default-export': 'error',
+        'no-restricted-globals': [
+            'error',
+            {
+                name: 'window',
+                message: 'Do not use global window',
+            },
+            {
+                name: 'document',
+                message: 'Do not use global document',
+            },
+        ],
     },
     overrides: [
         {
             files: ['roosterjs-color-utils/**/*.ts'],
             rules: {
                 'import/no-default-export': 'off',
+            },
+        },
+        {
+            files: [
+                'roosterjs-editor-adapter/**/*.ts',
+                'roosterjs-react/**/*.ts',
+                'roosterjs-react/**/*.tsx',
+            ],
+            rules: {
+                'no-restricted-globals': 'off',
             },
         },
     ],
--- a/packages/roosterjs-content-model-api/lib/modelApi/table/createTableStructure.ts
+++ b/packages/roosterjs-content-model-api/lib/modelApi/table/createTableStructure.ts
@@ -1,5 +1,9 @@
 import { addBlock, createTable, createTableCell } from 'roosterjs-content-model-dom';
-import type { ContentModelBlockGroup, ContentModelTable } from 'roosterjs-content-model-types';
+import type {
+    ContentModelBlockGroup,
+    ContentModelTable,
+    ContentModelTableCellFormat,
+} from 'roosterjs-content-model-types';
 
 /**
  * @internal
@@ -7,7 +11,8 @@ import type { ContentModelBlockGroup, ContentModelTable } from 'roosterjs-conten
 export function createTableStructure(
     parent: ContentModelBlockGroup,
     columns: number,
-    rows: number
+    rows: number,
+    cellFormat?: ContentModelTableCellFormat
 ): ContentModelTable {
     const table = createTable(rows);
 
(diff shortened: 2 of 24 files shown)


JiuqingSong:
Please fix the line-end so we can see what is changed

Dennis273:
> Please fix the line-end so we can see what is changed

Done. I've resolved the line-ending problem. You should be able to see the changes correctly now.

JiuqingSong:
Can you help my understand why it need to be handled using formatContainerProcessor?

We are handling BLOCKQUOTE with margin top and bottom to 0 as a regular paragraph, because if content was created using some other editor (especially when relies on browser function `execCommand`), indentation will be represented by blockquote with top and bottom margin set to 0. So we recognize it and transform it to be a normal indent (margin-left and right on DIV). This should be rendered the same with blockquote. So in theory we don't need to use formatContainer to represent this blockquote.

Dennis273:
Thanks for the explanation! I understand the current logic and agree that using `div` for visual consistency makes sense in most cases.

In our case, because our mail client has some specific logic for "Reply/Forward" that relies on the `blockquote` tag to identify quote threads, we are looking for a way to preserve it.

I realize this might be a more specific requirement rather than a universal one. However, by exporting `formatContainerProcessor`, it would allow us to use `processorOverride` to handle this at the application level without changing RoosterJS's default behavior or having to copy-paste the internal processor logic.

What do you think about opening up this capability for cases like ours?

PR #3239, opened 2025-12-16
Add minWidth cell to insertTable

juliaroldi:
Add a new parameter, `customCellFormat`, to the `insertTable` function to allow custom styles for tables. 
CellMinimunWidth

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -151,12 +151,33 @@ module.exports = {
         'no-var': 'error',
         'etc/no-const-enum': ['error', { allowLocal: true }],
         'import/no-default-export': 'error',
+        'no-restricted-globals': [
+            'error',
+            {
+                name: 'window',
+                message: 'Do not use global window',
+            },
+            {
+                name: 'document',
+                message: 'Do not use global document',
+            },
+        ],
     },
     overrides: [
         {
             files: ['roosterjs-color-utils/**/*.ts'],
             rules: {
                 'import/no-default-export': 'off',
+            },
+        },
+        {
+            files: [
+                'roosterjs-editor-adapter/**/*.ts',
+                'roosterjs-react/**/*.ts',
+                'roosterjs-react/**/*.tsx',
+            ],
+            rules: {
+                'no-restricted-globals': 'off',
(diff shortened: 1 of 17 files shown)


packages/roosterjs-content-model-api/lib/publicApi/table/insertTable.ts
```diff
@@ -41,7 +43,9 @@ export function insertTable(
 
             if (insertPosition) {
                 const doc = createContentModelDocument();
-                const table = createTableStructure(doc, columns, rows);
+                const table = createTableStructure(doc, columns, rows, {
+                    minWidth: cellMinWidth + 'px',
```
JiuqingSong: what if cellMinWidth is undefined?

packages/roosterjs-content-model-api/lib/publicApi/table/insertTable.ts
```diff
@@ -25,13 +26,15 @@ import type {
...
+ * @param cellFormat (Optional) custom format for table cells
  */
 export function insertTable(
     editor: IEditor,
     columns: number,
     rows: number,
     tableMetadataFormat?: Partial<TableMetadataFormat>,
-    format?: ContentModelTableFormat
+    format?: ContentModelTableFormat,
+    customCellFormat?: CustomTableCellFormat
```
JiuqingSong: Why not just use ContentModelTableCellFormat?

juliaroldi: ContentModelTableCellFormat has styles that conflict with TableMetadataFormat such as borders. If we set border styles in this parameter it will be replaced by `applyTableFormat` by the `tableMetadataFormat` parameter. We would have to parameter for the same styles.

JiuqingSong: I mean, keep existing `tableMetadataFormat` and `format` (you can rename it to `tableFormat` for clear), and add another parameter `tableCellFormat`. So we have 3 parameters, two for table and one for all cells

PR #3247, opened 2026-01-05
Fix table adaptive color

juliaroldi:
This PR fixes an issue where text could become invisible in table cells when the text color matched the background color (e.g., black text on black background or white text on white background).
TableHeaderTextColor


packages/roosterjs-content-model-dom/lib/modelApi/editing/setTableCellBackgroundColor.ts
```diff
@@ -140,3 +129,16 @@ function calculateLightness(color: string) {
         return 255;
     }
 }
+
+/**
+ * Check if two colors are the same by comparing their RGB values
+ */
+function isSameColor(color1: string, color2: string): boolean {
```
JiuqingSong: let's name it "areSameColors"

JiuqingSong (approved):

PR #3258, opened 2026-01-14
[Table Improvements] Insert table content

juliaroldi:
When inserting a table in a range selection, insert the selected content inside the table. 
image

### DIFF
--- a/packages/roosterjs-content-model-api/lib/modelApi/common/clearModelFormat.ts
+++ b/packages/roosterjs-content-model-api/lib/modelApi/common/clearModelFormat.ts
@@ -1,17 +1,17 @@
 import { adjustWordSelection } from '../selection/adjustWordSelection';
 import {
     applyTableFormat,
+    copyFormat,
     createFormatContainer,
     getClosestAncestorBlockGroupIndex,
     iterateSelections,
     mutateBlock,
     mutateSegments,
-    updateTableCellMetadata,
-    updateTableMetadata,
 } from 'roosterjs-content-model-dom';
 import type {
     ContentModelSegmentFormat,
     ContentModelTable,
+    ContentModelTableCellFormat,
     ReadonlyContentModelBlock,
     ReadonlyContentModelBlockGroup,
     ReadonlyContentModelDocument,
@@ -106,7 +106,6 @@ function createTablesFormat(tablesToClear: [ContentModelTable, boolean][]) {
                 useBorderBox: table.format.useBorderBox,
                 borderCollapse: table.format.borderCollapse,
             };
-            updateTableMetadata(table, () => null);
         }
 
         applyTableFormat(table, undefined /*newFormat*/, true);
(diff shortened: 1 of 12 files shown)


packages/roosterjs-content-model-api/lib/publicApi/table/insertTable.ts
```diff
@@ -40,15 +46,24 @@ export function insertTable(
 
     editor.formatContentModel(
         (model, context) => {
-            const insertPosition = deleteSelection(model, [], context).insertPoint;
+            const copiedModel = cloneModel(model);
```
JiuqingSong: curious why we need to clone a model?

packages/roosterjs-content-model-api/lib/publicApi/table/insertTable.ts
```diff
@@ -102,3 +117,42 @@ function getTableCellWidth(columns: number): number {
...
+                    undefined /* isHeader */,
+                    customCellFormat
+                );
+                row.cells.push(cell);
+            }
+            table.rows.push(row);
+        }
+
+        if (path.length == 1 && block) {
+            table.rows[index].cells[0].blocks = [block];
```
JiuqingSong: what if only part of a block is selected? Do we still expect to insert the whole block into table?

packages/roosterjs-content-model-api/lib/publicApi/table/insertTable.ts
```diff
@@ -102,3 +117,42 @@ function getTableCellWidth(columns: number): number {
         return 70;
     }
 }
+
+function insertTableContent(
```
JiuqingSong: This looks more like a model API, so you can create a separate file under modelAPI folder

packages/roosterjs-content-model-api/lib/publicApi/table/insertTable.ts
juliaroldi: I used the cloned model, because it returns the content model types, not readonly types, making it easier to insert the content into the table that is ContentModelTable type.

BryanValverdeU: Could we iterate the selected blocks, mutate them and then store them in an array？ If the model is very big we would need to iterate everything to clone it.

packages/roosterjs-content-model-api/lib/modelApi/table/tableContent.ts
```diff
@@ -0,0 +1,280 @@
...
+    ReadonlyContentModelBlockGroup,
+    ReadonlyContentModelSegment,
+    ReadonlyContentModelTable,
+    ShallowMutableContentModelDocument,
+} from 'roosterjs-content-model-types';
+
+/**
+ * @internal
+ */
+export function getSelectedContent(
```
JiuqingSong: This seems similar with packages\roosterjs-content-model-core\lib\command\cutCopy\getContentForCopy.ts, can you reuse?

packages/roosterjs-content-model-dom/lib/domUtils/selection/getSelectedModel.ts
```diff
@@ -0,0 +1,28 @@
...
+import { preprocessTable } from './preprocessTable';
+import { pruneUnselectedModel } from './pruneUnselectedModel';
+import type { ContentModelDocument, DOMSelection } from 'roosterjs-content-model-types';
+
+/**
+ * Get the content model containing only the selection
+ * @param model the model document
+ * @param selection The editor selection
+ * */
+export function getSelectedModel(
```
JiuqingSong: This name is confusing. Normally a function named "get**" will not modify the input object. Here it actually removes unselected contents, so better name it something like `trimModelForSelection`.

JiuqingSong: And then no need to return model since we can directly get it from the parameter.

PR #3268, opened 2026-01-22
Patch bump to 9.45.3

vinguyen12:
{
    "react": "9.0.4",
    "main": "9.45.3",
    "legacyAdapter": "8.65.2",
    "overrides": {}
}


versions.json
```diff
@@ -1,6 +1,6 @@
 {
     "react": "9.0.4",
-    "main": "9.45.2",
+    "main": "9.45.3",
```
BryanValverdeU, line 3: I dont think we need to change version for a publish script fix.

juliaroldi: I would like to push this change #3265 in this bump, then we would need a patch version.

PR #3284, opened 2026-02-06
Preserve cursor position when navigating tables 

juliaroldi:
When pressing key up or key down in table cells that have text, preserve the cursor position. 
UpDownTable

### DIFF
--- a/packages/roosterjs-content-model-core/test/corePlugin/selection/SelectionPluginTest.ts
+++ b/packages/roosterjs-content-model-core/test/corePlugin/selection/SelectionPluginTest.ts
@@ -1,4 +1,6 @@
 import * as findTableCellElement from '../../../lib/coreApi/setDOMSelection/findTableCellElement';
+import * as getDOMInsertPointRectFile from 'roosterjs-content-model-dom/lib/domUtils/selection/getDOMInsertPointRect';
+import * as getNodePositionFromEventFile from 'roosterjs-content-model-dom/lib/domUtils/event/getNodePositionFromEvent';
 import * as isSingleImageInSelection from '../../../lib/corePlugin/selection/isSingleImageInSelection';
 import * as parseTableCells from 'roosterjs-content-model-dom/lib/domUtils/table/parseTableCells';
 import { createDOMHelper } from '../../../lib/editor/core/DOMHelperImpl';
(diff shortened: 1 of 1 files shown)


packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
```diff
@@ -364,7 +364,14 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
                             this.handleSelectionInTable(this.getTabKey(rawEvent));
                             rawEvent.preventDefault();
                         } else {
-                            win?.requestAnimationFrame(() => this.handleSelectionInTable(key));
+                            const textOffset =
```
BryanValverdeU: If the text in the cell have multiple segments, like when bold/italic/underline. The textOffset could be wrong. 

We might have to get the whole string before cursor for this. 

Also, when setting the cursor in the target cell. We need to get all the segments, and then add a SelectionMarker based on the text offset. As target cell could also have bold/italic segments.

JiuqingSong: Agree.

You can also try use `getNodePositionFromEvent` to find a target selection from coordinates, see if browser can give you a good potential selection.

packages/roosterjs-content-model-dom/lib/domUtils/event/getNodePositionFromEvent.ts
```diff
@@ -1,16 +1,18 @@
...
+ * Get insertion point from coordinate.
+ * @param doc Parent document object
+ * @param domHelper The DOM helper of the editor
+ * @param x The cursor coordinate for the x-axis
+ * @param y The cursor coordinate for the y-axis
  */
 export function getNodePositionFromEvent(
-    editor: IEditor,
+    doc: Document,
+    domHelper: DOMHelper,
```
JiuqingSong, line 12: Is domHelper used?

juliaroldi: Yes, we use it get the isNodeInEditor function

JiuqingSong: I see.

Issue #3235, opened 2025-12-11
Content Model to HTML is stripping trailing space.


BryanValverdeU:
**Describe the bug**
When loading the a model like below
```
{
    "blockGroupType": "Document",
    "blocks": [
        {
            "blockType": "Paragraph",
            "segments": [
                {
                    "segmentType": "Text",
                    "text": "Could you please provide more information ",
                    "format": {
                        "fontFamily": "Corbel, Skia, sans-serif",
                        "fontSize": "12pt",
                        "textColor": "rgb(94, 50, 124)"
                    }
                }
            ],
            "format": {
                "isProofing": true
            },
            "segmentFormat": {
                "fontFamily": "Corbel, Skia, sans-serif",
                "fontSize": "12pt",
                "textColor": "rgb(94, 50, 124)"
            }
        },
        {
            "blockType": "Paragraph",
            "segments": [
                {
                    "segmentType": "SelectionMarker",
                    "isSelected": true,
                    "format": {
                        "fontFamily": "Corbel, Skia, sans-serif",
                        "fontSize": "12pt",
                        "textColor": "rgb(94, 50, 124)"
                    }
                }
            ],
            "format": {
                "isProofing": true
            },
            "segmentFormat": {
                "fontFamily": "Corbel, Skia, sans-serif",
                "fontSize": "12pt",
                "textColor": "rgb(94, 50, 124)"
            }
        }
    ],
    "format": {
        "fontFamily": "'Corbel', 'Skia', sans-serif",
        "fontSize": "12pt",
        "textColor": "#5e327c"
    }
}
```

As you can see the paragraph has an ending space in the Text property, however when rendered that space is lost.
<img width="810" height="262" alt="Image" src="https://github.com/user-attachments/assets/bcce961e-d840-4353-8772-6116bce7f016" />

<img width="1065" height="205" alt="Image" src="https://github.com/user-attachments/assets/62eeba5a-4703-4a34-9606-7a6b84f445d9" />

If instead of adding space adding a &nbsp fixes the issue:

<img width="790" height="65" alt="Image" src="https://github.com/user-attachments/assets/3eb70a76-5e3d-4fa0-80e8-4d71b8ca8926" />

<img width="992" height="247" alt="Image" src="https://github.com/user-attachments/assets/46022765-11e4-4fad-bbc5-c61a42c07554" />

**To Reproduce**
1. [Go to '...'](https://microsoft.github.io/roosterjs/index.html#contentModel)
[...]

PR #3287, opened 2026-02-23
Preserve trailing space in last paragraph segment by converting to nbsp (#3235)

BryanValverdeU:
When the last text segment in a paragraph ends with a regular space, browsers collapse it during rendering. This change detects that case in handleText and replaces the trailing space with a non-breaking space (\u00A0) so it is preserved in the output.

To support this, a new ModelToDomSegmentContext interface is introduced that extends ModelToDomContext with an isLastSegment flag. handleParagraph sets this flag for each segment before dispatching, and ContentModelSegmentHandler is updated to use ModelToDomSegmentContext as its context type, eliminating the need for type casts in the handlers.

import this model:
```
{
    "blockGroupType": "Document",
    "blocks": [
        {
            "blockType": "Paragraph",
            "segments": [
                {
                    "segmentType": "Text",
                    "text": "Could you please provide more information ",
                    "format": {
                        "fontFamily": "Corbel, Skia, sans-serif",
                        "fontSize": "12pt",
                        "textColor": "rgb(94, 50, 124)"
                    }
                }
            ],
            "format": {
                "isProofing": true
            },
            "segmentFormat": {
                "fontFamily": "Corbel, Skia, sans-serif",
                "fontSize": "12pt",
                "textColor": "rgb(94, 50, 124)"
            }
        },
        {
            "blockType": "Paragraph",
            "segments": [
                {
                    "segmentType": "SelectionMarker",
                    "isSelected": true,
                    "format": {
                        "fontFamily": "Corbel, Skia, sans-serif",
                        "fontSize": "12pt",
                        "textColor": "rgb(94, 50, 124)"
                    }
                }
            ],
            "format": {
                "isProofing": true
            },
            "segmentFormat": {
                "fontFamily": "Corbel, Skia, sans-serif",
                "fontSize": "12pt",
                "textColor": "rgb(94, 50, 124)"
            }
        }
    ],
    "format": {
        "fontFamily": "'Corbel', 'Skia', sans-serif",
        "fontSize": "12pt",
        "textColor": "#5e327c"
    }
}
```

Before
image 

After
image

### DIFF
--- a/packages/roosterjs-content-model-dom/lib/modelToDom/handlers/handleText.ts
+++ b/packages/roosterjs-content-model-dom/lib/modelToDom/handlers/handleText.ts
@@ -1,6 +1,8 @@
 import { handleSegmentCommon } from '../utils/handleSegmentCommon';
 import type { ContentModelSegmentHandler, ContentModelText } from 'roosterjs-content-model-types';
 
+const nonBreakingSpace = '\u00A0';
+
 /**
  * @internal
  */
@@ -11,7 +13,11 @@ export const handleText: ContentModelSegmentHandler<ContentModelText> = (
     context,
     segmentNodes
 ) => {
-    const txt = doc.createTextNode(segment.text);
+    const textContent =
+        context.isLastSegment && segment.text.endsWith(' ')
+            ? segment.text.slice(0, -1) + nonBreakingSpace
+            : segment.text;
+    const txt = doc.createTextNode(textContent);
     const element = doc.createElement('span');
 
     parent.appendChild(element);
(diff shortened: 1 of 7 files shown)


packages/roosterjs-content-model-dom/lib/modelToDom/handlers/handleText.ts
```diff
@@ -11,7 +13,11 @@ export const handleText: ContentModelSegmentHandler<ContentModelText> = (
     context,
     segmentNodes
 ) => {
-    const txt = doc.createTextNode(segment.text);
+    const textContent =
+        context.isLastSegment && segment.text.endsWith(' ')
```
JiuqingSong: I feel the "last segment" phrase is not very accurate here. What if there is an image after the selection? What should also do this conversion.

So I think the check should be, ignoring all selection marker, check if next segment is still text. As long as it is not text (image, BR, entity, or no other segments), we should do this.

JiuqingSong (approved):

PR #3300, opened 2026-03-05
FHL Add RewriteMutatedBlocksOnly experimental feature for optimized DOM write-back

BryanValverdeU:
Introduces a fast path in handleBlockGroupChildren that skips re-running block handlers for blocks that haven't been mutated (i.e. still have a valid cachedElement). Instead, their existing DOM elements are reused directly via reuseCachedElement, reducing unnecessary DOM operations when only a small portion of the content model was changed.

The motivation of this change is to prevent unneeded mutations from observers attached to the Editor div, which could cause uneeded effects on mutations.

Extra change
- karma.fast.conf.js updated with a FailOnlyReporter to suppress passing test output, making failures easier to spot

Attached a MutationObserver to the editorDiv, 
`new MutationObserver((asd) => console.log(asd)).observe($0, {childList:true,subtree: true })`
with the experimental feature on, less mutations are handled, on the gif, I am using Enter Key

Before
AnimationBefore

After
AnimationAfter


packages/roosterjs-content-model-dom/lib/modelToDom/utils/isBlockFullyCached.ts
```diff
@@ -0,0 +1,68 @@
...
+                                    cell.blocks.every(isBlockFullyCached))
+                        )
+                )
+            );
+
+        case 'Divider':
+            return !!block.cachedElement && !block.isSelected;
+
+        case 'Entity':
+            return false;
```
JiuqingSong, line 40: why return false for entity?

packages/roosterjs-content-model-dom/lib/modelToDom/utils/isBlockFullyCached.ts
```diff
@@ -0,0 +1,68 @@
...
+                        !!block.cachedElement &&
+                        block.blocks.length > 0 &&
+                        block.blocks.every(isBlockFullyCached)
+                    );
+
+                case 'ListItem':
+                    return (
+                        !!block.cachedElement &&
+                        block.levels.every(l => !!l.cachedElement) &&
+                        block.blocks.length > 0 &&
```
JiuqingSong, line 57: If block.length == 0, should we treat it as cached?

JiuqingSong: If not , add comment to explain why

JiuqingSong (commented):
Can you show what mutations are skipped?

Before this change, whether it should reuse cache is decided by each block handler. And we may still call format handlers even cache is still available (e.g. list items). I can see in your change list is excluded, which is right. But still curious what are impacted, what kind of mutations are skipped.

PR #3319, opened 2026-04-07
Prevent drop malicious content on the editor

juliaroldi:
Introduced protection to prevent potentially harmful HTML content from being added to the editor through drag-and-drop. The DragAndDrop Plugin was implemented to manage external content drops, block the default drop action, sanitize any dropped content, and insert only the sanitized content into the editor.


packages/roosterjs-content-model-core/lib/corePlugin/domEvent/DOMEventPlugin.ts
```diff
@@ -137,8 +139,24 @@ class DOMEventPlugin implements PluginWithState<DOMEventPluginState> {
...
-    private onDrop = () => {
-        const doc = this.editor?.getDocument();
+    private onDrop = (e: DragEvent) => {
+        if (!this.editor) {
+            return;
+        }
+        const doc = this.editor.getDocument();
+        const html = e.dataTransfer?.getData('text/html');
+
+        if (html && this.forbiddenDropElement.length > 0) {
```
BryanValverdeU: I think we should still allow the drop to proceed, but set the HTML data transfer with the sanitized content from htmlToDom. The trustedHTMLHandler should already strip the <script> tag from the content. And to be extra safe, we could use querySelector to remove all <iframe> elements from the content before inserting it.

With the current approach, if a user drags and drops HTML that contains both legitimate content and an <iframe>, the entire drop becomes a no-op, we'd lose the non-malicious HTML as well.

JiuqingSong: I agree

BryanValverdeU (approved):

JiuqingSong (approved):

PR #3329, opened 2026-04-28
Add back the check for handle enter key

JiuqingSong:
In my previous change I have removed  the check for handleEnterKey callback and always handle enter key using content model.

However, OutlookMobile still have the requirement to let browser handle enter key. So add back the check.


JiuqingSong:
@haven2world

BryanValverdeU (approved):

PR #3342, opened 2026-05-18
Fix #3341: Preserve selection marker when reconciling range across two text nodes

### LINKED ISSUE #3341: Failed to set selection in Content Model when the selection crosses 2 text nodes
**Describe the bug**
I’ve already debugged the code and identified the root cause, so let’s jump straight into it.

When reconciling a cached content model for a range selection that spans across two text nodes, `DomIndexerImpl.reconcileSelection()` inserts two SelectionMarkers by calling reconcileTextSelection() twice:
``` typescript
const marker1 = this.reconcileNodeSelection(startContainer, startOffset);
const marker2 = this.reconcileNodeSelection(endContainer, endOffset);

if (marker1 && marker2) {
    setSelection(model, marker1, marker2);
}
```

Inside `reconcileTextSelection()`, the replacement range is expanded to include adjacent `SelectionMarkers` before splicing the new segments back into paragraph.segments:
``` typescript
while (
    firstIndex > 0 &&
    paragraph.segments[firstIndex - 1].segmentType == 'SelectionMarker'
) {
    firstIndex--;
}

while (
    lastIndex < paragraph.segments.length - 1 &&
    paragraph.segments[lastIndex + 1].segmentType == 'SelectionMarker'
) {
    lastIndex++;
}

paragraph.segments.splice(firstIndex, lastIndex - firstIndex + 1, ...newSegments);
```

This cleanup logic can unintentionally include the SelectionMarker that was just inserted by the first `reconcileTextSelection()` call.

Example:

Initial paragraph.segments:
```
[
    Text("I was wondering if you’d like to join me for dinner this "),
    Text("Friday at 7:00 PM"),
    Text(". It would be great to catch up")
]
```
I selects "Frid", but for some reason in safari, the actual DOM range spans two text nodes:

- start: after "this "
- end: after "Frid" in "Friday at 7:00 PM"

After reconciling the start boundary, the model becomes:
```
[
    Text("I was wondering if you’d like to join me for dinner this "),
    SelectionMarker(marker1),
    Text("Friday at 7:00 PM"),
    Text(". It would be great to catch up")
]
```
Then the end boundary is reconciled against `Text("Friday at 7:00 PM")`. `reconcileTextSelection()` creates:
```
[
    Text("Frid"),
    SelectionMarker(marker2),
    Text("ay at 7:00 PM")
]
```
But before splicing, it expands firstIndex backward to include adjacent existing SelectionMarkers. Since marker1 is immediately before `Text("Friday at 7:00 PM")`, the replacement range includes `marker1`.

After the second splice, the model becomes:
```
[
    Text("I was wondering if you’d like to join me for dinner this "),
    Text("Frid"),
    SelectionMarker(marker2),
    Text("ay at 7:00 PM"),
    Text(". It would be great to catch up")
]
```
`marker1` has been removed from `paragraph.segments`.

The local `marker1` variable still points to the removed marker object, but that marker no longer exists in the model. Then `setSelection(model, marker1, marker2)` cannot find the start boundary, so `Text("Frid")` is not marked with `isSelected: true`.

Observed behavior:

- cachedSelection is correct.
- The range spans two text nodes.
- The expected selected text segment, for example `Text("Frid")`, does not get isSelected: true.
- `formatContentModel()` reuses the cached model, so the formatter receives a model without the expected selected segment.

JiuqingSong:
## Summary
- `DomIndexerImpl.reconcileSelection` was losing the `SelectionMarker` inserted by the first `reconcileNodeSelection` call when handling a range that spans two text nodes. The second call's adjacent-marker cleanup loop absorbed `marker1` if it ended up directly next to the second text node's segment (notably when `startOffset == startContainer.nodeValue.length`). `setSelection` then orphaned both markers and the selected segment never received `isSelected`.
- Fix: thread the first call's marker into the second call as `preserveMarker`; the cleanup loops now skip that specific reference while still absorbing other adjacent stale markers.
- Added a regression test (`Repro #3341: range across two text nodes with startOffset at end of first node`) that verifies the full paragraph model after the action.

## Test plan
- [x] `yarn eslint` — clean
- [x] `yarn test:fast` — 5840 passing (includes the new regression test)
- [x] `yarn b` — build succeeds

Fixes #3341

### DIFF
--- a/packages/roosterjs-content-model-core/lib/corePlugin/cache/domIndexerImpl.ts
+++ b/packages/roosterjs-content-model-core/lib/corePlugin/cache/domIndexerImpl.ts
@@ -325,7 +325,19 @@ export class DomIndexerImpl implements DomIndexer {
                         );
                     } else {
                         const marker1 = this.reconcileNodeSelection(startContainer, startOffset);
-                        const marker2 = this.reconcileNodeSelection(endContainer, endOffset);
+                        // Pass marker1 to the second call so its adjacent-marker cleanup
+                        // does not consume the SelectionMarker we just inserted. Without
+                        // this guard, when marker1 lands directly next to endContainer's
+                        // segment in paragraph.segments (e.g. startOffset == startContainer
+                        // text length), the second splice would absorb marker1 and leave
+                        // setSelection with a dangling reference. See issue #3341.
+                        const marker2 = this.reconcileNodeSelection(
+                            endContainer,
+                            endOffset,
+                            undefined,
+                            undefined,
+                            marker1
+                        );
 
                         if (marker1 && marker2) {
                             if (newSelection.isReverted) {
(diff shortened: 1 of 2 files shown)


haven2world (approved):
Looks great! Thanks for the quick fix!

PR #3344, opened 2026-05-19
Filter out invisible unicode characters from text segments

JiuqingSong:
## Summary
- Strip invisible Unicode tag characters (U+E0000–U+EFFFF) inside `createText` so they cannot survive paste/DOM-to-model conversion. These characters are used to hide instructions/text inside HTML (see https://embracethered.com/blog/posts/2024/hiding-and-finding-text-with-unicode-tags/) and otherwise leak into the model as normal text.
- Meaningful invisible characters that fall outside that range (e.g. ZWSP `U+200B`, ZWJ `U+200D`, RLO `U+202E`, PDF `U+202C`) are preserved.
- Unit tests in `creatorsTest.ts` cover mixed/boundary/only-invisible inputs and confirm meaningful invisible chars are untouched. An end-to-end test in `endToEndTest.ts` verifies a full DOM → Model → DOM/text round-trip strips only the tag range.

## Test plan
- [x] `yarn test:fast --testPathPattern=creatorsTest`
- [x] `yarn test:fast --testPathPattern=endToEndTest`

### DIFF
--- a/packages/roosterjs-content-model-dom/lib/modelApi/creators/createText.ts
+++ b/packages/roosterjs-content-model-dom/lib/modelApi/creators/createText.ts
@@ -19,9 +19,10 @@ export function createText(
     link?: ReadonlyContentModelLink,
     code?: ReadonlyContentModelCode
 ): ContentModelText {
+    const filterText = stripInvisibleUnicode(text);
     const result: ContentModelText = {
         segmentType: 'Text',
-        text: text,
+        text: filterText,
         format: { ...format },
     };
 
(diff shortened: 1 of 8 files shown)


packages/roosterjs-content-model-dom/lib/modelApi/creators/createText.ts
```diff
@@ -35,3 +36,17 @@ export function createText(
...
+// We need to strip them out before processing the pasted content, otherwise they will be treated as normal text and cause unexpected behavior.
+const INVISIBLE_UNICODE_REGEX = /[\u{E0000}-\u{EFFFF}]/gu;
+
+/**
+ * Strip invisible unicode characters from the given string
+ * @param value The string to be processed
+ * @returns The string with invisible unicode characters removed
+ */
+function stripInvisibleUnicode(value: string): string {
+    return value.replace(INVISIBLE_UNICODE_REGEX, '');
```
BryanValverdeU: I remember this was originally limited to the content was inserted as initial content in the editor and for Links. Do we have any perf concerns with applying the regex to all created text on every call?

JiuqingSong: This is a valid concern.

I investigated the original security issue, and realize that the attack can happen from any kind of source, as long as the content is put into editor. So manual operations (new editor, paste), or 3rd party code (call formatContentModel() can both trigger the result. Of cause the manual operation is easier to do. 

What do you think? Should we limit the check to manual operation only? I'm open to any suggestion.

@romanisa  fyi.

BryanValverdeU: Sounds good, we could add an experimental feature, to be extra careful, and that way we can measure whether the change can cause perf impact. 
Looks good to me.

BryanValverdeU (approved):

packages/roosterjs-content-model-dom/lib/modelApi/creators/createText.ts
JiuqingSong: Added experimental feature.

BryanValverdeU (approved):

PR #3351, opened 2026-05-26
[Draft] Test shadow dom in old browsers

BryanValverdeU:


JiuqingSong:
@BryanValverdeU what is this change? Do you plan to check it in?

PR #3360, opened 2026-06-04
Reconcile IMG src and data-* attribute changes in Content Model cache

JiuqingSong:
@
## Summary

Previously, any attribute change other than `id` on editor content set the cache mutation type to `unknown`, which **invalidated the whole cached Content Model** and forced a full reparse from DOM next time a model was requested.

This PR recognizes two more cases on an **indexed ``** and reconciles them in place instead of invalidating:

- **`src` change** → update `ContentModelImage.src`
- **`data-*` change** → update `ContentModelImage.dataset`

## Changes

- **`MutationType.ts`** — new `attribute` mutation type + `AttributeMutation` interface (`element`, `attributeName`).
- **`textMutationObserver.ts`** — the `attributes` case now recognizes `src` and `data-*` and emits an `attribute` mutation; everything else still falls through to `unknown`.
- **`DomIndexer.ts`** (types) — new `reconcileImageAttribute(element, attributeName)` method.
- **`domIndexerImpl.ts`** — implements it: `src` uses `getAttribute("src")` to match `imageProcessor` (port stripping); `data-*` rebuilds the whole `dataset` from `element.dataset` (covers add/modify/remove, including metadata stored as data attributes). Returns `false` for non-image / unindexed / unrelated attributes.
- **`CachePlugin.ts`** — dispatches the `attribute` mutation, invalidating only when reconcile returns `false`.

## Tests

- New observer tests (`src` / `data-*`), `reconcileImageAttribute` unit tests (indexed/unindexed, src set/removed, data-* add/modify/remove, unrelated attr, non-image), and `CachePlugin` dispatch tests.
- Updated the typed `DomIndexer` mock literals across dom-package tests to satisfy the interface.

All `yarn test:fast`, `yarn eslint`, and `yarn build` pass.

## Note

Adding a required member to the exported `DomIndexer` interface is technically a breaking change for any external implementer — same precedent as when `reconcileChildList` was added. It is effectively an internal interface.

@

### DIFF
--- a/packages/roosterjs-content-model-core/lib/corePlugin/cache/domIndexerImpl.ts
+++ b/packages/roosterjs-content-model-core/lib/corePlugin/cache/domIndexerImpl.ts
@@ -411,6 +411,38 @@ export class DomIndexerImpl implements DomIndexer {
         }
     }
 
+    reconcileImageAttribute(element: HTMLElement, attributeName: string) {
+        if (isElementOfType(element, 'img')) {
+            const image = getIndexedSegmentItem(element)?.segments[0];
+
+            if (image?.segmentType == 'Image') {
+                if (attributeName == 'src') {
+                    // Use getAttribute('src') instead of retrieving src directly, in case the src
+                    // has port and may be stripped by browser. This matches imageProcessor.
+                    image.src = element.getAttribute('src') ?? '';
+
+                    return true;
+                } else if (attributeName.indexOf('data-') == 0) {
+                    // A data-* attribute may be added, modified or removed. Rebuild the whole
+                    // dataset from DOM to keep it in sync, the same way imageProcessor builds it.
+                    const { dataset } = image;
+
+                    getObjectKeys(dataset).forEach(key => {
+                        delete dataset[key];
+                    });
+
+                    getObjectKeys(element.dataset).forEach(key => {
+                        dataset[key] = element.dataset[key] || '';
+                    });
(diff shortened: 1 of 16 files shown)


packages/roosterjs-content-model-core/lib/corePlugin/cache/domIndexerImpl.ts
```diff
@@ -411,6 +411,38 @@ export class DomIndexerImpl implements DomIndexer {
...
+            const image = getIndexedSegmentItem(element)?.segments[0];
+
+            if (image?.segmentType == 'Image') {
+                if (attributeName == 'src') {
+                    // Use getAttribute('src') instead of retrieving src directly, in case the src
+                    // has port and may be stripped by browser. This matches imageProcessor.
+                    image.src = element.getAttribute('src') ?? '';
+
+                    return true;
+                } else if (attributeName.indexOf('data-') == 0) {
```
vinguyen12, line 425: nit:  `attributeName.startsWith('data-')`

vinguyen12 (approved):

packages/roosterjs-content-model-core/lib/corePlugin/cache/domIndexerImpl.ts
JiuqingSong: I should teach AI to use this function next time. Haha.

PR #3362, opened 2026-06-09
Remove redundant leading/trailing selection markers within a paragraph

JiuqingSong:
## Summary

When `setSelection` applies a range selection, a leading or trailing `SelectionMarker` that sits next to a real (non-marker) selected segment **in the same paragraph** carries no selection information — the range is already fully defined by the selected content. This change removes those redundant boundary markers.

The cleanup is deliberately scoped to **within a single paragraph**. A marker at a paragraph edge must be preserved, because it is the only way to distinguish *"selection starts at the beginning of this line"* from *"selection starts at the end of the previous line"*. Collapsed-cursor markers (no other selected segment) are likewise preserved.

## Implementation

- `setSelection.ts`: a per-paragraph `ParagraphSelectionState` tracks whether any non-marker segment was selected (`hasSelectedNonMarker`) and which boundary markers were tentatively kept (`boundaryMarkers`). After the segment loop, the boundary markers are dropped if the paragraph contained real selected content. The delete list is sorted ascending before the pop/splice so the appended (lower-index) boundary markers don't corrupt indices.

## Tests

- Added 4 `setSelection` cases: both markers removed with content between; leading marker kept across a paragraph boundary; trailing marker kept across a paragraph boundary; collapsed marker preserved.
- Updated 3 `domIndexerImpl.reconcileSelection` snapshots (including Repro #3341). The #3341 `preserveMarker` guard is untouched and still hands `setSelection` two live markers, so the selected segment keeps `isSelected`; `setSelection` now additionally drops the redundant boundary markers.

## Validation

- `yarn test:fast` — 5901 passing, 0 failures
- `yarn eslint` — clean
- `yarn b` — builds clean

### DIFF
--- a/packages/roosterjs-content-model-core/test/corePlugin/cache/domIndexerImplTest.ts
+++ b/packages/roosterjs-content-model-core/test/corePlugin/cache/domIndexerImplTest.ts
@@ -595,7 +595,7 @@ describe('domIndexerImpl.reconcileSelection', () => {
         expect(paragraph).toEqual({
             blockType: 'Paragraph',
             format: {},
-            segments: [segment1, marker1, segment2, segment3, marker2, segment4],
+            segments: [segment1, segment2, segment3, segment4],
         });
         expect(setSelectionSpy).toHaveBeenCalledWith(model, marker1, marker2);
         expect(model).toEqual({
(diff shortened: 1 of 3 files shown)


haven2world (approved):
Verified on safari. Thanks for the fix!

PR #3363, opened 2026-06-12
Add Paste As Markdown support and isPastedContentMarkdown API

juliaroldi:
Adds two new public APIs to roosterjs-content-model-markdown and a "Paste as Markdown" demo button that uses them to decide whether to convert clipboard content with convertMarkdownToContentModel or fall back to the regular paste flow.

**New public APIs** `(roosterjs-content-model-markdown)`:

* `isContentMarkdown(text: string): boolean `— Heuristic check on plain text. Returns `true` if the text contains markdown markup, recognizing both block-level patterns (headings `#`, blockquotes `>`, ordered/unordered lists, horizontal rules `---`, table rows `| a | b |`) and inline patterns (bold `**text**`, italic `*text*`, strikethrough `~~text~~`, links `text`, images `!alt`).

* `isPastedContentMarkdown(editor: IEditor, clipboardData: ClipboardData): boolean `— Decides whether clipboard content should be treated as markdown. Returns `true` when the plain text contains markdown markup and the accompanying HTML (if any) is just a thin wrapper of the same plain text — i.e. only `<div>, <p>, <br>, <span> `elements with no attributes other than `class / style`, and the textual content matches. This avoids treating rich content (tables, images, lists, formatted text) as markdown, which would lose information when converted.

RoosterPasteAsMarkdown

### DIFF
--- a/demo/scripts/controlsV2/demoButtons/pasteAsMarkdownButton.ts
+++ b/demo/scripts/controlsV2/demoButtons/pasteAsMarkdownButton.ts
@@ -1,0 +1,47 @@
+import {
+    convertMarkdownToContentModel,
+    isPastedContentMarkdown,
+} from 'roosterjs-content-model-markdown';
+import { mergeModel } from 'roosterjs-content-model-dom';
+import { paste } from 'roosterjs-content-model-core';
+import { readClipboardData } from '../../utils/readClipboardData';
+import type { IEditor } from 'roosterjs-content-model-types';
+import type { RibbonButton } from 'roosterjs-react';
+
+/**
+ * @internal
+ * "Paste as Markdown" button on the format ribbon.
+ *
+ * Reads the clipboard, checks whether the content can be interpreted as markdown
+ * (via {@link isMarkdownContent}). If so, converts it with
+ * {@link convertMarkdownToContentModel} and merges the resulting model into the
+ * editor. Otherwise, falls back to the regular paste flow.
+ */
+export const pasteAsMarkdownButton: RibbonButton<'buttonNamePasteAsMarkdown'> = {
+    key: 'buttonNamePasteAsMarkdown',
+    unlocalizedText: 'Paste as Markdown',
+    iconName: 'Paste',
+    onClick: async editor => {
+        const doc = editor.getDocument();
+        const clipboardData = await readClipboardData(doc);
+        if (!clipboardData) {
+            return;
+        }
+
+        if (isPastedContentMarkdown(doc, clipboardData, editor.getDOMCreator())) {
+            insertMarkdown(editor, clipboardData.text);
+        } else {
+            paste(editor, clipboardData);
+        }
+    },
+};
+
+function insertMarkdown(editor: IEditor, text: string) {
+    const newModel = convertMarkdownToContentModel(text);
+    editor.formatContentModel((model, context) => {
+        mergeModel(model, newModel, context, {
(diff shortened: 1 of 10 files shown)


packages/roosterjs-content-model-markdown/lib/publicApi/isPastedContentMarkdown.ts
```diff
@@ -0,0 +1,79 @@
...
+    const { text, rawHtml } = clipboardData;
+
+    // There must be some plain text to interpret as markdown
+    if (!text || !text.trim()) {
+        return false;
+    }
+
+    // No HTML content at all (text/plain only), so the plain text is all we have
+    if (!rawHtml) {
+        return true;
```
cyanzhong: What happens if we copy markdown content from a plain text editor?

Does this mean "plain text is always markdown"? While this is theoretically true, it doesn't seem to work for our purpose in practical.

juliaroldi: When using convertMarkdownToContentModel, it goes through the plain text and applies any markdown styles it finds. If there are no markdown styles, it simply returns the plain text unchanged. So, if we paste plain text without markdown, it should return the text as is. Do we need to block the pasting if it has no markdown markup?

cyanzhong: In my Mac V1 demo, the "Paste Markdown" from the paste options floating button is dynamic.

image

It won't show up if the content is likely not markdown (the detection is heuristic and patterns are learned from `convertMarkdownToContentModel `).

I guess this is the point of `isPastedContentMarkdown`, since as you mentioned, `convertMarkdownToContentModel` works for plain text too, the actual paste action has only subtle difference.

demo/scripts/controlsV2/demoButtons/pasteAsMarkdownButton.ts
```diff
@@ -0,0 +1,47 @@
...
+    iconName: 'Paste',
+    onClick: async editor => {
+        const doc = editor.getDocument();
+        const clipboardData = await readClipboardData(doc);
+        if (!clipboardData) {
+            return;
+        }
+
+        if (isPastedContentMarkdown(doc, clipboardData, editor.getDOMCreator())) {
+            insertMarkdown(editor, clipboardData.text);
```
cyanzhong, line 32: I mean, if `isPastedContentMarkdown` cannot differentiate plain text and markdown, we can always call `insertMarkdown` here.

If we can differentiate, we can provide options conditionally.

juliaroldi: I was considering a different scenario where the user has rich text, such as tables, images, and bullet lists. In this case, we should block markdown since plain text content doesn't support these elements, and pasting could result in malformed content. For plain text only, we can choose to verify if it's strict markdown and convert it in that case, or always allow the user to paste.

cyanzhong: Fair point about the "blocking" behavior. Do we always provide "Paste Markdown" even the content isn't Markdown (well plain text is markdown... but you know this can create confusions)

Eventually, we may consider more aggressive approach: when we are confident about the pasted content is Markdown, it happens automatically (and there's a way to revert, the opposite of the current workflow)

JiuqingSong (approved):

PR #3381, opened 2026-06-23
[Markdown Plugin] Support auto-conversion on native paste and add undoConversion option

juliaroldi:
## Summary

Improves the Markdown paste plugin so that auto-conversion works for native (Ctrl+V) paste, not just the explicit "Paste as Markdown" path. `MarkdownPastePlugin` now reacts to the `contentChanged` event with source `Paste`: when `autoConversion` is enabled and the pasted content is detected as Markdown, it merges the converted Content Model into `modelBeforePaste` and re-applies it via `formatContentModel`. A new `undoConversion` option (default `false`) takes an extra undo snapshot before the conversion so the user can undo the Markdown conversion separately from the paste.

To distinguish genuine native paste events from programmatic clipboard reads, `extractClipboardItems` now accepts an `isPasteNative` flag that drives `ClipboardData.pasteNativeEvent` (previously hard-coded to `true`). `CopyPastePlugin` passes `true` for real paste events, while `demo/scripts/utils/readClipboardData.ts` leaves it `undefined`. The demo's editor options gain an "Undo auto-converted markdown" checkbox wired to the new option.

## How to test

1. Run the unit tests:
   `yarn test:fast --testPathPattern=MarkdownPastePlugin`
2. Manual: In the demo (`yarn builddemo`), enable the Markdown plugin with "Auto convert on paste" on. Copy Markdown text (e.g. `# Title` or `**bold**`) and paste with Ctrl+V.
   - Before this change: native paste did not auto-convert Markdown (only "Paste as Markdown" did).
   - After this change: the pasted Markdown is converted in place.
3. With "Undo auto-converted markdown" enabled, paste Markdown and press Ctrl+Z — the conversion is undone as its own step.

### DIFF
--- a/packages/roosterjs-content-model-markdown/lib/plugins/MarkdownPastePlugin.ts
+++ b/packages/roosterjs-content-model-markdown/lib/plugins/MarkdownPastePlugin.ts
@@ -81,7 +81,8 @@ export class MarkdownPastePlugin implements EditorPlugin {
             event.data
         ) {
             const clipboardData = event.data as ClipboardData;
-            const shouldConvert = this.options.autoConversion && clipboardData.pasteNativeEvent;
+            const shouldConvert =
+                clipboardData && clipboardData.pasteNativeEvent && this.options.autoConversion;
             if (
                 shouldConvert &&
                 isPastedContentMarkdown(this.editor, clipboardData) &&
--- a/packages/roosterjs-content-model-markdown/lib/plugins/MarkdownPastePlugin.ts
+++ b/packages/roosterjs-content-model-markdown/lib/plugins/MarkdownPastePlugin.ts
@@ -1,12 +1,11 @@
 import { convertMarkdownToContentModel } from '../markdownToModel/convertMarkdownToContentModel';
 import { isPastedContentMarkdown } from '../publicApi/isPastedContentMarkdown';
-
 import {
     contentModelToDom,
     createModelToDomContext,
     mergeModel,
     ChangeSource,
-    cloneModel,
+    cloneModelForPaste,
 } from 'roosterjs-content-model-dom';
 import type { MarkdownPasteOptions } from './MarkdownPasteOptions';
 import type {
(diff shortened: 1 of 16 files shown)


packages/roosterjs-content-model-markdown/lib/plugins/MarkdownPastePlugin.ts
```diff
@@ -58,14 +69,46 @@ export class MarkdownPastePlugin implements EditorPlugin {
...
-
-        if (shouldConvert && isPastedContentMarkdown(this.editor, event.clipboardData)) {
-            convertPastedTextToMarkdown(this.editor, event.fragment, event.clipboardData.text);
+        if (
+            event.eventType === 'contentChanged' &&
+            event.source === ChangeSource.Paste &&
+            this.options.autoConversion
+        ) {
+            const clipboardData = event.data as ClipboardData;
+            const shouldConvert = this.options.autoConversion && clipboardData.pasteNativeEvent;
```
JiuqingSong: Need null check for clipboardData

packages/roosterjs-content-model-markdown/lib/plugins/MarkdownPastePlugin.ts
```diff
@@ -58,14 +69,46 @@ export class MarkdownPastePlugin implements EditorPlugin {
...
+            if (
+                shouldConvert &&
+                isPastedContentMarkdown(this.editor, clipboardData) &&
+                clipboardData.modelBeforePaste
+            ) {
+                mergeModel(
```
JiuqingSong, line 91: Suggest clone a model from clipboardData.modelBeforePaste, but do not modify the original one.

packages/roosterjs-content-model-markdown/lib/plugins/MarkdownPastePlugin.ts
JiuqingSong: `clipboardData` itself needs null check. It comes from event.data. But event.data is not guaranteed to have valid value.

packages/roosterjs-content-model-markdown/lib/plugins/MarkdownPastePlugin.ts
```diff
@@ -58,14 +71,47 @@ export class MarkdownPastePlugin implements EditorPlugin {
...
+        ) {
+            const clipboardData = event.data as ClipboardData;
+            const shouldConvert =
+                clipboardData && clipboardData.pasteNativeEvent && this.options.autoConversion;
+            if (
+                shouldConvert &&
+                isPastedContentMarkdown(this.editor, clipboardData) &&
+                clipboardData.modelBeforePaste
+            ) {
+                const modelBeforePaste = cloneModel(clipboardData.modelBeforePaste);
```
JiuqingSong: You may need to use function `cloneModelForPaste` instead of `cloneModel`, they have different handling for entities

JiuqingSong (approved):

PR #3382, opened 2026-06-25
Move getDocumentSource to core and expose pasteSource on BeforePasteEvent

BryanValverdeU:
## Summary

Moves paste-source detection out of `roosterjs-content-model-plugins` and into `roosterjs-content-model-core`, and exposes the detected source on `BeforePasteEvent` so plugins consume it from the event rather than recomputing it. Additionally add an extra param to pasteTypeGetter, so we can consume the paste source and change the paste type if the contents are comming from a O365 document.

## Changes

- Move the `pasteSourceValidations` folder (`getDocumentSource`, all validators, `constants`) and their tests from `roosterjs-content-model-plugins` to `roosterjs-content-model-core` (core cannot depend on plugins).
- Add `KnownPasteSourceType` to `roosterjs-content-model-types` and export it.
- Add a `readonly pasteSource` property to `BeforePasteEvent` and a `pasteSource` argument to the `PasteTypeOrGetter` callback.
- `paste.ts` now computes the paste source (new `getPasteSource` helper) and passes it to the paste-type getter and to `generatePasteOptionFromPlugins`, which sets it on the event.
- `PastePlugin` consumes `event.pasteSource` instead of calling `getDocumentSource`; uses a local `GOOGLE_SHEET_NODE_NAME` constant.
- Make `documentContainWacElements` self-contained and drop now-unused WAC selector constants from `WacComponents/constants.ts`.
- Update tests to set `event.pasteSource` and repoint spies to the core module.

## Validation

- `yarn eslint` - pass
- `yarn test:fast` - 5890 passing (33 skipped)
- `yarn build` - pass

## Notes

- Adding the getter's 3rd param is **not** a breaking change: in TS a function with fewer params is assignable to a type expecting more, so existing 2-arg getters still compile.
- `BeforePasteEvent.pasteSource` is optional to match existing optional props and avoid breaking event mocks; core always populates it.


juliaroldi:
Can we not add all this code to the roosterjs core? Maybe we can create an option in EditorOptions to pass a callback to detect the paste source? Otherwise, every time we need to detect a paste source we will need to change the roosterJs core code.

JiuqingSong:
> Can we not add all this code to the roosterjs core? Maybe we can create an option in EditorOptions to pass a callback to detect the paste source? Otherwise, every time we need to detect a paste source we will need to change the roosterJs core code.

I have the same feeling with @juliaroldi . Ideally this kind of code should be in plugin package. And actually the API `paste` itself, should be in api package. But just because the build-in paste feature also need to trigger the BeforePaste event and do proper handling, I need to put it in core package. (Now I feel at least some of the functionality can be moved out such as re-paste hanlding)

Any special reason we need to move the code into core? If we want the API paste to be smart enough, we can split it into two parts:
1. Basic paste, only handle Ctrl+V and context menu, trigger BeforePaste event with most common options, then do basic content merge
2. Customized paste (can be in API package or plugin package), do more advanced handling, such as repaste (change paste format), specify paste format (html, merge format,  plain text, markdown, ...)

And allow things to be overwritten from the PastePlugin.

PR #3383, opened 2026-06-26
Pin aria-live announcer to top:0; left:0 to avoid scroll overflow

daxiang94:
## Problem

`createAriaLiveElement` sets `position: absolute` on the screen-reader announcer container but leaves `top`/`left` unset. CSS resolves them to the element's *static position* — its place in normal flow, which sits after every other child of the announcer's eventual container.

When that container has a non-trivial content height (typical app layouts), the absolute element contributes to the `documentElement`'s scrollable overflow region and pushes out a small but visible scrollbar even though nothing is laid out beyond the viewport.

## Repro

Mount the editor in any host page where document body has a layout height >= viewport. A 1px-tall scrollbar appears on `<html>` even with no editor content overflowing. Removing the announcer from DOM at runtime confirms it is the cause.

## Fix

Standard visually-hidden patterns anchor the element to a known viewport corner. Add `top: 0; left: 0` so the absolute element sits at the top-left of its containing block and never extends the scroll area.

- Element remains invisible: `clip-path: inset(100%)` + 1×1 size + `overflow: hidden`
- `aria-live` continues to announce `textContent` changes — no behavior change for screen readers

## Files changed

- `packages/roosterjs-content-model-core/lib/utils/createAriaLiveElement.ts` (+2)
- `packages/roosterjs-content-model-core/test/coreApi/announce/announceTest.ts` (+8, matching expected style in existing `toEqual` assertions)

## Checklist

- [x] Code passes linting
- [x] Pre-commit hooks passed
- [x] Follows existing code patterns and conventions


daxiang94:
@microsoft-github-policy-service agree company="Alibaba"

packages/roosterjs-content-model-core/lib/utils/createAriaLiveElement.ts
```diff
@@ -9,6 +9,8 @@ export function createAriaLiveElement(document: Document): HTMLDivElement {
     div.style.height = '1px';
     div.style.overflow = 'hidden';
     div.style.position = 'absolute';
+    div.style.top = '0';
```
JiuqingSong, line 12: @BryanValverdeU please take a look

BryanValverdeU (approved):

PR #3396, opened 2026-07-09
[Selection]  keep image selection when 'selectionchange' range is collapsed

liangdrime:
### Summary

This change is to prevent `SelectionPlugin` from converting an image selection into a range selection when the browser reports a collapsed range during "selectionchange".

### Issue

> [Fix selection with ctrl+a #2556](https://github.com/microsoft/roosterjs/pull/2556)
>
> // If am image selection changed to a wider range due a keyboard event, we should update the selection

The logic for converting image selection to a range was first introduced in #2566. According to its description, this change was intended to correct the selection to a range when it expands to cover a wider area, such as when triggered by keyboard shortcuts like Ctrl+A.

However, on Mac, right-clicking an image while it is in editing mode converts the collapsed image selection to a range selection, causing the subsequent 'Crop' action from the context menu to fail.

image

image

`onSelectionChange` currently converts image selection to range whenever `isSingleImageInSelection(range)` returns null.
For collapsed ranges, that helper correctly returns null, but this was treated as a wider-range transition and caused an unintended downgrade from image selection to collapsed range selection.

### Fix

The intended conversion is for wider range transitions only. **A collapsed range is an empty caret state, not a wider selection**, so converting image selection in that case is not semantically correct and causes product regressions.

So, this change updated `SelectionPlugin.onSelectionChange` and image to range conversion only happens when the current range is not collapsed.

// local host demo behaves same with demo site.
ImageSelectionFix

### DIFF
--- a/packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
+++ b/packages/roosterjs-content-model-core/lib/corePlugin/selection/SelectionPlugin.ts
@@ -738,11 +738,11 @@ class SelectionPlugin implements PluginWithState<SelectionPluginState> {
             const newSelection = this.editor.getDOMSelection();
             const domHelper = this.editor.getDOMHelper();
 
-            //If am image selection changed to a wider range due a keyboard event, we should update the selection
+            // If an image selection changed to a wider range due to a keyboard event, we should update the selection
             const range = domHelper.getSelectionRange();
             if (range) {
                 const image = isSingleImageInSelection(range);
-                if (newSelection?.type == 'image' && !image) {
+                if (newSelection?.type == 'image' && !image && !range.collapsed) {
                     const sel = this.editor.getDocument().defaultView?.getSelection();
                     const isReverted = sel
                         ? sel.focusNode != range.endContainer || sel.focusOffset != range.endOffset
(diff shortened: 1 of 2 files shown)


liangdrime:
> https://github.com/microsoft/roosterjs/actions/runs/29032470134/job/86168595237?pr=3396

```
Chrome 150.0.0.0 (Linux x86_64) retrieveHtmlInfo Has global CSS rule FAILED
	Error: Expected $.globalCssRules[2].text = 'border-width: medium; border-style: none; border-color: currentcolor; border-image: none;' to equal 'border-width: medium; border-style: none; border-color: currentcolor; border-image: initial;'.
	    at <Jasmine>
	    at runTest (webpack://roosterjs/packages/roosterjs-content-model-core/test/command/paste/retrieveHtmlInfoTest.ts:19:24 <- karma.test.all.2302904290.js:71236:24)
	    at UserContext.<anonymous> (webpack://roosterjs/packages/roosterjs-content-model-core/test/command/paste/retrieveHtmlInfoTest.ts:203:9 <- karma.test.all.2302904290.js:71355:9)
	    at <Jasmine>
```
The `test-on-chrome` failed in `retrieveHtmlInfoTest.ts`. Because the newer Chrome (the CI runner is on 150) expands the `border: none` shorthand's `border-image` longhand to `none` instead of `initial`.

The fix is simple: updates `border-image: initial` to `border-image: none`. However, if this test does not block the change, it should be addressed in a separate PR. Unfortunately, I'm unable to re-run it at this time.

juliaroldi (approved):

PR #3399, opened 2026-07-14
 Reset spanAbove when merging tables to avoid unexpected vertical merge on paste

BryanValverdeU:
## Summary

When pasting cells that came from a horizontal split located in the same row as a
vertically split cell, the pasted cells were unexpectedly getting vertically merged.
This resets `spanAbove` on all incoming source-table cells during a table-into-table
merge, so pasted cells become standalone cells instead of continuations of a vertical merge.

## Repro

1. New mail
2. Insert a table
3. Split a cell horizontally > Split another cell vertically in the same row
4. Input some content in the horizontal split cells
5. Copy the horizontal split cells and then paste
6. Observe
[Bug 414670](https://outlookweb.visualstudio.com/Outlook%20Web/_workitems/edit/414670): [Table] Pasting horizontally split cells should not cause the pasted content to merge cells

Actual: The pasted cells were vertically merged.
Expect: Pasting should not cause cells to merge.

## Changes

- `mergeModel.ts` (`mergeTables`): reset `spanAbove = false` on every source cell before merging.
- `mergeModelTest.ts`: added test `merge source table with spanAbove cells - should reset spanAbove on merged cells`.

## Testing

`yarn test:fast --testPathPattern=mergeModelTest` - 64 passing. Verified the new test fails without the fix and passes with it.

### DIFF
--- a/packages/roosterjs-content-model-dom/lib/modelApi/editing/mergeModel.ts
+++ b/packages/roosterjs-content-model-dom/lib/modelApi/editing/mergeModel.ts
@@ -198,6 +198,12 @@ function mergeTables(
         const lastTargetColIndex = getTargetColIndex(table, rowIndex, colIndex, newTableColCount);
         const extraColsNeeded = lastTargetColIndex - table.rows[0].cells.length;
 
+        newTable.rows.forEach(row => {
+            row.cells.forEach(cell => {
+                cell.spanAbove = false;
+            });
+        });
+
         if (extraColsNeeded > 0) {
             const currentColCount = table.rows[0].cells.length;
             for (let col = 0; col < extraColsNeeded; col++) {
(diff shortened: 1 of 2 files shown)


juliaroldi (approved):

packages/roosterjs-content-model-dom/lib/modelApi/editing/mergeModel.ts
```diff
@@ -198,6 +198,12 @@ function mergeTables(
         const lastTargetColIndex = getTargetColIndex(table, rowIndex, colIndex, newTableColCount);
         const extraColsNeeded = lastTargetColIndex - table.rows[0].cells.length;
 
+        newTable.rows.forEach(row => {
```
JiuqingSong, line 201: Is it possible that newTable really has spanned cells?

BryanValverdeU: It is possible, but the current approach follows what Word does, so I propose we go with this for now and listen user feedback in case they want to retain the spanned cells, that is still not the most common scenario though

JiuqingSong: hmm ok.

JiuqingSong (approved):

PR #3400, opened 2026-07-15
Do not scroll on paste images

juliaroldi:
Pasting large images can cause the editor to jump far into view, which forces the user to scroll back to where the cursor was. When images are pasted, the caret should not be scrolled into view.

### DIFF
--- a/packages/roosterjs-content-model-core/lib/command/paste/mergePasteContent.ts
+++ b/packages/roosterjs-content-model-core/lib/command/paste/mergePasteContent.ts
@@ -7,6 +7,8 @@ import {
     getSelectedSegments,
     mergeModel,
     cloneModelForPaste,
+    isNodeOfType,
+    isElementOfType,
 } from 'roosterjs-content-model-dom';
 import type {
     BeforePasteEvent,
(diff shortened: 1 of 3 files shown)


packages/roosterjs-content-model-core/lib/command/paste/mergePasteContent.ts
```diff
@@ -137,3 +141,22 @@ function getLastSegmentFormat(pasteModel: ContentModelDocument): ContentModelSeg
 
     return {};
 }
+
+function isImageFragment(pasteFragment: DocumentFragment): boolean {
```
JiuqingSong: So this check if the only pasted content is image? 

Then you can name it `isImageOnlyFragment`

And to simplify the check, you can first querySelectorAll('img') and check there is exactly one result, and check element.textContent.trim() == ''

JiuqingSong: btw, what if the pasted content is a table and its only content is an image? do you want to scroll in that case?

juliaroldi: I don't think we should allow scrolling when a table contains only an image, and the same goes for an anchor element that contains only an image.

JiuqingSong (approved):

PR #3401, opened 2026-07-16
[Table Borders] Pass experimental features to paste sanitizing context

juliaroldi:
## Summary

Forward the editor's experimental features into `createDomToModelContextForSanitizing` so that the `TransformTableBorderColors` experimental feature is honored while sanitizing pasted content. Previously `experimentalFeatures` was hard-coded to an empty array in the sanitizing context, so table border color transformation never ran during paste. `mergePasteContent now passes all editor experimental features.

## How to test

1. Run the affected unit tests:
   ```
   yarn test:fast --testPathPattern=createDomToModelContextForSanitizing
   ```
2. Manual check: enable the `TransformTableBorderColors` experimental feature on the editor, then copy a table with themed/`currentColor` borders from another source and paste it into the editor.
   - Before this fix: pasted table border colors are not transformed (feature ignored during paste).
   - After this fix: pasted table border colors are transformed as expected.

### DIFF
--- a/packages/roosterjs-content-model-core/lib/command/paste/mergePasteContent.ts
+++ b/packages/roosterjs-content-model-core/lib/command/paste/mergePasteContent.ts
@@ -49,7 +49,10 @@ export function mergePasteContent(
                 undefined /*defaultFormat*/,
                 editor.getEnvironment().domToModelSettings.customized,
                 domToModelOption,
-                editor.getDOMHelper()
+                editor.getDOMHelper(),
+                editor.isExperimentalFeatureEnabled('TransformTableBorderColors')
+                    ? ['TransformTableBorderColors']
+                    : undefined
             );
 
             domToModelContext.segmentFormat = getSegmentFormatForPaste(model, pasteType);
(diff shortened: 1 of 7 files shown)


packages/roosterjs-content-model-core/lib/command/paste/mergePasteContent.ts
```diff
@@ -49,7 +49,10 @@ export function mergePasteContent(
                 undefined /*defaultFormat*/,
                 editor.getEnvironment().domToModelSettings.customized,
                 domToModelOption,
-                editor.getDOMHelper()
+                editor.getDOMHelper(),
+                editor.isExperimentalFeatureEnabled('TransformTableBorderColors')
```
JiuqingSong: I think you can just directly pass in the whole  editor.isExperimentalFeatureEnabled, no need to make this feature special.

JiuqingSong: I mean, no need to check "TransformTableBorderColors", just directly pass the whole array in.

JiuqingSong (approved):

PR #3404, opened 2026-07-17
[Image Edit] Skip DOM selection during image edit to prevent editor jump

juliaroldi:
## Summary

Add a `skipDOMSelection` option to `FormatContentModelOptions` and honor it in
`formatContentModel`: when `true`, the DOM selection is not restored after the
DOM tree is written (`setContentModel` is always called with
`{ ignoreSelection: true }`, even when the editor has focus). `ImageEditPlugin`
now passes `skipDOMSelection: true` when entering image edit mode.

This replaces the previous workaround in `createImageWrapper` that pinned the
shadow host's `width`/`height` to the original image's footprint to stop the
editor from jumping. Before this change, wrapping the image re-applied the DOM
selection, causing the browser to scroll the caret back into view and the
editor to jump. After this change, the selection is left untouched during image
edit, so the shadow host no longer needs fixed dimensions.

## How to test

1. Run the affected unit tests:
   ```
   yarn test:fast --testPathPattern="formatContentModel|createImageWrapper"
   ```
   All tests should pass, including the new `skipDOMSelection` cases in
   `formatContentModelTest.ts`.
2. Manual check: in the editor demo, select an image to enter edit/resize mode.
   - Before this fix: the editor may scroll/jump when the resize wrapper is
     applied.
   - After this fix: selecting the image enters edit mode without the editor
     jumping.


JiuqingSong:
Found two bugs with this change.

1. Select image, then unselect (by clicking somewhere else), image is moved to next line:
 
bug1

2. Type some text before and after image, select image, then keep typing. Before: typed text will replace image. After: first the new text appeared at line end, then it will appear before image.

bug2

juliaroldi:
@JiuqingSong 

> 1. Select image, then unselect (by clicking somewhere else), image is moved to next line

This was happening because the shadow DOM wrapper had an id, which caused the content model to rewrite it as a div inside the editor. I removed the id since it wasn’t used anywhere.

> 2. Type some text before and after image, select image, then keep typing. Before: typed text will replace image. After: first the new text appeared at line end, then it will appear before image. 

That bug was caused by me not setting the selection when clicking outside the image. I updated the code so DOMSelection is only skipped when the image wrapper is added.

JiuqingSong (approved):

PR #3412, opened 2026-07-23
[DragAndDrop] Support dropping internal content within the same editor

juliaroldi:
## Summary

Adds support for **dropping content within the same editor** (internal drag-and-drop), gated behind the new `HandleDropInternalContent` experimental feature.

- `DragAndDropPlugin` now tracks internal drags (`internalDrag` flag set on `dragstart`) and, on `beforeDrop`, routes internal drops to the new `handleDroppedInternalContent` handler while external drops continue through the renamed `handleDroppedExternalContent` (formerly `handleDroppedContent`).
- `handleDroppedInternalContent` inserts the dragged content at the drop position using the Content Model: it resolves the drop point via `getNodePositionFromEvent`, clones and trims the current selection, deletes the original selection, then merges the cloned model at the insertion point and restores the selection across the newly inserted range.
- Adds the `HandleDropInternalContent` value to the `ExperimentalFeature` type and wires it into the demo site's editor options and experimental features panel.

## How to test

1. Run the drag-and-drop unit tests:
   ```
   yarn test:fast --testPathPattern=dragAndDrop
   ```
2. Manual: build the demo (`yarn builddemo`), enable the **HandleDropInternalContent** experimental feature, then select content in the editor and drag it to a different position within the same editor.
   - Before this change: internal drops were effectively no-ops (the internal drag flag was only cleared).
   - After this change: the dragged content is moved to the drop location and the moved content stays selected.


packages/roosterjs-content-model-plugins/lib/dragAndDrop/utils/handleDroppedInternalContent.ts
```diff
@@ -0,0 +1,55 @@
...
+    const doc = editor.getDocument();
+    const domPosition = getNodePositionFromEvent(doc, editor.getDOMHelper(), event.x, event.y);
+
+    if (domPosition) {
+        event.preventDefault();
+        event.stopPropagation();
+
+        const range = doc.createRange();
+        range.setStart(domPosition.node, domPosition.offset);
+        range.collapse(true);
```
JiuqingSong: It seems range is not used at all

packages/roosterjs-content-model-plugins/lib/dragAndDrop/DragAndDropPlugin.ts
```diff
@@ -51,7 +59,15 @@ export class DragAndDropPlugin implements EditorPlugin {
         this.disposer = editor.attachDomEvent({
             dragstart: {
                 beforeDispatch: _ev => {
-                    this.isInternalDragging = true;
+                    this.internalDrag = true;
+                    if (this.editor?.isExperimentalFeatureEnabled('HandleDropInternalContent')) {
+                        const model = editor.getContentModelCopy('disconnected');
```
JiuqingSong: be careful about the "disconnected" mode, make sure you test the case when selected content has entity, see after drop if the entity is correctly dropped and if the DOM element is still the same one

JiuqingSong (approved):

PR #3418, opened 2026-07-28
Fix image selection for context menu on Shift+F10

juliaroldi:
## Summary

When a range selection contains a single image, pressing **Shift+F10** (the keyboard shortcut that opens the context menu) now promotes the selection to an image selection before the menu opens. This ensures the context menu operates on the image rather than the surrounding text range. Implemented in `SelectionPlugin.ts` by detecting `Shift+F10` on a range selection and, when `isSingleImageInSelection` returns an image, calling `setDOMSelection` with an `image` selection.

ImageContextMenuF10

The demo `MainPane.tsx` change moves `imageEditPlugin` initialization into `render()` so it reflects current init state.

## How to test

1. Manual check: In the demo editor, click just before/around a single image so it's within a range selection, then press **Shift+F10**.
   - Before this fix: the context menu opens against the text range.
   - After this fix: the image becomes selected (image selection) before the context menu opens.

### DIFF
--- a/demo/scripts/controlsV2/mainPane/MainPane.tsx
+++ b/demo/scripts/controlsV2/mainPane/MainPane.tsx
@@ -176,16 +176,15 @@ export class MainPane extends React.Component<{}, MainPaneState> {
             },
             activeTab: 'all',
         };
+    }
 
+    render() {
+        const theme = getTheme(this.state.isDarkMode);
         this.imageEditPlugin = this.state.initState.pluginList.imageEditPlugin
             ? new ImageEditPlugin({
                   disableSideResize: this.state.initState.disableSideResize,
               })
             : undefined;
-    }
-
-    render() {
-        const theme = getTheme(this.state.isDarkMode);
 
         return (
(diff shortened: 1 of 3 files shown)


demo/scripts/controlsV2/mainPane/MainPane.tsx
```diff
@@ -176,16 +176,15 @@ export class MainPane extends React.Component<{}, MainPaneState> {
             },
             activeTab: 'all',
         };
+    }
 
+    render() {
+        const theme = getTheme(this.state.isDarkMode);
         this.imageEditPlugin = this.state.initState.pluginList.imageEditPlugin
```
JiuqingSong: You can recreate plugin in `getToggleablePlugins`

juliaroldi: Recreate the plugin in getToggleablePlugins will make the ribbons buttons lose the plugin reference. I added a change to reset the plugin when we reset the editor, so the plugin will be reset in the editor and for ribbon buttons.

JiuqingSong (approved):

PR #3424, opened 2026-08-03
[Drag and Drop] Handle drag-out deletion in beforeinput

juliaroldi:
## Summary

Update `DragAndDropPlugin` to handle the browser's `deleteByDrag` `beforeinput` event after an internal drag. The plugin now prevents the native deletion and removes the selection through the Content Model with the `DragOutOfEditor` change source, so drag-out deletions participate in the editor's normal change and undo flow.

## How to test

1. Run `yarn test:fast --testPathPattern=DragAndDropPluginTest`.
2. Enable `HandleDropInternalContent`, select editor content, and drag it outside the editor. Verify the selected content is removed and Undo restores it.
3. Trigger a non-`deleteByDrag` input after starting a drag and verify the selection is not deleted by the drag-and-drop plugin.


JiuqingSong:
When drag out of editor and drop, we don't delete from editor, right? Then why do we need to take a snapshot?

juliaroldi:
> When drag out of editor and drop, we don't delete from editor, right? Then why do we need to take a snapshot?

We do not delete if the content is dragged inside the same window. Ex: drag from the editor to the markdown editor, the content is deleted.

JiuqingSong:
> > When drag out of editor and drop, we don't delete from editor, right? Then why do we need to take a snapshot?
> 
> We do not delete if the content is dragged inside the same window. Ex: drag from the editor to the markdown editor, the content is deleted.


I just tried, yes the content is deleted in that case. But it seems this is done by browser, but not by our code. This may cause some problem since our code does not know something got deleted. I think we need to explore a way that we can know drop is happening and content is being deleted, so we can delete it with our code, and take snapshot, and trigger ContentChanged event

JiuqingSong (approved):

PR #3449, opened 2026-08-24
Merge pasted adjacent tables

juliaroldi:
## Summary

Merge a pasted table into the table immediately before the insertion point instead of creating a separate adjacent table. The merge supports different column counts and appends the pasted rows without reapplying destination table formatting, preserving source cell border colors.

MergeTables

## How to test

1. Run `yarn test:fast --testPathPattern=mergeModelTest`.
2. Place the caret directly after a table and paste another table. Verify the pasted rows join the preceding table for both matching and different column counts, and that the pasted cells retain their original border colors.

### DIFF
--- a/packages/roosterjs-content-model-dom/lib/modelApi/editing/mergeModel.ts
+++ b/packages/roosterjs-content-model-dom/lib/modelApi/editing/mergeModel.ts
@@ -332,9 +332,35 @@ function mergeTables(
 
         normalizeTable(table, markerPosition.marker.format);
         applyTableFormat(table, undefined /*newFormat*/, true /*keepCellShade*/);
-    } else {
+    } else if (!mergeTableAfterPreviousTable(markerPosition, newTable)) {
         insertBlock(markerPosition, newTable);
     }
+}
+
+function mergeTableAfterPreviousTable(
+    markerPosition: InsertPoint,
+    newTable: ContentModelTable
+): boolean {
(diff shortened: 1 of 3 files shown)


packages/roosterjs-content-model-dom/lib/modelApi/editing/mergeModel.ts
```diff
@@ -332,11 +332,37 @@ function mergeTables(
...
+    const parent = path[0];
+    const paragraphIndex = parent.blocks.indexOf(paragraph);
+    const previousBlock = parent.blocks[paragraphIndex - 1];
+
+    if (
+        paragraphIndex > 0 &&
+        paragraph.segments.indexOf(marker) == 0 &&
+        previousBlock?.blockType == 'Table' &&
+        previousBlock.rows[0]?.cells.length > 0 &&
+        newTable.rows[0]?.cells.length > 0
```
JiuqingSong, line 355: What if their cells count are different?

juliaroldi: Good point. At first, only the new rows were added to the end of the table, but after the model was updated, that changed the table’s appearance. I added code to expand the smaller table to match the size of the larger one.
MergeTablesDiffSize

JiuqingSong (approved):

PR #3450, opened 2026-08-24
Support partial shortcut overrides in ShortcutPlugin

vinguyen12:
Adds partial override support to `ShortcutPlugin`, allowing consumers to customize selected built-in shortcut keys without replacing the complete command list.

```ts
new ShortcutPlugin({
    shortcutOverrides: {
        bold: {
            modifierKey: 'ctrl',
            shiftKey: false,
            which: 78,
        },
    },
});
```
An override changes only the selected command's shortcut key. Its callback, environment, command identity, matching order, and all unaffected default shortcuts remain unchanged.

Backward compatibility

•  new ShortcutPlugin()  continues to use all built-in shortcuts.
•  new ShortcutPlugin(undefined)  continues to use all built-in shortcuts.
• Passing a  ShortcutCommand[]  continues to replace the complete built-in command list.
• Existing shortcut matching behavior and precedence are preserved.

API changes

Exports:

•  ShortcutPluginOptions 
•  DefaultShortcutCommandName

### DIFF
--- a/packages/roosterjs-content-model-plugins/lib/index.ts
+++ b/packages/roosterjs-content-model-plugins/lib/index.ts
@@ -26,7 +26,11 @@ export {
     ShortcutIndentList,
     ShortcutOutdentList,
 } from './shortcut/shortcuts';
-export { ShortcutPlugin } from './shortcut/ShortcutPlugin';
+export {
+    ShortcutPlugin,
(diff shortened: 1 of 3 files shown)


vinguyen12:
Shortcuts are exported so we can build entire shortcut list in consumer apps

PR #3457, opened 2026-09-03
Refactor table styles

juliaroldi:
## Summary

Refactor the demo table style presets to add reusable color templates and accent variants, and update `applyTableFormat` to correctly replace header-row and first-column text formatting when styles change.

## How to test

1. Run `yarn test:fast --testPathPattern="(formatTable|applyTableFormat|borderColorFormatHandler)"`.
2. In the demo, insert a table, apply multiple table presets and accent variants, then toggle header-row and first-column options. Confirm borders and shading match the selected preset, stale bold/italic/alignment formatting is removed when switching styles, and border colors adapt when dark mode is toggled.


packages/roosterjs-content-model-dom/lib/modelApi/editing/applyTableFormat.ts
```diff
@@ -388,17 +465,20 @@ function setHeaderRowFormat(
...
-                            } else if (cellSegment.format.italic) {
-                                delete cellSegment.format.italic;
-                            }
-                        });
-                    }
+        }
+
+        for (const block of cell.blocks) {
+            if (block.blockType == 'Paragraph') {
+                for (const segment of block.segments) {
```
JiuqingSong: nit: you only need to call mutateBlock() just one time:
```javascript
const mutablePara = mutateBlock(block);
for (const segment of mutablePara.segments) {
   ... // now segment is already mutable
}
```

JiuqingSong (approved):

BryanValverdeU (approved):

PR #3459, opened 2026-09-04
Avoid losing common ancestors on git history

ccasbre27:
When bumping rooster we might use a git squash approach, which ends losing the common ancestor, and during the commit history we might see an odd timelapse. As a result, improving this. On this PR, we add instructions to the agent used to bump by developers, and also adding new steps for the skill to verify, the branch is targeting release and not using squash approach.

In addition, this is the first step of the approach, since on another PR we will add a github action to check during the build, in case the agent wasn't used.

For example: below image you can see is showing commits from around 3 months ago, while the latest bump happened around 15 days ago. So, looking forward to display only the changes made since last bump. 

image

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,0 @@
-registry=https://packagefeedproxy.microsoft.io/npm/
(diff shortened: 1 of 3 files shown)


.npmrc
juliaroldi: Do not delete this file.

JiuqingSong, line 1: why delete this?

PR #3463, opened 2026-09-10
Place the cursor after a table when clicking beside it

juliaroldi:
## Summary

Update `SelectionPlugin` to place the cursor in the block after a table when the user clicks to the right of the table. If the table is the editor's final block, create an empty paragraph first so the cursor has a valid destination, while preserving normal selection behavior for clicks below the table or on following content.
ClickAfterTable

## How to test

1. Run `yarn test:fast --testPathPattern=SelectionPluginTest`.
2. In an editor containing a table, click to the right of a table row and confirm the cursor moves to the first line after the table. Repeat with a table as the final block and confirm an empty paragraph is created below it and receives the cursor.


JiuqingSong:
Is there a bug link for this issue? 

Normally, moving focus should not change the doc. We can put focus after table, and only create paragraph when user type something.

PR #3464, opened 2026-09-11
Toggle matching table borders

juliaroldi:
## Summary

Update `applyTableBorderFormat` to toggle the targeted table borders off when every selected border already matches the requested format, while preserving the existing apply behavior when any target differs. Normalize equivalent CSS border values before comparison and add coverage for border operations across selection shapes, directions, repeated operations, and table defaults.
borderToggle

## How to test

1. Run `yarn test:fast --testPathPattern=applyTableBorderFormat`.
2. In a table selection, apply a border format once and confirm the targeted borders appear; apply the same format again and confirm only those targeted borders are removed. Apply it again and confirm the borders reappear.


packages/roosterjs-content-model-api/lib/publicApi/table/applyTableBorderFormat.ts
```diff
@@ -386,6 +403,57 @@ export function applyTableBorderFormat(
...
+        for (const pos of positions) {
+            const value = cell.format[pos] || '';
+
+            // Exact matches need no CSS parsing. Equivalent values are parsed only once.
+            if (value == borderFormat || matchingBorders.has(value)) {
+                continue;
+            }
+
+            if (!comparisonStyle) {
+                comparisonStyle = editor.getDocument().createElement('div').style;
```
JiuqingSong: why need to create a DIV?

juliaroldi: It was to normalize the border color before comparing the border color being set with the previous one. I refactored the code to use parseColor before comparing the color, then we do not need the div anymore.

JiuqingSong (approved):

